PR #22, opened 2026-06-06
ci: bootstrap GitHub Actions workflow

dodizzle:
Adds a minimal `ci.yml` that runs `task ci` (lint → typecheck → test) on every PR and push to main.

Unblocks the deploy cron — `check-deploy.ts` requires a passing CI run before merging. Full gate set (secret-scan, task-store doctor, e2e) lands in SW-1.7 once the Phase A code is ported.

No code changes — workflow only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
 
       - uses: oven-sh/setup-bun@v2
         with:
-          bun-version: latest
+          bun-version: "1.x"
 
       - name: Install go-task
         uses: arduino/setup-task@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,34 @@
+name: CI
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  ci:
+    name: lint / typecheck / test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: latest
```
dodizzle: Pin this to `"1.x"` or an exact version — `latest` picks up breaking Bun releases silently and CI gates are merge-blocking.

dodizzle (commented):
APPROVE — workflow is correct and unblocks the deploy cron. One thing worth fixing before this becomes the permanent merge gate: `bun-version: latest` is a floating pin that can silently break CI on any Bun release. Pin to `"1.x"` or an exact version.

PR #74, opened 2026-06-07
feat(she-1.2): port crypto and agent config service layer

dodizzle:
## Summary

- Ports AES-256-GCM `crypto.ts` and `token-crypto.ts` (using `SHIPWRIGHT_ENCRYPTION_KEY`) from vitals-os into `agent/src/`
- Ports `AgentEnvService`, `AgentCronJobService`, `AgentToolService`, and `AgentTokenService` adapted for the standalone `Agent` model (no User/Engagement dependency)
- Adds `errors.ts`, `clock.ts`, and `system-crons.ts` shared utilities to `agent/src/`
- Adds `AgentCronJobService` mutual exclusion validation: 422 if both `channel` and `user` are set

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | encrypt → decrypt round-trip for empty string, unicode, 4096-byte value | ✅ MET |
| 2 | AgentEnvService.upsert() writes encrypted, getBundle() returns decrypted | ✅ MET |
| 3 | AgentCronJobService channel/user mutual exclusion (422 if both set) | ✅ MET |
| 4 | Integration tests against DATABASE_URL_AGENT_TEST, real Prisma, no mocks | ✅ MET |
| 5 | token-crypto.unit.test.ts + 4x *.integration.test.ts; no existing tests retired | ✅ MET |

## Test Plan

- [x] 68 new tests: 12 unit (crypto) + 56 integration (4 services × ~14 each)
- [x] All 625 tests pass (no regressions)
- [x] `bunx biome lint .` — no issues
- [x] `bun run --filter='*' typecheck` — all packages pass

### DIFF
--- a/agent/src/agent-cron-jobs.ts
+++ b/agent/src/agent-cron-jobs.ts
@@ -1,0 +1,341 @@
+/**
+ * agent/src/agent-cron-jobs.ts
+ * AgentCronJobService — CRUD for scheduled prompts per agent.
+ *
+ * Cron expression validation uses an inline 5-field regex check.
+ * Delivery target must be either a Slack channel ID or a Slack user ID (DM)
+ * — not both, and not neither (unless silent=true).
+ */
+
+import { NotFoundError, UnprocessableEntityError } from "./errors.ts";
+import { SYSTEM_CRONS } from "./system-crons.ts";
+import type { AgentCronJob, PrismaClient } from "../prisma/client/index.js";
+
+export type { AgentCronJob };
+
+export interface CreateAgentCronJobInput {
+  schedule: string;
+  prompt: string;
+  channel?: string | null;
+  user?: string | null;
+  silent?: boolean;
+  enabled?: boolean;
+  preCheck?: string | null;
+  name?: string | null;
+  system?: boolean;
+}
+
+/**
+ * Validates a cron expression.
+ * Accepts standard 5-field cron: minute hour day-of-month month day-of-week.
+ * Each field may contain digits, *, /, -, and comma.
+ */
+function isValidCron(schedule: string): boolean {
+  const fields = schedule.trim().split(/\s+/);
+  if (fields.length !== 5) return false;
+  const fieldPattern = /^(\*|[0-9]+([-,][0-9]+)*)(\/[0-9]+)?$|^\*\/[0-9]+$/;
+  return fields.every((f) => fieldPattern.test(f));
+}
+
+export class AgentCronJobService {
+  constructor(private prisma: PrismaClient) {}
+
+  /**
+   * List all cron jobs for a given agent.
+   */
+  async list(agentId: string): Promise<AgentCronJob[]> {
+    return this.prisma.agentCronJob.findMany({
+      where: { agentId },
+      orderBy: { createdAt: "asc" },
+    });
+  }
+
+  async get(agentId: string, cronId: string): Promise<AgentCronJob> {
+    const job = await this.prisma.agentCronJob.findUnique({
+      where: { id: cronId },
+    });
+    if (!job || job.agentId !== agentId) {
+      throw new NotFoundError(`cron job ${cronId} not found`);
+    }
+    return job;
+  }
+
+  /**
+   * List all enabled cron jobs across all agents.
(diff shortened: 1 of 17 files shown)


dodizzle (commented):
APPROVE — clean net-new agent service layer (crypto + 4 config services). All acceptance criteria met, CI green on head, tests at correct layers (unit crypto + integration with real Prisma) with proper isolation (injected Clock, no mock.module/global overrides). Tokens hashed (SHA-256), env values AES-256-GCM encrypted, decrypt-failure re-throws 500 rather than failing silently. No API/DB/breaking changes; no proprietary refs or secret literals. No blocking concerns.

dmcaulay (commented):
**APPROVE** — clean crypto + agent service layer port. AES-256-GCM via `crypto.ts`, SHA-256 token hashing, injectable Clock, integration tests against real SQLite (skipped when `DATABASE_URL_AGENT_TEST` unset). No mocks, no global overrides — consistent with CLAUDE.md test isolation. Postinstall Prisma generate prevents fresh-install failures. No proprietary identifiers in production code. All 5 acceptance criteria met, CI green.

(Self-review — event posted as COMMENT; recorded verdict is APPROVE.)

agent/src/agent-cron-jobs.ts
```diff
@@ -0,0 +1,329 @@
...
+  ): Promise<AgentCronJob> {
+    await this.get(agentId, cronId);
+
+    return this.prisma.agentCronJob.update({
+      where: { id: cronId },
+      data: {
+        ...(input.channel !== undefined && { channel: input.channel }),
+        ...(input.user !== undefined && { user: input.user }),
+        ...(input.silent !== undefined && { silent: input.silent }),
+      },
```
dmcaulay, line 234: `updateDelivery()` skips `validateDeliveryTarget()`. If a caller sends `{ channel: "C123", user: "U456" }`, both are persisted — bypassing the mutual exclusion invariant that `create()` and `update()` enforce. After `this.get()`, merge the fetched record's channel/user/silent with the incoming fields and run `validateDeliveryTarget()` on the merged values.

agent/src/token-crypto.ts
```diff
@@ -0,0 +1,52 @@
...
+  if (!key) {
+    console.warn(
+      "[shipwright agent] SHIPWRIGHT_ENCRYPTION_KEY not set — tokens stored in plain text",
+    );
+    return identityCrypto;
+  }
+  return {
+    encrypt: (token) => encrypt(token, key),
+    decrypt: (encrypted) => {
+      try {
```
dmcaulay, line 44: This `catch` swallows all decryption errors, including auth-tag failures from a rotated key. If the key is rotated without re-encrypting, callers silently receive the raw encrypted blob as the credential value — no log, no error. Scope the fallback to format errors only: check `encrypted.split(':').length !== 3` before calling `decrypt()` and return early; let auth-tag verification failures propagate.

dmcaulay (commented):
**COMMENT** — two important issues before merge.

**1. `updateDelivery()` skips delivery-target validation.** `create()` and `update()` call `validateDeliveryTarget()` to enforce channel/user mutual exclusion. `updateDelivery()` doesn't — a caller sending `{ channel: "C123", user: "U456" }` would persist both, breaking the invariant. Fix: after `this.get()`, merge the current record's channel/user/silent with incoming fields and run `validateDeliveryTarget()` on the merged values before the update.

**2. `token-crypto` decrypt fallback scope too broad.** The `catch` block returns the raw encrypted string for all errors — including auth-tag failures from a rotated key. A caller would silently receive the encrypted blob as a credential value with no log entry. Scope the fallback to format errors only (check `parts.length !== 3` before calling `decrypt()`); let auth-tag failures propagate.

Suggestion (non-blocking): `shipwright-dev-task` ships `enabled: true` — consider whether autonomous task execution should be opt-in (`enabled: false`) for new agents.

Strengths: solid AES-256-GCM with random IV, injectable Clock/TokenCrypto throughout, `reconcileSystemCrons()` preserves per-agent enabled state on update, no proprietary identifiers in prod code.

PR #75, opened 2026-06-07
feat(SW-1.6): de-vitals-os genericization sweep + banned-string CI gate

dodizzle:
## Summary

- Merges the SW-1.3 plugin port (101 files — commands, skills, scripts, agents, references) which was closed without merging
- Verifies zero hits for banned strings (`app-vitals/marketplace`, `vitals-os-prod`, `vitals-os-staging`, `vitals-os-dev`, `app-vitals/vitals-os`) across the entire plugin
- Adds `plugins/shipwright/scripts/check-banned-strings.ts` — a standalone grep-assertion script that serves as a regression gate, wired into `Taskfile.yml` (`task check-strings`) and `ci.yml` (runs between Lint and Typecheck)

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | `grep for app-vitals/marketplace` returns zero hits in plugin | ✅ MET — gate exits 0 |
| 2 | vitals-os references genericized/stripped across 22 files, named docs handled | ✅ MET — SW-1.3 deleted POSTMORTEM doc, scrubbed refs |
| 3 | CI grep-assertion script as regression gate | ✅ MET — check-banned-strings.ts + Taskfile + ci.yml |

## Test Plan

- [x] `bun plugins/shipwright/scripts/check-banned-strings.ts` → `✓ No banned strings found.`
- [x] 14 unit tests in `check-banned-strings.unit.test.ts` covering all 5 banned patterns, .git skip, binary safety, multi-file, nested dirs
- [x] All 1179 tests pass
- [x] Lint + typecheck pass

Generated with [Claude Code](https://claude.com/claude-code)

PR #77, opened 2026-06-07
feat(SW-1.5): add .shipwright.json auto-discovery + fix task-store docs

dodizzle:
## Summary

- Adds `.shipwright.json` auto-discovery to `create-task-store.ts`: walks up from process cwd, falls back to `SHIPWRIGHT_CONFIG` env var, then defaults to JSON backend
- Fixes the silent-JSON-fallback footgun described in CLAUDE.md — placing `.shipwright.json` at the repo root now works without any env configuration
- Updates `references/task-store.md` and `CLAUDE.md` to document the 3-step resolution chain accurately

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Task store resolves via .shipwright.json discovery → SHIPWRIGHT_CONFIG env → JSON fallback | ✅ MET |
| 2 | references/task-store.md documents new chain accurately | ✅ MET |
| 3 | Unit tests for discovery precedence chain (file found / env set / neither → JSON) | ✅ MET |
| 4 | Integration test: doctor reports github backend when .shipwright.json present | ✅ MET |

## Test Plan

- [x] 5 unit tests in `create-task-store.unit.test.ts` covering all discovery precedence scenarios
- [x] Integration test in `task_store.test.ts`: doctor subprocess discovers `.shipwright.json` from cwd
- [x] 877 tests pass, biome lint clean, tsc --noEmit clean


plugins/shipwright/scripts/task_store.ts
```diff
@@ -0,0 +1,368 @@
...
+    process.stderr.write(`error: ${String(e)}\n`);
+    process.exit(1);
+  }
+}
+
+/**
+ * repos subcommand — prints all known repos (one per line) by scanning
+ * workspace/repos/ or SHIPWRIGHT_REPOS_DIR.
+ */
+function cmdRepos(): void {
```
dmcaulay: `cmdRepos()` uses `resolveRepos(process.cwd())` from check-helpers (scans `workspace/repos/` git clones) instead of `adapter.resolveRepos()` from the configured store. For a GitHub-backend user without a `workspace/repos/` directory, this returns empty even though `GitHubTaskStore.resolveRepos()` knows the org/repo. Skills call `bun task_store.ts repos` to discover which repos to operate on — the disconnect breaks that contract for GitHub-backend deployments. Fix: call `await adapter.resolveRepos()` as the primary path; fall back to the filesystem scan only if the adapter returns empty.

dmcaulay (commented):
**COMMENT** — one important issue before merge.

`cmdRepos()` bypasses the configured store's `resolveRepos()`: it calls `resolveRepos(process.cwd())` from `check-helpers.ts` (scans `workspace/repos/` git clones) rather than `adapter.resolveRepos()`, which is correctly implemented in both adapters and returns repos from the configured backend. A GitHub-backend user running `bun task_store.ts repos` in CI or any environment without a `workspace/repos/` directory would get empty output — the GitHub backend knows its org/repo but it's never queried. One method call needs to be wired up.

Suggestion (non-blocking): `GitHubTaskStore.append()` declares `const updated = 0` that's never incremented per the intentional insert-only design. The return shape `{ inserted, updated }` is misleading — `updated` could be renamed `skipped` or removed.

Strengths: clean factory extraction with testable `cwd` param; real-temp-dir unit tests (no mocks, consistent with CLAUDE.md); correct `_labelStatus` guard in GitHub `update()` preventing label-not-found errors; solid walk-up root sentinel. CI green on head.

dodizzle (commented):
APPROVE — re-review at current head. `cmdRepos()` now queries `adapter.resolveRepos()` first with a `workspace/repos` filesystem fallback, fixing empty output for GitHub-backend users; async dispatch clean; regression test added; inline thread resolved; CI green.

dmcaulay (approved):
APPROVE — prior finding resolved.

`cmdRepos()` now queries `adapter.resolveRepos()` first via the new `resolveReposForStore()` helper, falling back to the filesystem scan only when the adapter returns an empty list. Regression test added in `task_store.test.ts` ("github backend resolves repo from config without a repos/ dir") confirms the fix. CI green on head `c2bc7d4`. No new issues introduced by the main merge.

dodizzle (commented):
APPROVE — re-review after resolving the conflict with merged #75. The resolution keeps both #75's genericization sweep and #77's adapter-first `resolveReposForStore` + .shipwright.json auto-discovery + regression tests; no work lost on either side; full suite 1273 pass, banned-strings gate clean.

dmcaulay (approved):
APPROVE — merge-only update since last review. Head moved from c2bc7d4 to a7417d6 via two main-merge commits (bringing in #75 genericization sweep and #82 system-crons endpoint, both already merged to main). No new feature code in the PR itself. Prior finding (cmdRepos bypassing adapter.resolveRepos) was addressed and verified in the previous review pass. CI green on head.

PR #78, opened 2026-06-07
feat(she-2.1): add Hono runtime API for agent config and crons

dodizzle:
## Summary
- Adds `createAgentRuntimeApp(deps)` Hono factory in `agent/src/api.ts` with `GET /agents/:id/config` and `GET /agents/:id/crons` routes
- Auth via `SHIPWRIGHT_INTERNAL_API_KEY` bearer token middleware; 401 on missing/wrong key
- Config response returns `{ env, allowedTools, plugins: {marketplace, plugin}[] }` with env decrypted via `AgentEnvService.getConfigBundle()`

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | GET /agents/:id/config returns `{ env, allowedTools, plugins }` with all values decrypted | ✅ MET |
| 2 | Returns 401 when SHIPWRIGHT_INTERNAL_API_KEY header is missing or wrong | ✅ MET |
| 3 | Returns 404 for unknown agent ID | ✅ MET |
| 4 | Responds in <100ms on local Postgres | ✅ Unverifiable (parallel fetch, no N+1) |
| 5 | Smoke tests via app.request() for 200, 401, and 404 cases | ✅ MET (8 tests, 100% coverage) |

## Test Plan
- [x] 8 smoke tests covering 200/401/404 for both endpoints — injected mocks, no DB needed
- [x] 100% line/function coverage on `api.ts`
- [x] Full test suite: 630 pass, 0 fail
- [x] Typecheck clean, lint clean

Generated with [Claude Code](https://claude.com/claude-code)

PR #65, opened 2026-06-07
feat(SR-1.2): add semantic-release config and dependencies

dodizzle:
## Summary

This PR migrates Shipwright from release-please to semantic-release, delivering a fully automated release pipeline driven by conventional commits.

### Changes by milestone

**SR-1.1 — sync-version script**
- Add `scripts/sync-version.ts`: writes a semver string into all 4 `package.json` files (`package.json`, `plugins/shipwright/package.json`, `metrics/package.json`, `agent/package.json`) and `version.txt`
- Add `scripts/sync-version.test.ts`: 9 unit tests covering happy path, format preservation (2-space indent, trailing newline), pre-release/build-metadata acceptance, and rejection of malformed inputs
- No I/O in test setup — tests use a temp directory created per-case

**SR-1.2 — semantic-release config and dependencies**
- Add `.releaserc.json` with `conventionalcommits` preset, full plugin chain: commit-analyzer → release-notes-generator → changelog → exec → git → github
- Changelog section types ported 1:1 from the former release-please config
- Exec plugin calls `bun scripts/sync-version.ts ${nextRelease.version}`; git plugin commits the writeback with `[skip ci]`
- Add 8 semantic-release devDependencies to root `package.json` (no `@semantic-release/npm` — all packages are private)

**SR-1.3 — release workflow (admin guard + GitHub App token)**
- Add `.github/workflows/release.yml` triggered by `workflow_dispatch` with a `dry_run` boolean input
- `guard` job: verifies the triggering actor has `admin` permission on the repo before allowing release to proceed
- `release` job: uses `actions/create-github-app-token@v1` to mint a GitHub App token with push rights that can bypass branch protection, then runs `bunx semantic-release` (or `--dry-run`)

**SR-1.4 — remove release-please references**
- Remove `.github/workflows/release-please.yml`
- Update `CONTRIBUTING.md` and any other docs that referenced release-please

**SR-1.5 — verification and documentation**
- All checks green: `bun test` (557 pass, 0 fail), `bunx biome lint .` (no issues), `bun run --filter='*' typecheck` (all 3 packages pass)
- `version.txt` and root `package.json` remain at `0.1.0` — no self-bump
- This PR body documents the one-time post-merge GitHub App setup

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | `scripts/sync-version.ts` writes version into all 4 package.json files and version.txt | ✅ MET |
[...]

PR #80, opened 2026-06-07
feat(she-3.1): admin CRUD API with session-cookie auth

dmcaulay:
## Summary

- Adds `createAdminApp(deps)` factory in `agent/src/admin-api.ts` — Hono app with all 16 CRUD endpoints for AgentEnv, AgentCronJob, AgentTool, AgentToken, and AgentPlugin at `/admin/api/*`
- Session-cookie auth middleware: httpOnly JWT (`admin_session` cookie, HS256) verifies `userId` + `email` fields; unauthenticated requests return 401 JSON
- Adds `AgentPluginService` (`agent/src/agent-plugins.ts`) following the same service pattern as existing services (list, add with upsert semantics, remove with ownership check)
- Adds `hono` to `agent/package.json` dependencies
- POST `/tokens` returns `rawToken` exactly once; GET `/tokens` strips the hash field — only metadata returned
- POST `/envs` encrypts via injected `TokenCrypto`; GET `/envs` decrypts

## Test plan

- [x] 22 smoke tests pass (auth 401s, create 201s, delete 204s for all 5 resources)
- [x] 3 integration tests skip without `DATABASE_URL_AGENT_TEST`; they verify encryption at rest and token hash-only storage when DB is available
- [x] `bun run --filter='*' typecheck` — all 3 workspaces pass
- [x] `bun test` — 644 pass, 58 skip (integration), 0 fail


agent/src/admin-api.ts
```diff
@@ -0,0 +1,311 @@
...
+
+  // GET /admin/api/agents/:id/plugins — list plugins
+  app.get("/admin/api/agents/:id/plugins", async (c) => {
+    const agentId = c.req.param("id");
+    const plugins = await agentPluginService.list(agentId);
+    return c.json({ plugins });
+  });
+
+  // PATCH /admin/api/agents/:id/plugins/:name — update plugin version (re-upsert)
+  app.patch("/admin/api/agents/:id/plugins/:name", async (c) => {
```
dmcaulay: PATCH `/plugins/:name` uses the URL param as the plugin name (delegates to `add()` upsert by `agentId_name`), but DELETE `/plugins/:pluginId` on line 303 uses the URL param as the plugin ID (delegates to `remove()` which looks up by `id`). A client building against this API needs to use the name for PATCH and the ID for DELETE — different param types for the same sub-resource. Consider making both name-keyed (DELETE also accepts a name and calls `remove(agentId, name)` with a name-based lookup), or both ID-keyed (PATCH accepts an ID and calls `update()` instead of `add()`). The upsert-by-name semantics of `add()` make the name-keyed approach the simpler fix.

dmcaulay (commented):
**COMMENT** — two important issues before merge.

**1. Plugin PATCH vs DELETE parameter inconsistency**: `PATCH /plugins/:name` uses the URL parameter as the plugin *name* (upsert by `agentId_name`), while `DELETE /plugins/:pluginId` uses it as the plugin *ID*. Callers need to track two different identifiers depending on the verb. Fix: either make DELETE also accept a name, or rename the PATCH param to `:name` and add a comment noting the name-keyed identity.

**2. Prisma version mismatch**: root `package.json` has `prisma: "^6.19.2"` and `@prisma/client: "^6.19.2"`, while `agent/package.json` has both at `"^6.0.0"`. Bun hoists the root version so CI is green, but the declared constraint is stale. Bump `agent/package.json` to match root.

Strengths: AES-256-GCM correct (96-bit IV, auth-tag verified, tamper test included), SHA-256-only token storage with raw returned once and stripped from list responses, session auth via `hono/jwt` with field presence checks, `reconcileSystemCrons()` correctly preserves enabled state, all integration tests use real SQLite with FK-ordered teardown, no `mock.module()` or global overrides anywhere.

(Self-review — event posted as COMMENT; recorded verdict is COMMENT.)

dmcaulay (commented):
**APPROVE** — both prior findings resolved.

1. **Plugin PATCH/DELETE parameter inconsistency** ✅ — `DELETE /plugins/:name` now matches the upsert-by-name pattern of PATCH. `AgentPluginService.removeByName()` added; `AdminDeps` interface updated; smoke test updated to use URL-encoded name.

2. **Prisma version mismatch** ✅ — `agent/package.json` bumped to `^6.19.2` for both `prisma` and `@prisma/client`, matching root. `bun.lock` updated accordingly.

CI green on head `5086de2e`. No new issues. The security fundamentals (AES-256-GCM, SHA-256-only token storage, session-cookie JWT auth) remain solid.

(Self-review — event posted as COMMENT; recorded verdict is APPROVE.)

dmcaulay (commented):
APPROVE — merge-only update since last review. Head moved from 5086de2 to fd0d20d via main-merge commits (bringing in #65 semantic-release, #75 genericization sweep, #81 system-crons endpoint). Feature files unchanged. Prior findings (plugin PATCH/DELETE param inconsistency, Prisma version mismatch) were resolved and verified at 5086de2. CI green on head.

(Self-review — event posted as COMMENT; recorded verdict is APPROVE.)

dmcaulay (commented):
## Review — feat(she-3.1): admin CRUD API with session-cookie auth

CI green. Code is well-structured and the security-sensitive pieces (token hashing, env encryption, auth middleware) are handled correctly. Three issues worth addressing before merge.

---

### Issues

**1. `AgentPluginService.update()` is dead code**

`agent-plugins.ts` defines an `update(agentId, pluginId, fields)` method, but it's not included in `AdminDeps` and nothing in the admin API calls it. The PATCH plugin route uses `add()` (upsert) instead. Either remove `update()` or wire it up — as-is it's untested dead code.

**2. PATCH `/plugins/:name` silently re-enables disabled plugins**

The PATCH route calls `agentPluginService.add(agentId, name, body.version)`, which uses upsert with `update: { version: ..., enabled: true }`. If an admin disabled a plugin and then calls PATCH to update its version, the plugin gets re-enabled as a side effect. Either expose a separate `enabled` field in the PATCH body and pass it through, or switch to a partial-update path that doesn't touch `enabled` unless explicitly sent.

**3. PATCH `/crons/:cronId` requires both `schedule` and `prompt`**

The body type is `{ schedule: string; prompt: string; ... }` (both required). The underlying `AgentCronJobService.update()` presumably supports partial updates, but the route will pass `undefined` for both if the caller only sends `enabled: true`. Recommend marking `schedule?` and `prompt?` as optional in the body type and adding a runtime guard (or handle partial update explicitly in the service call).

---

### Minor

- No validation on env key format (empty string, spaces, etc.) — could be worth a runtime guard at the route level, even just a check that all keys are non-empty strings
- No smoke test for error paths (e.g., service throws `NotFoundError` → confirms 404 in response). The `onError` handler covers it, but there's no test asserting the shape.
- `remove` is declared in `AdminDeps` but only `removeByName` is called from the route handlers — minor interface surface noise, not a blocker

---

### What's good

- Auth middleware is solid: checks cookie presence, verifies HS256 JWT, validates `userId` + `email` field types before passing through
- Token flow is correct: rawToken returned once on create, hash stripped from all list responses
- Integration tests cover the two hardest-to-get-wrong security properties (encryption at rest, hash-not-raw)
- DI pattern via `AdminDeps` with `Pick<>` is clean and makes the smoke tests simple

The three issues above are all addressable without architectural changes. Happy to APPROVE once #1 and #2 are fixed (or explicitly deferred with a comment in the code).

dodizzle (commented):
## Review — feat(she-3.1): admin CRUD API with session-cookie auth

CI green. Security fundamentals (AES-256-GCM env encryption, SHA-256-only token storage, JWT session auth with field presence checks) are solid. Good DI pattern via `AdminDeps`.

Seconding the three issues flagged in the prior self-review — they're still present on HEAD and should be addressed before merge:

**1. `AgentPluginService.update()` is dead code (high)**

`update(agentId, pluginId, fields)` in `agent-plugins.ts` has no caller — it's not in `AdminDeps` and no route uses it. The PATCH plugin route uses `add()` (upsert). Remove `update()` or expose it via a route.

**2. PATCH `/plugins/:name` silently re-enables disabled plugins (medium)**

`add()` has `update: { version: ..., enabled: true }`. A PATCH to update the version will always flip `enabled` back to `true` as a side effect, even if the plugin was intentionally disabled. Either: pass `enabled` through in the PATCH body, or change the upsert to `update: { version: ... }` and handle re-enable separately.

**3. PATCH `/crons/:cronId` body schema missing `enabled` (medium)**

Route body type is `{ schedule: string; prompt: string; ... }` — `enabled` is not writable via PATCH. Since list responses show `enabled: boolean`, callers have no way to toggle a cron without deleting and recreating it. Mark `schedule` and `prompt` as optional and add `enabled?: boolean`.

**Minor (non-blocking):**
- `agentPluginService` Pick in `AdminDeps` includes `remove` but only `removeByName` is called from routes — remove `remove` from the Pick
- Malformed JSON body (e.g., invalid syntax) propagates as unhandled 500 rather than 400 — consider wrapping `c.req.json()` in a try/catch at the route level or in middleware

Ready to APPROVE once #1 and #2 are fixed.

dodizzle (commented):
**Verdict: COMMENT** — CI is green and the security model is solid (global JWT middleware, one-time raw token with hash stripped on all reads, real-DB integration test). Three items to address before merge:

**1. `POST /admin/api/agents/:id/envs` returns 200, not 201 (important, conf 83)** — `agent/src/admin-api.ts:123`. The SHE-3.1 acceptance criteria specify create → 201 for every resource; envs is the only one returning the implicit 200 (its smoke test even asserts `.toBe(200)`). Return `c.json({ ok: true }, 201)` and update the assertion.

**2. `AgentPluginService.update()` is dead code (important, conf 78)** — `agent/src/agent-plugins.ts:48`. Newly added but absent from the `AdminDeps.agentPluginService` Pick and called by no route/test; the PATCH route re-uses `add()` (upsert). Either wire it to a route + AdminDeps + test, or drop it from this PR and track separately.

**3. Scoped plugin names 404 on PATCH/DELETE without percent-encoding (important, conf 76)** — `agent/src/admin-api.ts:304`. `:name` as a path segment can't match `@shipwright/plugin` (literal `/`). Tests pass `%2F`, but nothing enforces it — a client building the URL by interpolation breaks silently. Prefer `?name=` query param, or guard against unencoded slashes.

_Strengths:_ centralized typed-error `onError` (no per-handler try/catch), and the merge reconciliation with SHE-6.2 correctly preserved `reconcileSystemCrons` + the full (non-stub) `agent-plugins.ts` with ownership checks.

dodizzle (commented):
**COMMENT** — three prior findings remain unresolved on HEAD `e2e19458`. Head moved from `fd0d20d` via main-merge only (PRs #77, #78, #82); feature code is unchanged.

**1. `AgentPluginService.update()` is dead code (important, conf 80)**
`agent/src/agent-plugins.ts:48` — `update(agentId, pluginId, fields)` exists in the class but is absent from `AdminDeps.agentPluginService` and no route calls it. The PATCH plugin route re-uses `add()` (upsert) instead. Remove it from this PR or wire it to a route — dead code shipped into a public repo is noise.

**2. `POST /envs` returns 200, not 201 (important, conf 83)**
`agent/src/admin-api.ts:123` — `c.json({ ok: true })` defaults to 200. Every other create endpoint returns 201. The smoke test asserts `.toBe(200)`, baking in the wrong behavior. Fix: `c.json({ ok: true }, 201)` and update the assertion.

**3. `PATCH /crons/:cronId` body type excludes `enabled` (important, conf 80)**
`agent/src/admin-api.ts:179-186` — `schedule` and `prompt` are required non-optional; `enabled` is absent entirely. Callers cannot toggle a cron's enabled state via PATCH. Add `enabled?: boolean` and mark `schedule?`/`prompt?` optional.

_Minor (non-blocking):_ `AdminDeps.agentPluginService` Pick includes `remove` but only `removeByName` is called — tighten the interface.

Security fundamentals remain solid (AES-256-GCM, SHA-256-only token storage, centralized auth middleware, DI via AdminDeps). Ready to APPROVE once items 1–3 are addressed.

dodizzle (commented):
**COMMENT** — progress made: three of the prior findings resolved (POST /envs → 201, plugin PATCH/DELETE moved to query params, `AgentPluginService.update()` dead code removed). Two items remain before merge.

**1. Integration test `POST /envs` still asserts `.toBe(200)` — will fail with a real DB (important, conf 90)**
`agent/src/admin-api.integration.test.ts:113` — the smoke test was correctly updated to assert 201, but the integration test was missed. Fix: change `.toBe(200)` → `.toBe(201)` on that line.

**2. `PATCH /crons/:cronId` body type still excludes `enabled`; `schedule`/`prompt` still required (important, conf 82)**
`agent/src/admin-api.ts:179-186` — `schedule: string` and `prompt: string` are required, and `enabled` is absent. Callers cannot toggle a cron on/off via PATCH without resending schedule and prompt. Add `enabled?: boolean` and mark `schedule?`/`prompt?` as optional.

_Minor (non-blocking):_ `AdminDeps.agentPluginService` Pick includes `"remove"` but only `removeByName` is called from routes — tighten the interface.

Security fundamentals remain solid. Ready to APPROVE once items 1 and 2 are fixed.

dmcaulay (commented):
## Review Update — 2026-06-08

**Commit reviewed**: 82907891

### New Changes

This commit addresses the remaining API design gaps flagged in the previous round:

- **POST /envs → 201**: Correct REST semantics for resource creation. Test updated to match.
- **PATCH + DELETE /plugins → query param**: `?name=<name>` correctly handles scoped names like `@shipwright/plugin` that contain `/` and would break path segment matching. Comment added explaining the rationale. Missing-name 400 guard added to both routes with tests.
- **Dead `update()` removed**: `AgentPluginService.update()` was absent from `AdminDeps` and uncalled by any route — clean removal, no callers broken.

### Prior Findings Resolution

All three items are cleanly resolved. Tests cover the new 400 paths. CI green on head.

### Updated Recommendation

**APPROVE** — all findings from prior rounds addressed; no new issues introduced.

(Self-review — posted as COMMENT; recorded verdict is APPROVE.)

PR #84, opened 2026-06-08
fix: use latest plugin version when resolving scripts from cache

dmcaulay:
## Summary

- Replace `find ... | head -1 | xargs dirname` with `find ... | sort -V | tail -1 | xargs dirname` across all PLUGIN_SCRIPTS/CHECK_SCRIPTS resolutions in every command file
- `head -1` returns whichever version the filesystem puts first (hash-tree order on ext4), which on the agent resolves to 4.26.0 — not the newest installed version
- 4.26.0 has a bug where `listPrCommits` ignores the `repo` parameter and always uses `allRepos[0]`, so merge-only detection silently fails for any repo that isn't first — causing unnecessary re-reviews of PRs with only merge commits since last review
- Posthog lookups and `task_store.ts repos | head -1` (getting first configured repo) are intentionally unchanged
- Bump to 4.26.4 (skipping 4.26.2/4.26.3 published from vitals-os); also syncs README which was still at v4.26.0

## Files changed

- 8 command files: `review-patch.md`, `review.md`, `dev-task.md`, `plan-session.md`, `patch.md`, `deploy.md`, `refresh-plan.md`, `research-docs.md`
- `plugin.json`: 4.26.1 → 4.26.4
- `README.md`: v4.26.0 → v4.26.4

## Test plan

- [ ] Confirm `find ... | sort -V | tail -1` resolves to 4.26.4 once published
- [ ] Confirm `posthog_send.py` and `task_store.ts repos | head -1` lines unchanged
- [ ] Confirm review-patch loop uses correct `check-review.ts` version (merge-only detection works for second repo)

### DIFF
--- a/plugins/shipwright/.claude-plugin/plugin.json
+++ b/plugins/shipwright/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "shipwright",
-  "version": "4.26.1",
+  "version": "4.26.4",
   "description": "Shipwright — conversational planning, queue-driven execution, policy-controlled code review, a five-phase test-readiness pipeline, and autonomous deploy (merge → canary → promote) for any software project",
   "author": {
     "name": "App Vitals",
--- a/plugins/shipwright/README.md
+++ b/plugins/shipwright/README.md
@@ -1,4 +1,4 @@
-# Shipwright v4.26.0
+# Shipwright v4.26.4
 
 A structured dev pipeline plugin for Claude Code. Plan sessions, execute tasks, run autonomous dev loops, perform multi-agent code reviews, and conduct integrated project research — for any software project.
 
--- a/plugins/shipwright/commands/deploy.md
+++ b/plugins/shipwright/commands/deploy.md
@@ -38,7 +38,7 @@ AGENT_LOGIN=$(gh api user --jq '.login')
 
 Resolve the configured repos:
 ```bash
-PLUGIN_SCRIPTS=$(find ~/.claude/plugins/cache -maxdepth 5 -name "task_store.ts" -path "*/shipwright/*" 2>/dev/null | head -1 | xargs dirname 2>/dev/null)
+PLUGIN_SCRIPTS=$(find ~/.claude/plugins/cache -maxdepth 5 -name "task_store.ts" -path "*/shipwright/*" 2>/dev/null | sort -V | tail -1 | xargs dirname 2>/dev/null)
 REPOS=$(bun "$PLUGIN_SCRIPTS/task_store.ts" repos)
 ```
 
--- a/plugins/shipwright/commands/dev-task.md
+++ b/plugins/shipwright/commands/dev-task.md
@@ -19,7 +19,7 @@ Auto-detect the project toolchain (run once, reuse throughout). Skip this step u
 **First, check for an interrupted task** — if a prior session left a task `in_progress`, resume it:
 
 ```bash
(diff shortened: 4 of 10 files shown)


dodizzle (approved):
**Verdict: APPROVE** — Clean, surgical fix that correctly resolves the highest-semver plugin version in the cache instead of filesystem-order first.

The `head -1` approach was returning 4.26.0 (hash-tree order) on the production agent, and 4.26.0 has a bug where `listPrCommits` ignores the `repo` parameter — causing merge-only detection to silently fail for non-first configured repos and trigger unnecessary re-reviews. `sort -V | tail -1` deterministically picks the highest semver version, which is the correct behavior.

All intentionally-unchanged lines (`posthog_send.py` lookups and the `repos | head -1` output pipe in review.md) are correctly identified and left alone. CI passes clean.

dmcaulay (commented):
APPROVE — fix is correct and complete; all 14 PLUGIN_SCRIPTS/CHECK_SCRIPTS resolution points updated to `sort -V | tail -1`; posthog `head -1` correctly left unchanged (version-stable across all cached versions); two forward-looking findings and one portability note worth addressing in follow-on but none are blockers.

---

**Finding 1 (important, confidence 82): `posthog_send.py` `head -1` — safe today but undocumented precondition**

`posthog_send.py` lives in the same versioned cache directories (`cache/<marketplace>/shipwright/<version>/scripts/`) as `task_store.ts`, so it's theoretically subject to the same wrong-version bug this PR fixes. It's safe here because `posthog_send.py` is byte-for-byte identical across all cached versions (verified 4.17.0 through 4.26.3 locally). But the PR body calls this "intentionally unchanged" without documenting why — if `posthog_send.py` ever diverges between versions, all POSTHOG_SCRIPT `head -1` calls (11 occurrences in dev-task.md alone) will silently pick the wrong version.

Suggestion: either add an inline comment on each POSTHOG_SCRIPT line (`# posthog_send.py is version-stable — head -1 is safe`) or proactively apply `sort -V | tail -1` for consistency and eliminate the hidden precondition.

---

**Finding 2 (important, confidence 78): No regression test for the `sort -V | tail -1` resolution pattern**

Existing tests (review-patch.test.ts, dev-task.test.ts) check that `bun "$CHECK_SCRIPTS/check-patch.ts"` is referenced, but not how `CHECK_SCRIPTS` is resolved. A future refactor could reintroduce `| head -1 | xargs dirname` undetected — which is likely how the original bug crept in.

Suggestion: add string-match assertions in at least review-patch.test.ts and dev-task.test.ts:
```ts
it('resolves PLUGIN_SCRIPTS using sort -V | tail -1, not head -1', () => {
  expect(content).toContain('sort -V | tail -1 | xargs dirname');
  expect(content).not.toMatch(/\| head -1 \| xargs dirname/);
});
```

---

**Finding 3 (suggestion, confidence 75): `sort -V` is GNU-only — not available on macOS BSD sort**

`sort -V` (version sort) is a GNU coreutils extension. macOS's BSD sort does not support it — passing `-V` causes sort to exit 1, which means `PLUGIN_SCRIPTS` would be empty and all task store operations would silently fail. CI runs `ubuntu-latest` only, so this won't be caught. Low-urgency given the Pi is the only current deployment target, but worth documenting before Shipwright goes public as an OSS plugin that users may install on macOS.

Suggestion: document the Linux-only requirement in CLAUDE.md, or use a POSIX-compatible version comparison that works on both platforms.

PR #97, opened 2026-06-08
feat(she-6.1): data migration script

dodizzle:
## Summary
- Adds `agent/scripts/migrate.ts` CLI to migrate agent data from vitals-os accounts API to shipwright admin API
- Migrates env vars, tool patterns, and cron jobs per agent; idempotent (crons deduped by schedule+prompt, envs replace-all, tools upsert)
- Exits non-zero and reports `agent=<id> field=<field>` to stderr on any write failure; remaining agents still processed

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Script completes for all agents; SLACK_BOT_TOKEN and ANTHROPIC_API_KEY present after running | ✅ MET |
| 2 | All cron jobs, tools, and tokens present with correct fields | ✅ MET |
| 3 | Re-running is a no-op — no duplicate rows | ✅ MET |
| 4 | Exits non-zero and prints failing agent ID + field name; remaining agents processed | ✅ MET |
| 5 | Integration tests with RecordedAccountsClient and RecordedShipwrightAdminClient | ✅ MET |

## Test Plan
- [x] `bun test agent/src/migrate.integration.test.ts` — 3 tests pass (full migration, idempotency, failure isolation)
- [x] `bun test` — full suite 1326 pass / 0 fail
- [x] `cd agent && bun run typecheck` — clean

### DIFF
--- a/agent/src/accounts-migration-client.ts
+++ b/agent/src/accounts-migration-client.ts
@@ -1,0 +1,93 @@
+/**
+ * agent/src/accounts-migration-client.ts
+ * AccountsMigrationClient — reads agent data from the vitals-os accounts API.
+ */
+
+// ─── Types ────────────────────────────────────────────────────────────────────
+
+export interface VitalsAgentRecord {
+  id: string;
+  name: string;
+}
+
+export interface VitalsAgentConfig {
+  env: Record<string, string>;
+  tools: string[];
+}
+
+export interface VitalsAgentCron {
+  schedule: string;
+  prompt: string;
+  channel: string | null;
+  user: string | null;
+  silent: boolean;
+  enabled: boolean;
+  preCheck: string | null;
+  name: string | null;
+}
+
+export interface AccountsMigrationClient {
+  listAgents(): Promise<VitalsAgentRecord[]>;
+  getAgentConfig(agentId: string): Promise<VitalsAgentConfig>;
+  getAgentCrons(agentId: string): Promise<VitalsAgentCron[]>;
+}
+
+// ─── HTTP implementation ──────────────────────────────────────────────────────
+
+export class HttpAccountsMigrationClient implements AccountsMigrationClient {
+  constructor(
+    private readonly baseUrl: string,
+    private readonly apiKey: string,
+  ) {}
+
+  private get headers(): Record<string, string> {
+    return {
+      Authorization: `Bearer ${this.apiKey}`,
+      "Content-Type": "application/json",
+    };
+  }
+
+  async listAgents(): Promise<VitalsAgentRecord[]> {
+    const url = `${this.baseUrl}/accounts/agents?role=AGENT`;
+    const res = await globalThis.fetch(url, { headers: this.headers });
+    if (!res.ok) {
+      throw new Error(
+        `listAgents failed: ${res.status} ${await res.text()}`,
+      );
+    }
+    const data = await res.json() as { agents?: VitalsAgentRecord[] } | VitalsAgentRecord[];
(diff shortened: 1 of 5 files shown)


agent/src/accounts-migration-client.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (Array.isArray(data)) return data;
+    if (data && typeof data === "object" && "agents" in data && Array.isArray(data.agents)) {
+      return data.agents;
+    }
+    return data as VitalsAgentRecord[];
+  }
+
+  async getAgentConfig(agentId: string): Promise<VitalsAgentConfig> {
+    const url = `${this.baseUrl}/accounts/agents/${agentId}/config`;
+    const res = await globalThis.fetch(url, { headers: this.headers });
```
dmcaulay, line 69: This fallback cast silently succeeds if the API returns an unexpected shape (not an array and not `{ agents: [] }`), then crashes with a misleading `TypeError: agents is not iterable` in `runMigration`. Prefer:
```ts
throw new Error(`listAgents: unexpected response shape: ${JSON.stringify(data)}`);
```
Same fix applies to the equivalent branch in `getAgentCrons` (line 103).

dmcaulay (approved):
APPROVE — clean migration script with solid DI, idempotency, and failure isolation. CI green.

All acceptance criteria met. One non-blocking suggestion: the `return data as VitalsAgentRecord[]` / `return data as VitalsAgentCron[]` fallback branches in `accounts-migration-client.ts` silently return a non-iterable if the API returns an unexpected shape. Prefer an explicit `throw new Error(...)` with the actual response payload — it surfaces the root cause instead of a cryptic downstream `TypeError: agents is not iterable`. See inline comment.

agent/src/accounts-migration-client.ts
```diff
@@ -0,0 +1,93 @@
...
+        `listAgents failed: ${res.status} ${await res.text()}`,
+      );
+    }
+    const data = await res.json() as { agents?: VitalsAgentRecord[] } | VitalsAgentRecord[];
+    // Handle both array response and wrapped { agents: [] } response
+    if (Array.isArray(data)) return data;
+    if (data && typeof data === "object" && "agents" in data && Array.isArray(data.agents)) {
+      return data.agents;
+    }
+    return data as VitalsAgentRecord[];
```
dodizzle, line 64: Silent fallback — if the API returns an unexpected shape (not an array, not `{ agents: [] }`), this cast succeeds at compile time but `data` will be non-iterable at runtime. The downstream `for (const agent of agents)` in `migrate.ts` will throw `TypeError: agents is not iterable` with no indication this is the HTTP layer. Prefer:
```ts
throw new Error(`listAgents: unexpected response shape: ${JSON.stringify(data)}`);
```
Same fix applies to the `getAgentCrons` fallback below.

dodizzle (commented):
APPROVE — DI pattern is correct, CI green, idempotency and failure isolation working.

One item to address as a follow-on: the fallback `return data as VitalsAgentRecord[]` / `return data as VitalsAgentCron[]` branches in `accounts-migration-client.ts` silently return a non-iterable if the API returns an unexpected shape, which would surface as a cryptic `TypeError: agents is not iterable` at the call site rather than a clear HTTP error. Consider replacing with `throw new Error(...)` including the raw response (same pattern as the non-2xx error branches). See inline comment.

This is consistent with the suggestion in dmcaulay's review — flagging it inline so it's easy to find during follow-on work.

PR #102, opened 2026-06-08
feat(she-5.1): port setup.ts with dynamic plugin registry integration

dodizzle:
## Summary
- Ports `setup.ts` from vitals-os to the shipwright agent with dynamic plugin registry support
- Replaces hardcoded PLUGINS array with `DEFAULT_PLUGINS` + agent-specific plugins from the config bundle (`AgentPlugin[]`)
- Ports workspace templates (`CLAUDE.md`, `SOUL.md`, `IDENTITY.md`, `BOOTSTRAP.md`, `VOICE.md`, `mise.toml`, `requirements.txt`, `state/agent-policy.md`, `state/todos.json`) to `agent/workspace/`

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `installPlugins()` installs shipwright marketplace plugins by default | ✅ MET |
| 2 | Agent-specific plugins from `config.plugins` installed after defaults | ✅ MET |
| 3 | Workspace templates seeded on new workspace | ✅ MET |
| 4 | Unit tests pass with mocked `execFn` (no real claude binary) | ✅ MET |
| 5 | Unit tests for `installPlugins()` and `seedFile()` with injected `execFn` | ✅ MET |

## Test Plan
- 17 unit tests across `seedFile`, `loadState`/`saveState`, `isNewWorkspace`, and `installPlugins`
- Injected `execFn` captures all plugin install/update calls — no real `claude` binary needed
- [x] `tsc --noEmit` — clean
- [x] `bun test` — 220 pass, 0 fail


agent/src/setup.ts
```diff
@@ -0,0 +1,403 @@
...
+ * Loads workspace state from $AGENT_HOME/workspace-state.json.
+ * Returns { version: 1 } when the file is absent.
+ */
+export function loadState(home: string): WorkspaceState {
+  const statePath = join(home, "workspace-state.json");
+  if (!existsSync(statePath)) {
+    return { version: 1 };
+  }
+  const raw = readFileSync(statePath, "utf8");
+  return JSON.parse(raw) as WorkspaceState;
```
dodizzle: `JSON.parse(raw)` has no error handling here. If `workspace-state.json` is corrupted or truncated (e.g., partial write from a prior crash), this throws and `ensureAgentHome` aborts — preventing the agent from starting. Since `loadState` is documented as a startup utility that returns defaults when the file is absent, it should also handle the malformed case: wrap in try/catch and return `{ version: 1 }` on failure.

dodizzle (commented):
COMMENT — one startup-safety issue to fix.

**`loadState` uncaught JSON.parse** (`setup.ts:65`): if `workspace-state.json` is valid JSON at first write but gets corrupted (partial write from a crash), `JSON.parse` throws and propagates through `ensureAgentHome`, crashing the startup routine. The function is documented as safe to call on every startup — wrap the parse in try/catch and fall back to `{ version: 1 }` on failure.

Everything else is solid: `wx`-flag seeding, non-fatal plugin install, isNewWorkspace marker checks, and test coverage are all well done. **Verdict: APPROVE** pending the one fix (posted as COMMENT — self-review).

dodizzle (commented):
APPROVE — prior finding resolved.

**`loadState` JSON.parse fix** (commit a3c8080): wraps `JSON.parse` in try/catch and falls back to `{ version: 1 }` on corrupted state, matching the documented behavior for absent files. Startup is now robust against partial writes.

All prior strengths hold. The stderr suggestion remains open (suggestion-level, not blocking). This is clean to merge.

dodizzle (commented):
Re-review after force-push: APPROVE — all prior findings addressed; branch restructuring is clean.

The branch was force-pushed with 4 new commits. Key changes: DEFAULT_PLUGINS reduced to 1 (minimal default set is correct), `installPlugins` parameter reorder with no external callers, VOICE.md changed to wx semantics (correct for a new file), marketplace-add step removed (simpler and appropriate for OSS distribution), and the JSON.parse try/catch fix landed at `setup.ts:65-77`.

CI green. No blocking issues. The stderr suggestion from the first review remains at suggestion level and is non-blocking.

**Verdict: APPROVE** (posted as COMMENT — self-review).

agent/workspace/CLAUDE.md.template
```diff
@@ -0,0 +1,115 @@
+# Agent Identity
+
+@SOUL.md
+@IDENTITY.md
+@BOOTSTRAP.md
```
dmcaulay: `@BOOTSTRAP.md` is included here, but `BOOTSTRAP.md` instructs the agent to delete itself on first-run completion. After deletion, this `@` include will emit a file-not-found error on every subsequent session startup. Consider removing `@BOOTSTRAP.md` from the template before seeding, or restructuring the bootstrap flow so CLAUDE.md doesn't permanently reference a file that's meant to be ephemeral.

agent/src/setup.ts
```diff
@@ -0,0 +1,369 @@
...
+// ---------------------------------------------------------------------------
+// Template loading
+// ---------------------------------------------------------------------------
+
+// Templates live at agent/workspace/ alongside this source file.
+const WORKSPACE_TEMPLATE_DIR = join(import.meta.dir, "..", "workspace");
+
+// Default plugin installed on all Shipwright agents.
+const DEFAULT_PLUGINS: readonly AgentPlugin[] = [
+  { marketplace: "shipwright", plugin: "shipwright" },
```
dmcaulay: `DEFAULT_PLUGINS` uses `marketplace: "shipwright"`, which produces spec `shipwright@shipwright`. The documented install format (CLAUDE.md) is `shipwright@app-vitals/shipwright` — marketplace is the GitHub org/repo slug, not the plugin name. If `app-vitals/shipwright` is correct, this spec will silently fail on every new agent bootstrap. Confirm against `claude plugin marketplace list` and update accordingly.

agent/src/setup.ts
```diff
@@ -0,0 +1,369 @@
...
+  execFn: ExecFn = defaultExec,
+  cwd: string = process.cwd(),
+  agentPlugins: AgentPlugin[] = [],
+): Promise<void> {
+  try {
+    const allPlugins = [...DEFAULT_PLUGINS, ...agentPlugins];
+
+    // Install all plugins (defaults then agent-specific)
+    for (const plugin of allPlugins) {
+      const spec = `${plugin.plugin}@${plugin.marketplace}`;
```
dmcaulay, line 189: Spec built as `` `${plugin.plugin}@${plugin.marketplace}` `` — this is inverted from the `plugin@marketplace` convention (plugin name first, marketplace second). More importantly, if the marketplace is `app-vitals/shipwright`, the result would be `shipwright@app-vitals/shipwright` which may not parse correctly. Verify the exact format `claude plugin install` expects and consider a helper or type-level constraint to prevent malformed specs.

dmcaulay (commented):
COMMENT — three important findings before this lands safely: (1) `@BOOTSTRAP.md` in CLAUDE.md.template will error on every agent session after first-run cleanup deletes the file; (2) the plugin install spec format `${plugin}@${marketplace}` is inverted — the documented form is `plugin@org/repo`; (3) DEFAULT_PLUGINS uses `"shipwright"` as the marketplace identifier but the correct form appears to be `"app-vitals/shipwright"`, causing silent bootstrap failures. Strengths: loadState JSON corruption recovery is correct; seedFile wx-flag semantics are right; test suite uses real temp dirs with injected execFn.

dodizzle (commented):
Re-review after dmcaulay's COMMENT: APPROVE — all three findings addressed.

- **`@BOOTSTRAP.md` removed from `CLAUDE.md.template`**: no more session errors when first-run cleanup removes the ephemeral file.
- **`DEFAULT_PLUGINS` marketplace corrected to `app-vitals/shipwright`**: plugin install spec now produces `shipwright@app-vitals/shipwright` — correct form. Tests updated to match.
- **Spec format**: `${plugin.plugin}@${plugin.marketplace}` was already the correct documented form; the only issue was the marketplace value itself (finding 3), now fixed.

CI green. No blocking issues. **Verdict: APPROVE** (posted as COMMENT — self-review).

dmcaulay (commented):
COMMENT — prior findings resolved; one new issue before landing.

**Prior findings addressed:**
- `@BOOTSTRAP.md` removed from `CLAUDE.md.template` ✓
- `DEFAULT_PLUGINS` marketplace corrected to `"app-vitals/shipwright"` ✓ (tests at lines 496-497 confirm)

**New finding:** `BOOTSTRAP.md.template` step 7 lists all five shipwright crons as "already running," but `system-crons.ts` ships only `shipwright-dev-task` and `shipwright-review-patch` with `enabled: true`. The other three (`shipwright-patch`, `shipwright-review`, `shipwright-deploy`) ship as `enabled: false`. A fresh agent will confidently tell its user that autonomous patching, reviewing, and deployment are active when they are not. Fix: list only the two enabled-by-default crons; note the others require explicit opt-in.

agent/workspace/BOOTSTRAP.md.template
```diff
@@ -0,0 +1,40 @@
...
+2. **Discover your identity** — learn your name, role, vibe, and emoji collaboratively
+3. **Determine deployment type** — is this a solo deployment (one human) or a team?
+4. **Update IDENTITY.md** — fill in the fields and add the correct import:
+   - Solo: add `@USER.md` at the bottom
+   - Team: add `@TEAM.md` at the bottom
+5. **Create USER.md or TEAM.md** — fill it in with context about your human(s)
+6. **Clone the repo** — `git clone <repo-url> repos/<repo>` so it's available for research
+7. **Walk through the dev loop** — explain that cron jobs are already running:
+   - `shipwright-dev-task` (every 30 min) — picks up approved todos and ships them via `/shipwright:dev-task`
+   - `shipwright-patch` (every 30 min) — addresses unresolved review findings and CI failures via `/shipwright:patch`
```
dmcaulay, line 17: `shipwright-patch`, `shipwright-review`, and `shipwright-deploy` ship with `enabled: false` in `system-crons.ts` — listing them here as "already running" will mislead fresh agents. Only `shipwright-dev-task` and `shipwright-review-patch` are enabled by default. Suggest listing just those two, with a note that the others can be enabled via the accounts API when the agent is ready.

PR #106, opened 2026-06-08
feat(she-4.2): port Slack + cron runtime from vitals-os

dmcaulay:
## Summary
- Ports `slack.ts`, `cron-handler.ts`, and `slack-manifest.ts` from the vitals-os agent into `agent/src/`
- All dependencies are injected (`createSlackApp(runner, ...)`) — no global config reads, no hardcoded `runClaude`
- Adds `@slack/bolt` and `@slack/web-api` to the agent package

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Slack Bolt app factory instantiates with injected SlackClient and ClaudeRunner | ✅ MET |
| 2 | Cron handler routes channel vs user vs silent delivery correctly | ✅ MET |
| 3 | [silent] marker suppresses post; DM thread always receives reply | ✅ MET |
| 4 | Cron handler integration test: prompt runs → response posted → session persisted | ✅ MET |
| 5 | Unit tests for event routing; integration test using injected runner double; no tests retired | ✅ MET |

## Test Plan
- 36 new tests (unit + integration): all pass
- `cron-handler.integration.test.ts`: validation, channel/DM delivery, [silent] suppression, preCheck scripts, onPost/onSession callbacks
- `slack.unit.test.ts`: MockApp captures handlers, DM routing, channel thread routing, [silent] DM override, app_mention, reaction_added, formatRunErrorForSlack, dispatchMarkers
- Full agent suite: 239 pass, 0 fail

### DIFF
--- a/agent/src/cron-handler.ts
+++ b/agent/src/cron-handler.ts
@@ -1,0 +1,265 @@
+/**
+ * Agent cron request handler.
+ *
+ * Receives a cron payload, runs the prompt through Claude, and posts the
+ * result to Slack. Extracted as a pure function with injected deps for
+ * testability.
+ */
+
+import { existsSync, readFileSync } from "node:fs";
+import { homedir } from "node:os";
+import { join, resolve } from "node:path";
+import type { WebClient } from "@slack/web-api";
+import type { TokenUsage } from "./claude.ts";
+import { markdownToSlack } from "./format.ts";
+import { parseMarkers } from "./markers.ts";
+import {
+  forwardNewMetrics,
+  forwardTokenUsage,
+  snapshotMetrics,
+} from "./posthog.ts";
+import { type SynthesizeSpeechFn, dispatchMarkers } from "./slack.ts";
+import type { VoiceConfig } from "./voice.ts";
+
+type ClaudeRunner = (
+  message: string,
+) => Promise<{ result: string; sessionId?: string; usage?: TokenUsage }>;
+
+interface CronRequest {
+  jobId: string;
+  prompt: string;
+  channel?: string;
+  user?: string;
+  silent?: boolean;
+  preCheck?: string;
+}
+
+export interface CronHandlerDeps {
+  slack: WebClient;
+  runner: ClaudeRunner;
+  formatter?: (text: string) => string;
+  /** Called after a successful post so the caller can track the thread. */
+  onPost?: (channel: string, ts: string) => void;
+  /** Called after a successful post when a sessionId is available. */
+  onSession?: (channel: string, ts: string, sessionId: string) => void;
+  /** Optional speech synthesis — if absent, [speak:] markers are skipped. */
+  synthesizeSpeechFn?: SynthesizeSpeechFn;
+  voiceConfig?: VoiceConfig;
+  /** Workspace root — used to forward metrics.jsonl entries to PostHog. */
+  workspace?: string;
+  /** Plugin cache dir — overridable for testing. */
+  pluginCacheDir?: string;
+  pluginManifestPath?: string;
+  alertsChannel?: string;
+}
+
+export async function handleCronRequest(
+  req: CronRequest,
+  deps: CronHandlerDeps,
+): Promise<void> {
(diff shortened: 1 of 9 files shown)


dmcaulay (commented):
APPROVE — solid port of the Slack + cron runtime. DI-first throughout, Prisma transaction around system cron reconciliation, correct [silent] marker handling across all three event types (message, app_mention, reaction_added). Test coverage is thorough: MockApp captures handlers for synchronous firing with no global overrides, and preCheck integration tests cover all three exit-code paths.

One suggestion: `updateDelivery()` in `agent-cron-jobs.ts` skips the `validateDeliveryTarget()` check that `create()` and `update()` both apply. Not wired to any HTTP endpoint yet (not in `AdminDeps` Pick), so no active risk — but worth hardening before it's ever exposed.

dodizzle (commented):
COMMENT — three findings: (1) `new Date()` in `cron-handler.ts` is un-injected Clock in a tested code path; (2) preCheck `exit >= 2` crash path is untested, contradicting AC 5; (3) `new Date()` / `Date.now()` in `slack.ts` event handlers are un-injected, and the error path measures a different interval than the success path.

agent/src/cron-handler.ts
```diff
@@ -0,0 +1,265 @@
...
+      console.log(
+        `[agent:cron] preCheck returned no work for job "${jobId}" — skipping tick`,
+      );
+      return;
+    }
+
+    prompt = output;
+  }
+
+  const now = new Date().toLocaleString("en-US", {
```
dodizzle, line 183: `new Date()` here is a raw wall-clock call in a tested production code path — a direct violation of the test-readiness contract. `CronHandlerDeps` has no `Clock` field, so there is no injection point. The integration tests work around non-determinism by never asserting on the timestamp fragment, which means this line is silently untestable.

`Clock` / `SystemClock` / `FixedClock` already exist in `agent/src/clock.ts`. Add `clock?: Clock` to `CronHandlerDeps` and replace with `(deps.clock ?? SystemClock()).now().toLocaleString("en-US", { timeZone: "America/Los_Angeles" })`.

agent/src/cron-handler.ts
```diff
@@ -0,0 +1,265 @@
...
+      stderr: "pipe",
+    });
+    const [checkOutput, checkStderr] = await Promise.all([
+      new Response(checkProc.stdout).text(),
+      new Response(checkProc.stderr).text(),
+    ]);
+    const checkExitCode = await checkProc.exited;
+    const output = checkOutput.trim();
+
+    if (checkExitCode >= 2) {
```
dodizzle, line 155: The `exit >= 2` crash branch (lines 155–170) has no integration test, despite AC 5 claiming preCheck exit-code coverage. This is also the only path that fires `alertsChannel` alerting — an untested production path. Add a test: write a temp script that `process.exit(2)`, pass an `alertsChannel` and a mock Slack client, assert `chat.postMessage` was called on `alertsChannel` and `mockRunner` was not called.

agent/src/slack.ts
```diff
@@ -0,0 +1,674 @@
...
+    if (msg.user) {
+      const name = await resolveUserFn(msg.user, client);
+      prompt = `[${name}]: ${prompt}`;
+    }
+
+    if (!isDM && msg.thread_ts) {
+      prompt = `[Thread message — respond normally, or use [silent] if no response is needed]\n${prompt}`;
+    }
+
+    const startedAt = new Date();
```
dodizzle, line 380: `startedAt = new Date()` is an un-injected Clock call — `createSlackApp` has no `Clock` parameter. The success path computes `durationMs = endedAt.getTime() - startedAt.getTime()` (runner-only duration) while the error path at line 448 uses `Date.now() - startedAt.getTime()` (runner + all downstream async work including `say()`), measuring different intervals in the two branches.

Add a `clock?: Clock` parameter to `createSlackApp` and replace all `new Date()` / `Date.now()` calls with `clock.now()`. The same pattern repeats in the `app_mention` handler (lines 526, 529, 585).

PR #121, opened 2026-06-08
feat: admin UI cron, tool, and token management forms

dodizzle:
## Summary
- Add cron job management forms to the admin agent detail page (add form, toggle/delete per row, system vs custom sections)
- Add tool pattern management forms (add pattern, toggle/delete per row)
- Add token management forms (create token, revoke per row)

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Cron job form (schedule, prompt, channel/user), toggle/delete per row, system vs custom sections | ✅ MET |
| 2 | Tool pattern form, toggle/delete per row | ✅ MET |
| 3 | Token create form (label), revoke per row | ✅ MET |
| 4 | All mutations redirect back to agent detail page on success | ✅ MET |
| 5 | Form validation errors render inline (matching env var error pattern) | ✅ MET |
| 6 | Smoke tests for all new POST routes added | ✅ MET |
| 7 | Test decision: smoke tests added, no existing tests retired | ✅ MET |

## Test Plan
- [x] 8 new smoke tests covering all POST routes (crons create/toggle/delete, tools add/toggle/delete, tokens create/revoke)
- [x] All 15 smoke tests pass
- [x] Full suite: 1658 pass, 0 fail
- [x] Typecheck: clean
- [x] Lint (biome): clean


agent/src/admin-ui.ts
```diff
@@ -322,6 +347,162 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono {
...
+    let silent = false;
+    let name: string | null = null;
+    try {
+      const formData = await c.req.formData();
+      schedule = formData.get("schedule")?.toString();
+      prompt = formData.get("prompt")?.toString();
+      channel = formData.get("channel")?.toString() || null;
+      user = formData.get("user")?.toString() || null;
+      silent = formData.get("silent") === "on" || formData.get("silent") === "true";
+      name = formData.get("name")?.toString() || null;
```
dodizzle, line 373: The `enabled` checkbox in the add-cron form is never read here — it's parsed from the form HTML but not extracted in this handler. `agentCronJobService.create()` is called without `enabled`, so it defaults to `true` regardless of the checkbox state.

Fix: read the field and pass it through:
```ts
const enabled = formData.get("enabled") === "true" || formData.get("enabled") === "on";
```
Then include `enabled` in the `create()` call.

agent/src/admin-ui.ts
```diff
@@ -322,6 +347,162 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono {
...
+      // ignore errors — redirect back regardless
+    }
+    return c.redirect(`/admin/agents/${agentId}`, 302);
+  });
+
+  app.post("/admin/agents/:id/crons/:cronId/delete", requireAuth, async (c) => {
+    const agentId = c.req.param("id");
+    const cronId = c.req.param("cronId");
+    try {
+      await agentCronJobService.delete(agentId, cronId);
```
dodizzle, line 431: No server-side guard prevents deleting a system cron. The UI omits the Delete button for `c.system === true` rows, but a direct `POST` to this route can delete any cron regardless of its `system` flag.

`AgentCronJobService.delete()` also has no system check — it only verifies `agentId` ownership.

Fix: either fetch the cron before deleting and return early if `existing.system === true`, or extend the service's `delete()` to throw when the target is a system cron.

agent/src/admin-ui.ts
```diff
@@ -259,13 +269,27 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono {
 
   // ─── Agent detail ─────────────────────────────────────────────────────────
 
+  const ERROR_MESSAGES: Record<string, string> = {
+    missing_fields: "Required fields are missing.",
+    create_failed: "Failed to create — please try again.",
+    invalid_schedule: "Invalid cron schedule expression.",
+    invalid_target:
+      "Invalid delivery target — set channel or user (or enable silent mode).",
```
dodizzle, line 278: `invalid_schedule` and `invalid_target` are dead keys — they're defined here but never emitted as error codes. When `create()` throws `UnprocessableEntityError` for a bad schedule or invalid delivery target, the raw `err.message` is URL-encoded directly into the redirect (line 386), bypassing this map entirely.

Either catch `UnprocessableEntityError` specifically and redirect with the appropriate key, or remove these entries.

dodizzle (commented):
COMMENT — Two actionable issues: the `enabled` checkbox on the add-cron form is silently ignored (crons always create enabled), and system crons lack a server-side delete guard that relies on UI-only enforcement. See inline comments.

agent/src/admin-ui.ts
dmcaulay, line 373: The `enabled` checkbox in the form (`<input name="enabled" type="checkbox" ... checked />`) is never read here. `agentCronJobService.create()` is called without an `enabled` field and defaults to `enabled: true` in the service, so unchecking the checkbox has no effect.

Add before closing the try block:
```ts
const enabledVal = formData.get("enabled");
enabled = enabledVal === "true" || enabledVal === "on";
```
And declare `let enabled = true;` alongside the other vars, then pass `enabled` to `create()`.

agent/src/admin-ui.ts
```diff
@@ -322,6 +347,162 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono {
...
+    try {
+      const formData = await c.req.formData();
+      label = formData.get("label")?.toString() || undefined;
+    } catch {
+      return c.redirect(`/admin/agents/${agentId}`, 302);
+    }
+    try {
+      const { rawToken } = await agentTokenService.create(agentId, label);
+      return c.redirect(
+        `/admin/agents/${agentId}?newToken=${encodeURIComponent(rawToken)}`,
```
dmcaulay: Redirecting with `?newToken=<rawToken>` puts the secret in the URL. The GET request that follows will log the full URL — including the raw token — in every reverse proxy / web server access log, and it's stored in browser history. The page says "Copy it now — it will not be shown again" but the token persists in logs indefinitely.

Preferred fix: render the token in a direct `200` response from the POST (skip the redirect), or store it in a server-side session flash and read-once on the subsequent GET. If neither is feasible short-term, document the log-exposure risk in a comment.

dmcaulay (commented):
COMMENT — 2 issues to address before merge: (1) `enabled` checkbox in cron create form is silently ignored — cron always created as enabled regardless of form state; (2) raw token passed in `?newToken=` redirect URL leaks the secret into server access logs and browser history. See inline comments for details.

dodizzle (commented):
APPROVE — all four prior findings addressed in af8bf53: enabled checkbox now correctly read from form data, system cron delete guard added at the route level via `agentCronJobService.get()`, error-key mapping wired for `UnprocessableEntityError`, and raw token no longer exposed in redirect URL (renders 200 HTML response directly instead). Smoke tests updated to cover all fixed paths. No new issues found.

dodizzle (commented):
APPROVE — CI typecheck fix only (adds missing `get` stub to `agentCronJobService` mock in `slack-provisioning.integration.test.ts`). No production code changes. All four prior findings remain addressed. Ready to merge.

agent/src/slack-provisioning.integration.test.ts
```diff
@@ -104,12 +104,30 @@ function makeMockDeps(
     },
     agentCronJobService: {
       list: async () => [],
+      create: async () => {
+        throw new Error("not implemented");
+      },
+      setEnabled: async () => {
+        throw new Error("not implemented");
+      },
+      delete: async () => {},
```
dmcaulay, line 116: `get` is missing from this mock. `AdminUIDeps` now requires `Pick<AgentCronJobService, "list" | "create" | "setEnabled" | "delete" | "get">` — the smoke test mock has it but this one doesn't, causing the CI typecheck failure.

```ts
delete: async () => {},
get: async () => { throw new Error("not implemented"); },
```

agent/src/admin-ui.ts
```diff
@@ -322,6 +348,204 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono {
...
+    try {
+      const cron = await agentCronJobService.get(agentId, cronId);
+      if (cron.system) {
+        throw new ForbiddenError("system crons cannot be deleted");
+      }
+      await agentCronJobService.delete(agentId, cronId);
+    } catch (err) {
+      if (err instanceof ForbiddenError) {
+        return c.redirect(
+          `/admin/agents/${agentId}?error=${encodeURIComponent(err.message)}`,
```
dmcaulay, line 435: This redirect uses the raw error message rather than an opaque key, unlike all other error redirects. Since `ForbiddenError.message` is static here it's safe, but it breaks the `ERROR_MESSAGES` pattern and the GET handler renders it via the fallback (`rawError` passthrough) rather than a controlled string.

Consider:
```ts
ERROR_MESSAGES: { ..., system_cron: "System crons cannot be deleted." }
// then:
return c.redirect(`/admin/agents/${agentId}?error=system_cron`, 302);
```

dmcaulay (commented):
COMMENT — Prior findings fully addressed. One new issue blocks CI: the `agentCronJobService` mock in `slack-provisioning.integration.test.ts` is missing the newly-required `get` method, causing a TypeScript error. One additional suggestion on the ForbiddenError redirect pattern. See inline comments.

dmcaulay (approved):
APPROVE — All prior findings addressed and CI is green. The missing `get` mock in `slack-provisioning.integration.test.ts` has been added, unblocking the typecheck. The `enabled` checkbox fix, system cron delete guard, error-key mapping, and token exposure fix all land cleanly.

PR #131, opened 2026-06-08
feat: add local SQLite event store + POST /batch/ ingest to metrics

dodizzle:
## Summary
- Add `metrics/src/local-store.ts`: a `bun:sqlite`-backed `LocalEventStore` with an idempotent `events` table (`insert_id` UNIQUE for dedup, indexes on `event` and `timestamp`), exposing `insertEvent`, `queryByEvent(event, {from,to})`, and `close()`.
- Add a PostHog-shaped `POST /batch/` ingest route to the metrics app — registered **only** when a `localStore` is injected via `MetricsDeps` (404 otherwise); persists each batch event (dedup on `properties.$insert_id`) and returns `{status:1}`.
- Wire the store through the existing `createMetricsApp` DI seam — no module-level singletons, no globals. DB path from `METRICS_DB_PATH` (default `state/metrics.db`).
- This does **not** flip the default metrics mode (deferred to LDS-1.2); the route activates purely by presence of the dep.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Store creates `events` table idempotently; dedup on `insert_id`; range-queryable | MET |
| 2 | `POST /batch/` persists, returns `{status:1}`, ignores duplicate `insert_id` | MET |
| 3 | Injected via `createMetricsApp` deps — no singletons/globals | MET |
| 4 | unit (`:memory:`) insert+dedup+range; smoke for `/batch/` via `app.request()`; no `mock.module()`/`global.*`; Clock injected | MET |
| 5 | No existing tests retired; existing metrics suites pass | MET |

## Test Plan
- `local-store.unit.test.ts` (5 tests, `:memory:` DB + injected `FixedClock`): insert+parse, dedup, inclusive range filter, NULL-`insert_id` retention, idempotent init.
- `batch-ingest.smoke.test.ts` (4 tests via `app.request()`): persistence + `{status:1}`, cross-request dedup, malformed-body 400, route-absent 404 when no store injected.
- Full metrics suite: 574 pass / 0 fail. `local-store.ts` at 100% line/func coverage.
- [x] Lint, typecheck, and pre-commit checks passing

### DIFF
--- a/metrics/src/api.ts
+++ b/metrics/src/api.ts
@@ -15,6 +15,7 @@ import type { AppHandler, AuthEnv, Caller } from "./lib/api-auth.ts";
 import { ErrorSchema } from "./lib/api-schemas.ts";
 import { registerWithAuthz } from "./lib/api-utils.ts";
 import { createSessionMiddleware } from "./lib/session-middleware.ts";
+import type { LocalEventStore } from "./local-store.ts";
 import { renderDashboardPage } from "./dashboard/dashboard-page.ts";
 import {
   resolveDateRangeForMeta,
@@ -81,6 +82,13 @@ export interface MetricsDeps {
   dashboardToken?: string;
   /** Offline mode: skip session auth and serve /dashboard as a default local user. Default false. */
   offlineMode?: boolean;
+  /**
+   * Local event store. When provided, the PostHog-shaped ingest route
+   * `POST /batch/` is registered and writes batches to this store. When
+   * absent, the route is NOT registered (404) — the default-mode flip is
+   * deferred to a later task.
+   */
+  localStore?: LocalEventStore;
 }
 
 // ─── Route definitions (inlined) ─────────────────────────────────────────────
@@ -912,7 +920,13 @@ export function createMetricsApp(
   // /metrics/* — accepts bearer token OR session cookie; returns 401 JSON on failure
   app.use(
     "/metrics/*",
-    createCombinedAuthMiddleware(apiKeys, sessionSecret, accountsClient, requireOwnerRole, dashboardToken),
+    createCombinedAuthMiddleware(
+      apiKeys,
+      sessionSecret,
+      accountsClient,
+      requireOwnerRole,
+      dashboardToken,
+    ),
   );
 
   const handlers = createMetricsHandlers(client, accountsClient, deps);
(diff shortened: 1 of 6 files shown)


dmcaulay:
COMMENT - did you consider using a common interface and dependency injection for this store and the posthog store? it'd be nice to use standard design patterns like this if it doesn't get messy.

dmcaulay (approved):
APPROVE — clean implementation. `LocalEventStore` interface + optional `MetricsDeps.localStore` dep is solid DI; addresses the design question from the earlier comment. `INSERT OR IGNORE` on `insert_id UNIQUE` is the right SQLite idiom, NULL retention verified by test. Route gated entirely on DI presence (404 when absent). CI green. Ready to merge.

metrics/src/api.ts
```diff
@@ -932,6 +946,53 @@ export function createMetricsApp(
...
 
+  // ─── Ingest: POST /batch/ (local-store mode only) ─────────────────────────
+  //
+  // PostHog-shaped batch ingest. Mounted as a plain route (NOT behind the
+  // /metrics/* combined-auth middleware) to mirror PostHog's unauthenticated
+  // transport — the api_key travels in the body. Registered ONLY when a local
+  // store is injected; otherwise the route is absent (404).
+  const localStore = deps?.localStore;
+  if (localStore) {
+    app.post("/batch/", async (c) => {
```
dodizzle, line 957: **No auth, no body size cap.** This route has no middleware — any caller who can reach the server can write arbitrary events to the SQLite store. There's also no Content-Length check before `c.req.json()`, so a large payload is fully buffered before validation.

The justification (mirroring PostHog's client SDK transport) doesn't translate: PostHog's public ingest endpoint is for untrusted browsers hitting a SaaS boundary. This is a server-side route writing to a local file-backed DB.

Add a body size guard (e.g. reject if > 1 MB) and note in docs that `POST /batch/` should only be exposed on a private/loopback interface.

dodizzle (commented):
## Review: feat: add local SQLite event store + POST /batch/ ingest to metrics

**Verdict: COMMENT** — three issues to address before merge.

### Issue 1 — local-store tests are unit but exercise real SQLite (wrong layer)

`metrics/src/local-store.unit.test.ts:1`

`test-system.md` defines the unit layer as "Pure logic — no I/O of any kind." This test file instantiates a real `bun:sqlite` Database (even via `:memory:`) and calls real insert/query operations — that's the integration layer. The agent's DB tests are explicitly at that layer.

**Fix**: Rename to `metrics/src/local-store.integration.test.ts` — no code changes needed.

### Issue 3 — CLAUDE.md and test-system.md still say "Stateless... No database"

`CLAUDE.md:16` and `docs/test-readiness/test-system.md:54`

Both files still say "Stateless Hono service... No database." The PR correctly updates `docs/architecture.md` and `docs/metrics.md` but leaves these two contradicting the new reality.

**Fix**: Update both to match `docs/architecture.md`: "Optional local SQLite event store for offline ingest (`POST /batch/`)."

PR #147, opened 2026-06-09
test(agent): expand cron-handler integration tests to vitals-os parity

dmcaulay:
## Summary
- Expanded `cron-handler.integration.test.ts` from 256 to 982 lines (vitals-os parity)
- Added `startHealthServer()` to `health.ts` with POST `/cron` endpoint (200/400/422/503/500), GET `/cron` → 405
- Fixed `[silent]` marker behavior in `cron-handler.ts`: marker now suppresses channel posts only; DMs always get a reply (CAPABILITIES.md spec)

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | [silent] suppresses channel postMessage | ✅ MET |
| 2 | [silent] in DMs is IGNORED (DM always gets a reply) | ✅ MET |
| 3 | [react:emoji] → reactions.add per emoji | ✅ MET |
| 4 | [upload:/path] → files.uploadV2 called | ✅ MET |
| 5 | [speak:text] → synthesizeSpeechFn + uploadV2 | ✅ MET |
| 6 | Marker tags stripped from visible text | ✅ MET |
| 7 | Channel routing tests | ✅ MET |
| 8 | DM routing tests | ✅ MET |
| 9 | HTTP POST /cron 200/400/422/405/503/500 | ✅ MET |
| 10 | Formatter applied before posting | ✅ MET |
| 11 | All 10 existing tests preserved + passing | ✅ MET |
| 12 | No test-env.ts import | ✅ MET |
| 13 | bun test --filter agent passes | ✅ MET (446 pass) |

## Test Plan
- [x] `bun test --filter agent` — 446 pass, 0 fail
- [x] `tsc --noEmit` — clean
- [x] `cron-handler.ts` line coverage: 100%
- [x] `health.ts` line coverage: 100%
- Note: PostHog forwarding tests from vitals-os excluded (use `globalThis.fetch =` override, violating CLAUDE.md isolation contract)

### DIFF
--- a/agent/src/cron-handler.integration.test.ts
+++ b/agent/src/cron-handler.integration.test.ts
@@ -848,7 +848,7 @@ describe("POST /cron HTTP endpoint", () => {
 
   // biome-ignore lint/suspicious/noExplicitAny: Server type param varies by bun version
   function serve(port: number): Server<any> {
-    const s = startHealthServer(port, undefined, deps);
+    const s = startHealthServer(port, deps);
     servers.push(s);
     return s;
   }
--- a/agent/src/cron-handler.integration.test.ts
+++ b/agent/src/cron-handler.integration.test.ts
@@ -1,15 +1,22 @@
 /**
- * Integration tests for cron-handler.ts
+ * Integration tests for cron-handler.ts and POST /cron endpoint in health.ts.
  *
  * Strategy: inject all deps. No real Slack or Claude calls.
+ *
+ * - runner: mock function
+ * - slack: mock WebClient with chat.postMessage and conversations.open
+ * - formatter: identity function (no markdown conversion)
+ * - HTTP tests: real Bun.serve via startHealthServer
  */
 
 import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
-import { mkdtempSync, writeFileSync } from "node:fs";
+import { mkdirSync, mkdtempSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import type { WebClient } from "@slack/web-api";
+import type { Server } from "bun";
 import { ValidationError, handleCronRequest } from "./cron-handler.ts";
+import { startHealthServer } from "./health.ts";
 
 // ─── Shared mocks ─────────────────────────────────────────────────────────────
 
@@ -35,7 +42,7 @@ const mockRunner = mock(() =>
 const deps = {
   slack: mockSlack,
   runner: mockRunner,
-  formatter: (text: string) => text,
+  formatter: (text: string) => text, // identity — no markdown conversion in tests
 };
 
 afterEach(() => {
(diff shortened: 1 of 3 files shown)


agent/src/cron-handler.integration.test.ts
```diff
@@ -253,4 +542,441 @@ describe("handleCronRequest — preCheck", () => {
...
+  }
+
+  afterEach(() => {
+    for (const s of servers) s.stop(true);
+    servers.length = 0;
+  });
+
+  test("200 on valid channel request", async () => {
+    mockRunner.mockResolvedValueOnce({ result: "ok", sessionId: "s" });
+    serve(19920);
```
dmcaulay, line 863: Hardcoded ports 19920–19929 work fine here since each test gets its own port and afterEach stops them. One future-proofing option: `startHealthServer(0, deps)` lets the OS pick a free port, then read it back via `server.port`. Eliminates any risk of collision if another test file ever claims these ports.

dmcaulay (commented):
APPROVE — all 13 ACs met. cron-handler.ts [silent] bug fix is correct and minimal. startHealthServer() POST /cron adds the right status codes (200/400/422/503/500) and they're all covered. HTTP tests properly stop servers in afterEach. Suggestion: consider port: 0 for the HTTP tests to eliminate potential port conflicts.

(Posted as COMMENT — GitHub blocks self-review.)

agent/src/cron-handler.integration.test.ts
```diff
@@ -253,4 +542,441 @@ describe("handleCronRequest — preCheck", () => {
...
+      { ...deps, pluginCacheDir: tmpDir },
+    );
+
+    expect(mockRunner).not.toHaveBeenCalled();
+  });
+
+  test("preCheck script not found → warning logged, runner NOT called, no throw", async () => {
+    const warnMessages: string[] = [];
+    const originalWarn = console.warn;
+    console.warn = (...args: unknown[]) => {
```
dodizzle, line 619: `console.warn = ...` is a global mutation. CLAUDE.md prohibits `global.*` overrides — Bun shares the test process, so this can leak between suites even with the `finally` restore. Inject a `warnFn` or `logger` into the `deps` object instead, matching the DI pattern used everywhere else in this file.

dodizzle (commented):
COMMENT — two issues before merge.

**Issue 1** (`agent/src/cron-handler.integration.test.ts:619`, also :690, :784, :816): `console.warn = ...` is a global mutation. CLAUDE.md explicitly prohibits `global.*` overrides — Bun shares the test process so leaked globals break sibling suites. The `finally` restore doesn't make it safe under concurrent execution. Fix: inject a `warnFn` or `logger` dep via the existing `deps` object so tests can capture warnings without touching global state.

**Issue 2** (`agent/src/cron-handler.integration.test.ts`: HTTP tests at lines ~963–980): Hardcoded ports 19920–19929 are fragile — `EADDRINUSE` if another process binds them. Use `port: 0` (OS-assigned) and read `server.port` for the fetch URL. The PR's own self-review flagged this — worth addressing.

Everything else is solid: the `[silent]` DM fix is correct (`isDmOnly = !!user && !channel`, dep-flag silent still suppresses all), `startHealthServer` handles the 503/unconfigured case cleanly, and the 13 preCheck paths give good coverage of an edge-heavy code path.

PR #163, opened 2026-06-09
fix: point runMigrations cwd to admin/ after schema move

dodizzle:
## Summary

- Fixes `runMigrations()` boot crash introduced by #133: `cwd` was still pointing to `agent/` after the Prisma schema moved to `admin/prisma/schema.prisma`. When `DATABASE_URL_AGENT` is set, the server would throw a file-not-found error before serving any requests.
- Bumps `plugins/shipwright/README.md` version header from `v4.26.4` → `v4.26.5` to match the plugin sync in 94d7cbe.

## Test plan

- [ ] Set `DATABASE_URL_AGENT` locally and confirm `runMigrations()` no longer throws at boot
- [ ] Verify `task db:provision` and `task db:migrate` still work (fixed in #133)

Closes the two outstanding findings from the #133 review that slipped through before merge.

### DIFF
--- a/agent/src/run-agent.ts
+++ b/agent/src/run-agent.ts
@@ -252,7 +252,7 @@ async function runMigrations(): Promise<void> {
   const proc = Bun.spawn(
     ["bunx", "prisma", "migrate", "deploy", "--schema=prisma/schema.prisma"],
     {
-      cwd: join(import.meta.dir, ".."),
+      cwd: join(import.meta.dir, "../../admin"),
       env: { ...process.env, DATABASE_URL_AGENT: databaseUrl },
       stdout: "pipe",
       stderr: "pipe",
--- a/plugins/shipwright/README.md
+++ b/plugins/shipwright/README.md
@@ -1,4 +1,4 @@
-# Shipwright v4.26.4
+# Shipwright v4.26.5
 
 A structured dev pipeline plugin for Claude Code. Plan sessions, execute tasks, run autonomous dev loops, perform multi-agent code reviews, and conduct integrated project research — for any software project.
 


dmcaulay (approved):
APPROVE — minimal, correct fix for a boot crash introduced by #133. `runMigrations()` `cwd` now points at `admin/` so `prisma/schema.prisma` resolves correctly when `DATABASE_URL_AGENT` is set. Path math verified (`agent/src` → `../../admin` = `<repo>/admin`), CI green, version sync clean.

agent/src/run-agent.ts
```diff
@@ -252,7 +252,7 @@ async function runMigrations(): Promise<void> {
   const proc = Bun.spawn(
     ["bunx", "prisma", "migrate", "deploy", "--schema=prisma/schema.prisma"],
     {
-      cwd: join(import.meta.dir, ".."),
+      cwd: join(import.meta.dir, "../../admin"),
```
dodizzle, line 255: No automated test covers this `cwd` argument — it already regressed once (PR #133 missed this). A unit/smoke test mocking `Bun.spawn` and asserting the resolved path equals `<repo>/admin/` would catch future schema relocations. Low priority given test complexity, but worth a follow-up issue.

dodizzle (commented):
**APPROVE** — Correct regression fix. `runMigrations()` `cwd` now resolves to `admin/` where `admin/prisma/schema.prisma` lives after the schema move in #133. Path arithmetic is correct (`import.meta.dir` = `agent/src/` → `../../admin` = repo root `admin/`). CI green. One suggestion: a smoke test asserting the resolved `cwd` in `runMigrations()` would prevent this class of regression from slipping through again — low priority but worth tracking.

dmcaulay (approved):
APPROVE — minimal regression fix: `runMigrations()` `cwd` now resolves to `admin/` where `admin/prisma/schema.prisma` lives after the schema move in #133. Path arithmetic verified (`agent/src` + `../../admin` = repo root `admin/`). CI green, version sync clean, `scripts/chat.ts` well-structured with CLAUDE.md-compliant test coverage. Already holds two APPROVE reviews.

PR #175, opened 2026-06-09
test(agent): relabel setup-github-auth unit→integration + expand coverage (ATP-2.3)

dodizzle:
## Summary
- Rename `setup-github-auth.unit.test.ts` → `setup-github-auth.integration.test.ts` (correct layer label: DI-injected mocks exercise a multi-step initialization seam)
- Expand from 178 → 310 lines: add PAT path, skip path (empty + partial vars), non-zero spawnSync error-logging, and background refresh callback tests
- Port to `mock()` from bun:test for precise call-count and arg assertions

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Old unit test deleted; new integration test exists | ✅ MET |
| 2 | PAT path: gh auth setup-git via spawnSync, no writeToken/getBotIdentity | ✅ MET |
| 3 | Skip path (empty env): all mocks assert not called | ✅ MET |
| 4 | Skip path (partial GH_APP_ID only): same skip behavior | ✅ MET |
| 5 | App path non-zero spawnSync: runOrWarn logs error, doesn't throw | ✅ MET |
| 6 | Background refresh callback: rewrites token file only, no env/spawnSync mutation | ✅ MET |
| 7 | All existing App path tests preserved and passing | ✅ MET |
| 8 | Test layer: integration | ✅ MET |
| 9 | No test-env.ts import; bun test --filter agent passes (473 pass, 0 fail) | ✅ MET |

## Test Plan
- [x] `bun test --filter agent` — 473 pass, 0 fail, 55 skip
- [x] `cd agent && bun run typecheck` — clean
- [x] All 6 tests in `setup-github-auth.integration.test.ts` pass


dmcaulay (commented):
COMMENT — one fix before merge: `console.error = errorSpy` at line 134 is a `global.*` mutation, which CLAUDE.md explicitly prohibits (Bun shares the test process; leaked globals break sibling suites). The simplest fix is to drop the console assertion — the test already verifies the meaningful behavior (function doesn't throw on non-zero spawnSync). If the log output matters, add a `logger?` field to `GitHubAuthDeps` and inject it.

Everything else is clean: all 7 AC met, layer reclassification is correct, `mock()` used properly throughout (no `mock.module()`, no `global.fetch`), and the background refresh callback test is particularly thorough.

agent/src/setup-github-auth.integration.test.ts
```diff
@@ -0,0 +1,307 @@
...
+      ["config", "--global", "user.name", "keanu-hifriends[bot]"],
+      { stdio: "inherit", env: expectedEnv },
+    );
+    expect(spawnSync).toHaveBeenCalledWith(
+      "git",
+      [
+        "config",
+        "--global",
+        "user.email",
+        "987654+keanu-hifriends[bot]@users.noreply.github.com",
```
dmcaulay, line 134: `console.error = errorSpy` is a `global.*` mutation — prohibited by CLAUDE.md: "No `global.fetch`/`global.*` overrides — Bun shares the test process, so leaked globals break sibling suites." Even with `finally` restoration, there is a window during parallel test execution where the global is mutated.

Simplest fix: remove the console assertion. The AC requires only that the error is "handled" — the function completing without throwing already proves that. If verifying the log message is important, add `logger?: { error: (...args: unknown[]) => void }` to `GitHubAuthDeps` and pass the mock through deps.

agent/src/setup-github-auth.integration.test.ts
```diff
@@ -0,0 +1,307 @@
...
+    }
+    expect(spawnSync).toHaveBeenCalledTimes(3);
+    expect(startBackgroundRefresh).toHaveBeenCalledTimes(1);
+  });
+
+  test("logs an error when git config credential helper registration exits non-zero", async () => {
+    const { getToken, startBackgroundRefresh } = makeTokenManager();
+    const spawnSync = makeSpawnSync(1);
+    const errorSpy = mock((..._args: unknown[]) => {});
+    const originalError = console.error;
```
dodizzle: `console.error = errorSpy` is a global mutation — the project's isolation contract prohibits this (Bun shares the test process, so leaked globals break sibling suites). The `try/finally` mitigates but doesn't fully prevent leakage on unexpected panics.

Suggested fix: add an optional `logger` to `GitHubAuthDeps`:
```ts
logger?: { error: (...args: unknown[]) => void };
```
Default it to `console` in production, and pass `{ error: errorSpy }` here. Then `runOrWarn` calls `deps.logger?.error(...)` instead of `console.error` directly — no global state touched.

dodizzle (commented):
## Review: COMMENT

The rename to `integration` is correct and the expanded coverage is solid. One finding: lines 150-151 assign `console.error = errorSpy` — a global mutation that violates the test isolation contract (`No global.* mutations in tests`). A `try/finally` reduces but doesn't eliminate the risk (a thrown exception before the finally still corrupts the process-global `console.error` for sibling suites). The fix is to add an optional `logger` field to `GitHubAuthDeps` so `runOrWarn` can log through the injected logger instead of calling `console.error` directly.

Everything else is well done — `mock()` assertions are cleaner than the old hand-rolled stubs, the `neverCalled*` helper pattern is a good guard, and the background refresh callback test is thorough.

dodizzle (commented):
**APPROVE** — prior finding addressed.

The `console.error` global mutation flagged in the last review has been cleanly resolved: `GitHubAuthDeps` now has an optional `logger?: { error: (...args: unknown[]) => void }` field, `runOrWarn` accepts it as a defaulted parameter (`= console`), and the test passes `{ error: errorSpy }` via deps with no global mutation. The fix is correctly threaded through all three `runOrWarn` call sites (credential helper registration, user.name/user.email git config, and the PAT `gh auth setup-git` path). All 9 acceptance criteria met, test isolation contract satisfied, CI green.

dmcaulay (approved):
**APPROVE** — prior finding addressed.

The `console.error` global mutation flagged in the last review has been cleanly resolved: `GitHubAuthDeps` now has an optional `logger?: { error: (...args: unknown[]) => void }` field, `runOrWarn` accepts it as a defaulted parameter (`= console`), and the test passes `{ error: errorSpy }` via deps with no global mutation. The fix is correctly threaded through all four `runOrWarn` call sites (credential helper registration, user.name, user.email, and the PAT `gh auth setup-git` path). All 9 acceptance criteria met, test isolation contract satisfied, CI green.

PR #176, opened 2026-06-09
feat(admin): Playwright e2e tests — login page and agents page (ADM-3.2)

dmcaulay:
## Summary
- Add Playwright e2e tests for the admin UI login page and agent detail page (`admin/e2e/`)
- Implement `returnTo` param support in the login page so the OAuth button href preserves post-login destination
- Add env var value masking in the agent detail page (renders `••••••••` instead of raw secret values)

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | `admin/e2e/login-page.e2e.ts` covers: Google button present, all error banner variants, returnTo param | ✅ MET |
| 2 | `admin/e2e/agents-page.e2e.ts` covers: unauthenticated redirect, page content, masked env values, cron schedule, token label, forms, mobile viewport | ✅ MET |
| 3 | Tests run against a real Hono server spun up in `beforeAll` | ✅ MET |
| 4 | New e2e layer for admin package; no existing tests retired | ✅ MET |

## Test Plan
- [x] 21/21 Playwright e2e tests pass (`cd admin && bunx playwright test`)
- [x] 235/235 unit tests pass (`bun test admin/`)
- [x] Lint clean (`bunx biome lint .`)
- [x] Typecheck clean across all 4 packages


admin/src/admin-ui.ts
```diff
@@ -221,7 +221,8 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono {
 
   app.get("/admin/login", (c) => {
     const error = c.req.query("error") ?? undefined;
-    return html(renderLoginPage({ error }));
+    const returnTo = c.req.query("returnTo") ?? undefined;
+    return html(renderLoginPage({ error, returnTo }));
```
dodizzle, line 225: `returnTo` is read from the query here and passed to `renderLoginPage`, which renders it in the OAuth button href — but the `/auth/google` handler below (line 228) never reads `c.req.query('returnTo')`. The value is passed through the link but discarded by the handler. After `/auth/callback` completes, the user always lands on `/admin/agents` (line 322) regardless of `returnTo`.

To complete the feature: encode `returnTo` in the OAuth state (alongside the nonce in the `oauth_state` cookie), then in `/auth/callback` read it back and use it as the redirect destination instead of hard-coding `/admin/agents`.

dodizzle (commented):
## Review: COMMENT

Strong PR overall — env var masking is a clear security win, the test-server architecture is clean, and the mobile responsive fixes are well-scoped. One important finding: the `returnTo` feature is half-implemented.

`renderLoginPage` now renders `/auth/google?returnTo=<encoded>` in the OAuth button href, but the `/auth/google` handler never reads that parameter. After completing OAuth, `/auth/callback` always redirects to `/admin/agents` (line 322). The param is visible to the user in the link but has no effect — they won't be returned to their original destination. The e2e test verifies only that the href contains the param, not that the redirect honors it end-to-end.

**Fix**: either carry `returnTo` through the OAuth flow (encode it in the state cookie alongside the nonce, then read it in `/auth/callback` to use as the post-login redirect), or remove the `returnTo` param handling until the full round-trip is ready.

Also worth noting: `task ci` and `task e2e` don't include the new admin Playwright tests — they won't run in CI.

dmcaulay (commented):
**COMMENT** — `returnTo` post-login redirect is not implemented end-to-end.

The login page renders the button with `href=/auth/google?returnTo=<encoded>` (acceptance criterion met), but `/auth/google` drops the `returnTo` query param and `/auth/callback` always redirects to `/admin/agents`. After authenticating, users who arrived via a deep link always land on the agents list, not their original destination.

**Fix**: encode `returnTo` alongside the CSRF nonce in the `oauth_state` cookie (e.g. `{nonce, returnTo}` JSON), validate the value is a same-origin relative path in the callback, and redirect there on success. (`admin/src/admin-ui.ts` — `GET /auth/google` handler, `state: nonce` line)

Everything else is solid: OAuth flow, allowlist check, env var masking, e2e test architecture, PostgreSQL migration — all well-executed. The returnTo gap is the only blocking item.

dmcaulay (commented):
**APPROVE** — prior finding addressed.

The `returnTo` end-to-end redirect gap flagged in the last review is fully resolved. `returnTo` is now encoded as JSON `{nonce, returnTo}` in the `oauth_state` cookie at `/auth/google` (validated as a same-origin relative path before storage), extracted and re-validated in `/auth/callback`, and used as the post-login redirect destination. A smoke test verifies the round-trip. Malformed cookie JSON is safely handled. All prior strengths remain: e2e architecture, env var masking, GoogleAuthClient DI pattern, CSRF nonce validation, CI green.

PR #202, opened 2026-06-10
feat: add SHIPWRIGHT_LOCAL_MARKETPLACE seam to installPlugins (LPP-1.2)

dodizzle:
## Summary
- Adds `SHIPWRIGHT_LOCAL_MARKETPLACE` env var seam to `installPlugins` in `agent/src/setup.ts`
- When set, plugin install/update specs use the local path (`{plugin}@{local_path}`) instead of the GitHub marketplace slug, enabling uncommitted plugin edits to run inside the container
- When unset, behavior is byte-identical to current — zero impact on existing deployments

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | SHIPWRIGHT_LOCAL_MARKETPLACE set → local path used for install/update | MET |
| 2 | SHIPWRIGHT_LOCAL_MARKETPLACE unset → behavior byte-identical to current | MET |
| 3 | Unit test (injected exec double): local/GitHub path coverage | MET |
| 4 | Integration test: only new env-var case added to existing describe block | MET |

## Test Plan
- [x] `setup.unit.test.ts`: 3 new unit tests covering set/unset/both-commands with injected exec doubles
- [x] `setup.integration.test.ts`: 1 new env-var case added to existing `installPlugins` describe block
- [x] 48 tests pass (0 fail) across setup unit + integration suites
- [x] Biome lint clean on all modified files


agent/src/setup.unit.test.ts
```diff
@@ -0,0 +1,89 @@
+/**
+ * Unit tests for agent/src/setup.ts — SHIPWRIGHT_LOCAL_MARKETPLACE seam
+ *
+ * Pure logic tests: no I/O, no real binaries, injected exec doubles only.
+ */
+
+import { afterEach, beforeEach, describe, expect, it } from "bun:test";
```
dmcaulay: `beforeEach` is imported but never used. Remove it:
```ts
import { afterEach, describe, expect, it } from "bun:test";
```

agent/src/setup.unit.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    // Both install and update should use the local path
+    const specs = calls.map((c) => c.args[2]);
+    expect(specs).toEqual([
+      "shipwright@/local/path",
+      "shipwright@/local/path",
+    ]);
+  });
+
+  it("uses GitHub marketplace slug when SHIPWRIGHT_LOCAL_MARKETPLACE is unset", async () => {
+    process.env.SHIPWRIGHT_LOCAL_MARKETPLACE = undefined;
```
dmcaulay: `process.env.X = undefined` coerces to the string `"undefined"` in Node.js, making the env var truthy and breaking the fallback logic. Bun deletes the key, which is why tests pass. Use the portable form:
```ts
delete process.env.SHIPWRIGHT_LOCAL_MARKETPLACE;
```
Apply the same fix in `afterEach` when restoring an originally-unset var.

dmcaulay (commented):
COMMENT — Two issues in the new unit test file before merge:

**1. Unused `beforeEach` import** (`setup.unit.test.ts:7`): `beforeEach` is imported but never called — remove it.

**2. Non-portable `= undefined` env-var reset** (`setup.unit.test.ts:42`): Assigning `undefined` to `process.env.SHIPWRIGHT_LOCAL_MARKETPLACE` works in Bun (deletes the key) but in Node.js coerces to the string `"undefined"`, which is truthy. `localMarketplace ?? plugin.marketplace` would then yield `"undefined"` instead of the fallback, silently breaking the 'unset' test path. Use `delete process.env.SHIPWRIGHT_LOCAL_MARKETPLACE` in both the inline test and `afterEach` restoration.

Production logic is correct — the `??` fallback and single-read pattern are clean. CI green.

agent/src/setup.integration.test.ts
```diff
@@ -629,4 +597,30 @@ describe("installPlugins", () => {
...
+      };
+
+      await installPlugins(mockExec, testHome, []);
+
+      // install and update should both use the local path
+      expect(calls[0].args).toEqual(["plugin", "install", "shipwright@/local/marketplace"]);
+      expect(calls[1].args).toEqual(["plugin", "update", "shipwright@/local/marketplace"]);
+    } finally {
+      // Restore env var regardless of test outcome
+      process.env.SHIPWRIGHT_LOCAL_MARKETPLACE = originalLocalMarketplace;
```
dodizzle: This `finally` block has the same non-portable env-var restore that was fixed in the unit test's `afterEach`. When `SHIPWRIGHT_LOCAL_MARKETPLACE` was unset before the test, `originalLocalMarketplace` is `undefined`, and `process.env.X = undefined` coerces to the string `"undefined"` rather than removing the key — leaving subsequent tests with a truthy string instead of the missing env var.

Apply the same guard used in `afterEach`:
```ts
finally {
  if (originalLocalMarketplace === undefined) {
    delete process.env.SHIPWRIGHT_LOCAL_MARKETPLACE;
  } else {
    process.env.SHIPWRIGHT_LOCAL_MARKETPLACE = originalLocalMarketplace;
  }
}
```

dodizzle (commented):
**COMMENT** — One finding from the prior review was addressed (unused `beforeEach` import removed, `afterEach` in unit test uses portable `delete` guard) but the same non-portable restore pattern remains in the integration test's `finally` block.

**Integration test `finally` still uses `= originalLocalMarketplace`** (`setup.integration.test.ts`, `finally` block): when `SHIPWRIGHT_LOCAL_MARKETPLACE` is unset at test time, `originalLocalMarketplace` is `undefined`, and assigning that to `process.env` coerces it to the string `"undefined"` rather than removing the key. Any code reading the env var after this test will see a truthy string instead of `undefined`, silently breaking the `??` fallback. The fix applied to `afterEach` in the unit test should be applied here too:
```ts
finally {
  if (originalLocalMarketplace === undefined) {
    delete process.env.SHIPWRIGHT_LOCAL_MARKETPLACE;
  } else {
    process.env.SHIPWRIGHT_LOCAL_MARKETPLACE = originalLocalMarketplace;
  }
}
```

Production logic in `setup.ts` is correct. Unit test suite is now clean. One-line fix and this is ready.

agent/src/setup.integration.test.ts
```diff
@@ -629,4 +629,30 @@ describe("installPlugins", () => {
...
```
dmcaulay: Non-portable env var restore: when `SHIPWRIGHT_LOCAL_MARKETPLACE` is unset at test time, `originalLocalMarketplace` is `undefined`, and assigning it to `process.env` coerces to the string `"undefined"` in Node.js-compatible runtimes. Any subsequent test reading this key will see a truthy string, silently breaking the `??` fallback.

Apply the same fix used in the unit test's `afterEach`:
```ts
finally {
  if (originalLocalMarketplace === undefined) {
    delete process.env.SHIPWRIGHT_LOCAL_MARKETPLACE;
  } else {
    process.env.SHIPWRIGHT_LOCAL_MARKETPLACE = originalLocalMarketplace;
  }
}
```

dmcaulay (commented):
COMMENT — One prior finding remains unaddressed: the integration test `finally` block still restores the env var with `process.env.SHIPWRIGHT_LOCAL_MARKETPLACE = originalLocalMarketplace`. When the var is unset at test time, `originalLocalMarketplace` is `undefined`, which Node.js-compatible runtimes coerce to the string `"undefined"` rather than removing the key — silently breaking the `??` fallback for any test that runs after.

The fix applied in `afterEach` of the unit test applies here too:
```ts
finally {
  if (originalLocalMarketplace === undefined) {
    delete process.env.SHIPWRIGHT_LOCAL_MARKETPLACE;
  } else {
    process.env.SHIPWRIGHT_LOCAL_MARKETPLACE = originalLocalMarketplace;
  }
}
```

Production logic in `setup.ts` is correct. Unit test suite is clean. One-line fix and this is ready.

dmcaulay (approved):
APPROVE — all prior findings resolved. The integration test `finally` block now uses the portable `if (originalLocalMarketplace === undefined) { delete ... }` guard, identical to the unit test's `afterEach`. Production logic in `setup.ts` is correct, CI is green, and the `SHIPWRIGHT_LOCAL_MARKETPLACE` seam is clean.

dodizzle (commented):
APPROVE — All prior findings addressed. The `finally` block in the integration test now uses the portable `delete`-vs-assign guard (matching the unit test's `afterEach`). Production logic is correct, tests cover both env-var paths, CI is green. Ready to merge.

PR #204, opened 2026-06-10
feat: add Claude Code and gh CLI to agent/Dockerfile (LPP-1.1)

dodizzle:
## Summary
- Add `nodejs`, `npm`, and the `gh` CLI to `agent/Dockerfile` base stage so the image ships both tools without relying on the host
- Install Claude Code globally (`npm install -g @anthropic-ai/claude-code`) in the runtime stage, placed after the `COPY --from=build` layers to preserve bun install cache
- Add `agent-docker-build` CI job to verify `docker build -f agent/Dockerfile` passes on every PR

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `docker build -f agent/Dockerfile .` produces image where `claude --version` and `gh --version` succeed | ✅ MET |
| 2 | Entrypoint boots without a `claude not found` warning | ✅ MET |
| 3 | No layer caching regressions — `npm install` runs after `bun install` layer | ✅ MET |
| 4 | CI docker build step passes | ✅ MET |

## Test Plan
- [x] New `agent-docker-build` CI job runs `docker build -f agent/Dockerfile -t shipwright-agent:ci .`
- [x] `bunx biome lint .` passes (259 files, no fixes)
- [x] Layer ordering: `bun install` in build stage → `npm install -g @anthropic-ai/claude-code` in runtime stage (after COPY lines)
- [x] gh CLI installed via official GitHub apt keyring (same pattern as root Dockerfile)

### DIFF
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -76,13 +76,13 @@ WORKDIR /app
 # Set AGENT_HOME — the PVC should be mounted here in Kubernetes
 ENV AGENT_HOME=/data/agent-home
 
-# Copy installed node_modules and generated Prisma client from build stage
+# Copy installed node_modules from build stage
 COPY --from=build /app/node_modules ./node_modules
 COPY --from=build /app/agent ./agent
 
-# Install Claude Code globally — placed after COPY layers so bun install
-# cache is not invalidated by changes to Claude Code version
-RUN npm install -g @anthropic-ai/claude-code
+# Install Claude Code globally — after COPY layers so a Claude Code version bump
+# does not invalidate the node_modules/agent copy in this stage
+RUN npm install -g @anthropic-ai/claude-code@2.1.170
 
 # Add scripts/bin to PATH so the git-credential-shipwright.sh wrapper
 # and gh wrapper are available to subprocesses
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -58,6 +58,7 @@ COPY package.json bunfig.toml* bun.lock ./
 COPY agent/package.json ./agent/
 COPY metrics/package.json ./metrics/
 COPY admin/package.json ./admin/
+COPY admin/prisma ./admin/prisma/
 COPY plugins/shipwright/package.json ./plugins/shipwright/
 
 # Install all dependencies (including devDependencies for typecheck)
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -24,14 +24,11 @@ FROM oven/bun:1.3 AS base
 #   - curl         — used by mise installer and gh keyring setup
 #   - ca-certificates — TLS trust for API calls
 #   - unzip        — needed by some mise-managed tools
-#   - nodejs npm   — required to install Claude Code via npm
 RUN apt-get update && apt-get install -y --no-install-recommends \
     git \
     curl \
     ca-certificates \
     unzip \
-    nodejs \
-    npm \
     && rm -rf /var/lib/apt/lists/*
 
 # Install gh CLI via official GitHub apt keyring
@@ -46,6 +43,11 @@ RUN curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg \
 # Install mise (polyglot runtime manager)
 RUN curl https://mise.run | sh && \
     ln -s /root/.local/bin/mise /usr/local/bin/mise
+
+# Install Node.js 20 LTS via mise (Debian bookworm's apt provides Node 18 which is EOL)
+# mise shims live in ~/.local/share/mise/shims — add to PATH so npm/node are found
+ENV PATH="/root/.local/share/mise/shims:${PATH}"
+RUN mise use --global node@20
 
 # ─── Build stage ──────────────────────────────────────────────────────────────
 
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -54,8 +54,11 @@ FROM base AS build
 WORKDIR /app
 
 # Copy workspace manifests first (layer caching for deps)
-COPY package.json bunfig.toml* ./
+COPY package.json bunfig.toml* bun.lock ./
 COPY agent/package.json ./agent/
+COPY metrics/package.json ./metrics/
+COPY admin/package.json ./admin/
+COPY plugins/shipwright/package.json ./plugins/shipwright/
 
 # Install all dependencies (including devDependencies for typecheck)
 RUN bun install --frozen-lockfile
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -56,16 +56,12 @@ WORKDIR /app
 # Copy workspace manifests first (layer caching for deps)
 COPY package.json bunfig.toml* ./
 COPY agent/package.json ./agent/
-COPY agent/prisma ./agent/prisma/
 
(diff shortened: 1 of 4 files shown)


agent/Dockerfile
```diff
@@ -42,19 +54,18 @@ FROM base AS build
 WORKDIR /app
 
 # Copy workspace manifests first (layer caching for deps)
-COPY package.json bunfig.toml* ./
+COPY package.json bunfig.toml* bun.lock ./
 COPY agent/package.json ./agent/
-COPY agent/prisma ./agent/prisma/
+COPY metrics/package.json ./metrics/
+COPY admin/package.json ./admin/
```
dmcaulay, line 62: `@shipwright/admin` has `"postinstall": "prisma generate --schema=prisma/schema.prisma"`. The schema file is not copied before `bun install --frozen-lockfile` runs, so the postinstall fails with `Could not load --schema from provided path prisma/schema.prisma: file or directory not found`. Add a COPY for the schema right after this line:
```dockerfile
COPY admin/prisma ./admin/prisma/
```

dmcaulay (commented):
COMMENT — CI-blocking issue before merge: `@shipwright/admin` postinstall runs `prisma generate --schema=prisma/schema.prisma`, but `admin/prisma/schema.prisma` is not copied before `RUN bun install --frozen-lockfile`. Add `COPY admin/prisma ./admin/prisma/` immediately after `COPY admin/package.json ./admin/`. All other changes are correct — gh CLI install, Claude Code placement, bun.lock addition, and layer ordering are all solid.

dmcaulay (approved):
APPROVE — all prior findings resolved. `admin/prisma` is now copied before `bun install` (fixing the postinstall failure), the stale `COPY agent/prisma` + `prisma generate` is removed (agent has no Prisma schema), and `bun.lock` + all workspace package manifests are in place for `--frozen-lockfile`. CI is green.

agent/Dockerfile
```diff
@@ -68,6 +80,10 @@ ENV AGENT_HOME=/data/agent-home
 COPY --from=build /app/node_modules ./node_modules
 COPY --from=build /app/agent ./agent
 
+# Install Claude Code globally — placed after COPY layers so bun install
+# cache is not invalidated by changes to Claude Code version
+RUN npm install -g @anthropic-ai/claude-code
```
dodizzle: No version pin on the global Claude Code install. `bun install --frozen-lockfile` is used earlier in this file to lock bun deps, but `npm install -g` has no lockfile equivalent — each rebuild may pull a different version. Consider pinning: `npm install -g @anthropic-ai/claude-code@<version>`

agent/Dockerfile
```diff
@@ -68,6 +80,10 @@ ENV AGENT_HOME=/data/agent-home
 COPY --from=build /app/node_modules ./node_modules
 COPY --from=build /app/agent ./agent
 
+# Install Claude Code globally — placed after COPY layers so bun install
+# cache is not invalidated by changes to Claude Code version
```
dodizzle: This comment's reasoning is off: `bun install` runs in the `build` stage, which is entirely separate from `runtime`. The actual benefit is that placing `npm install` *after* the `COPY --from=build` lines means a Claude Code version bump only invalidates the cache from this line forward, not the COPY layers above it.

dodizzle (commented):
COMMENT — One important issue before merge: `npm install -g @anthropic-ai/claude-code` installs the latest version at build time, while the rest of the Dockerfile uses `--frozen-lockfile` for reproducibility. Pin a specific version to avoid silent breakage on future rebuilds.

Two minor comment issues in `agent/Dockerfile`: (1) line 79's comment still references "generated Prisma client" even though the `bunx prisma generate` step was removed; (2) lines 83-84's layer-caching reasoning is incorrect — `bun install` runs in the separate `build` stage, not `runtime`.

The functional changes are correct: gh CLI installation, layer ordering, workspace manifest COPY expansion, and the new CI job are all solid.

dodizzle (commented):
APPROVE — All prior findings from the first review are addressed. Claude Code is now pinned to `@anthropic-ai/claude-code@2.1.170`, and both stale Dockerfile comments have been corrected. CI is green. No new issues.

dmcaulay (commented):
COMMENT — All three prior self-review findings are resolved and CI is green. One new issue: the Debian bookworm apt repository provides Node.js 18.19.x, which reached end-of-life in April 2025. `mise` is already present in the image — using it to install Node 20 LTS would avoid baking an EOL runtime into the production container.

agent/Dockerfile
```diff
@@ -21,14 +21,26 @@ FROM oven/bun:1.3 AS base
...
+#   - curl         — used by mise installer and gh keyring setup
 #   - ca-certificates — TLS trust for API calls
 #   - unzip        — needed by some mise-managed tools
+#   - nodejs npm   — required to install Claude Code via npm
 RUN apt-get update && apt-get install -y --no-install-recommends \
     git \
     curl \
     ca-certificates \
     unzip \
+    nodejs \
```
dmcaulay: `nodejs` from Debian bookworm's apt repository is version 18.19.x — Node.js 18 reached EOL in April 2025 and no longer receives security patches. `mise` is already in the image; you could replace this with `RUN mise use --global node@20` (or add a NodeSource 20.x apt source) to get a supported LTS runtime.

dodizzle (commented):
APPROVE — All prior findings resolved: EOL Node 18 fixed via `mise use --global node@20` (removes the Debian apt `nodejs`/`npm` packages), CI green. The `SHIPWRIGHT_LOCAL_MARKETPLACE` seam (merged from main) is well-tested with unit and integration tests following CLAUDE.md conventions.

**Suggestion** (`agent/src/setup.ts:189`): `localMarketplace` overrides the marketplace for every plugin in `allPlugins`, including future `agentPlugins` entries. If an agent-specific plugin ever lives in a different marketplace, this env var would silently redirect all installs to the same local path. Consider scoping to `DEFAULT_PLUGINS` only, or making the blanket override explicit in the var name. Non-blocking for the current single-plugin setup.

dodizzle:
All review findings have been addressed in the latest commits:

- **dmcaulay's Prisma schema finding** → fixed in `5f48db2` (COPY admin/prisma before bun install)
- **dmcaulay's Node 18 EOL finding** → fixed in `c18cfcd` (mise use --global node@20)
- **dodizzle's Claude Code version pin** → fixed in `778f109` (@anthropic-ai/claude-code@2.1.170)
- **dodizzle's stale Dockerfile comments** → fixed in `778f109`

CI is green across all checks. Requesting re-review from @dmcaulay since the last approval was on an older commit.

dodizzle (commented):
APPROVE — Prior suggestion addressed: `SHIPWRIGHT_LOCAL_MARKETPLACE` override is now scoped to `DEFAULT_PLUGINS` only via `resolvedDefaultPlugins`, so `agentPlugins` always use their own marketplace. Unit test added covering the isolation guarantee. CI green. No new issues.

dmcaulay (approved):
APPROVE — Both prior findings resolved. Node 20 LTS via mise is correct (shims PATH set via ENV before RUN, inherited by runtime stage). LOCAL_MARKETPLACE scoping is clean (spread map, no mutation, both install and update loops updated). CI green.

PR #205, opened 2026-06-10
feat: add bearer token auth to admin CRUD middleware (ADMD-3.1)

dmcaulay:
## Summary
- Adds `api-auth.ts` — combined admin auth middleware that accepts either a session cookie or a bearer token
- Updates `admin-api.ts` to use the new combined middleware (previously session-cookie-only)
- Adds unit tests for bearer middleware and smoke tests for bearer auth on CRUD routes

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | AdminToken Prisma model added; migration runs clean | MET |
| 2 | POST /admin/api/tokens creates token with hashed storage, returns raw token once | MET |
| 3 | Bearer token accepted on all admin CRUD endpoints alongside session cookie auth | MET |
| 4 | Token management UI section on agents page (create, list, revoke) | MET |
| 5 | Copied crypto/auth helpers have no vitals-os-specific dependencies | MET |
| 6 | Unit tests cover token create/validate/revoke and bearer middleware | MET |

## Test Plan
- [x] `api-auth.unit.test.ts` — 167 lines covering bearer valid/invalid/revoked/malformed, session valid/invalid/missing
- [x] `admin-api.smoke.test.ts` — bearer auth paths on CRUD routes (70 lines added)
- [x] 264 tests pass, 0 fail (`bun test admin/`)
- [x] `api-auth.ts` coverage: 100% functions, 97.50% lines
- [x] Biome lint clean

### DIFF
--- a/admin/src/api-auth.ts
+++ b/admin/src/api-auth.ts
@@ -17,6 +17,9 @@ import type { AgentTokenService } from "./agent-tokens.ts";
 
 const SESSION_COOKIE = "admin_session";
 
+// Matches /admin/api/agents/{agentId}/... routes — used for per-agent scope enforcement.
+const AGENT_ROUTE_RE = /^\/admin\/api\/agents\/([^/]+)/;
+
 // ─── Middleware factory ───────────────────────────────────────────────────────
 
 /**
@@ -46,12 +49,23 @@ export function createAdminAuthMiddleware(deps: {
       // Authorization header is present — bearer token path.
       // Do NOT fall through to cookie check on failure.
       if (!authHeader.startsWith("Bearer ")) {
-        return c.json({ error: "Unauthorized" }, 401);
+        return c.json({ error: "Unauthorized" }, 401, {
+          "WWW-Authenticate": "Bearer",
+        });
       }
       const raw = authHeader.slice(7);
       const result = await agentTokenService.validate(raw);
       if (!result) {
-        return c.json({ error: "Unauthorized" }, 401);
+        return c.json({ error: "Unauthorized" }, 401, {
+          "WWW-Authenticate": 'Bearer error="invalid_token"',
+        });
+      }
+      // Enforce per-agent scope: token must belong to the agent being accessed.
+      // c.req.param() is not available in global middleware (params resolve only at
+      // handler dispatch), so we extract the agent ID directly from the URL path.
+      const match = AGENT_ROUTE_RE.exec(c.req.path);
+      if (match && result.agentId !== match[1]) {
+        return c.json({ error: "Forbidden" }, 403);
       }
       return next();
     }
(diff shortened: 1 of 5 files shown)


admin/src/api-auth.ts
```diff
@@ -0,0 +1,83 @@
...
+      // Do NOT fall through to cookie check on failure.
+      if (!authHeader.startsWith("Bearer ")) {
+        return c.json({ error: "Unauthorized" }, 401);
+      }
+      const raw = authHeader.slice(7);
+      const result = await agentTokenService.validate(raw);
+      if (!result) {
+        return c.json({ error: "Unauthorized" }, 401);
+      }
+      return next();
```
dmcaulay, line 70: Bearer token validated but `result.agentId` is discarded — any valid token authenticates requests for any agent in the URL. Compare to `c.req.param('id')` before calling `next()`, or document that tokens are intentionally global admin credentials.

dmcaulay (commented):
COMMENT — Bearer token's `agentId` is discarded after validation; any valid token authenticates requests for any agent.

After `agentTokenService.validate(raw)` returns non-null, the middleware calls `next()` unconditionally without comparing `result.agentId` to the `:id` route parameter. A token minted for agent-A can read/write agent-B's env vars, cron jobs, tools, and plugins. Session cookies intentionally have no per-agent scope, but tokens are stored per-agent — the mismatch between the storage model and enforcement is the gap.

Fix: either (a) enforce scope: `if (result.agentId !== c.req.param('id')) return c.json({ error: 'Forbidden' }, 403)`, or (b) explicitly document tokens as global admin credentials and drop the per-agent foreign key.

Suggestion (non-blocking): RFC 6750 §3 requires `WWW-Authenticate: Bearer` on 401 responses from Bearer-protected endpoints. All rejection paths currently return bare 401 JSON without this header.

Fallthrough logic, DI design, and test coverage are solid.

dmcaulay (commented):
APPROVE — both prior findings fully addressed.

**Prior Findings Resolution:**

| Finding | Status | Evidence |
|---------|--------|----------|
| Bearer token not scoped to agent in URL | Addressed | `AGENT_ROUTE_RE` regex extracts `:id` from path; returns 403 when `result.agentId !== match[1]`. Three scope enforcement unit tests added (match → 200, mismatch → 403, unscoped route → 200). |
| WWW-Authenticate header missing on 401 | Addressed | Both rejection paths now return `WWW-Authenticate: Bearer` / `Bearer error="invalid_token"` headers. Unit tests verify header presence. |

The scope enforcement approach is correct: using a direct regex against `c.req.path` rather than `c.req.param()` is the right call in global middleware — Hono only populates route params at handler dispatch. The `if (match && ...)` guard also correctly passes tokens on unscoped routes.

CI green. Clean.

PR #221, opened 2026-06-10
feat(agent): expand config.ts with slack/voice/alerts/owner sections (AVP-1.2)

dmcaulay:
## Summary
- Add `slack`, `alerts`, `owner`, and `voice` config sections to `agent/src/config.ts`, matching the Vitals OS config structure
- All env vars follow SHIPWRIGHT_* namespacing; SLACK_* and third-party keys (GROQ, ELEVENLABS, WHISPER) keep conventional names per CLAUDE.md
- AGENT_HOME default (`~/.shipwright-agent/`) confirmed at call site in `entrypoint-main.ts` — no change needed

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Config has all sections: claude, shipwright, slack, alerts, owner, voice, paths | ✓ MET |
| 2 | AGENT_HOME defaults to ~/.shipwright-agent/ (not ~/.vitals-agent/) | ✓ MET |
| 3 | All env var names follow SHIPWRIGHT_* namespacing (SLACK_* vars keep conventional names) | ✓ MET |
| 4 | config.unit.test.ts covers new sections with set/unset pairs for each field | ✓ MET |
| 5 | bun test agent passes clean | ✓ MET |

## Test Plan
- 35 unit tests pass covering all new config fields (set + undefined-when-unset for each)
- 100% line coverage on `agent/src/config.ts`
- Full agent suite: 588 pass, 0 fail
- [x] Pre-commit checks passing (biome lint clean)

### DIFF
--- a/agent/src/shipwright-runtime-client.ts
+++ b/agent/src/shipwright-runtime-client.ts
@@ -1,0 +1,103 @@
+/**
+ * agent/src/shipwright-runtime-client.ts
+ *
+ * Unified runtime client for the three agent-facing Shipwright API methods.
+ *
+ * - ShipwrightClientError — typed error with statusCode
+ * - ShipwrightRuntimeClient — interface for DI
+ * - HttpShipwrightRuntimeClient — real HTTP implementation with injected fetchFn
+ */
+
+import type { AgentConfigResponse, AgentCronJob } from "@shipwright/admin";
+
+// ─── Error ────────────────────────────────────────────────────────────────────
+
+export class ShipwrightClientError extends Error {
+  constructor(
+    public readonly statusCode: number,
+    message: string,
+  ) {
+    super(message);
+    this.name = "ShipwrightClientError";
+  }
+}
+
+// ─── Interface ────────────────────────────────────────────────────────────────
+
+export interface ShipwrightRuntimeClient {
+  getAgentConfigBundle(agentId: string): Promise<AgentConfigResponse>;
+  listAgentCronJobs(agentId: string): Promise<AgentCronJob[]>;
+  reconcileSystemCrons(agentId: string): Promise<void>;
+}
+
+// ─── HttpShipwrightRuntimeClient ──────────────────────────────────────────────
+
+export class HttpShipwrightRuntimeClient implements ShipwrightRuntimeClient {
+  private readonly apiUrl: string;
+  private readonly apiKey: string;
+  private readonly fetchFn: typeof fetch;
+
+  constructor(opts: {
+    apiUrl: string;
+    apiKey: string;
+    fetchFn?: typeof fetch;
+  }) {
+    this.apiUrl = opts.apiUrl;
+    this.apiKey = opts.apiKey;
+    this.fetchFn = opts.fetchFn ?? fetch;
+  }
+
+  private get authHeaders(): Record<string, string> {
+    return {
+      Authorization: `Bearer ${this.apiKey}`,
+      "Content-Type": "application/json",
+    };
+  }
+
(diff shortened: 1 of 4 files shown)


dmcaulay (commented):
APPROVE — clean additive change. All acceptance criteria met: all six config sections present (claude, shipwright, slack, alerts, owner, voice, paths), AGENT_HOME default confirmed at call site, SHIPWRIGHT_* namespacing clean, 35 new unit tests covering every field with set/unset pairs. CI green. No issues.

agent/src/shipwright-runtime-client.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  async reconcileSystemCrons(agentId: string): Promise<void> {
+    const url = `${this.apiUrl}/admin/api/agents/${encodeURIComponent(agentId)}/crons/reconcile`;
+    const response = await this.fetchFn(url, {
+      method: "POST",
+      headers: this.authHeaders,
+    });
+
+    if (!response.ok) {
```
dodizzle, line 101: `/admin/api/` prefix hardcoded here but the other two methods use `/agents/...` against the same `apiUrl`. If the URL is ever scoped to an agent-facing-only endpoint, this silently 404s. Consider either a separate `adminApiUrl` param or moving this method to `HttpShipwrightAdminClient`.

agent/src/shipwright-runtime-client.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  private get authHeaders(): Record<string, string> {
+    return {
+      Authorization: `Bearer ${this.apiKey}`,
+      "Content-Type": "application/json",
+    };
+  }
+
+  async getAgentConfigBundle(agentId: string): Promise<AgentConfigResponse> {
```
dodizzle, line 62: `HttpShipwrightConfigClient` in `shipwright-config-client.ts` already calls this exact endpoint with the same logic. `entrypoint-main.ts` still uses the old client — no callsites exist for this method. Either deprecate the old client and migrate the callsite, or remove this method and keep the two clients with separate concerns.

dodizzle (commented):
COMMENT — two issues in the new `HttpShipwrightRuntimeClient` to resolve before wiring it into the entrypoint.

**1. `reconcileSystemCrons` mixes admin-tier path into the agent-facing URL base** (`shipwright-runtime-client.ts:101`): the other two methods call `${this.apiUrl}/agents/...`, but `reconcileSystemCrons` hardcodes `/admin/api/` against the same `apiUrl`. Works today since `SHIPWRIGHT_API_URL` is the unified admin service, but `ShipwrightRuntimeClient` gives callers no indication that one method hits a different namespace — and a scope-limited URL would silently 404. Either accept a separate `adminApiUrl` for reconcile, or move `reconcileSystemCrons` to `HttpShipwrightAdminClient` where the other `/admin/api/` calls already live.

**2. `getAgentConfigBundle` duplicates `HttpShipwrightConfigClient`** (`shipwright-runtime-client.ts:62`): identical endpoint, divergent error handling (`Error` vs `ShipwrightClientError`), no callsites for the new method — `entrypoint-main.ts` still wires the old client. Without a migration or explicit deprecation, callers won't know which to use and the two will drift. Remove `getAgentConfigBundle` from this PR and migrate in a follow-up, or deprecate the old client here.

---

Config additions are clean — env var naming is correct per CLAUDE.md, all 35 unit tests follow set/unset conventions, and the integration tests use injected `FetchFn` doubles with no global overrides. No issues there.

PR #242, opened 2026-06-10
docs(task-store): add docs/task-store.md for all three backends (JTS-2.2)

dodizzle:
## Summary
- Adds `docs/task-store.md` as the authoritative reference for all three task store backends (json, github, jira)
- Jira section documents all env vars, full config example, default status map (all 13 entries), custom `statusMap`/`readyJql` overrides, and troubleshooting for common auth errors
- Includes quick-start sections for each backend so a developer can go from zero to `task_store.ts doctor` passing

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | docs/task-store.md exists with sections for all three backends | ✅ MET |
| 2 | Jira section covers env vars, full config, status map table, setup command | ✅ MET |
| 3 | Developer can follow docs to run `task_store.ts doctor` against real Jira | ✅ MET |
| 4 | No test changes — docs only | ✅ MET |

## Test Plan
- [x] Lint passing (biome: 0 issues)
- [x] Docs verified against source: env vars, status map, config keys all cross-checked against `jira.ts` and `store.ts`

### DIFF
--- a/docs/task-store.md
+++ b/docs/task-store.md
@@ -1,0 +1,295 @@
+# Task Store
+
+The Shipwright task store is the backing database for the plan-execute-review loop. It holds all tasks, their statuses, dependencies, and metadata. Three backends are available:
+
+| Backend | Where tasks live | Best for |
+|---------|-----------------|---------|
+| `json` | `state/todos.json` (local file) | Local development, offline use |
+| `github` | GitHub Issues in a repo | Team collaboration with GitHub as the source of truth |
+| `jira` | Jira project issues | Teams already using Jira for project tracking |
+
+Config is resolved at startup using a three-step discovery chain:
+
+1. Walk up from `cwd` looking for `.shipwright.json`
+2. Fall back to the `SHIPWRIGHT_CONFIG` env var (path to a JSON config file)
+3. Default to the JSON backend (no config needed)
+
+> **Note for agent deployments:** Standard agent deployments configure the task store via env vars managed through the admin service — `.shipwright.json` is a local fallback for running the plugin outside the agent (e.g. manual CLI use or local development). You do not need a `.shipwright.json` for a production agent.
+
+---
+
+## JSON backend (default)
+
+The JSON backend requires no configuration. When no `.shipwright.json` is found and `SHIPWRIGHT_CONFIG` is not set, Shipwright automatically uses `state/todos.json` in the process working directory.
+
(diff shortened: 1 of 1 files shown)


docs/task-store.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Default JQL query
+
+When `readyJql` is not set, the adapter fetches issues using:
+
+```
+project = "PROJECT_KEY" AND labels = "shipwright-session" ORDER BY created ASC
+```
+
+The `readyJql` field overrides this entire query. Use it to narrow scope (e.g. a specific sprint or fix version), add status filters, or change the sort order.
```
dmcaulay, line 208: Minor accuracy issue: the default JQL has no status clause — status filtering is applied client-side in `query()` after the fetch. This matters because a `readyJql` without a status filter will cause ALL `shipwright-session` issues (regardless of status) to be fetched and filtered in memory. For large Jira instances this can be expensive. Consider adding: "If you provide `readyJql`, include explicit status clauses — the default has no status filter."

dmcaulay (commented):
COMMENT — 1 finding (important): `readyJql` is described as overriding a "status-based query" but the default JQL has no status filter (`project = "..." AND labels = "shipwright-session" ORDER BY created ASC` — status is applied client-side). A `readyJql` without explicit status clauses will fetch all project issues tagged with `shipwright-session`, which can be a large set on real Jira instances. The description should warn that callers must add status filters themselves when overriding. Everything else cross-checks cleanly against source: the default status map (13 entries), the default JQL, config keys, and troubleshooting cases all match `jira.ts` exactly.

dodizzle (commented):
APPROVE — Docs verified against source: all 13 default Jira status map entries match `jira.ts` exactly, error messages in Troubleshooting match `setup()` error strings exactly, config discovery chain matches `loadConfig()` in `create-task-store.ts`, and ADF `codeBlock` storage claim matches `jira.ts:101`. Quick starts are accurate. No issues.

dmcaulay (approved):
APPROVE — Prior finding addressed: `readyJql` now includes a clear warning that callers must add explicit status clauses, with an example. All source cross-checks clean: default JQL, 13 default status map entries, ADF codeBlock metadata storage, troubleshooting error messages, and config discovery chain all match source exactly. CI passing.

PR #255, opened 2026-06-10
feat(admin): add admin API key parsing to auth middleware (VOP-1.1)

dmcaulay:
## Summary
- Added `parseAdminApiKeys()` to `admin/src/api-auth.ts` — parses `SHIPWRIGHT_ADMIN_API_KEYS` env var (comma-separated `name:token:scope` tuples) into `Map<token, AdminApiKey>`, same logic as Vitals OS `parseApiKeys`
- Updated `createAdminAuthMiddleware` to check env API keys before falling through to `AgentTokenService` DB validation — `scope=*` keys bypass all scope enforcement; `scope=<agentId>` keys enforce route-level agent ID matching
- Threaded `SHIPWRIGHT_ADMIN_API_KEYS` from `main.ts` env through `AdminDeps` → `createAdminApp` → `createAdminAuthMiddleware`

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | SHIPWRIGHT_ADMIN_API_KEYS parsed into Map<token,{name,scope}> using same logic as Vitals OS parseApiKeys | ✅ MET |
| 2 | Bearer matching scope=* key sets isAdmin=true and bypasses all per-agent scope enforcement | ✅ MET |
| 3 | Bearer matching scope=<agentId> key enforces route agentId matches scope value | ✅ MET |
| 4 | Existing AgentTokenService DB token path unchanged and still works when no env key matches | ✅ MET |
| 5 | Unit tests: admin key accepted, admin bypasses scope check, scoped key enforces agentId, invalid bearer 401, absent env var is a no-op | ✅ MET |

## Test Plan
- `bun test admin/src/api-auth.unit.test.ts` — 28 tests pass (100% line coverage on api-auth.ts)
- `bun test admin/src/ --testPathPattern "unit"` — 1345 tests pass across 60 files
- `bunx biome lint .` — clean
- `bun run --filter='*' typecheck` — all 4 packages pass

### DIFF
--- a/admin/src/api-auth.ts
+++ b/admin/src/api-auth.ts
@@ -117,6 +117,8 @@ export function createAdminAuthMiddleware(deps: {
           if (match && envKey.scope !== match[1]) {
             return c.json({ error: "Forbidden" }, 403);
           }
+          // Non-agent routes: scoped keys are permitted (no agentId to enforce against).
+          // All current /admin/api/* routes match AGENT_ROUTE_RE — revisit if that changes.
           return next();
         }
       }
--- a/admin/src/api-auth.ts
+++ b/admin/src/api-auth.ts
@@ -20,16 +20,64 @@ const SESSION_COOKIE = "admin_session";
 // Matches /admin/api/agents/{agentId}/... routes — used for per-agent scope enforcement.
 const AGENT_ROUTE_RE = /^\/admin\/api\/agents\/([^/]+)/;
 
+// ─── Admin API key parsing ────────────────────────────────────────────────────
+
+/**
+ * Represents a parsed admin API key entry.
+ * scope === "*" → admin (bypasses all per-agent scope checks)
+ * scope === "<agentId>" → scoped to a single agent
+ */
+export interface AdminApiKey {
+  name: string;
+  scope: string;
+}
+
+/**
+ * Parse the SHIPWRIGHT_ADMIN_API_KEYS env var into a Map<token, AdminApiKey>.
+ * Format: comma-separated "name:token:scope" tuples.
+ * Example: "bodhi:sk_bodhi_abc:*,svc:sk_svc_xyz:agent-id-123"
+ *
+ * Tokens may contain embedded colons — first segment is name, last is scope,
+ * everything in between is the token.
+ *
+ * Returns an empty map if the env var is missing or empty.
+ */
+export function parseAdminApiKeys(
+  envStr: string | undefined,
+): Map<string, AdminApiKey> {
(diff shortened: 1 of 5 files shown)


admin/src/api-auth.ts
```diff
@@ -54,6 +103,25 @@ export function createAdminAuthMiddleware(deps: {
...
+          if (envKey.scope === "*") {
+            // Admin key — bypass all scope checks.
+            return next();
+          }
+          // Scoped key — enforce route agentId matches key scope.
+          const match = AGENT_ROUTE_RE.exec(c.req.path);
+          if (match && envKey.scope !== match[1]) {
+            return c.json({ error: "Forbidden" }, 403);
+          }
+          return next();
```
dodizzle, line 122: Minor: when a scoped key (`scope !== "*"`) is used on a non-agent route (path doesn't match `AGENT_ROUTE_RE`), scope enforcement is silently skipped and the request is allowed. All current `/admin/api/*` routes are agent-scoped so there's no exposure today, but if non-agent admin routes are added later this will silently over-permit. A one-line comment here would flag this for the next reader:
```ts
// Non-agent routes: scoped keys are permitted (no agentId to enforce against).
// All current /admin/api/* routes match AGENT_ROUTE_RE — revisit if that changes.
return next();
```
Not a blocker.

dodizzle (approved):
APPROVE — Clean implementation of env-based admin API key auth. All acceptance criteria met. `parseAdminApiKeys` handles embedded colons and malformed entries correctly; middleware checks env keys before DB with a proper guard to skip the lookup when unconfigured. Test coverage is thorough (empty/undefined, multi-key, embedded colons, scope=* bypass, scoped enforcement, 401 fallback, no-op). One minor note: scoped keys on non-agent routes silently allow access — no current routes are affected, but worth a comment if non-agent admin routes are added later.

dmcaulay (commented):
APPROVE — all 5 acceptance criteria met, 28 unit tests (admin bypass, scoped enforcement, invalid bearer 401, no-op when absent), env key check before DB round-trip, optional adminApiKeys is a clean no-op for existing deployments. Two sub-threshold suggestions: (1) scoped key passes on non-agent-prefixed routes (no practical impact today, latent footgun), (2) Map.get() is not constant-time (HTTP timing attacks are hard to exploit on this surface). No blocking issues.

dodizzle (approved):
**APPROVE** — Prior suggestion addressed: `4396269a` adds the scoped-key-on-non-agent-route comment as suggested. CI green, two approvals already. Clean, merge-ready.

Issue #42, opened 2026-06-07
SHE-2.1: Agent runtime API


dodizzle:
Hono app mounted at /agents/*. Implement GET /agents/:id/config (returns decrypted env + allowedTools + plugin list) and GET /agents/:id/crons. Auth via SHIPWRIGHT_INTERNAL_API_KEY bearer token. This is the endpoint the harness polls every 60s.

## Acceptance Criteria
- [ ] GET /agents/:id/config returns { env: Record<string, string>, allowedTools: string[], plugins: {marketplace, plugin}[] } with all values decrypted
- [ ] Returns 401 when SHIPWRIGHT_INTERNAL_API_KEY header is missing or wrong
- [ ] Returns 404 for unknown agent ID
- [ ] Responds in <100ms on local Postgres
- [ ] Test decision: smoke tests via app.request() for 200, 401, and 404 cases; no existing tests retired

```shipwright
{
  "id": "SHE-2.1",
  "title": "Agent runtime API",
  "status": "merged",
  "source": "gh:app-vitals/shipwright#38",
  "session": "shipwright-harness-extraction",
  "repo": "shipwright",
  "description": "Hono app mounted at /agents/*. Implement GET /agents/:id/config (returns decrypted env + allowedTools + plugin list) and GET /agents/:id/crons. Auth via SHIPWRIGHT_INTERNAL_API_KEY bearer token. This is the endpoint the harness polls every 60s.",
  "acceptanceCriteria": [
    "GET /agents/:id/config returns { env: Record<string, string>, allowedTools: string[], plugins: {marketplace, plugin}[] } with all values decrypted",
    "Returns 401 when SHIPWRIGHT_INTERNAL_API_KEY header is missing or wrong",
    "Returns 404 for unknown agent ID",
    "Responds in <100ms on local Postgres",
    "Test decision: smoke tests via app.request() for 200, 401, and 404 cases; no existing tests retired"
  ],
  "layer": "API",
  "branch": "feat/she-2-1-agent-runtime-api",
  "dependencies": [
    "SHE-1.2"
  ],
  "pr": 78,
  "addedAt": "2026-06-07T04:10:00Z",
  "hours": 3,
  "startedAt": "2026-06-07T18:03:17Z",
  "prCreatedAt": "2026-06-07T18:13:12Z"
}
```

PR #256, opened 2026-06-10
fix(plugin): sync vitals-os bug fixes — assignee fallback + multi-repo review keying (v4.26.7)

dmcaulay:
## Summary

- **`adapters/github.ts`**: fall back to GitHub issue assignee when the YAML block has no `assignee` field; adds `assignees` to the `gh issue list --json` field list and populates `task.assignee` from the first assignee in `issueToTask()`. Fixes `--assignee` filtering silently dropping tasks owned via GitHub assignment rather than YAML.
- **`check-review.ts`**: key the `reviewByPr` map on `"repo:pr-number"` instead of bare PR number. Bare number causes cross-repo collisions when multiple repos are configured (two repos can share the same PR number), causing one repo's review state to shadow the other's.
- Version bumped to **4.26.7** (patch).

Backported from `app-vitals/vitals-os` plugin v4.26.7 (PRs #1452, #1453, and the check-review fix).

## Test plan

- [x] `bun test plugins/shipwright/scripts/adapters/github.unit.test.ts plugins/shipwright/scripts/check-review.test.ts` — 90 pass, 0 fail
- [x] `bunx biome check` — clean on changed files

### DIFF
--- a/plugins/shipwright/scripts/check-review.ts
+++ b/plugins/shipwright/scripts/check-review.ts
@@ -95,12 +95,10 @@ export async function run(deps: Deps): Promise<RunResult> {
   const currentUser = await deps.getCurrentUser();
   const reviews = deps.readReviews();
 
-  // Build a lookup from "repo:pr-number" → review entry.
-  // Keying on number alone causes cross-repo collisions when multiple repos
-  // are configured — two repos can share PR numbers. Include the repo in the key.
-  const reviewByPr = new Map<string, ReviewEntry>();
+  // Build a lookup from PR number → review entry
+  const reviewByPr = new Map<number, ReviewEntry>();
   for (const entry of reviews) {
-    reviewByPr.set(`${entry.repo ?? ""}:${entry.pr}`, entry);
+    reviewByPr.set(entry.pr, entry);
   }
 
   // We don't know the repo here without a context — use a placeholder for tests.
@@ -110,7 +108,7 @@ export async function run(deps: Deps): Promise<RunResult> {
   for (const pr of prs) {
     if (!deps.isSelfReviewAllowed && pr.author.login === currentUser) continue;
 
-    const entry = reviewByPr.get(`${pr.repo ?? ""}:${pr.number}`);
+    const entry = reviewByPr.get(pr.number);
 
     // No entry → eligible
     if (!entry) {
--- a/plugins/shipwright/scripts/check-review.ts
+++ b/plugins/shipwright/scripts/check-review.ts
@@ -95,10 +95,12 @@ export async function run(deps: Deps): Promise<RunResult> {
   const currentUser = await deps.getCurrentUser();
   const reviews = deps.readReviews();
 
-  // Build a lookup from PR number → review entry
-  const reviewByPr = new Map<number, ReviewEntry>();
+  // Build a lookup from "repo:pr-number" → review entry.
+  // Keying on number alone causes cross-repo collisions when multiple repos
+  // are configured — two repos can share PR numbers. Include the repo in the key.
+  const reviewByPr = new Map<string, ReviewEntry>();
   for (const entry of reviews) {
-    reviewByPr.set(entry.pr, entry);
+    reviewByPr.set(`${entry.repo ?? ""}:${entry.pr}`, entry);
   }
 
   // We don't know the repo here without a context — use a placeholder for tests.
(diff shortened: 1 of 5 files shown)


plugins/shipwright/scripts/check-review.ts
```diff
@@ -95,12 +95,10 @@ export async function run(deps: Deps): Promise<RunResult> {
...
 
-  // Build a lookup from "repo:pr-number" → review entry.
-  // Keying on number alone causes cross-repo collisions when multiple repos
-  // are configured — two repos can share PR numbers. Include the repo in the key.
-  const reviewByPr = new Map<string, ReviewEntry>();
+  // Build a lookup from PR number → review entry
+  const reviewByPr = new Map<number, ReviewEntry>();
   for (const entry of reviews) {
-    reviewByPr.set(`${entry.repo ?? ""}:${entry.pr}`, entry);
+    reviewByPr.set(entry.pr, entry);
```
dodizzle: This reverts the `repo:pr-number` fix that's already on `main`. `main` currently uses `Map<string, ReviewEntry>` keyed on `` `${entry.repo ?? ""}:${entry.pr}` `` to prevent cross-repo collisions. This PR switches back to bare `Map<number, ReviewEntry>`, which re-introduces the bug the description claims to be fixing.

The `check-review.test.ts` multi-repo collision test (line 311) still passes by coincidence — the SHA mismatch (`"sha-A"` vs `"sha-B"`) happens to route through the merge-only check path and return the right exit code for the wrong reason. The actual collision scenario (same SHA across repos) is not covered.

Suggestion: Drop the `check-review.ts` hunk from this PR entirely — `main` is already correct.

dodizzle (commented):
COMMENT — The assignee fallback in `adapters/github.ts` is correct and well-tested. However, the `check-review.ts` change reverts the multi-repo `repo:pr-number` keying that `main` already has right, re-introducing a cross-repo collision bug. The PR description says it adds the fix — the diff does the opposite. Drop or reverse the `check-review.ts` change before merge.

dodizzle (approved):
**APPROVE** — Both fixes are correct and clean. The repo:pr-number keying regression flagged in the prior review is fixed in `65abe49d` (restores `Map<string, ReviewEntry>` with `` `${entry.repo ?? ""}:${entry.pr}` `` keying). The optional chaining guard on `assignees?.[0]` is in place. Tests are thorough. Good to merge once CI goes green.

plugins/shipwright/scripts/check-review.ts
```diff
@@ -95,12 +95,10 @@ export async function run(deps: Deps): Promise<RunResult> {
   const currentUser = await deps.getCurrentUser();
   const reviews = deps.readReviews();
 
-  // Build a lookup from "repo:pr-number" → review entry.
-  // Keying on number alone causes cross-repo collisions when multiple repos
-  // are configured — two repos can share PR numbers. Include the repo in the key.
-  const reviewByPr = new Map<string, ReviewEntry>();
+  // Build a lookup from PR number → review entry
+  const reviewByPr = new Map<number, ReviewEntry>();
```
dmcaulay: This regresses the `repo:pr-number` composite key that `main` already has. `main` uses `Map<string, ReviewEntry>` keyed on `` `${entry.repo ?? ''}:${entry.pr}` `` — this PR replaces it with bare `Map<number, ReviewEntry>`. When two repos each have PR #42 open, the second lookup collides with the first repo's entry, silently suppressing or misattributing review state. Drop this hunk entirely — `main` is already correct.

dmcaulay (commented):
COMMENT — The `check-review.ts` change reverts the `repo:pr-number` composite key that `main` already has correct, re-introducing the cross-repo collision bug. The commit message is inverted vs the diff. Drop the `check-review.ts` hunk entirely — keep only the `github.ts` assignee fallback (which is correct and well-tested).

**check-review.ts regression** (`plugins/shipwright/scripts/check-review.ts:99`): `main` has `Map<string, ReviewEntry>` keyed on `${entry.repo ?? ''}:${entry.pr}`. This PR replaces it with `Map<number, ReviewEntry>` keyed on bare `entry.pr`. When two repos both have PR #42 open, the second repo's entry shadows the first — causing missed or incorrect review triggers. The correct code is already on `main`.

**The multi-repo test doesn't catch this** (`check-review.test.ts:322`): prB has `headRefOid: 'sha-B'` which differs from repo-A's `lastReviewedCommit: 'sha-A'`, so the regressed code still returns exit 0 via sha-mismatch path (not via "no entry found"). To make this test a regression sentinel, change prB's `headRefOid` to `'sha-A'` — then bare-number keying suppresses prB (SHA match on wrong entry → skip → exit 1), while composite keying correctly finds no entry for repo-b and returns exit 0.

PR #258, opened 2026-06-10
fix(admin): navbar highlight reflects current path

dodizzle:
## Summary
- `renderAdminToolbar` was hardcoding `active` on the Agents link — regardless of which page you were on
- Added an `activePath: string` param and apply the class via `startsWith` prefix match
- Updated all 5 call sites: agents pages pass `/admin/agents`, provision pages pass `/admin/provision`

## Test plan
- [ ] Navigate to `/admin/agents` → Agents link is highlighted
- [ ] Click into an agent detail page → Agents link stays highlighted
- [ ] Navigate to `/admin/provision` (any step) → Provision link is highlighted

### DIFF
--- a/admin/src/admin-ui-styles.ts
+++ b/admin/src/admin-ui-styles.ts
@@ -367,12 +367,14 @@ export function escapeHtml(str: string): string {
     .replace(/"/g, "&quot;");
 }
 
-export function renderAdminToolbar(userName: string): string {
+export function renderAdminToolbar(userName: string, activePath: string): string {
+  const active = (prefix: string) =>
+    activePath.startsWith(prefix) ? " active" : "";
   return `<nav class="vos-toolbar" aria-label="Site navigation">
     <a href="/admin/agents" class="vos-wordmark">Shipwright Admin</a>
     <div class="vos-nav">
-      <a href="/admin/agents" class="vos-nav-link active">Agents</a>
-      <a href="/admin/provision" class="vos-nav-link">Provision</a>
+      <a href="/admin/agents" class="vos-nav-link${active("/admin/agents")}">Agents</a>
+      <a href="/admin/provision" class="vos-nav-link${active("/admin/provision")}">Provision</a>
     </div>
     <div class="vos-user">
       <span class="vos-username">${escapeHtml(userName)}</span>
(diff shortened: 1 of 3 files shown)


admin/src/admin-ui-styles.ts
```diff
@@ -367,12 +367,14 @@ export function escapeHtml(str: string): string {
     .replace(/"/g, "&quot;");
 }
 
-export function renderAdminToolbar(userName: string): string {
+export function renderAdminToolbar(userName: string, activePath: string): string {
+  const active = (prefix: string) =>
+    activePath.startsWith(prefix) ? " active" : "";
```
dmcaulay, line 372: The active-state logic here is pure string computation — exactly what `admin-ui-pages.unit.test.ts` is designed to cover. CLAUDE.md: "Tests land with the code, at the correct layer — same PR, no add tests later tasks."

In the existing describe blocks for each render function, add:
```typescript
expect(html).toContain('href="/admin/agents" class="vos-nav-link active"');
```
(and the inverse — that Provision is not active on agent pages). A sub-path case (`/admin/agents/agent-id`) is also worth a test since that's the main motivation for `startsWith` over strict equality.

dmcaulay (commented):
COMMENT — the active-state fix is correct and all 5 call sites are updated, but per CLAUDE.md's test policy tests must land in the same PR. `admin-ui-pages.unit.test.ts` already imports and exercises all the render functions with string assertions — adding active-class coverage is a one-liner per describe block. Also worth considering: making `activePath` optional with a default of `""` to reduce merge-fix friction on in-flight branches.

dodizzle:
Both findings addressed:
- `activePath` now defaults to `""` — no changes needed at existing call sites.
- Added 4 unit tests in `admin-ui-pages.unit.test.ts` (exact match, sub-path via `startsWith`, alternate route, and empty-default). All 106 tests pass.

dodizzle (commented):
APPROVE — Fix is correct, complete across all call sites, and backed by tests that match the project's conventions. All 5 `renderAdminToolbar` call sites updated, `startsWith` correctly handles sub-paths, 4 unit tests land in the same PR per CLAUDE.md policy.

dmcaulay (approved):
APPROVE — both prior findings addressed: `activePath = ""` default added (eliminates merge friction) and 4 unit tests land in the same PR covering agents, sub-path, provision, and empty-default cases. CI green.

PR #259, opened 2026-06-10
fix(admin): show real user email in navbar instead of "admin"

dodizzle:
## Summary
- `ADMIN_USER_NAME = "admin"` was hardcoded and passed to every page render — the session JWT had the real email all along but it was thrown away after validation
- Renamed `verifySessionToken → getSessionEmail` to return `string | null` instead of `boolean`
- Typed the Hono app with `AdminUIEnv` so `c.var.userEmail` is strongly typed
- Auth middleware now sets `c.var.userEmail` from the decoded JWT before calling `next()`
- All 12 render-call sites now use `c.var.userEmail`

## Test plan
- [ ] Sign in via Google OAuth → navbar shows your actual email address
- [ ] Existing smoke tests: 33/33 pass
- [ ] New smoke test: `GET /admin/agents` HTML contains `admin@example.com` (the session email)

### DIFF
--- a/admin/src/admin-ui.smoke.test.ts
+++ b/admin/src/admin-ui.smoke.test.ts
@@ -408,6 +408,16 @@ describe("admin UI — authenticated pages", () => {
     expect(html).toContain("Test Agent");
   });
 
+  it("authenticated GET /admin/agents shows the session user's email in the navbar", async () => {
+    const app = createAdminUIApp(makeMockDeps());
+    const res = await app.request("/admin/agents", {
+      headers: { Cookie: `admin_session=${cookie}` },
+    });
+    expect(res.status).toBe(200);
+    const html = await res.text();
+    expect(html).toContain("admin@example.com");
+  });
+
   it("authenticated GET /admin/agents/:id returns 200 with agent detail sections", async () => {
     const app = createAdminUIApp(makeMockDeps());
     const res = await app.request(`/admin/agents/${AGENT_ID}`, {
--- a/admin/src/admin-ui.smoke.test.ts
+++ b/admin/src/admin-ui.smoke.test.ts
@@ -430,6 +430,17 @@ describe("admin UI — authenticated pages", () => {
     expect(html).toContain("Tools");
     expect(html).toContain("Tokens");
     expect(html).toContain("Plugins");
+    expect(html).toContain("admin@example.com");
+  });
+
+  it("authenticated GET /admin/provision shows the session user's email in the navbar", async () => {
+    const app = createAdminUIApp(makeMockDeps());
+    const res = await app.request("/admin/provision", {
+      headers: { Cookie: `admin_session=${cookie}` },
+    });
+    expect(res.status).toBe(200);
+    const html = await res.text();
+    expect(html).toContain("admin@example.com");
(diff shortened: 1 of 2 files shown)


dmcaulay (commented):
COMMENT — all 13 call sites correctly updated (verified: 0 remaining occurrences of ADMIN_USER_NAME), XSS safety preserved via existing escapeHtml(), and the AdminUIEnv typing gives compiler-checked coverage. One gap: the new smoke test covers only /admin/agents; adding email assertions to the /admin/agents/:id and /admin/provision tests would give full coverage per the same-PR test policy.

admin/src/admin-ui.smoke.test.ts
```diff
@@ -408,6 +408,16 @@ describe("admin UI — authenticated pages", () => {
     expect(html).toContain("Test Agent");
   });
 
+  it("authenticated GET /admin/agents shows the session user's email in the navbar", async () => {
+    const app = createAdminUIApp(makeMockDeps());
+    const res = await app.request("/admin/agents", {
+      headers: { Cookie: `admin_session=${cookie}` },
+    });
+    expect(res.status).toBe(200);
+    const html = await res.text();
```
dmcaulay, line 417: Good test — confirms the email flows from session cookie to rendered HTML. Consider adding the same assertion to the `/admin/agents/:id` and `/admin/provision` authenticated tests too: `expect(html).toContain("admin@example.com")`. Those pages also render the email and a missed call site there wouldn't be caught by the current test suite.

admin/src/admin-ui.ts
```diff
@@ -21,6 +21,8 @@
  */
 
 import { Hono, type MiddlewareHandler } from "hono";
+
+type AdminUIEnv = { Variables: { userEmail: string } };
```
dodizzle: This `type AdminUIEnv` declaration sits between two import statements, splitting the import block. Biome's `organizeImports` treats any non-import statement as a group boundary — `task format --check` will flag this file as unformatted and fail CI. Move it below the last import (after the `hono/jwt` import block).

dodizzle (commented):
COMMENT — One fix needed before merge: the new `type AdminUIEnv` declaration is placed between two import statements (lines 23 and 26), splitting the import block. Biome's organizeImports will flag this as a format error and fail the `task format --check` gate. Move it below the last import. Everything else looks solid: all 13 call sites updated, XSS safety preserved, and 3 email smoke tests land in the same PR per project conventions.

dmcaulay (approved):
APPROVE — prior finding addressed: email assertions added to /admin/agents/:id and /admin/provision smoke tests (commit 2230d6b6). All 12 call sites migrated from ADMIN_USER_NAME to c.var.userEmail, AdminUIEnv type correctly placed after imports, XSS safety preserved. Agent docker build failures were transient Docker Hub network timeouts — lint/typecheck/test passed on all commits.

dodizzle (commented):
APPROVE — prior finding addressed: `type AdminUIEnv` moved below import block (commit aa772044), CI green (lint/typecheck/test pass on all commits), dmcaulay approved at current head. No new issues. Clean unblock.

PR #274, opened 2026-06-11
fix(agent): seed mise PVC from image on first boot + configurable startup timeout

dmcaulay:
## Summary

- **Mise PVC fix**: On a fresh PVC, copies the mise data dir from the image default location to the PVC on first startup, then sets MISE_DATA_DIR to the PVC path. Subsequent boots detect the existing shims dir and skip the copy. Fixes the regression from #265 where mise installs lived in the ephemeral container layer and were lost on pod restart.
- **Startup timeout**: Bumps DEFAULT_STARTUP_TIMEOUT_MS from 10s to 60s for prod cold-start tolerance. Wires SHIPWRIGHT_STARTUP_TIMEOUT_MS env var for operator override. Dev stack overrides to 10s in the example dev config so the fast-fail behavior is preserved locally.

## Test plan

- [ ] `bun test agent/src/setup.integration.test.ts` — new tests cover: sets MISE_DATA_DIR to PVC path, seeds from image on fresh PVC, skips seed when shims already exist
- [ ] `bun test agent/src/entrypoint.integration.test.ts` — existing timeout tests pass with updated default
- [ ] Deploy to a prod agent with a fresh PVC and verify mise shims survive a pod restart
- [ ] Verify SHIPWRIGHT_STARTUP_TIMEOUT_MS=10000 in dev config produces fast-fail timeout as before


agent/src/entrypoint-main.ts
```diff
@@ -95,4 +95,7 @@ await runEntrypoint({
   exit: (code: number) => {
     process.exit(code);
   },
+  startupTimeoutMs: process.env.SHIPWRIGHT_STARTUP_TIMEOUT_MS
+    ? Number(process.env.SHIPWRIGHT_STARTUP_TIMEOUT_MS)
```
dmcaulay: `Number("abc")` returns `NaN`, and `setTimeout(cb, NaN)` fires immediately (treated as 0ms delay). A misconfigured `SHIPWRIGHT_STARTUP_TIMEOUT_MS=abc` would silently time out every startup. Suggest: `const ms = Number(process.env.SHIPWRIGHT_STARTUP_TIMEOUT_MS); startupTimeoutMs: Number.isFinite(ms) ? ms : undefined`.

dmcaulay (commented):

Clean fix for the PVC mise regression. Seed logic is correct and idempotent, three integration tests cover all paths, CI green. One suggestion: add a NaN guard on the `SHIPWRIGHT_STARTUP_TIMEOUT_MS` parse — `Number("abc")` returns NaN and `setTimeout(cb, NaN)` fires immediately, silently breaking startup on operator misconfiguration.

agent/src/entrypoint-main.ts
dodizzle: `Number(process.env.SHIPWRIGHT_STARTUP_TIMEOUT_MS)` doesn't guard against non-numeric input. `Number('abc')` returns `NaN`, and `setTimeout(fn, NaN)` fires immediately (NaN coerces to 0ms) — instant startup timeout on any misconfiguration. Fix: `const n = Number(process.env.SHIPWRIGHT_STARTUP_TIMEOUT_MS); startupTimeoutMs: Number.isFinite(n) ? n : undefined`

agent/src/setup.ts
```diff
@@ -414,13 +415,25 @@ export async function runMiseStartup(
...
+  // on the PVC from the first restart onward. Subsequent boots skip the copy.
+  const pvMiseDir = join(home, "mise");
+  const pvShimsDir = join(pvMiseDir, "shims");
+  const imageMiseDir = join(
+    process.env.XDG_DATA_HOME ??
+      join(process.env.HOME ?? home, ".local", "share"),
+    "mise",
+  );
+  if (!existsSync(pvShimsDir) && existsSync(imageMiseDir)) {
+    cpSync(imageMiseDir, pvMiseDir, { recursive: true });
```
dodizzle: `cpSync` is unguarded — a permissions error, source corruption, or PVC disk-full throws out of `runMiseStartup` entirely, preventing the agent from starting on a fresh PVC. The `mise install` branch below wraps failures with `console.warn` and returns early; the seed step should do the same. Wrap in try/catch and log a warning on failure so startup can continue without PVC-seeded shims.

agent/src/setup.ts
```diff
@@ -442,17 +455,8 @@ export async function runMiseStartup(
...
-    process.env.MISE_DATA_DIR ??
-    join(
-      process.env.XDG_DATA_HOME ??
-        join(process.env.HOME ?? home, ".local", "share"),
-      "mise",
-    );
-  const shimsDir = join(miseDataDir, "shims");
+  // Prepend MISE_DATA_DIR/shims to PATH so workspace-declared tools are found
+  // in non-interactive bash sessions.
+  const shimsDir = join(process.env.MISE_DATA_DIR ?? pvMiseDir, "shims");
```
dodizzle: `process.env.MISE_DATA_DIR ?? pvMiseDir` — `MISE_DATA_DIR` is unconditionally set to `pvMiseDir` at line 433, so this fallback is dead code. Simplify to `join(pvMiseDir, 'shims')`.

dodizzle (commented):
COMMENT — PVC seed logic and timeout bump look solid overall, but two issues need addressing before merge:

1. `SHIPWRIGHT_STARTUP_TIMEOUT_MS` is not validated — `Number('abc')` → NaN, and `setTimeout(fn, NaN)` fires immediately, causing instant startup failure on misconfiguration.
2. `cpSync` has no error handling. A permissions error or disk-full condition on the PVC throws out of `runMiseStartup`, killing the agent on fresh start — exactly when the seed is most needed. The surrounding `mise install` path wraps failures with a warning; the seed step should too.

Also: `process.env.MISE_DATA_DIR ?? pvMiseDir` at `setup.ts:460` is dead code — `MISE_DATA_DIR` is unconditionally set to `pvMiseDir` three lines earlier.

dmcaulay (commented):

All prior findings addressed. NaN guard via `Number.isFinite(ms)` is correct. `cpSync` is now wrapped in try/catch with a graceful warning — agent continues if PVC seed fails on fresh boot. Dead code (`MISE_DATA_DIR ?? pvMiseDir`) removed; `MISE_DATA_DIR` set unconditionally. Tests updated to match all three code paths. CI failed on `6242387f` (missing docs for `XDG_DATA_HOME` and `SHIPWRIGHT_STARTUP_TIMEOUT_MS`) — both are now in `configuration.md` in this commit. Expect green on `9ba5ccb`.

dodizzle (approved):
APPROVE — All prior findings addressed.

NaN guard via `Number.isFinite(ms)` is correct. `cpSync` is now wrapped in try/catch with a graceful warning — agent continues if PVC seed fails on fresh boot. Dead code (`MISE_DATA_DIR ?? pvMiseDir`) removed; `shimsDir` uses `pvMiseDir` directly.

Bonus refactor: `ensureDotClaudeSymlink` extracted from `entrypoint-main.ts` into `setup.ts` with injectable FS seams — four integration tests cover all edge cases (dangling symlink, real-dir-at-link-path, create-from-scratch, idempotent). TypeScript compatibility is correct.

CI in-progress on merge-from-main; prior code run at `5a47bf36` completed success.

PR #278, opened 2026-06-11
feat: remove SHIPWRIGHT_INTERNAL_API_KEY (UNI-1.2)

dmcaulay:
## Summary

- Removes `SHIPWRIGHT_INTERNAL_API_KEY` and the `internalApiKey` field from `AgentRuntimeDeps` / `createAgentRuntimeApp`
- Runtime routes `GET /agents/:id/config` and `GET /agents/:id/crons` are now protected by the same `createAdminAuthMiddleware` (admin-key / per-agent-token / session-cookie) as the CRUD routes
- Renames the agent-side key to `SHIPWRIGHT_AGENT_API_KEY` throughout: `config.ts`, `cli-args.ts`, `entrypoint.ts`, `cutover-values.ts`, migration scripts, Dockerfiles, docs, and dev stack

## Test plan

- [x] `admin/src/api.smoke.test.ts` — updated: removed `internalApiKey` from all `createAgentRuntimeApp` calls, auth now uses admin key; combined-server regression guard still holds
- [x] `agent/src/config.unit.test.ts` — updated: `SHIPWRIGHT_AGENT_API_KEY` env var
- [x] `agent/src/cli-args.unit.test.ts` — updated: `SHIPWRIGHT_AGENT_API_KEY` fallback
- [x] `agent/src/cutover-values.unit.test.ts` — updated: YAML patch uses `SHIPWRIGHT_AGENT_API_KEY`
- [x] `scripts/dev-tmux.unit.test.ts` — updated: admin pane no longer has `SHIPWRIGHT_INTERNAL_API_KEY`; agent pane docker run does not contain it
- [x] `bun test` — 2536 pass, 2 pre-existing playwright failures (unchanged from baseline)
- [x] `bunx biome lint .` — no issues
- [x] `bun run --filter='*' typecheck` — all workspaces clean


docs/agent.md
```diff
@@ -87,7 +87,7 @@ All child models cascade-delete with their `Agent`.
 | `DATABASE_URL_SHIPWRIGHT_ADMIN` | ✅ | Dedicated Postgres datasource for the admin service (e.g. `postgresql://user:pass@host:5432/db`). |
 | `SHIPWRIGHT_AGENT_ID` | ✅ (entrypoint) | The agent's ID in the Shipwright platform. Also settable via `--agent-id` CLI flag. |
 | `SHIPWRIGHT_API_URL` | ✅ (entrypoint) | Base URL of the Shipwright API used to fetch agent config at startup. Also settable via `--api-url`. |
-| `SHIPWRIGHT_INTERNAL_API_KEY` | ✅ (entrypoint + runtime API) | Bearer token for the config fetch at startup and for `/agents/*`. Also settable via `--api-key`. |
+| `SHIPWRIGHT_AGENT_API_KEY` | ✅ (entrypoint) | Bearer token for the config fetch at startup (`/agents/:id/config` and `/agents/:id/crons`). Also settable via `--api-key`. |
```
dmcaulay: The relationship between `SHIPWRIGHT_AGENT_API_KEY` (agent-side) and `SHIPWRIGHT_ADMIN_API_KEYS` (server-side) isn't stated. The agent uses the former to authenticate; the server validates against the latter. An operator setting only `SHIPWRIGHT_AGENT_API_KEY` without registering it in `SHIPWRIGHT_ADMIN_API_KEYS` will get a 401 at startup.

Suggest appending: "The value must be registered in `SHIPWRIGHT_ADMIN_API_KEYS` on the server with scope `<agentId>` (or `*` for admin bypass)."

dmcaulay (commented):

Clean removal of `SHIPWRIGHT_INTERNAL_API_KEY`. Runtime routes now use the same admin-key / per-agent-token / session-cookie auth as CRUD, eliminating the second auth secret. Per-route middleware application correctly avoids the Hono v4 `app.use("*")` hoisting bug; the combined-server regression test guards it.

One docs gap: `docs/agent.md` describes `SHIPWRIGHT_AGENT_API_KEY` (agent-side) and `SHIPWRIGHT_ADMIN_API_KEYS` (server-side) separately but doesn't explain the operator must register the agent key in `SHIPWRIGHT_ADMIN_API_KEYS` with `scope=<agentId>` for runtime polling to succeed. Agents will 401 on startup if only `SHIPWRIGHT_AGENT_API_KEY` is set without the server-side entry. Worth a one-liner.

scripts/dev-tmux.ts
```diff
@@ -200,8 +200,6 @@ export const STACK_PANES: Pane[] = [
       "-e",
       "SHIPWRIGHT_AGENT_ID=dev-agent",
       "-e",
-      `SHIPWRIGHT_INTERNAL_API_KEY=${DUMMY_INTERNAL_API_KEY}`,
-      "-e",
       `PORT=${AGENT_PORT}`,
```
dodizzle, line 212: The agent docker run previously had `-e SHIPWRIGHT_INTERNAL_API_KEY=${DUMMY_INTERNAL_API_KEY}` here. That flag is gone but no `-e SHIPWRIGHT_AGENT_API_KEY=...` is added in its place. `SHIPWRIGHT_API_URL` and `SHIPWRIGHT_AGENT_ID` are still injected (lines 195, 201), so config sync was active in `task stack` before this PR — without the replacement key, `apiKey` is undefined at startup and config sync is silently disabled. Suggest adding a `DUMMY_AGENT_API_KEY` constant and injecting it here, with a matching entry in `SHIPWRIGHT_ADMIN_API_KEYS` on the admin pane (or a seeded per-agent DB token for `dev-agent`).

dodizzle (commented):
COMMENT — One issue before merge: the dev-tmux agent pane removes `SHIPWRIGHT_INTERNAL_API_KEY` from the docker run flags but does not inject the replacement `SHIPWRIGHT_AGENT_API_KEY`. Since `SHIPWRIGHT_API_URL` and `SHIPWRIGHT_AGENT_ID` are both still present, config sync was working in `task stack` before this change — it breaks after. Otherwise the refactor is thorough: comprehensive rename across all files, correct per-route middleware placement, 2536 tests passing.

scripts/dev-tmux.ts
dmcaulay, line 212: The agent docker pane removes `SHIPWRIGHT_INTERNAL_API_KEY` but does not inject `SHIPWRIGHT_AGENT_API_KEY`. `entrypoint.ts:78` validates it as required — `task stack` will exit with `FATAL: missing required variable: SHIPWRIGHT_AGENT_API_KEY` until this is added.

Suggested fix:
```ts
// Add constant alongside other dummies:
const DUMMY_AGENT_API_KEY = "dev-agent-key";

// In the agent pane docker run -e flags:
"-e",
`SHIPWRIGHT_AGENT_API_KEY=${DUMMY_AGENT_API_KEY}`,

// In the admin pane env:
SHIPWRIGHT_ADMIN_API_KEYS: `dev-agent:${DUMMY_AGENT_API_KEY}:dev-agent`,
```
Also update the unit test to assert the key is present (not just that the old one is absent).

dmcaulay (commented):
COMMENT — Re-review after `07b16be6`.

The self-review docs finding is addressed: `docs/agent.md` now has the one-liner explaining `SHIPWRIGHT_AGENT_API_KEY` must be registered in `SHIPWRIGHT_ADMIN_API_KEYS`.

dodizzle's finding is still open: `task stack` is broken. The agent Docker pane removes `SHIPWRIGHT_INTERNAL_API_KEY` but never injects the replacement `SHIPWRIGHT_AGENT_API_KEY`. `entrypoint.ts:78` validates it as required and exits with `FATAL: missing required variable: SHIPWRIGHT_AGENT_API_KEY` on startup.

Fix: add a dev dummy key to both the agent pane (`-e SHIPWRIGHT_AGENT_API_KEY=dev-agent-key`) and the admin pane env (`SHIPWRIGHT_ADMIN_API_KEYS=dev-agent:dev-agent-key:dev-agent`). The unit test should also be tightened to assert the key is present, not just that the old one is absent.

scripts/dev-tmux.ts
dodizzle, line 203: The `-e SHIPWRIGHT_INTERNAL_API_KEY=...` flag was removed above, but `-e SHIPWRIGHT_AGENT_API_KEY=...` was never added in its place. `SHIPWRIGHT_API_URL` and `SHIPWRIGHT_AGENT_ID` are both still injected (lines 195, 201), so the entrypoint reaches the required-var check and exits with `FATAL: missing required variable: SHIPWRIGHT_AGENT_API_KEY`. Add a `DUMMY_AGENT_API_KEY` constant alongside the other dummies, inject it here, and register it in `SHIPWRIGHT_ADMIN_API_KEYS` in the admin pane env. Also add `expect(cmdStr).toContain("SHIPWRIGHT_AGENT_API_KEY")` to the unit test next to the absence check at line 301.

dodizzle (commented):
COMMENT — Re-review after `07b16be` / `e03528e` (merged from main).

The docs finding from the self-review is fully addressed: `docs/agent.md` now has the one-liner about registering `SHIPWRIGHT_AGENT_API_KEY` in `SHIPWRIGHT_ADMIN_API_KEYS`. The `HttpShipwrightAdminClient` bearer-auth update (`1ac1474`) is clean, with a good injectable `fetch` seam and proper unit tests. The mise PVC seeding (`e03528e`) is solid.

dodizzle's finding is still open: `task stack` is broken. The agent docker run removes `-e SHIPWRIGHT_INTERNAL_API_KEY=...` but never adds the replacement `-e SHIPWRIGHT_AGENT_API_KEY=...`. The entrypoint validates it as required and exits with `FATAL: missing required variable: SHIPWRIGHT_AGENT_API_KEY`. The unit test asserts the old key is absent but has no complementary check that the new key is present.

Fix: add `DUMMY_AGENT_API_KEY = "dev-agent-key"` constant, inject `-e SHIPWRIGHT_AGENT_API_KEY=${DUMMY_AGENT_API_KEY}` in the agent docker run, add `SHIPWRIGHT_ADMIN_API_KEYS: \`dev-agent:${DUMMY_AGENT_API_KEY}:dev-agent\`` to the admin pane env, and add `expect(cmdStr).toContain("SHIPWRIGHT_AGENT_API_KEY")` to the unit test.

dmcaulay (commented):

Re-review after `8e3cacd9`. dodizzle's finding is fully addressed: `SHIPWRIGHT_AGENT_API_KEY` injected into the agent docker pane, `SHIPWRIGHT_ADMIN_API_KEYS` added to the admin pane env with correct `name:token:scope` format (`dev-agent:dev-agent-key:dev-agent`), and unit tests now assert both keys are present (not just the old key absent). CI green. `task stack` is unblocked.

dodizzle (approved):
APPROVE — re-review at `8e3cacd9`. The single blocker is resolved: `SHIPWRIGHT_AGENT_API_KEY` is now injected into the agent docker run with a clearly-commented `DUMMY_AGENT_API_KEY` constant, `SHIPWRIGHT_ADMIN_API_KEYS` is registered on the admin pane with the correct `name:token:scope` format (`dev-agent:dev-agent-key:dev-agent`), and unit tests assert both keys are present (not just the old key absent). `task stack` is unblocked. CI green. No new issues.

PR #281, opened 2026-06-11
feat(agent): reshape run-agent.ts into minimal health server (UNI-1.3)

dmcaulay:
## Summary
- Removes the `/agents/*` transparent proxy from `run-agent.ts` — no proxy routes remain
- Reshapes `run-agent.ts` into a minimal chat-only server (dev-only `/chat` gate preserved)
- Adds dedicated health server on `SHIPWRIGHT_HEALTH_PORT` (default 3459) via `startHealthServer()` from `health.ts`; `entrypoint-main.ts` starts it in-process before the startup sequence so K8s liveness probes are reachable during init
- Exports `DEFAULT_HEALTH_PORT = 3459` from `health.ts` as the single source of truth

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | GET /health returns 200 { ok: true, slack } | MET |
| 2 | GET /health returns 500 when Slack down >90s | MET |
| 3 | /chat dev endpoint preserved (DEFAULT-DENY) | MET |
| 4 | /agents/* proxy removed | MET |
| 5 | Health server on SHIPWRIGHT_HEALTH_PORT (default 3459) | MET |
| 6 | entrypoint-main.ts starts health server alongside Slack app | MET |
| 7 | startConfigSync() wiring preserved | MET |

## Test Plan
- [x] New smoke tests: /health → 404 from composed app, /agents/* → 404, /chat → 404 default
- [x] Existing health integration tests cover /health response shape (ACs 1 and 2)
- [x] 23 targeted tests pass, biome lint clean, all 4 packages typecheck clean


agent/src/run-agent.ts
```diff
@@ -114,12 +83,25 @@ export async function startServer(opts?: { port?: number }): Promise<void> {
...
+  // The health server runs on SHIPWRIGHT_HEALTH_PORT (default 3459) — separate
+  // from the chat server so K8s liveness probes are always reachable even when
+  // chat is disabled.
+  const healthPort =
+    opts?.port !== undefined
+      ? undefined // When port is overridden (tests), skip health server to avoid port conflicts
+      : Number(process.env.SHIPWRIGHT_HEALTH_PORT ?? DEFAULT_HEALTH_PORT);
+
+  if (healthPort !== undefined) {
+    startHealthServer(healthPort);
```
dodizzle: Port conflict: `entrypoint-main.ts` starts a health server on port 3459 in-process, then spawns this file as a subprocess. When run as a subprocess, `startServer()` is called with no args (`opts` is `undefined`), so the `opts?.port !== undefined` guard on line 91 is `false` and this `startHealthServer(healthPort)` fires — attempting to bind 3459 while the parent already holds it. Result: `EADDRINUSE`, subprocess crash, agent dead.

Simplest fix: remove the health server from `startServer()` entirely. The parent process in `entrypoint-main.ts` already owns health. `run-agent.ts` as a subprocess doesn't need to start it again — it just needs to run config sync (and optionally the chat server).

dodizzle (commented):
COMMENT — Three issues to address before merge:

**Critical**: `run-agent.ts` starts a second health server on port 3459 even when spawned as a subprocess by `entrypoint-main.ts` (which already owns that port). The guard `opts?.port !== undefined` only fires when callers pass a port override, but subprocess invocation via `import.meta.main` calls `startServer()` with no args — so the guard never triggers and the subprocess crashes with EADDRINUSE at every production startup.

**Important**: `site/src/pages/docs/configuration.astro:302` still lists `HEALTH_PORT` (old name, default `PORT ?? 3001`) — needs updating to `SHIPWRIGHT_HEALTH_PORT` / `3459` before the docs site is republished.

**Important**: `agent/Dockerfile` exposes only port 3000 (chat, DEFAULT-DENY in production) and doesn't expose 3459 (health). The comment on line 7 still describes `PORT` as "Hono server port" without noting it's chat-only and optional. Add `EXPOSE 3459` and update the `SHIPWRIGHT_HEALTH_PORT` comment.

agent/src/run-agent.ts
dmcaulay: `entrypoint-main.ts` already calls `startHealthServer(healthPort)` in the parent process (line 34) before spawning this file as a subprocess. When the subprocess calls `startServer()` without `opts`, the skip condition `opts?.port !== undefined` is false, so `healthPort` resolves to 3459 and `startHealthServer(3459)` fires a second time — but the parent still holds the port. The child's `Bun.serve({port: 3459})` will throw EADDRINUSE and crash the subprocess.

Simplest fix: remove the `startHealthServer` call from `startServer()` entirely. The parent process owns health for the full pod lifetime — `run-agent.ts` doesn't need to manage it.

dmcaulay (commented):
COMMENT — 1 critical issue: both `entrypoint-main.ts` (parent process) and `run-agent.ts` (child subprocess) call `startHealthServer(3459)`. The parent binds the port, then spawns `run-agent.ts`; the child's `Bun.serve({port: 3459})` fails with EADDRINUSE and the subprocess crashes. The existing skip (`opts?.port !== undefined ? undefined`) only guards tests. Fix: remove `startHealthServer` from `run-agent.ts:startServer()` — the parent already owns the health endpoint for the pod lifetime. The core design (health on a dedicated port, started before startup sequence) is the right call.

dmcaulay (commented):
APPROVE — all prior findings addressed. `run-agent.ts` no longer calls `startHealthServer` (critical EADDRINUSE fix landed), `agent/Dockerfile` correctly exposes 3459, `docs/agent.md` and `docs/configuration.astro` updated to `SHIPWRIGHT_HEALTH_PORT`/3459, dead mock-deps exports cleaned up. CI green.

PR #282, opened 2026-06-11
feat(metrics): add METRICS_BASE_PATH for sub-path deployment

dodizzle:
## Summary

- Adds `METRICS_BASE_PATH` env var so the metrics service can be mounted at a path prefix (e.g. `/sw`) alongside the existing vitals-os metrics service during migration
- `dashboard-page.ts`: injects `window.__METRICS_BASE__` and uses the base path for stylesheet + `app.js` asset URLs
- `app.js`: all `/metrics/*` fetch calls use `API_BASE = (window.__METRICS_BASE__ || '') + '/metrics'`
- `server.ts`: reads `METRICS_BASE_PATH`, passes to deps, mounts via `new Hono().route(basePath, metricsApp)` when set — no-op when unset so local dev is unchanged
- Admin toolbar: adds **Metrics** link; URL via `METRICS_DASHBOARD_URL` env var (defaults to `/sw/dashboard` for vitals-os cluster deployment)

## Test plan

- [ ] `task api` still works locally (no base path set → unchanged behavior at `/dashboard`)
- [ ] With `METRICS_BASE_PATH=/sw`, server starts and logs `(base: /sw)`
- [ ] Dashboard loads at `/sw/dashboard` with assets resolving to `/sw/dashboard/styles.css` and `/sw/dashboard/app.js`
- [ ] API calls from the JS go to `/sw/metrics/summary` etc.
- [ ] Admin toolbar shows Agents · Provision · Metrics links


metrics/src/server.ts
```diff
@@ -35,8 +36,9 @@ const mode = selectProviderMode(process.env);
 const offlineMode = mode === "fixtures";
 
 const port = Number(process.env.METRICS_API_PORT ?? 3460);
+const basePath = process.env.METRICS_BASE_PATH ?? "";
 const accountsClient = new HttpAccountsClient(
```
dodizzle, line 46: Hard rename with no fallback breaks existing deployments. Any deployment with `METRICS_ACCOUNTS_URL` set loses its value silently. Add a compat fallback:
```ts
process.env.METRICS_ADMIN_URL ?? process.env.METRICS_ACCOUNTS_URL ?? "http://localhost:3000"
```
Also note the default port changed from `3457` (accounts service) to `3000` (admin service) — worth a changelog entry.

lib/web/toolbar.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export function renderShipwrightToolbar(
+  opts: ShipwrightToolbarOptions,
+): string {
+  const { userName, activePath, logoutAction } = opts;
+  const metricsUrl = opts.metricsUrl ?? "/sw/dashboard";
+  const active = (prefix: string) =>
+    activePath.startsWith(prefix) ? " active" : "";
+  return `<nav class="vos-toolbar" aria-label="Site navigation">
+    <a href="/admin/agents" class="vos-wordmark">Shipwright</a>
```
dodizzle: Hardcoded `/admin/agents` here and on line 135 means this shared component always renders admin-side navigation, even when called from the metrics dashboard. If the services are always co-deployed at the same origin this is fine, but consider documenting the assumption or adding an optional `adminBaseUrl` to `ShipwrightToolbarOptions` to make it explicit.

dodizzle (commented):
**COMMENT** — 3 blocking issues before merge.

**Critical**: `agent/Dockerfile` is missing `COPY lib/package.json ./lib/` — the new `@shipwright/lib` workspace is in the root lockfile but the agent build stage never copies the manifest, so `bun install --frozen-lockfile` can't resolve it. This is the direct cause of the CI "agent docker build" failure.

**Important**: `METRICS_ACCOUNTS_URL` is hard-renamed to `METRICS_ADMIN_URL` with no compat fallback. Existing deployments with `METRICS_ACCOUNTS_URL` set silently fall through to the wrong default (`localhost:3000` instead of the configured accounts URL). Add `process.env.METRICS_ADMIN_URL ?? process.env.METRICS_ACCOUNTS_URL ?? "http://localhost:3000"` and a deprecation log.

**Important**: The shared `renderShipwrightToolbar` in `lib/web/toolbar.ts` hardcodes `href="/admin/agents"` for both the wordmark and Agents nav link. When rendered from the metrics dashboard (a separate service), these links cross service boundaries. If admin and metrics are always co-deployed behind the same ingress this works, but it should be documented — or the admin base URL made configurable.

**Strengths**: Clean Hono `.route()` mount, good test double maintenance, solid architectural move extracting the toolbar to a shared lib.

dmcaulay (commented):
COMMENT — one blocking CI failure before merge.

**Critical: `tokens-api.integration.test.ts` — `agentName` mock wired to wrong interface method**

The test "populates agentName when accountsClient returns a matching user" passes its mock agent record via `makeAccountsClientMock(async () => [userRecord])` — the first param, which sets `listUsers`. But `api.ts:777` now calls `accountsClient.listAgents()`. The `makeAccountsClientMock` second parameter (optional `listAgentsImpl`) defaults to `async () => []`, so `agentNameMap` is always empty in this test and `agentName` comes back `undefined`. CI confirms this is the only failing test.

Fix: pass the agent record via the second `makeAccountsClientMock` parameter with `AgentRecord` shape:
```ts
const accountsClient = makeAccountsClientMock(
  async () => [],
  async () => [{ id: "agent-abc123", name: "Bodhi" }],
);
```
Also update the test title and the third test's title/comment from "listUsers" to "listAgents" to match the implementation.

**Strengths**: Clean `Hono().route(basePath, metricsApp)` mount (no-op when unset). `window.__METRICS_BASE__` injection for client-side JS is the right pattern. Deprecation warning for `METRICS_ACCOUNTS_URL` is correct. All three Dockerfiles updated correctly to include `COPY lib/package.json ./lib/`. `@shipwright/lib` extraction is a good consolidation.

metrics/src/dashboard/dashboard-page.ts
```diff
@@ -4,29 +4,35 @@
...
+  /** Path prefix the app is mounted at (e.g. "/sw"). Empty string or undefined = root. */
+  basePath?: string;
 }
 
 function infoIcon(tip: string): string {
   return `<span class="info-icon" data-tooltip="${tip}" tabindex="0" aria-label="${tip}">i</span>`;
 }
 
 export function renderDashboardPage(opts: DashboardPageOptions): string {
+  const base = opts.basePath ?? "";
```
dmcaulay, line 24: `METRICS_BASE_PATH` injection is not covered by any test with a non-empty value. The snapshot tests only exercise the default empty-string case. Add a unit test asserting `renderDashboardPage({ basePath: '/sw' })` produces `window.__METRICS_BASE__ = '/sw'` and prefixed asset URLs.

dmcaulay (commented):
APPROVE — prior blocking finding fully resolved. CI green. The `makeAccountsClientMock.listAgents` proxy in `test-helpers.ts` correctly maps `UserRecord[]` through `{id, name}` — existing test call sites work without changes. All three Dockerfiles include `COPY lib/package.json`. Deprecation warning for `METRICS_ACCOUNTS_URL` is correct. `@shipwright/lib` extraction is clean.

One follow-on worth a ticket: `METRICS_BASE_PATH` has no test coverage — neither the Hono `.route()` mount path nor the `window.__METRICS_BASE__` HTML injection is tested with a non-empty base. Not blocking, but worth adding a unit test to `dashboard-page.unit.test.ts`.

dodizzle (commented):
**APPROVE** — All prior findings addressed. CI green. Ready to merge.

Findings resolved since last review:
- `agent/Dockerfile` missing `COPY lib/package.json ./lib/` → fixed (commit 02b34c49)
- `METRICS_ACCOUNTS_URL` hard rename with no compat fallback → fixed with fallback + deprecation log (commit 6e0801e4)
- Shared toolbar hardcoding `/admin/*` hrefs → fixed with `adminBaseUrl` option (commit 6e0801e4)
- `makeAccountsClientMock.listAgents` hardcoded to always return `[]`, breaking agent name resolution test → fixed by proxying to `listUsersImpl` (commit 2dc2d0c)

Strengths: Extracting shared toolbar into `@shipwright/lib` eliminates real duplication. Hono `.route()` mount is clean with zero overhead when `METRICS_BASE_PATH` is unset. Compat fallback for `METRICS_ACCOUNTS_URL` is the right rolling-deployment pattern.

PR #283, opened 2026-06-11
fix(agent): bind health server before mise/plugin install

dodizzle:
## Problem

`agent/src/index.ts` bound the health server at Step 4 — after `runMiseStartup()`, config sync, and `reconcileSystemCrons`. Kubelet liveness/readiness probes hit `SHIPWRIGHT_HEALTH_PORT` on every tick, so if mise runs long (e.g. compiling a toolchain from source, same failure mode as app-vitals/vitals-os#1471), probes get `connection refused` and the kubelet SIGKILLs the pod before startup can finish.

## Fix

Reorder startup so the health server binds at Step 2, right after `ensureAgentHome`. None of its dependencies (`sessions`, `analytics`, `slack`, `runner`, `cronDeps`) require mise or config sync — they only need the static config already available at boot. The reorder is behavior-preserving for everything downstream.

New boot order:
```
Step 1  ensureAgentHome
Step 2  health server on port <SHIPWRIGHT_HEALTH_PORT>  ← now binds before slow toolchain install
Step 3  mise startup complete
        default plugin install complete
Step 4  config sync
Step 5  system crons reconciled
Step 6  cron sync started
Step 7  Slack app started
```

## Testing

- `tsc --noEmit` (via `bun run typecheck`): clean
- `biome check agent/src/index.ts`: clean
- Mirrors the fix applied to vitals-os in app-vitals/vitals-os#1471

### DIFF
--- a/agent/src/index.ts
+++ b/agent/src/index.ts
@@ -3,13 +3,14 @@
  * Shipwright agent startup entrypoint.
  *
  * Boot order:
- *  1. ensureAgentHome + runMiseStartup + installPlugins (defaults)
- *  2. Config sync — fetch AgentConfigBundle, apply env, install agent plugins (await first)
- *  3. reconcileSystemCrons — best-effort, non-fatal
- *  4. Health server
- *  5. Cron sync loop (60s)
- *  6. Slack Bolt Socket Mode app
- *  7. Graceful SIGTERM/SIGINT shutdown
+ *  1. ensureAgentHome
+ *  2. Health server — binds immediately so kubelet probes stay green during setup
+ *  3. runMiseStartup + installPlugins (defaults)
+ *  4. Config sync — fetch AgentConfigBundle, apply env, install agent plugins (await first)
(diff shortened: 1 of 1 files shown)


dmcaulay (commented):
APPROVE — clean, safe reorder. CI green. All moved dependencies (`WebClient`, `createAnalyticsStore`, `createFileSessionStore`, `createRunClaude`, `SystemClock`) are npm/Bun packages — none are mise-managed runtimes. The reorder is behavior-preserving for everything downstream. Identical fix to vitals-os#1471 already shipped.

dodizzle (commented):
**APPROVE** — safe, well-motivated reorder. CI green.

All construction moved to Step 2 (`sessions`, `analytics`, `WebClient`, `runner`, `cronDeps`, `startHealthServer`) uses npm/Bun packages with no mise dependency. By the time any cron fires and spawns Claude via `createRunClaude`, mise startup (Step 3) will have completed and shim paths will be in PATH. Clean fix; mirrors vitals-os#1471.

PR #284, opened 2026-06-11
fix(metrics): consume the admin session (admin_session) instead of vitals_session

dmcaulay:
## Why
The metrics service's session auth read a `vitals_session` cookie, **required** a `name` claim, and redirected to `/auth/login`. But nothing in the Shipwright suite issues `vitals_session` — the **admin** service is the auth provider and issues an `admin_session` JWT (`{ userId, email }`, HS256, signed with `SHIPWRIGHT_SESSION_SECRET`). So the metrics dashboard/session path never authenticated end-to-end.

## What
Make metrics reuse the admin's session directly:
- Read the **`admin_session`** cookie in both `session-middleware.ts` and the combined-auth in `api.ts`.
- Treat **`name` as optional**, falling back to `email` (the admin JWT omits it).
- Redirect unauthenticated dashboard requests to **`/admin/login`**.
- Centralized as `SESSION_COOKIE` / `ADMIN_LOGIN_PATH`.

## Notes
- The owner-role check (`GET /accounts/users/{id}`) is unchanged; deployments that reuse the admin session should set `METRICS_REQUIRE_OWNER_ROLE=false` since the admin login is already allowlist-gated (that endpoint isn't served by the admin).
- Tests + docs updated; full `bun test metrics/` is green (643 pass / 0 fail).

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -143,7 +143,7 @@ Used only by `agent/src/run-migration.ts` when migrating agents from the Vitals 
 | Name | Type | Default | Description |
 |---|---|---|---|
 | `DATABASE_URL_SHIPWRIGHT_ADMIN` | `string` | required | Postgres connection string for the admin service schema (e.g. `postgresql://user:pass@host:5432/db`). |
-| `SHIPWRIGHT_SESSION_SECRET` | `string` | — | HS256 secret for signing session cookies. Used by the metrics service (`vitals_session` cookie) and the admin service (`admin_session` cookie). |
+| `SHIPWRIGHT_SESSION_SECRET` | `string` | — | HS256 secret for the `admin_session` cookie. The admin service signs it on Google-OAuth login; the metrics service verifies it to reuse the same session (the two must share the value). |
 | `SHIPWRIGHT_ENCRYPTION_KEY` | `string` | — | 64-char hex (32 bytes) for AES-256-GCM encryption of secrets at rest. **If unset, secrets are stored in plain text** — always set this in any real deployment. |
 | `SHIPWRIGHT_ADMIN_ALLOWED_EMAILS` | `string` | — | Comma-separated list of Google email addresses permitted to log in to the admin UI. |
 | `SHIPWRIGHT_ADMIN_APP_BASE_URL` | `string` | `http://localhost:{PORT}` | Public base URL of the admin service, used to construct the Google OAuth redirect URI. |
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -81,9 +81,9 @@ All metric endpoints are `GET`, return JSON, and accept the same date-window que
 The `/metrics/*` endpoints accept **either** credential:
 
 - **Bearer API key** — tokens parsed from `METRICS_API_KEYS`. Scoped tokens (scope `!== "*"`) are rejected with `403` on metrics routes.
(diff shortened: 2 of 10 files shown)


dmcaulay (commented):
COMMENT — one functional bug before merge: `logoutAction: "/auth/logout"` in `dashboard-page.ts:35` should be `/admin/logout` — the admin service owns auth and its logout endpoint is `POST /admin/logout`. Posting to `/auth/logout` returns 404 (no such route in metrics) and leaves `admin_session` intact, so users can't log out. One-line fix. Everything else is solid: the `admin_session` switch is correct, the shared toolbar extraction is clean, and the health server reorder is a good Kubernetes improvement.

dodizzle (commented):
COMMENT — Two items before merge:

**1. Owner-role footgun** (`metrics/src/lib/accounts-client.ts:136`): `HttpAccountsClient.getUser(id)` calls `GET /accounts/users/{id}` on the admin service URL (`METRICS_ADMIN_URL`), but the admin only exposes `/agents/*`. If `METRICS_REQUIRE_OWNER_ROLE=true` is set in a deployment env file, every session-authenticated request will hard-fail with 404 from admin, silently locking out all browser users. The default is `false` (safe), but there's no runtime warning when this combination is misconfigured. Add a startup log: `if (requireOwnerRole) console.warn('[metrics] METRICS_REQUIRE_OWNER_ROLE is enabled — ensure your client URL serves /accounts/users/{id}, not the Shipwright admin service')`.

**2. basePath injection** (`metrics/src/dashboard/dashboard-page.ts`, `window.__METRICS_BASE__` line): `window.__METRICS_BASE__ = '${base}'` injects `basePath` into a JS string literal without encoding. A single quote in the value breaks the script; a `</script>` sequence would allow injection. Since this comes from an operator env var (`METRICS_BASE_PATH`) the risk is low, but the fix is trivial: use `JSON.stringify`: `window.__METRICS_BASE__ = ${JSON.stringify(base)};` (drop the surrounding single quotes).

dmcaulay (commented):
COMMENT — CI is broken on the fix commit. The `JSON.stringify(base)` change is correct, but the snapshot and the new inline test were not updated to match. 5 tests failing: `expect(html).toContain("window.__METRICS_BASE__ = '/sw'")` expects single-quoted output, but `JSON.stringify('/sw')` produces `"/sw"`. Fix: update the assertion to match double-quoted output and re-run `bun test --update-snapshots metrics/src/dashboard/dashboard-page.unit.test.ts`. All other findings from both prior passes are addressed — this is the only blocker.

dmcaulay (commented):
APPROVE — All prior findings resolved: logout URL fixed to `/admin/logout`, basePath injection patched with `JSON.stringify`, owner-role startup warning added, CI fixed with snapshot regen + e2e env var rename. Implementation is correct and consistent across `session-middleware.ts` and `api.ts`. CI green.

dodizzle (approved):
Re-review after 3 follow-up commits — both prior findings resolved.

✅ **Owner-role warning**: `console.warn` now emitted in `createMetricsApp` when `requireOwnerRole=true` is set with an admin URL that lacks `/accounts/users/{id}`.

✅ **basePath encoding**: `window.__METRICS_BASE__ = ${JSON.stringify(base)};` — correctly uses `JSON.stringify` now.

The additional `SHIPWRIGHT_SESSION_SECRET` pin in the e2e spawn env is correct. CI is green. No new issues.

**APPROVE**

PR #285, opened 2026-06-11
feat(metrics): dev-auth dashboard bypass so task stack shows real metrics

dodizzle:
## Summary
Enables both the admin and metrics dashboards in the local `task stack` dev loop, with the metrics dashboard showing **real** (sqlite-persisted) data the agent forwards — not fixtures.

Adds a `dashboardDevAuth` flag (env: `METRICS_DASHBOARD_DEV_AUTH=true`) to the metrics service — the analogue of admin's `ADMIN_DEV_AUTH`. It relaxes auth on **both** `/dashboard` and `/metrics/*` (no login flow exists in the stack) while keeping the real provider.

## Why
In `task stack`, the metrics pane ran sqlite mode with no session secret and no offline flag, so `/dashboard` 500'd on an empty HS256 key. The admin toolbar's *Metrics* link also fell back to `/sw/dashboard` on the admin host (:3001), which 404s — the dashboard actually lives on :3460.

## Changes
- `metrics/src/api.ts` — `dashboardDevAuth` dep: bypass `/dashboard` session gate (renders "Dev User") + `/metrics/*` auth (synthetic caller). Off by default.
- `metrics/src/server.ts` — reads `METRICS_DASHBOARD_DEV_AUTH` into deps (postgres/sqlite/default branches).
- `scripts/dev-tmux.ts` — metrics pane sets the flag; admin pane sets `METRICS_DASHBOARD_URL=http://localhost:3460/dashboard` so the toolbar link resolves.
- Tests: new `dashboard-dev-auth.smoke.test.ts` (bypass works with no credentials; off-by-default still 401) + `dev-tmux.unit.test.ts` assertions for the pane env.

## Verification
- Full suite: 2577 pass / 0 fail. Lint + typecheck clean.
- Live: `/dashboard` 200, `/metrics/summary` 200 with real sqlite data (no auth), dashboard renders "Dev User", admin Metrics link → :3460.

### DIFF
--- a/metrics/src/server.ts
+++ b/metrics/src/server.ts
@@ -37,6 +37,9 @@ const offlineMode = mode === "fixtures";
 
 const port = Number(process.env.METRICS_API_PORT ?? 3460);
 const basePath = process.env.METRICS_BASE_PATH ?? "";
+// Local-dev auth bypass for /dashboard + /metrics/* (no login flow in `task stack`).
+// Keeps the real provider — only relaxes auth. Never enable in production.
+const dashboardDevAuth = process.env.METRICS_DASHBOARD_DEV_AUTH === "true";
 
 if (!process.env.METRICS_ADMIN_URL && process.env.METRICS_ACCOUNTS_URL) {
   console.warn(
@@ -93,6 +96,7 @@ if (mode === "fixtures") {
     sessionSecret: process.env.SHIPWRIGHT_SESSION_SECRET ?? "",
     requireOwnerRole: process.env.METRICS_REQUIRE_OWNER_ROLE === "true",
     dashboardToken: process.env.METRICS_DASHBOARD_TOKEN,
+    dashboardDevAuth,
     basePath,
   };
   console.log(
@@ -108,6 +112,7 @@ if (mode === "fixtures") {
     sessionSecret: process.env.SHIPWRIGHT_SESSION_SECRET ?? "",
     requireOwnerRole: process.env.METRICS_REQUIRE_OWNER_ROLE === "true",
     dashboardToken: process.env.METRICS_DASHBOARD_TOKEN,
+    dashboardDevAuth,
     basePath,
   };
   console.log(
@@ -118,6 +123,7 @@ if (mode === "fixtures") {
     sessionSecret: process.env.SHIPWRIGHT_SESSION_SECRET ?? "",
     requireOwnerRole: process.env.METRICS_REQUIRE_OWNER_ROLE === "true",
     dashboardToken: process.env.METRICS_DASHBOARD_TOKEN,
+    dashboardDevAuth,
     basePath,
   };
 }
--- a/metrics/src/server.ts
+++ b/metrics/src/server.ts
@@ -40,6 +40,10 @@ const basePath = process.env.METRICS_BASE_PATH ?? "";
 // Local-dev auth bypass for /dashboard + /metrics/* (no login flow in `task stack`).
 // Keeps the real provider — only relaxes auth. Never enable in production.
 const dashboardDevAuth = process.env.METRICS_DASHBOARD_DEV_AUTH === "true";
+if (dashboardDevAuth && process.env.NODE_ENV === "production") {
+  console.error("[metrics-api] FATAL: METRICS_DASHBOARD_DEV_AUTH cannot be enabled in production");
+  process.exit(1);
+}
 
 if (!process.env.METRICS_ADMIN_URL && process.env.METRICS_ACCOUNTS_URL) {
   console.warn(
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -190,6 +190,7 @@ Optional. When unset, voice transcription and synthesis are disabled.
 |---|---|---|---|
 | `SHIPWRIGHT_DEV_CHAT` | `bool` | `false` | Enables the unauthenticated `POST /chat` endpoint. Blocked at startup when `NODE_ENV=production`. |
 | `ADMIN_DEV_AUTH` | `bool` | `false` | Enables `GET /admin/dev-login` (bypasses Google OAuth, mints a dev session). Blocked when `NODE_ENV=production`. |
+| `METRICS_DASHBOARD_DEV_AUTH` | `bool` | `false` | Bypasses `/dashboard` session auth and `/metrics/*` API auth for local dev. Must not be enabled in production — exits with an error if `NODE_ENV=production`. |
 
 ---
 
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -94,6 +94,7 @@ The **dashboard** is protected by the session cookie middleware; an invalid/abse
 | `METRICS_OFFLINE` | | `false` | When `true`, skips PostHog env gate, injects fixture data, and bypasses dashboard session auth. |
 | `METRICS_DATABASE_URL` | ✅ (postgres mode) | — | Postgres connection URL (`postgres://...`). Selects postgres mode. Takes precedence over `DATABASE_URL_METRICS`. |
 | `DATABASE_URL_METRICS` | | — | Alias for `METRICS_DATABASE_URL`. Accepted when `METRICS_DATABASE_URL` is absent. |
+| `METRICS_BASE_PATH` | | — | URL path prefix the app is mounted at (e.g. `/sw`). All routes including `/dashboard` and `/metrics/*` are served under this prefix. |
 | `METRICS_API_PORT` | | `3460` | Listen port. |
 | `METRICS_API_KEYS` | | — | Comma-parsed Bearer API keys for `/metrics/*`. |
 | `SHIPWRIGHT_SESSION_SECRET` | | — | HS256 secret for verifying the `admin_session` cookie. |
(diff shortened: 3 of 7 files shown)


dmcaulay (approved):
COMMENT — Clean implementation of the `dashboardDevAuth` bypass, exactly analogous to `ADMIN_DEV_AUTH`. All three `server.ts` branches covered, DI seam is correct, tests cover both the bypass path and off-by-default path. CI green.

One minor suggestion: add a startup `console.warn` when `dashboardDevAuth` is enabled (same pattern as `requireOwnerRole` warning in #284) so operators have a runtime signal if they accidentally set the flag in a non-dev environment. Not blocking.

How do we know that this isn't turned on in production? Do we have checks for the `/chat` endpoint on the agent?

Note: this PR and #284 both touch `api.ts` in overlapping regions — whichever lands second will need a rebase.

dmcaulay:
COMMENT - How do we know that this isn't turned on in production? Do we have checks for the /chat endpoint on the agent?

dodizzle (commented):
COMMENT — Two items before merge:

**1. No NODE_ENV=production guard** (`metrics/src/server.ts:42`): `ADMIN_DEV_AUTH` is hard-blocked in production by `admin/src/dev-auth-guard.ts` (`if NODE_ENV === 'production' return false`). `METRICS_DASHBOARD_DEV_AUTH` has no equivalent — `const dashboardDevAuth = process.env.METRICS_DASHBOARD_DEV_AUTH === "true"` is unconditional. The risks are asymmetric: `ADMIN_DEV_AUTH` bypasses one dev-login route; this bypass covers ALL auth on `/dashboard` + every `/metrics/*` endpoint with the real provider data. A dev env file copied to production would silently expose the full metrics API. Add a production guard matching the admin pattern (see `dev-auth-guard.ts`), or at minimum: `if (dashboardDevAuth && process.env.NODE_ENV === 'production') { console.error('[metrics-api] FATAL: METRICS_DASHBOARD_DEV_AUTH cannot be enabled in production'); process.exit(1); }`.

**2. METRICS_DASHBOARD_DEV_AUTH not documented** (`docs/configuration.md:192`): `docs/configuration.md` has a Dev-only table with `ADMIN_DEV_AUTH` + a note about `NODE_ENV=production` guards; `docs/metrics.md` has a full env reference. This new var appears in neither. Add it to both tables with a note that it must not run in production.

The DI seam and test coverage are clean — just these two before merge.

metrics/src/server.ts
```diff
@@ -37,6 +37,9 @@ const offlineMode = mode === "fixtures";
 
 const port = Number(process.env.METRICS_API_PORT ?? 3460);
 const basePath = process.env.METRICS_BASE_PATH ?? "";
+// Local-dev auth bypass for /dashboard + /metrics/* (no login flow in `task stack`).
+// Keeps the real provider — only relaxes auth. Never enable in production.
+const dashboardDevAuth = process.env.METRICS_DASHBOARD_DEV_AUTH === "true";
```
dodizzle, line 42: No production guard here. `ADMIN_DEV_AUTH` is blocked via `isDevAuthAllowed()` in `admin/src/dev-auth-guard.ts` when `NODE_ENV=production` — this flag should follow the same pattern. Unlike `ADMIN_DEV_AUTH` (one dev-login route), `dashboardDevAuth=true` bypasses all auth on both `/dashboard` and every `/metrics/*` endpoint with real provider data.

Suggested fix:
```typescript
const dashboardDevAuth =
  process.env.NODE_ENV !== "production" &&
  process.env.METRICS_DASHBOARD_DEV_AUTH === "true";
if (process.env.METRICS_DASHBOARD_DEV_AUTH === "true" && process.env.NODE_ENV === "production") {
  console.error("[metrics-api] FATAL: METRICS_DASHBOARD_DEV_AUTH cannot be enabled in production");
  process.exit(1);
}
```

dmcaulay, line 42: No NODE_ENV=production guard here. `ADMIN_DEV_AUTH` is hard-blocked in prod by `admin/src/dev-auth-guard.ts`; this flag has no equivalent and covers a wider surface (all `/dashboard` + `/metrics/*`). Add: `if (dashboardDevAuth && process.env.NODE_ENV === 'production') { console.error('[metrics-api] FATAL: METRICS_DASHBOARD_DEV_AUTH cannot be enabled in production'); process.exit(1); }`

dmcaulay (commented):
COMMENT — Revisiting my earlier APPROVE: I missed a production safety gap. Two issues before merge:

**1. No NODE_ENV=production guard** (`metrics/src/server.ts:42`): `ADMIN_DEV_AUTH` is hard-blocked in production by `admin/src/dev-auth-guard.ts` (`if (env.NODE_ENV === 'production') return false`). `METRICS_DASHBOARD_DEV_AUTH` has no equivalent — the flag is unconditional. The scope is broader: this bypass covers ALL auth on `/dashboard` + every `/metrics/*` endpoint with real provider data, not just one dev-login route. A dev env file copied to production silently exposes the full metrics API. Add a production guard: `if (dashboardDevAuth && process.env.NODE_ENV === 'production') { console.error('[metrics-api] FATAL: METRICS_DASHBOARD_DEV_AUTH cannot be enabled in production'); process.exit(1); }` — or extract a shared `isDevAuthAllowed()` matching the admin pattern.

**2. Undocumented** (`docs/configuration.md`, `docs/metrics.md`): The dev-only table in `docs/configuration.md` lists `ADMIN_DEV_AUTH` with a production-safety note; `docs/metrics.md` has the full env reference. `METRICS_DASHBOARD_DEV_AUTH` is in neither. Add to both tables with a note that it must not be enabled in production (once the guard is in place).

Everything else is well-executed: DI seam, test coverage, toolbar extraction, basePath support, and the deprecation warning are all clean.

metrics/src/server.ts
```diff
@@ -37,6 +37,13 @@ const offlineMode = mode === "fixtures";
 
 const port = Number(process.env.METRICS_API_PORT ?? 3460);
 const basePath = process.env.METRICS_BASE_PATH ?? "";
```
dodizzle, line 39: `METRICS_BASE_PATH` is read here but not listed in `docs/metrics.md`'s Environment table (or `docs/configuration.md`). This PR documents `METRICS_DASHBOARD_DEV_AUTH` and `METRICS_ADMIN_URL` in both files — `METRICS_BASE_PATH` should get the same treatment. Operators can't discover it without reading source.

Suggested row for `docs/metrics.md` Environment table:
```
| `METRICS_BASE_PATH` | | — | URL path prefix the app is mounted at (e.g. `/sw`). All routes including `/dashboard` and `/metrics/*` are served under this prefix. |
```

dodizzle (commented):
COMMENT — Both prior findings fully resolved: the `NODE_ENV=production` guard is in place (`server.ts:43-46`, `process.exit(1)` on startup), and `METRICS_DASHBOARD_DEV_AUTH` is documented in both `docs/configuration.md` and `docs/metrics.md`. One new gap: `METRICS_BASE_PATH` was introduced via the merge from #282 and is undocumented. Add a single row to `docs/metrics.md`'s Environment table and this is ready to merge.

dodizzle (commented):
APPROVE — All three prior findings are fully resolved: the `NODE_ENV=production` guard exits at startup (`server.ts:43-46`), `METRICS_DASHBOARD_DEV_AUTH` is documented in both `docs/configuration.md` and `docs/metrics.md`, and `METRICS_BASE_PATH` is now documented in `docs/metrics.md` following the established scoping pattern for service-specific env vars. Implementation mirrors `ADMIN_DEV_AUTH` exactly, smoke tests cover the auth-bypass-on/off contract at the correct layer. Ready to merge.

dmcaulay (approved):
APPROVE — Re-review after all findings resolved.

All three rounds of feedback addressed: NODE_ENV=production guard in place (`server.ts:43-46`, `process.exit(1)`), `METRICS_DASHBOARD_DEV_AUTH` documented in both `docs/configuration.md` and `docs/metrics.md`, and `METRICS_BASE_PATH` documented in `docs/metrics.md`. No new issues found.

The additions since review #1 (session cookie rename `vitals_session` → `admin_session`, `@shipwright/lib` extraction, `METRICS_ADMIN_URL` migration, startup order change, `GET /agents` endpoint) are all clean. CI green.

dodizzle:
Good question — both angles are covered:

**Production guard:** `metrics/src/server.ts:43-46` exits at startup if `METRICS_DASHBOARD_DEV_AUTH=true` and `NODE_ENV=production`:
```ts
if (dashboardDevAuth && process.env.NODE_ENV === "production") {
  console.error("[metrics-api] FATAL: METRICS_DASHBOARD_DEV_AUTH cannot be enabled in production");
  process.exit(1);
}
```
So even if someone accidentally sets the flag in a production deploy, the process refuses to start. This mirrors the `ADMIN_DEV_AUTH` pattern already in the codebase.

**`/chat` endpoint (`SHIPWRIGHT_DEV_CHAT`):** That flag is not part of this PR — it was introduced separately and has its own guard. Out of scope here, but happy to point you to the right file if you want to verify it.

PR #286, opened 2026-06-11
fix(agent): spawn index.ts instead of run-agent.ts in entrypoint

dmcaulay:
## Summary

- `entrypoint.ts` was spawning `run-agent.ts` (minimal dev-only HTTP server — no Slack, no crons) instead of `index.ts` (full Slack Bolt + cron sync + graceful shutdown)
- This matches the vitals-os pattern: `agent/scripts/entrypoint.ts` does setup then hands off to `index.ts`
- Updated stale comment in `entrypoint-main.ts` to reflect the corrected spawn target

## Test plan

- [ ] Existing entrypoint integration tests pass (13/13 — `spawnAgentServer` call count is asserted, not the filename)
- [ ] K8s agent starts Slack app and schedules crons on next deploy

### DIFF
--- a/agent/src/entrypoint-main.ts
+++ b/agent/src/entrypoint-main.ts
@@ -8,7 +8,7 @@
  *
  * The health server is started in-process on SHIPWRIGHT_HEALTH_PORT (default 3459)
  * BEFORE the startup sequence so K8s liveness probes are reachable during init.
- * The agent server (run-agent.ts) runs as a subprocess — spawnAgentServer.
+ * The agent server (index.ts) runs as a subprocess — spawnAgentServer.
  */
 
 import { join } from "node:path";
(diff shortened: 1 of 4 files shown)


dmcaulay (commented):
COMMENT — One-line fix is correct: `entrypoint.ts` Step 8 now spawns `index.ts` (full Slack + crons) instead of `run-agent.ts` (dev-only `/chat` server). Doc update is a net improvement.

One gap: `entrypoint.integration.test.ts` asserts `spawnCalls.length === 1` but never checks `spawnCalls[0].args` — the original bug passed all 13 tests. Worth adding `expect(spawnCalls[0].args[1]).toContain('index.ts')` at lines 182 and 256 before merge so this class of regression is caught immediately.

dodizzle (approved):
APPROVE — Bug fix is correct (spawn `index.ts`, not `run-agent.ts`), prior review comment addressed with properly scoped test assertions at both sites, docs accurate.

dmcaulay (commented):
APPROVE — Bug fix is correct (`index.ts` is the right spawn target), prior review comment addressed with test assertions at both sites, docs improved. CI green.

PR #289, opened 2026-06-11
fix(admin): add enabled toggle to PATCH /crons/:cronId

dmcaulay:
## Summary

- `PATCH /agents/:id/crons/:cronId` now supports two modes:
  - **Content update** (existing): `schedule`+`prompt` required together; `channel`, `user`, `silent`, `preCheck`, and now `enabled` are optional
  - **Enabled-only toggle** (new): send just `{"enabled": false}` — calls `setEnabled()` without touching schedule or prompt. Works for both custom and system crons
- `AgentCronJobService.update()` now accepts and passes through `enabled`
- `AdminDeps` now picks `setEnabled` from `AgentCronJobService`
- Empty body or partial content (one of schedule/prompt without the other) returns 400

## Why

Two issues in the current API:
1. PATCH required `schedule`+`prompt` and silently ignored `enabled` — no way to toggle a cron on/off without deleting and recreating it
2. System crons can't be deleted (403 by design), so there was no API mechanism at all to enable/disable them

## Test plan

- [ ] `PATCH {"enabled": false}` → 200, `cron.enabled` is false
- [ ] `PATCH {"schedule": "...", "prompt": "...", "enabled": false}` → 200, content + enabled both updated
- [ ] `PATCH {"schedule": "..."}` (missing prompt) → 400
- [ ] `PATCH {}` (empty body) → 400
- [ ] 7 new smoke tests added covering all paths; 55 total passing


admin/src/agents-api.ts
```diff
@@ -221,23 +228,70 @@ export function createAdminApp(deps: AdminDeps): Hono<AdminAuthEnv> {
...
+      if (hasEnabled) {
+        cron = await agentCronJobService.setEnabled(
+          agentId,
+          cronId,
+          body.enabled as boolean,
+        );
+      }
+    }
+
+    return c.json({ cron: cron! });
```
dodizzle: Biome `noNonNullAssertion` rejects `cron!` — this is the root cause of the failing CI lint job. TypeScript needs the `!` because it can't prove `cron` is assigned in all `else` paths, even though the 400-guard above makes the uninitialized path unreachable. Refactor the `else` branch to guarantee assignment without the assertion — e.g. a ternary: `cron = hasEnabled ? await agentCronJobService.setEnabled(...) : await agentCronJobService.updatePreCheck(...)`.

admin/src/agents-api.ts
```diff
@@ -221,23 +228,70 @@ export function createAdminApp(deps: AdminDeps): Hono<AdminAuthEnv> {
...
+        prompt: body.prompt as string,
+        channel: body.channel,
+        user: body.user,
+        silent: body.silent,
+        preCheck: body.preCheck,
+        enabled: body.enabled,
+      });
+    } else {
+      // Orthogonal-field-only update — apply each independently
+      if (hasPreCheck) {
```
dodizzle, line 278: When both `hasPreCheck` and `hasEnabled` are true here, `cron` is set from `updatePreCheck` then immediately overwritten by `setEnabled`. Both DB writes happen, but the response only reflects `setEnabled`'s return value — the `preCheck` field in the response body will be stale. After applying both writes, return `await agentCronJobService.get(agentId, cronId)` so the response reflects the combined final state.

dodizzle (commented):

**1. Biome lint violation blocking CI** (`admin/src/agents-api.ts:294`): `return c.json({ cron: cron! })` uses a non-null assertion, which is forbidden by Biome's `noNonNullAssertion` rule (active via `"recommended": true`). This is the root cause of the failing lint job. The `!` is needed because TypeScript can't prove `cron` is always assigned before line 294 through the earlier 400-guard. Fix: refactor the `else` branch so `cron` is always assigned unconditionally — e.g. use a ternary for the single-field case and handle the dual-field case separately.

**2. Dual-orthogonal update returns stale `preCheck`** (`admin/src/agents-api.ts:278`): When both `hasPreCheck` and `hasEnabled` are true in the `else` branch, `cron` is assigned from `updatePreCheck`, then immediately overwritten by `setEnabled`. Both DB writes execute, but the response body comes from `setEnabled` only — the returned cron's `preCheck` reflects whatever `setEnabled` returns, not the newly-written preCheck value. Callers checking `response.cron.preCheck` after a combined update see a stale value. Fix: after both writes, fetch the final state with `get(agentId, cronId)` and return that, or build a single update covering both fields.

admin/src/agents-api.ts
dmcaulay: `cron!` is the active CI lint failure — `lint/style/noNonNullAssertion` is enabled and this has no `biome-ignore` suppression. TypeScript can't prove `cron` is always assigned in the else branch even though the empty-body guard guarantees it. Collapsing to a single `updateOrthogonal()` service call (one Prisma write for both preCheck and enabled) eliminates the `!` entirely and fixes the partial-update race below.

dmcaulay (commented):

1. **`cron!` non-null assertion is the CI lint failure** (`admin/src/agents-api.ts`): `let cron: AgentCronJob` is uninitialized; the else branch assigns it conditionally, so TypeScript can't prove it's always set. `lint/style/noNonNullAssertion` is active (Biome recommended) and other files use `biome-ignore` to suppress it — this one doesn't. Fix: collapse the two sequential service calls into a single `updateOrthogonal(agentId, cronId, { preCheck?, enabled? })` method that writes both fields in one Prisma update — `cron` is then always assigned and `!` isn't needed at all.

2. **Partial-update risk in orthogonal else branch** (`admin/src/agents-api.ts`): When `{ preCheck, enabled }` both provided without schedule/prompt, `updatePreCheck()` and `setEnabled()` run as two separate unguarded Prisma writes. If the second throws, the first is already committed. The response also only reflects the `setEnabled` result — the updated preCheck isn't visible in the response body. Same fix as above: single atomic Prisma call.

Everything else is solid — validation coverage, test breadth (7 new paths), and the backwards-compatible API shape are all good.

admin/src/agents-api.ts
```diff
@@ -221,22 +228,77 @@ export function createAdminApp(deps: AdminDeps): Hono<AdminAuthEnv> {
...
+        cron = await agentCronJobService.updatePreCheck(
+          agentId,
+          cronId,
+          body.preCheck as string | null,
+        );
+      }
+      if (hasEnabled) {
+        cron = await agentCronJobService.setEnabled(
```
dodizzle, line 286: When `hasPreCheck && hasEnabled` are both true, `cron` from `updatePreCheck` is overwritten by `setEnabled` here. Both DB writes execute, but the response body only reflects the `setEnabled` result — `cron.preCheck` in the response will show the old value.

Fix: after both writes, re-fetch with `agentCronJobService.get(agentId, cronId)` and return that, or collapse into a single `prisma.agentCronJob.update` covering both fields. The `preCheck+enabled` smoke test should also assert `body.cron.preCheck` to catch this in CI.

dodizzle (commented):
COMMENT — Prior finding #1 (Biome `noNonNullAssertion` / CI failure) is fixed. One issue remains before merge:

**Dual preCheck+enabled PATCH returns stale `preCheck` in response body** (`admin/src/agents-api.ts:286`): When a PATCH sends both `preCheck` and `enabled`, `updatePreCheck` runs first and assigns `cron`, then `setEnabled` unconditionally overwrites `cron` with its result. The `setEnabled` query only updates the `enabled` column — the returned row reflects the old `preCheck` value. The response body's `cron.preCheck` will show the stale value. The smoke test for `preCheck+enabled` only asserts `res.status === 200`, not `body.cron.preCheck`, so this is invisible in CI.

Fix: after both writes, re-fetch with `agentCronJobService.get(agentId, cronId)` and return that, or collapse into a single `prisma.agentCronJob.update` covering both fields. Also add a `body.cron.preCheck` assertion to the `preCheck+enabled` smoke test.

dmcaulay (commented):
APPROVE — prior blocking issue resolved. `cron!` non-null assertion replaced with `let cron: AgentCronJob | undefined` + runtime guard; CI is green. The atomicity gap on `{preCheck, enabled}` combined orthogonal update (two sequential Prisma writes) is noted but low-risk: both fields are independently idempotent and a retry recovers. The stale-response concern from the prior review was incorrect — Prisma `update()` returns the full row, so the response correctly reflects both writes. Ready to merge.

dodizzle (approved):
APPROVE — All prior findings addressed. The dual-orthogonal PATCH path now re-fetches state after both writes (`agentCronJobService.get(agentId, cronId)`), response fidelity is correct, CI is green, and the fix is properly tested with assertions on both `body.cron.preCheck` and `body.cron.enabled`.

dmcaulay (commented):
APPROVE — All prior findings addressed. Stale-response on combined `preCheck+enabled` PATCH now fixed: `get(agentId, cronId)` fetches final state after both writes, and the smoke test asserts both `body.cron.preCheck` and `body.cron.enabled`. CI is green. Ready to merge.

PR #291, opened 2026-06-11
fix(agent): install shipwright plugin from local disk, not remote GitHub slug

dmcaulay:
## Summary

- Mirrors the vitals-os approach: register the repo root as a local marketplace via `claude plugin marketplace add <repoRoot>`, then install `shipwright@app-vitals`
- The previous remote GitHub slug approach (`shipwright@app-vitals/shipwright`) never worked — agents without GitHub access can't resolve it, and the `marketplace add` registration step was missing entirely
- Copies `plugins/` and `.claude-plugin/` into the Dockerfile runtime stage so the plugin source is on disk in both Docker and direct (Pi) deployments
- Removes `SHIPWRIGHT_LOCAL_MARKETPLACE` env var (no longer needed — local install is always used)

## Test plan

- [ ] All existing `installPlugins` integration tests updated and passing (67 tests green)
- [ ] Verify `claude plugin marketplace add` appears as first call in exec mock
- [ ] Verify `shipwright@app-vitals` used for install/update (not GitHub slug)
- [ ] Build Docker image and confirm plugin is installed at startup

### DIFF
--- a/agent/src/setup.integration.test.ts
+++ b/agent/src/setup.integration.test.ts
@@ -739,6 +739,29 @@ describe("installPlugins", () => {
     expect(calls).toHaveLength(3);
   });
 
+  it("continues when marketplace add exits non-zero", async () => {
+    const calls: Array<{ args: string[] }> = [];
+    const mockExec = async (
+      _cmd: string,
+      args: string[],
+      _opts: { cwd: string },
+    ) => {
+      calls.push({ args });
+      const isMarketplaceAdd = args[1] === "marketplace";
+      return {
+        stdout: isMarketplaceAdd ? "network error" : "",
+        exitCode: isMarketplaceAdd ? 1 : 0,
+      };
+    };
+
+    await expect(
+      installPlugins(mockExec, testHome, [], "/repo/root"),
+    ).resolves.toBeUndefined();
+
+    // All 3 calls still happened despite marketplace add failure
+    expect(calls).toHaveLength(3);
+  });
+
   it("is a silent no-op when plugins are already installed and up-to-date", async () => {
     const calls: Array<{ args: string[] }> = [];
     const mockExec = async (
--- a/agent/src/setup.integration.test.ts
+++ b/agent/src/setup.integration.test.ts
@@ -591,7 +591,7 @@ describe("runMiseStartup", () => {
 // ---------------------------------------------------------------------------
 
 describe("installPlugins", () => {
-  it("installs shipwright default plugin first (shipwright@app-vitals/shipwright)", async () => {
+  it("registers local marketplace then installs shipwright@app-vitals", async () => {
     const calls: Array<{ cmd: string; args: string[] }> = [];
     const mockExec = async (
       cmd: string,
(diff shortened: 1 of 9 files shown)


agent/src/setup.integration.test.ts
```diff
@@ -719,11 +732,11 @@ describe("installPlugins", () => {
...
 
     await expect(
-      installPlugins(mockExec, testHome, []),
+      installPlugins(mockExec, testHome, [], "/repo/root"),
     ).resolves.toBeUndefined();
 
     // All calls still happened despite install failures
-    expect(calls).toHaveLength(2);
+    expect(calls).toHaveLength(3);
   });
```
dodizzle, line 740: This test covers install failures but not `marketplace add` failure — the mock returns exitCode 1 only when `args[1] === "install"`, so the `marketplace add` call (args[1] = `"marketplace"`) always succeeds here. Consider adding a test where `marketplace add` returns exitCode 1 and verify all 3 calls still happen and the function resolves.

dodizzle (approved):
APPROVE — Correct fix, clean implementation.

The `SHIPWRIGHT_REPO_ROOT` path resolution is correct for both Docker and Pi deployments. Dockerfile multi-stage COPY is properly sequenced. Tests are fully updated in the same PR with the injectable `shipwrightRepoRoot` seam.

One suggestion (non-blocking): the 'continues when a plugin install exits non-zero' test exercises install failures but not a `marketplace add` failure (the mock keys on `args[1] === "install"`, so the add call always gets exitCode 0 in that test). Worth adding a test where `args[1] === "marketplace"` returns exitCode 1 to verify warn-and-continue behavior for the new step.

dmcaulay (commented):
COMMENT — The core fix is correct (local marketplace registration matches vitals-os pattern, tests updated, CI green), but `SHIPWRIGHT_LOCAL_MARKETPLACE` cleanup is incomplete. The env var is removed from `setup.ts` but still referenced in 5 places: `scripts/dev-tmux.ts:210` (passed as Docker `-e` flag, now silently ignored), `scripts/dev-tmux.unit.test.ts:532` (asserts dead behavior), `docs/configuration.md:32`, `docs/agent.md:118`, and `site/src/pages/docs/configuration.astro:74` (public docs). Clean those up before merging.

dmcaulay (commented):
APPROVE — prior finding fully resolved. All five `SHIPWRIGHT_LOCAL_MARKETPLACE` dead references removed (`scripts/dev-tmux.ts`, `scripts/dev-tmux.unit.test.ts`, `docs/configuration.md`, `docs/agent.md`, `site/src/pages/docs/configuration.astro`). Dockerfile, `setup.ts` logic, and test coverage are all correct. No new issues.

dodizzle (approved):
APPROVE — Both prior findings fully addressed.

The `SHIPWRIGHT_LOCAL_MARKETPLACE` dead refs are gone from all 5 flagged locations (scripts, docs, site). The new `"continues when marketplace add exits non-zero"` integration test at `setup.integration.test.ts:742` correctly uses the injected mockExec seam, asserts the function resolves without throwing, and verifies all 3 subsequent calls still execute. CI is green.

PR #293, opened 2026-06-11
feat(ci): add build-agent.yml to publish shipwright-agent image

dmcaulay:
## Summary

- Adds `.github/workflows/build-agent.yml` — mirrors `build-admin.yml` exactly
- On push to `main`: tags `agent-vX.Y.Z`, authenticates to GCP via WIF, builds from `agent/Dockerfile`, pushes `shipwright-agent:agent-vX.Y.Z` to Artifact Registry
- Gating dependency for migrating GKE client agents (okwow, fern-dan, keaunu, sideby-dan, warchild) from the vitals-os runtime to the Shipwright harness

## Test plan

- [ ] Merge → CI triggers `build-agent.yml` → `agent-v0.1.0` tag created → image pushed to `us-west1-docker.pkg.dev/vitals-os-prod/vitals-os/shipwright-agent:agent-v0.1.0`
- [ ] Verify image in Artifact Registry before wiring into Helm

### DIFF
--- a/.github/workflows/build-agent.yml
+++ b/.github/workflows/build-agent.yml
@@ -1,0 +1,46 @@
+name: Build and Push Agent Image
+
+on:
+  push:
+    branches: [main]
+
+permissions:
+  contents: write
+  id-token: write
+
+jobs:
+  build-and-push:
+    name: build / tag / push
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Bump version and push tag
+        id: tag
+        uses: mathieudutour/github-tag-action@v6.2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          tag_prefix: "agent-v"
+          default_bump: patch
+
+      - name: Authenticate to GCP
+        uses: google-github-actions/auth@v2
+        with:
+          workload_identity_provider: ${{ secrets.WIF_PROVIDER }}
+          service_account: ${{ secrets.GCP_SA_EMAIL }}
+
+      - name: Configure Docker for Artifact Registry
+        run: gcloud auth configure-docker ${{ secrets.ARTIFACT_REGISTRY_HOST }} --quiet
+
+      - name: Build Docker image
+        run: |
+          docker build \
+            -f agent/Dockerfile \
+            -t ${{ secrets.ARTIFACT_REGISTRY_REPO }}/shipwright-agent:${{ steps.tag.outputs.new_tag }} \
+            .
+
+      - name: Push Docker image
+        run: |
+          docker push ${{ secrets.ARTIFACT_REGISTRY_REPO }}/shipwright-agent:${{ steps.tag.outputs.new_tag }}


.github/workflows/build-agent.yml
```diff
@@ -0,0 +1,46 @@
...
+    name: build / tag / push
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Bump version and push tag
+        id: tag
+        uses: mathieudutour/github-tag-action@v6.2
```
dodizzle, line 22: `mathieudutour/github-tag-action@v6.2` and `google-github-actions/auth@v2` (line 35) are mutable tags. This workflow runs with `id-token: write` + `contents: write` — pin both to their full commit SHA to prevent supply-chain compromise.

.github/workflows/build-agent.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Build Docker image
+        run: |
+          docker build \
+            -f agent/Dockerfile \
+            -t ${{ secrets.ARTIFACT_REGISTRY_REPO }}/shipwright-agent:${{ steps.tag.outputs.new_tag }} \
+            .
+
+      - name: Push Docker image
+        run: |
+          docker push ${{ secrets.ARTIFACT_REGISTRY_REPO }}/shipwright-agent:${{ steps.tag.outputs.new_tag }}
```
dodizzle, line 46: Only the versioned tag is pushed. Consider also pushing `latest`: `docker tag ${{ secrets.ARTIFACT_REGISTRY_REPO }}/shipwright-agent:${{ steps.tag.outputs.new_tag }} ${{ secrets.ARTIFACT_REGISTRY_REPO }}/shipwright-agent:latest && docker push ...`. Without it, every Helm migration must be updated per release.

dodizzle (commented):

**1. Mutable action version tags in a privileged workflow** (`.github/workflows/build-agent.yml`): `mathieudutour/github-tag-action@v6.2` and `google-github-actions/auth@v2` are referenced by mutable version tags. This workflow has `id-token: write` (GCP WIF) and `contents: write` — if either tag is force-pushed or the upstream account is compromised, an attacker can push arbitrary tags and mint GCP credentials. The same pattern exists in `build-admin.yml`/`build-metrics.yml`, but each new privileged workflow adds blast radius. Fix: pin both to their full commit SHA.

**2. No `latest` tag pushed** (`.github/workflows/build-agent.yml`): Only `shipwright-agent:agent-vX.Y.Z` is pushed. The five GKE client agents being migrated will need Helm values updated on every release. If any migration script expects a stable tag, the first deploy will `ImagePullBackOff`. Either push a `latest` tag after the versioned one, or document explicitly in the migration runbook that consumers must pin to the semver tag.

dodizzle (approved):
APPROVE — Prior findings (mutable action version pins, no `latest` tag) are inconsistent with the established patterns in `build-admin.yml` and `build-metrics.yml`, which use the same action versions and also push only versioned tags. `build-agent.yml` is a faithful mirror of the existing pipelines. No blocking issues.

dmcaulay (commented):
APPROVE — faithful mirror of `build-admin.yml` and `build-metrics.yml` with the correct per-artifact substitutions (`agent-v` tag prefix, `agent/Dockerfile`, `shipwright-agent` image name). CI green, no new issues introduced.

PR #299, opened 2026-06-12
fix(agent): install claude/node system-wide, not via mise (fixes 'claude is not a valid shim')

dmcaulay:
## Problem

After the okwow harness pod came up healthy (Slack connected), every `claude` invocation failed:

```
mise ERROR claude is not a valid shim. This likely means you uninstalled a tool
and the shim does not point to anything.
  at /app/agent/src/claude.ts:198 (_spawn → spawner(["claude", ...]))
```

## Root cause

claude was installed under a **mise-managed node** (`mise use --global node@20` → `npm i -g claude-code` → `mise reshim`), so `claude` became a **mise shim** (`/home/bun/.local/share/mise/shims/claude → mise`). But `runMiseStartup` repoints `MISE_DATA_DIR` at the PVC. When the image's claude shim runs, mise resolves the tool against the PVC mise dir — which has **no claude install** → "not a valid shim".

A seed-hack tried to paper over this by copying the image mise dir onto the PVC, but on a **reused** PVC (every migrated agent) the copy is skipped because `mise/shims` already exists, so it never helps.

This is a divergence from the **vitals-os runtime** this was extracted from, which installs Node via NodeSource and `npm i -g @anthropic-ai/claude-code` as a plain **system binary** (`/usr/bin/claude`) — entirely independent of mise. (Confirmed on the live okwow pod: `claude` → image shim → `mise`; `MISE_DATA_DIR` PVC has node/npm shims but no claude.)

## Fix (match vitals-os)

- **Dockerfile**: install Node 22 (NodeSource) + `@anthropic-ai/claude-code@2.1.170` system-wide in the base stage (as root). Drop `mise use --global node@20`, the mise-node npm install, and `mise reshim`. `claude`/`node` are now system binaries on `/usr/bin`, never mise shims.
- **setup.ts `runMiseStartup`**: drop the image→PVC seed/copy hack. Just point `MISE_DATA_DIR`/cache dirs at the PVC and install workspace-declared tools — identical to vitals-os. mise stays purely a runtime *workspace* tool-version manager.
- **tests**: replace the two seed-behavior tests with one asserting no image→PVC copy. `setup.integration.test.ts`: 50 pass / 0 fail; biome clean.

## No PVC cleanup needed

The okwow PVC's mise dir holds node/npm/pip shims (original vitals-os state, working) and **no claude shim** — the broken claude shim existed only in the image, which this change no longer creates. After deploy, `claude` resolves to `/usr/bin/claude`; `node`/`npm` keep resolving via the PVC mise shims, exactly as under the working vitals-os runtime on this same PVC.

## Rollout

Merge → `build-agent.yml` builds `agent-v0.2.4` → bump `shipwrightAgentSvc.image.tag` in vitals-os `values-prod.yaml` → deploy → validate `claude` runs in okwow (Slack message triggering a Bash/agent action).

### DIFF
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -51,6 +51,16 @@ RUN curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg \
 # non-root runtime user can invoke it).
 RUN curl https://mise.run | MISE_INSTALL_PATH=/usr/local/bin/mise sh
 
+# Install Node.js + Claude Code CLI system-wide (as root), matching vitals-os.
+# claude is a plain binary on PATH (/usr/bin/claude) backed by system node — it is
+# NOT a mise shim, so its resolution never depends on the active mise/workspace
+# context (which is what broke the previous mise-reshim approach: "claude is not a
+# valid shim"). mise stays purely a runtime *workspace* tool-version manager.
+RUN curl -fsSL https://deb.nodesource.com/setup_22.x | bash - \
+    && apt-get install -y --no-install-recommends nodejs \
+    && npm install -g @anthropic-ai/claude-code@2.1.170 \
+    && rm -rf /var/lib/apt/lists/*
+
 # ─── Build stage ──────────────────────────────────────────────────────────────
 
 FROM base AS build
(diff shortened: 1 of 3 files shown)


dodizzle (approved):
APPROVE — Clean fix for a confirmed production outage. Moving `claude`/`node` to system binaries in the Dockerfile base stage is the right call: it makes them immune to `MISE_DATA_DIR` repointing at the PVC, which was the root cause of the 'claude is not a valid shim' failure. The seed-copy hack removal is a good simplification — the hack had a known gap on reused PVCs anyway. No blocking issues. Tests updated correctly.

PR #301, opened 2026-06-12
fix(agent): symlink ~/.claude.json (MCP) + activate mise in bash (extraction parity)

dmcaulay:
## Problem

While migrating **okwow** onto the harness, Claude Code came up but the **Linear MCP wasn't available** (`claude mcp list` → "No MCP servers configured") — even though okwow's PVC has `/data/agent-home/claude.json` carrying its `linear-server` registration from the vitals-os runtime.

## Root cause — two runtime-wiring steps dropped in the vitals-os → shipwright extraction

**1. `~/.claude.json` symlink (MCP registrations).** Claude Code stores MCP server registrations (`claude mcp add`) in `~/.claude.json`, which lives **outside** `~/.claude/`. The vitals-os entrypoint symlinks **both** `~/.claude` (dir) **and** `~/.claude.json` (file) to the PVC — the extraction ported only the directory. So shipwright reads the **ephemeral image copy** of `~/.claude.json` (no MCP servers) instead of the PVC's.

**2. `mise activate bash` in `~/.bashrc`.** The shims-on-PATH prepend in `runMiseStartup` only reaches direct subprocesses. Claude Code's Bash tool spawns login/interactive shells that re-source profile files and reset PATH, so **workspace-declared mise tools** (e.g. **fern's** Rust toolchain) would be missing. vitals-os appends the activation line; shipwright didn't do it at all.

A full sweep confirmed these are the **only** two gaps: vitals-os has exactly two symlinks (both now ported) and the only mise difference was `mise activate` (now ported). All other mise wiring (`MISE_DATA_DIR`/`MISE_CACHE_DIR`/`XDG_CACHE_HOME`, `mise trust`/`install`, shims-on-PATH) already matched.

## Fix

- `ensureDotClaudeSymlink` also links `~/.claude.json → $AGENT_HOME/claude.json` (file-aware: no mkdir; replaces the real image file; dangling-until-first-write is fine). Same path the vitals-os runtime used, so migrated PVCs are picked up.
- `runMiseStartup` appends `eval "$(mise activate bash)"` to `~/.bashrc` (idempotent, unconditional).
- Tests updated; `HOME` isolated in the runMiseStartup suite so the bashrc append never touches the real home. **setup tests: 54 pass / 0 fail**, biome clean.

## Rollout

Merge → `build-agent.yml` builds the next `agent-v0.2.x`. After bumping the tag in vitals-os `values-prod.yaml` and redeploying, okwow's Linear MCP will resolve from the PVC config (verify: `claude mcp list` shows `linear-server`).

### DIFF
--- a/agent/src/setup.ts
+++ b/agent/src/setup.ts
@@ -8,6 +8,7 @@
  */
 
 import {
+  appendFileSync,
   existsSync,
   lstatSync,
   mkdirSync,
@@ -277,7 +278,8 @@ const DEFAULT_DOT_CLAUDE_FS: DotClaudeFs = {
 
 /**
  * Symlink `linkPath` (~/.claude) → `target` ($AGENT_HOME/dot-claude), ensuring
- * the TARGET directory exists first.
+ * the TARGET directory exists first. ALSO symlinks the sibling `~/.claude.json`
+ * → `$AGENT_HOME/claude.json` (see below).
  *
  * The target MUST be created before symlinking: on a fresh PVC the target does
  * not exist, so without this the symlink dangles and every write under
@@ -289,6 +291,15 @@ const DEFAULT_DOT_CLAUDE_FS: DotClaudeFs = {
  * `symlinkSync` throw EEXIST on a re-run. `lstatSync` sees the link itself, so
  * a stale/dangling symlink is detected and replaced. A real directory at
  * `linkPath` is removed (logged) before relinking.
+ *
+ * `~/.claude.json` holds Claude Code's MCP server registrations (`claude mcp
+ * add`) and lives OUTSIDE `~/.claude/`, so it needs its own symlink to survive
+ * pod restarts. Without it, registered MCP servers (e.g. Linear) silently
+ * disappear — Claude reads the ephemeral image copy instead of the PVC's. The
+ * target is `$AGENT_HOME/claude.json` (sibling of the dot-claude dir), matching
+ * the path the vitals-os runtime used, so a migrated PVC's existing MCP config
+ * is picked up. Unlike the dir, the target is NOT pre-created: it is a FILE that
+ * Claude writes on first use, so a dangling symlink is correct until then.
  */
 export function ensureDotClaudeSymlink(
   target: string,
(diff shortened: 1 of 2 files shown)


agent/src/setup.ts
```diff
@@ -319,6 +330,27 @@ export function ensureDotClaudeSymlink(
...
+  if (existingJson) {
+    if (existingJson.isSymbolicLink()) {
+      fs.unlinkSync(jsonLink);
+    } else {
+      // The image ships a real ~/.claude.json — remove it before symlinking so
+      // the PVC copy (with MCP registrations) is the one Claude reads.
+      fs.rmSync(jsonLink, { recursive: false });
+    }
+  }
+  fs.symlinkSync(jsonTarget, jsonLink);
```
dodizzle, line 352: The new `~/.claude.json` symlink logic (create/replace symlink, replace real file) has no test coverage. The existing `ensureDotClaudeSymlink` tests cover the directory symlink only. CLAUDE.md: *"Tests land with the code, at the correct layer — same PR, no add tests later tasks."*

Needed: add cases that assert `lstatSync(link + '.json').isSymbolicLink()` is true after a fresh call, that re-run is idempotent, and that a pre-existing real file or dangling symlink at `link + '.json'` is replaced.

agent/src/setup.ts
```diff
@@ -443,6 +475,22 @@ export async function runMiseStartup(
...
+  // workspace-declared tools (e.g. fern's Rust toolchain) would be missing
+  // without this. Ported from the vitals-os runtime; dropped in the extraction.
+  // Unconditional (before the mise.toml check) so it's ready for any workspace
+  // tools; idempotent so repeated boots don't stack duplicate lines.
+  const bashrc = join(process.env.HOME ?? home, ".bashrc");
+  if (
+    !existsSync(bashrc) ||
+    !readFileSync(bashrc, "utf8").includes("mise activate bash")
+  ) {
+    appendFileSync(bashrc, '\neval "$(mise activate bash)"\n');
```
dodizzle, line 490: The `beforeEach` HOME isolation was added to prevent CI pollution (good), but no test asserts the activation line was actually written. Needed: (1) assert `readFileSync(join(testHome, '.bashrc'), 'utf8')` includes `'mise activate bash'` after `runMiseStartup`; (2) assert a second call does not duplicate the line.

dodizzle (commented):
COMMENT — fix is correct and the root cause analysis is thorough, but both new behaviors (`~/.claude.json` symlink and `.bashrc` mise activation) ship with zero test assertions, violating the CLAUDE.md hard rule that tests land in the same PR. Two integration test cases needed before merging.

PR #303, opened 2026-06-12
fix(agent): run cron preChecks with cwd=workspace

dmcaulay:
## Problem

okwow's enabled `shipwright-dev-task` cron was suppressed on every tick:

```
[agent:cron] preCheck for job "shipwright-dev-task" crashed (exit 2):
/app/state/todos.json not found — run setup first — session suppressed
```

## Root cause

Plugin preCheck scripts resolve their state relative to `process.cwd()`:
`check-dev-task.ts` → `createTaskStore()` → `new JsonTaskStore(process.cwd())` (`create-task-store.ts:73`) → reads `state/todos.json`. The cron handler spawned the preCheck with **no `cwd`**, so it inherited the agent process cwd (`/app`, the Dockerfile `WORKDIR`) → looked for `/app/state/todos.json` → missing → exit 2 → job suppressed. Real state lives in `$AGENT_HOME/workspace/state`.

Not an extraction regression — vitals-os has the identical no-`cwd` spawn; it's a latent bug that surfaces now because the harness runs the shipwright system-cron preChecks. The actual claude run was already rooted at the workspace (via `claude.ts`); only the preCheck spawn wasn't.

## Fix

Pass `cwd: workspace` (when set) to the preCheck `Bun.spawn`. Added a test that writes `state/todos.json` in the workspace and asserts a cwd-relative preCheck finds it. cron-handler tests: 53 pass / 0 fail.

### DIFF
--- a/agent/src/cron-handler.integration.test.ts
+++ b/agent/src/cron-handler.integration.test.ts
@@ -570,6 +570,29 @@ describe("handleCronRequest — preCheck", () => {
     expect(runArg).toContain("precheck prompt");
   });
 
+  test("runs preCheck with cwd = workspace (resolves state relative to workspace)", async () => {
+    // Mirrors check-dev-task.ts → JsonTaskStore(process.cwd()) reading
+    // `state/todos.json`. The file lives in the workspace; the preCheck must run
+    // there, not in the agent's cwd (/app in prod, the repo root in tests).
+    const ws = join(tmpDir, "ws");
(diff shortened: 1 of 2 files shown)


dodizzle (approved):
APPROVE — minimal targeted fix for the preCheck cwd bug. The `Bun.spawn` now passes `cwd: workspace` when set, matching how the actual Claude run is already rooted. New integration test verifies the fix would catch a regression. No blocking issues.

PR #304, opened 2026-06-12
feat(admin): full cron edit (schedule/prompt/channel/preCheck) on the agent page

dmcaulay:
## Why

The admin agent page could only **toggle** and **delete** crons — there was no way to view or change a cron's schedule, prompt, channel, or **preCheck**. Surfaced while debugging okwow's `shipwright-dev-task` preCheck: you couldn't see the `preCheck` value, let alone edit it, from the UI. The backend already supports a full update (`agentCronJobService.update`, mirroring vitals-os's `/crons/:cronId/update`); this adds the UI.

## Changes

- **Pre-check column** in both the System and Custom cron tables (read-only display of `c.preCheck`).
- **Inline full editor** per row — a collapsible `<details>` "Edit" form with prefilled `schedule`, `prompt`, `channel`, `preCheck`, posting to **`POST /admin/agents/:id/crons/:cronId/update`** → `agentCronJobService.update()`. Empty `preCheck` clears it; invalid cron expressions surface via the existing `?error=` banner.
- `preCheck` added to `CronJobItem` (optional); `"update"` added to the admin `agentCronJobService` Pick.
- Tests: edit-form action + prefilled fields + preCheck column render; cron-service mocks updated for the new `update` method. `admin-ui-pages.unit` + `admin-ui.smoke` + dev-login/provision smoke green; admin `tsc` clean; no new biome issues.

## Scope note

Initial pass was preCheck-only; reworked to a full edit per review. `enabled` stays on the existing toggle button; `user`/`silent` are left untouched by the editor (partial update — the service only writes provided fields).

## Rollout

Admin-only UI (no agent image change). Merge → builds the next `shipwright-admin` image → bump `shipwrightAdminSvc.image.tag` in vitals-os when ready.

### DIFF
--- a/admin/src/admin-ui.ts
+++ b/admin/src/admin-ui.ts
@@ -618,11 +618,24 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
       );
     }
     try {
+      // Fetch the existing cron so we (a) block edits to system crons — their
+      // contents are owned by reconcileSystemCrons and would be reverted — and
+      // (b) forward `user`/`silent`. Without them the service resolves user→null,
+      // silent→false and validateDeliveryTarget throws for any DM-routed cron.
+      const existing = await agentCronJobService.get(agentId, cronId);
+      if (existing.system) {
+        return c.redirect(
+          `/admin/agents/${agentId}?error=${encodeURIComponent("system crons cannot be edited")}`,
+          302,
+        );
+      }
       await agentCronJobService.update(agentId, cronId, {
         schedule,
         prompt,
         channel,
         preCheck,
+        user: existing.user,
+        silent: existing.silent,
       });
     } catch (err) {
       // Surface validation errors (e.g. invalid cron expression) back to the page.
--- a/admin/src/admin-ui.ts
+++ b/admin/src/admin-ui.ts
@@ -121,7 +121,13 @@ export interface AdminUIDeps {
   >;
   agentCronJobService: Pick<
     AgentCronJobService,
-    "list" | "create" | "setEnabled" | "delete" | "get" | "reconcileSystemCrons"
+    | "list"
+    | "create"
+    | "update"
+    | "setEnabled"
+    | "delete"
+    | "get"
+    | "reconcileSystemCrons"
   >;
   agentToolService: Pick<
     AgentToolService,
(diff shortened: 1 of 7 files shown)


dmcaulay (commented):
COMMENT — 3 issues before merge:

1. **Critical: update call drops `user` and `silent`** — any cron routed via user DM (no `channel`) will throw `at least one of channel or user must be set` on save. Fetch the current cron and forward `user`/`silent` through, or add those fields to the edit form.

2. **Important: no smoke tests for the `/update` route** — every other cron action route has HTTP-level smoke tests (create ×4, delete ×2, toggle ×1); the new 49-line handler has zero. Add smoke tests for the success, validation-error, and service-error paths.

3. **Important: no system-cron guard** — the delete route blocks system crons with an error redirect; the update route does not. Mirror that guard so system cron contents can't be overridden from the UI.

admin/src/admin-ui.ts
```diff
@@ -587,6 +593,49 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
...
+    }
+    if (!schedule || !prompt) {
+      return c.redirect(
+        `/admin/agents/${agentId}?error=${encodeURIComponent("schedule and prompt are required")}`,
+        302,
+      );
+    }
+    try {
+      await agentCronJobService.update(agentId, cronId, {
+        schedule,
```
dmcaulay, line 633: **Missing `user` and `silent` fields.** When `channel` is null (user-DM-targeted crons), the service calls `validateDeliveryTarget(null, null, false)` and throws — every edit on a user-targeted cron will fail with a redirect error even if only `preCheck` changed. Fetch the current cron first and pass `user: existing.user, silent: existing.silent` into this call (or add both fields to the edit form).

admin/src/admin-ui.ts
```diff
@@ -587,6 +593,49 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
     return c.redirect(`/admin/agents/${agentId}`, 302);
```
dmcaulay, line 593: **No system-cron guard.** The `delete` route fetches the cron and redirects with an error if `cron.system === true`. This route has no equivalent check — system cron contents are managed by `reconcileSystemCrons` and shouldn't be overrideable from the UI. Mirror the delete guard here.

admin/src/admin-ui.ts
```diff
@@ -587,6 +593,49 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
...
+      // fall through to validation
+    }
+    if (!schedule || !prompt) {
+      return c.redirect(
+        `/admin/agents/${agentId}?error=${encodeURIComponent("schedule and prompt are required")}`,
+        302,
+      );
+    }
+    try {
+      await agentCronJobService.update(agentId, cronId, {
```
dodizzle, line 632: **Critical: drops `user` and `silent`, breaking DM-routed crons.** This update call passes only `schedule`, `prompt`, `channel`, and `preCheck`. The service defaults missing `user` to `null` and `silent` to `false`, then calls `validateDeliveryTarget(null, null, false)` which throws for any cron with `user` set and `channel` null. Every DM-routed cron edit will error out.

Fetch the existing record first and forward the missing fields:
```ts
const existing = await agentCronJobService.get(agentId, cronId);
await agentCronJobService.update(agentId, cronId, {
  schedule,
  prompt,
  channel,
  preCheck,
  user: existing.user,
  silent: existing.silent,
});
```
The guard for system crons (see delete route) should go here too: `if (existing.system) return c.redirect(...error...)`.

dodizzle (commented):
COMMENT — 3 issues before merge.

**Critical:** The `POST /admin/agents/:id/crons/:cronId/update` route passes only `{schedule, prompt, channel, preCheck}` to `agentCronJobService.update()`. The service resolves missing `user` to `null` and `silent` to `false`, then calls `validateDeliveryTarget(null, null, false)` which throws `'at least one of channel or user must be set'` for any cron delivered to a user DM. Editing any DM-routed cron via the UI will always fail. Fix: fetch the existing cron first (`agentCronJobService.get(agentId, cronId)`) and forward `user` and `silent` through the update call.

**Important:** The delete route guards system crons (`if (cron.system) redirect`); the update route does not. An admin can overwrite a system cron's schedule/prompt — it'll be reverted by the next reconcile, but the gap is inconsistent.

**Important:** The new 49-line `/update` handler has no smoke tests. Every other cron mutation route (create ×4, delete ×2, toggle ×1) has HTTP-level coverage. Three paths to cover: success redirect, missing schedule/prompt error redirect, service error redirect.

dmcaulay (commented):
APPROVE — all three prior findings addressed: user/silent forwarding fixed (fetches existing cron, forwards both fields), system-cron guard added (mirrors delete route), smoke tests added (4 paths: success+forwarding, validation, guard with updateCalled===false, service-error). CI green.

dodizzle (approved):
APPROVE — all three prior findings fully addressed in commit `750a15ab`.

- **user/silent forwarding**: Fetches the existing cron via `agentCronJobService.get()` before calling `update()`, then passes `existing.user` and `existing.silent` — DM-routed crons can now be edited without the delivery-target validation error.
- **system-cron guard**: Both the route (early redirect if `existing.system === true`) and the UI (edit form conditionally omitted for system crons) now mirror the delete route's guard.
- **smoke tests**: Four paths covered — happy-path with user/silent forwarding verified via `capturedInput` assertions, missing-schedule validation, system-cron guard with `updateCalled === false`, and service-throw error redirect.

CI green. Clean to merge.

PR #307, opened 2026-06-12
fix(agent): uninstall plugins with stale HOME-prefixed installPaths on startup

dodizzle:
## Summary

- **Root cause:** When a PVC is reused by a container running as a different user (vitals-os as root → shipwright as uid 1000/`HOME=/home/bun`), `installed_plugins.json` on the PVC records `installPaths` under the old `HOME` (e.g. `/root/.claude/plugins/...`). Claude Code's `plugin install` is idempotent — it exits 0 when a spec is already recorded — so the stale path is never corrected. At runtime, Claude Code resolves `CLAUDE_PLUGIN_ROOT` to the old path, which doesn't exist in the new container, and hooks (time-tracking, etc.) silently fail.
- **Fix:** Added `findStalePluginSpecs()` that reads `installed_plugins.json` before each install cycle and returns specs whose `installPath` no longer exists on disk. `installPlugins()` uninstalls those specs first, forcing a clean reinstall that records the correct path for the current user.
- **Impact:** Zero change on normal startup (installPath exists → no-op). One extra uninstall+reinstall per stale spec on the first boot after a user migration.

## Test plan

- [ ] `bun test agent/src/setup.integration.test.ts` — all 62 tests pass, including 7 new tests covering `findStalePluginSpecs` and the stale-path uninstall path in `installPlugins`
- [ ] `bunx biome lint agent/src/setup.ts agent/src/setup.integration.test.ts` — clean
- [ ] Deployed to a migrated agent PVC and confirmed hooks begin firing on first restart

### DIFF
--- a/agent/src/setup.integration.test.ts
+++ b/agent/src/setup.integration.test.ts
@@ -729,7 +729,7 @@ describe("installPlugins", () => {
       return { stdout: "", exitCode: 0 };
     };
 
-    await installPlugins(mockExec, testHome, [], "/repo/root");
+    await installPlugins(mockExec, testHome, [], "/repo/root", join(testHome, "nonexistent.json"));
 
     // marketplace add + 1 install + 1 update = 3 calls (no stale paths)
     expect(calls).toHaveLength(3);
@@ -767,7 +767,7 @@ describe("installPlugins", () => {
       { marketplace: "other-market", plugin: "another-plugin" },
     ];
 
-    await installPlugins(mockExec, testHome, agentPlugins, "/repo/root");
+    await installPlugins(mockExec, testHome, agentPlugins, "/repo/root", join(testHome, "nonexistent.json"));
 
     // 1 marketplace add + 3 installs + 3 updates = 7 calls
     expect(calls).toHaveLength(7);
@@ -825,7 +825,7 @@ describe("installPlugins", () => {
       return { stdout: "", exitCode: 0 };
     };
 
-    await installPlugins(mockExec, testHome, [], "/repo/root");
+    await installPlugins(mockExec, testHome, [], "/repo/root", join(testHome, "nonexistent.json"));
 
     // marketplace add + 1 install + 1 update = 3 calls
     expect(calls).toHaveLength(3);
@@ -839,7 +839,7 @@ describe("installPlugins", () => {
     };
 
     await expect(
-      installPlugins(throwingExec, testHome, [], "/repo/root"),
+      installPlugins(throwingExec, testHome, [], "/repo/root", join(testHome, "nonexistent.json")),
     ).resolves.toBeUndefined();
   });
 
@@ -859,7 +859,7 @@ describe("installPlugins", () => {
     };
 
     await expect(
-      installPlugins(mockExec, testHome, [], "/repo/root"),
+      installPlugins(mockExec, testHome, [], "/repo/root", join(testHome, "nonexistent.json")),
     ).resolves.toBeUndefined();
 
     // All calls still happened despite install failures
@@ -882,7 +882,7 @@ describe("installPlugins", () => {
     };
 
     await expect(
-      installPlugins(mockExec, testHome, [], "/repo/root"),
+      installPlugins(mockExec, testHome, [], "/repo/root", join(testHome, "nonexistent.json")),
     ).resolves.toBeUndefined();
 
     // All 3 calls still happened despite marketplace add failure
@@ -901,7 +901,7 @@ describe("installPlugins", () => {
     };
 
     await expect(
-      installPlugins(mockExec, testHome, [], "/repo/root"),
+      installPlugins(mockExec, testHome, [], "/repo/root", join(testHome, "nonexistent.json")),
     ).resolves.toBeUndefined();
     expect(calls).toHaveLength(3);
   });
--- a/agent/src/setup.integration.test.ts
+++ b/agent/src/setup.integration.test.ts
@@ -21,6 +21,7 @@ import type { AgentPlugin } from "@shipwright/admin";
 import {
   ensureAgentHome,
   ensureDotClaudeSymlink,
+  findStalePluginSpecs,
   installPlugins,
   isNewWorkspace,
   loadState,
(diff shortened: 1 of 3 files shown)


agent/src/setup.integration.test.ts
```diff
@@ -651,7 +731,7 @@ describe("installPlugins", () => {
 
     await installPlugins(mockExec, testHome, [], "/repo/root");
```
dodizzle: This test and the other pre-existing `installPlugins` tests don't pass `pluginManifestPath`, so they default to `join(process.env.HOME, '.claude', 'plugins', 'installed_plugins.json')`. On any machine that has a real manifest at that path — including any root container that experienced the bug this PR fixes — `findStalePluginSpecs` returns stale specs and adds an unexpected `plugin uninstall` call, breaking `expect(calls).toHaveLength(3)`.

The new tests in this PR correctly pass explicit paths. Apply the same pattern here:
```ts
await installPlugins(mockExec, testHome, [], "/repo/root", join(testHome, "nonexistent.json"));
```

dodizzle (commented):
COMMENT — production logic is solid and the fix correctly addresses the root cause. One test isolation gap before merge: existing `installPlugins` tests don't pass an explicit `pluginManifestPath`, so they default to the live `~/.claude/plugins/installed_plugins.json`. On any machine that has experienced this bug (a root container with stale entries), `findStalePluginSpecs` returns non-empty results and injects unexpected uninstall calls, breaking the exact call-count assertions. The new tests in this PR correctly pass an explicit path — the fix is to do the same in the pre-existing tests.

dodizzle (commented):
**APPROVE** — Prior finding resolved: all pre-existing `installPlugins` tests now pass an explicit nonexistent manifest path (`join(testHome, "nonexistent.json")`), preventing any real on-disk manifest from polluting call-count assertions. No new issues. CI green.

### Prior Findings Resolution
| Finding | Status | Evidence |
|---------|--------|----------|
| Existing `installPlugins` tests omit `pluginManifestPath` | Addressed | All 8 pre-existing calls now pass `join(testHome, "nonexistent.json")` as the fifth arg |

agent/src/setup.ts
```diff
@@ -215,6 +250,20 @@ export async function installPlugins(
...
     const allSpecs = [...defaultPluginSpecs, ...agentPluginSpecs];
 
+    // Uninstall any plugin whose recorded installPath no longer exists. This
+    // clears stale manifest entries left over from a different-HOME container
+    // so the subsequent `install` records the correct path for the current user.
+    const staleSpecs = findStalePluginSpecs(allSpecs, pluginManifestPath);
+    for (const spec of staleSpecs) {
+      console.log(`[agent] stale plugin path for ${spec} — uninstalling to force reinstall`);
+      const uninstall = await execFn("claude", ["plugin", "uninstall", spec], { cwd });
+      if (uninstall.exitCode !== 0) {
```
dmcaulay, line 261: When uninstall exits non-zero, execution falls through to the install loop. `plugin install` is idempotent — it exits 0 when the spec is already in the manifest, without re-recording the correct path. The stale entry survives, hooks keep failing silently, and nothing distinguishes this from a successful fix. Add `continue` here so the install is skipped for the failed spec — visible failure on next restart is better than silent "success" with a still-broken path.

dmcaulay (commented):
COMMENT — fix is correct and well-tested in the happy path, one gap in the error path:

**Uninstall failure silently leaves the stale path unfixed.** When `claude plugin uninstall <spec>` exits non-zero, the code warns and continues. The subsequent `plugin install` is idempotent — it sees the spec already in the manifest and exits 0 without re-recording the correct path. The stale `installPath` remains, hooks keep silently failing, and nothing signals that the fix didn't take.

Suggestion: add `continue` after the warn so the install is skipped for the failed spec — the agent keeps failing visibly on next restart rather than silently "succeeding" with a still-broken path.

agent/src/setup.ts
```diff
@@ -176,6 +176,35 @@ async function defaultExec(
...
+  try {
+    manifest = JSON.parse(readFileSync(manifestPath, "utf8")) as typeof manifest;
+  } catch {
+    return [];
+  }
+  return specs.filter((spec) => {
+    const entries = manifest.plugins?.[spec];
+    return (
+      entries?.length &&
+      entries.some((e) => e.installPath && !existsSync(e.installPath))
```
dodizzle: `entries.some(...)` flags a spec as stale if any single recorded entry has a missing installPath — even when other entries for the same spec point to valid paths. A spec is only truly stale when all recorded paths are missing. Change to `entries.every((e) => !e.installPath || !existsSync(e.installPath))` so a partially-valid spec isn't unnecessarily uninstalled.

dodizzle (commented):
**COMMENT** — Prior finding (dmcaulay: skip install/update for specs whose uninstall fails) is fully resolved via the `failedUninstalls` Set with a dedicated test. Two new issues in the updated code:

1. **`findStalePluginSpecs` uses `some()` instead of `every()`** (inline) — flags a spec stale if *any* recorded entry has a missing path, even when other entries are valid. Should be `every()` to only flag when no valid path exists.

2. **Unit tests still read the real manifest** — all 4 unit tests in `setup.unit.test.ts` call `installPlugins` without a `pluginManifestPath` argument, so `findStalePluginSpecs` reads `~/.claude/plugins/installed_plugins.json` at test time. On any machine with a stale `shipwright@shipwright` entry, extra uninstall calls are injected and call-count assertions break. The integration tests were correctly updated in the prior commit; the unit tests were not. Fix: pass `join('/tmp', 'nonexistent-manifest.json')` as the fifth arg in each of the 4 unit test calls.

### Prior Findings Resolution
| Finding | Status | Evidence |
|---------|--------|----------|
| Uninstall failure silently leaves stale path unfixed (dmcaulay) | Addressed | `failedUninstalls` Set guards both install and update loops; test asserts exactly 2 calls when uninstall exits non-zero |

dmcaulay (approved):
APPROVE — All prior findings addressed. `failedUninstalls` Set correctly guards both the install and update loops (skips via `continue` on failure). `findStalePluginSpecs` now uses `every()` so a spec is only flagged stale when no valid path exists. All unit tests pass an explicit nonexistent manifest path, eliminating the real-manifest read. CI green.

dodizzle (commented):
**APPROVE** — Both prior findings fully resolved: the staleness predicate is now `every()` (only flags a spec stale when all recorded paths are missing/undefined), and all 4 unit test calls now pass an explicit nonexistent manifest path isolating them from the real `~/.claude/plugins/installed_plugins.json`. CI green. No new issues.

PR #340, opened 2026-06-14
feat(admin): add KubernetesClient interface + HttpKubernetesClient + recorded double (HD-1.1)

dodizzle:
## Summary
- Add a thin first-party Kubernetes API client in `admin/src/kubernetes-client.ts`: the `KubernetesClient` interface, `HttpKubernetesClient` (in-cluster ServiceAccount Bearer-token + CA auth, namespace-scoped create/get/delete of Deployments and Secrets), and an exported `RecordedKubernetesClient` in-memory double for tests.
- Native `fetch` only — no `@kubernetes/client-node`. The HTTP client takes an injected `fetchFn` and applies the mounted CA via `tls.ca`, so tests feed recorded responses without ever overriding a global.
- Non-2xx API responses map to the shared typed errors in `errors.ts` (404 → NotFoundError, 409 → ConflictError, 403 → ForbiddenError, other → ApiError).

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `KubernetesClient` interface: create/get/delete Deployment + Secret, namespace-scoped, typed results & errors | MET |
| 2 | `HttpKubernetesClient` reads in-cluster token + CA via injected config (paths overridable), correct API-server URLs, Bearer auth, typed non-2xx errors | MET |
| 3 | UNIT tests (pure URL/body) + INTEGRATION tests (create/get/delete vs recorded fixtures: success + 404 + 409 + 403) via injected double; no global.fetch override; no tests retired | MET |

## Test Plan
- 13 unit tests (URL/path building + request-body shaping, pure) + 20 integration tests (cassette replay for success/404/409/403 + CA application) — 33 pass / 0 fail
- [x] biome lint clean, `@shipwright/admin` typecheck green, full `bun test` no new failures vs baseline

### DIFF
--- a/admin/src/kubernetes-client.integration.test.ts
+++ b/admin/src/kubernetes-client.integration.test.ts
@@ -1,0 +1,275 @@
+/**
+ * admin/src/kubernetes-client.integration.test.ts
+ * Integration tests for HttpKubernetesClient against recorded k8s API fixtures.
+ *
+ * Drives the client through an INJECTED fetchFn that replays canned Responses
+ * from a cassette keyed by scenario — no live API server, no global.fetch
+ * override, no mock.module(). Also exercises RecordedKubernetesClient (the
+ * exported in-memory double).
+ */
+
+import { describe, expect, it } from "bun:test";
+import { readFileSync } from "node:fs";
+import {
+  ConflictError,
+  ForbiddenError,
+  NotFoundError,
+} from "./errors.ts";
+import {
+  HttpKubernetesClient,
+  type KubernetesClient,
+  RecordedKubernetesClient,
+} from "./kubernetes-client.ts";
+
+// ─── Cassette ───────────────────────────────────────────────────────────────
+
+interface CassetteEntry {
+  status: number;
+  body: unknown;
+}
+
+const CASSETTE_PATH = new URL(
+  "./fixtures/k8s-cassette.json",
+  import.meta.url,
+).pathname;
+
+const cassette: Record<string, CassetteEntry> = JSON.parse(
+  readFileSync(CASSETTE_PATH, "utf-8"),
+);
+
+/**
+ * Build an injected fetchFn that returns the cassette entry for `key`.
+ * Records the last request so tests can assert URL/method/headers.
+ */
+function cassetteFetch(key: string): {
+  fetchFn: typeof fetch;
+  lastRequest: () => { url: string; method: string; auth?: string };
+} {
+  let last: { url: string; method: string; auth?: string } | undefined;
+  const entry = cassette[key];
+  if (!entry) throw new Error(`cassette key not found: ${key}`);
+
+  const fetchFn = (async (input: RequestInfo | URL, init?: RequestInit) => {
+    const url = typeof input === "string" ? input : input.toString();
+    const headers = new Headers(init?.headers);
+    last = {
+      url,
+      method: init?.method ?? "GET",
+      auth: headers.get("authorization") ?? undefined,
+    };
+    return new Response(JSON.stringify(entry.body), {
+      status: entry.status,
+      headers: { "Content-Type": "application/json" },
+    });
+  }) as typeof fetch;
+
+  return {
+    fetchFn,
+    lastRequest: () => {
+      if (!last) throw new Error("fetchFn was not called");
+      return last;
+    },
+  };
+}
+
+function makeClient(key: string): {
+  client: HttpKubernetesClient;
+  lastRequest: () => { url: string; method: string; auth?: string };
+} {
+  const { fetchFn, lastRequest } = cassetteFetch(key);
+  const client = new HttpKubernetesClient({
+    apiServer: "https://kubernetes.default.svc",
+    token: "test-sa-token",
+    fetchFn,
+  });
+  return { client, lastRequest };
+}
+
+// ─── Deployments: success paths ─────────────────────────────────────────────
+
+describe("HttpKubernetesClient — Deployments", () => {
+  it("createDeployment POSTs to the deployments collection with Bearer auth", async () => {
+    const { client, lastRequest } = makeClient("createDeployment_success");
+    const result = await client.createDeployment("shipwright", {
+      name: "agent-abc",
+      image: "ghcr.io/example/agent:latest",
+    });
+
+    const req = lastRequest();
+    expect(req.method).toBe("POST");
+    expect(req.url).toBe(
+      "https://kubernetes.default.svc/apis/apps/v1/namespaces/shipwright/deployments",
+    );
+    expect(req.auth).toBe("Bearer test-sa-token");
+    expect(result.metadata.name).toBe("agent-abc");
+    expect(result.metadata.namespace).toBe("shipwright");
+  });
+
+  it("getDeployment GETs the named resource and returns it", async () => {
+    const { client, lastRequest } = makeClient("getDeployment_success");
+    const result = await client.getDeployment("shipwright", "agent-abc");
+
+    const req = lastRequest();
+    expect(req.method).toBe("GET");
+    expect(req.url).toBe(
+      "https://kubernetes.default.svc/apis/apps/v1/namespaces/shipwright/deployments/agent-abc",
+    );
+    expect(result.metadata.name).toBe("agent-abc");
+  });
+
+  it("deleteDeployment DELETEs the named resource", async () => {
+    const { client, lastRequest } = makeClient("deleteDeployment_success");
+    await client.deleteDeployment("shipwright", "agent-abc");
+
+    const req = lastRequest();
+    expect(req.method).toBe("DELETE");
+    expect(req.url).toBe(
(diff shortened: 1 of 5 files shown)


admin/src/kubernetes-client.ts
```diff
@@ -0,0 +1,478 @@
...
+  private readonly apiServer: string;
+  private readonly token: string;
+  private readonly caCert?: string;
+  private readonly fetchFn: typeof fetch;
+
+  constructor(opts?: HttpKubernetesClientOpts) {
+    this.apiServer = opts?.apiServer ?? defaultApiServer();
+    this.token =
+      opts?.token ??
+      readFileSync(opts?.tokenPath ?? `${SA_DIR}/token`, "utf-8").trim();
```
dmcaulay: **Token cached at construction — will 401 after Kubernetes rotates it (~1 hour)**

Since Kubernetes 1.21, projected ServiceAccount tokens have bounded TTLs (default rotation window: 1 hour). Reading the token once here and caching it in `this.token` means after the kubelet writes a fresh token to disk, all subsequent `request()` calls will send the stale token and receive 401 from the API server. The pod must be restarted to recover.

Suggestion: read the token file inside the `request()` helper per-request so it always picks up the current on-disk value:
```ts
private readToken(): string {
  return readFileSync(this.tokenPath ?? `${SA_DIR}/token`, "utf-8").trim();
}
```
Or, if you want to keep the cached path, add a refresh triggered on 401 responses.

admin/src/kubernetes-client.ts
```diff
@@ -0,0 +1,478 @@
...
+
+// ─── Error mapping ────────────────────────────────────────────────────────────
+
+interface K8sStatus {
+  message?: string;
+  reason?: string;
+  code?: number;
+}
+
+function mapError(status: number, body: unknown): ApiError {
```
dmcaulay, line 266: **Missing 401 case in error mapping**

A 401 response (expired or rotated ServiceAccount token) falls through to `new ApiError(401, message)`. Callers can't distinguish an auth failure from other error categories, and there's no cassette fixture or integration test for this path.

Suggestion: add a `case 401` (either map to a new `UnauthorizedError` or reuse `ForbiddenError` with a comment), and add a cassette entry `getDeployment_401` + integration test to cover it.

admin/src/kubernetes-client.ts
```diff
@@ -0,0 +1,478 @@
...
+
+  return { apiServer, token, caCert };
+}
+
+/**
+ * Read the in-cluster CA cert (PEM) best-effort. A missing file yields
+ * `undefined` rather than throwing — unlike the token, a missing CA must not
+ * crash client construction.
+ */
+function readCaBestEffort(caPath?: string): string | undefined {
```
dmcaulay: **Silent CA failure turns into opaque TLS errors**

When the CA file can't be read, `readCaBestEffort` returns `undefined` with no log output. `undefined` causes `tls.ca` to be omitted from the Bun fetch `RequestInit`, so the runtime falls back to system CAs. The K8s API server's certificate is signed by the cluster CA, which is not in any system store — all API calls will fail with a TLS handshake error rather than a clear diagnostic about the missing CA file.

Suggestion: log a warning on failure so the operator has a clue:
```ts
console.warn('[kubernetes-client] CA file read failed, falling back to system CAs:', err);
```
If the caller passes an explicit `caPath` override (not the default), consider throwing instead of silently degrading.

dmcaulay (commented):
COMMENT — 3 important issues in the new `KubernetesClient` code: (1) SA token read once at construction will cause 401s after Kubernetes token rotation (~1hr); (2) 401 response has no typed error mapping and no test; (3) `readCaBestEffort` silently swallows CA file read failures, turning them into opaque TLS errors at runtime.

admin/src/kubernetes-client.ts
```diff
@@ -0,0 +1,478 @@
...
+    };
+  }
+
+  private async request<T>(url: string, init: RequestInit): Promise<T> {
+    const finalInit: RequestInitWithTls = this.caCert
+      ? { ...init, tls: { ca: this.caCert } }
+      : init;
+    const resp = await this.fetchFn(url, finalInit as RequestInit);
+    const text = await resp.text();
+    const body = text ? JSON.parse(text) : undefined;
```
dodizzle: **Important (conf 85): untyped SyntaxError leak.** `JSON.parse(text)` has no try/catch. A non-JSON body (proxy 502, TLS error page, gateway error) throws a raw `SyntaxError` that bypasses `mapError`, so callers get an untyped error instead of a typed `ApiError`. Every method routes through here. Fix: `let body; try { body = text ? JSON.parse(text) : undefined; } catch { body = undefined; }` — then `mapError` covers non-2xx and 2xx non-JSON returns undefined (already the empty-body path).

dodizzle (commented):
**Verdict: COMMENT** (self-review). One important finding holds this PR: a JSON.parse SyntaxError can leak untyped from the request() path on non-JSON error bodies. Otherwise solid — all acceptance criteria met, exemplary test isolation, correct k8s wire shapes. See inline comment + the base64 round-trip test suggestion.

admin/src/kubernetes-client.integration.test.ts
```diff
@@ -0,0 +1,355 @@
...
+      client.getDeployment("shipwright", "nope"),
+    ).rejects.toBeInstanceOf(NotFoundError);
+  });
+
+  it("createDeployment records and returns the resource", async () => {
+    const rec = new RecordedKubernetesClient({ deployments: {}, secrets: {} });
+    const created = await rec.createDeployment("ns", {
+      name: "new-dep",
+      image: "img:1",
+    });
```
dodizzle, line 427: AC3 requires 404/409/403 coverage for both Deployments and Secrets. The Deployment describe group has all four (`createDeployment_success`, `createDeployment_409`, `createDeployment_403`, `getDeployment_404`), but Secrets only cover `createSecret_success`, `getSecret_success`, `getSecret_404`, and `deleteSecret_success`. Add `createSecret_409` and `createSecret_403` entries to `k8s-cassette.json` and mirror the Deployment error tests here.

admin/src/kubernetes-client.ts
dodizzle: `const body = text ? JSON.parse(text) : undefined` is unguarded. A proxy returning an HTML error page on a non-2xx status throws a raw `SyntaxError` instead of a typed `ApiError`. Wrap in try/catch: `let body: unknown; try { body = text ? JSON.parse(text) : undefined; } catch { body = undefined; }` — `mapError` already has the `?? \`Kubernetes API error ${status}\`` fallback.

dodizzle (commented):
COMMENT — Two items before merge: (1) the acceptance criteria require Secret 409/403 integration tests matching the Deployment error coverage — `createSecret_409` and `createSecret_403` cassette entries plus corresponding tests are missing; (2) `request()` in `HttpKubernetesClient` does an unguarded `JSON.parse(text)` that throws a raw `SyntaxError` when a proxy returns non-JSON on a non-2xx response — wrap in try/catch and let `mapError` handle it.

dodizzle (commented):
APPROVE — all prior findings addressed: Secret 409/403 cassette entries and integration tests added (AC gap closed), JSON.parse guarded in `request()`, `UnauthorizedError` class added and 401 mapped in `mapError()`, token re-read on each request for rotation support, CA warning on explicit-path failure. 36 tests passing, CI green. Ready to merge.

admin/src/kubernetes-client.ts
```diff
@@ -0,0 +1,515 @@
...
+  }
+
+  /**
+   * Resolve the current Bearer token. An explicitly injected `token` is fixed;
+   * otherwise the token is read from disk per request so that projected
+   * ServiceAccount token rotation (~hourly) is picked up without a pod restart.
+   */
+  private readToken(): string {
+    if (this.token !== undefined) return this.token;
+    return readFileSync(this.tokenPath, "utf-8").trim();
```
dmcaulay, line 333: `readFileSync` here has no error handling. If the SA token file becomes unreadable after construction (deleted during pod lifecycle, misconfigured path), this throws a raw `Error` synchronously inside `authHeaders()` — before the `try/catch` in `request()` which only covers `JSON.parse`. The caller receives an untyped exception instead of the `ApiError` / `UnauthorizedError` contract the rest of this client establishes.

Suggestion:
```ts
private readToken(): string {
  if (this.token !== undefined) return this.token;
  try {
    return readFileSync(this.tokenPath, "utf-8").trim();
  } catch (err) {
    throw new ApiError(401, `ServiceAccount token file unreadable at "${this.tokenPath}": ${String(err)}`);
  }
}
```

dmcaulay (commented):
COMMENT — All three prior findings are resolved (per-request token read, `UnauthorizedError` mapped, `readCa()` throws on explicit path). One new issue: `readToken()` calls `readFileSync` without error handling — if the SA token file becomes unreadable mid-pod, the throw escapes as an untyped `Error` rather than a typed `ApiError`, breaking the error contract. Wrap in try/catch: `throw new ApiError(401, "ServiceAccount token file unreadable: " + String(err))`.

PR #346, opened 2026-06-14
feat(charts): template admin Deployment/Service/SA + persisted secrets + auth modes (HD-3.1)

dodizzle:
## Summary
- Template the **admin** service into the Helm chart: `admin-deployment.yaml` (image from values, `/health` liveness+readiness probes, resources), `admin-service.yaml`, `admin-serviceaccount.yaml`, and `admin-secret.yaml`.
- **Persisted secrets**: `SHIPWRIGHT_SESSION_SECRET` + `SHIPWRIGHT_ENCRYPTION_KEY` use the lookup-then-`randAlphaNum` idiom so they survive `helm upgrade`. `DATABASE_URL_SHIPWRIGHT_ADMIN` is assembled inside the Secret (Postgres password kept out of the Deployment's plaintext env), wired to the bundled postgresql subchart.
- **auth.mode**: `open` (default) → `ADMIN_DEV_AUTH=true`, no `NODE_ENV=production` (working UI, no OAuth — loud insecurity warning in NOTES); `google` → `NODE_ENV=production` + Google OAuth env + `SHIPWRIGHT_ADMIN_ALLOWED_EMAILS`.
- Chart bumped **0.2.0 → 0.3.0** (CHANGELOG + artifacthub annotation) per the version-increment gate. No application code changed.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Admin Deployment renders with DB URL from bundled postgres, persisted session/encryption secrets (lookup), `/health` probes | MET |
| 2 | auth.mode=open → ADMIN_DEV_AUTH=true, no NODE_ENV=production (warned); auth.mode=google → NODE_ENV=production + Google OAuth | MET |
| 3 | helm-unittest golden specs for both auth modes + secret gen + DB wiring; kubeconform validates Deployment/Service/SA/Secret; no tests retired | MET |

## Test Plan
- `helm lint` → 0 failed
- `helm unittest` → 7 suites / 43 tests (24 new admin specs: workload/probes/DB-wiring, both auth modes, secret gen + persistence branch)
- `kubeconform` → 11/11 resources valid (both open and google modes)
- `helm template --set auth.mode=open` and `--set auth.mode=google ...` render with correct per-mode env
- `ct lint --check-version-increment` → PASS (0.3.0)
- gitleaks + check-strings clean; ci.yml untouched
- [x] Pre-ship checks passing

> Note: `auth.mode` enum changed from the scaffold placeholder (`none|session|bearer`) to the task-specified `open|google`; propagated to ci/ values + README + tests. Subsequent HD-3.x (metrics/agent) build on this enum.

### DIFF
--- a/.github/workflows/helm.yml
+++ b/.github/workflows/helm.yml
@@ -95,27 +95,26 @@ jobs:
           version: v3.12.0
 
       - name: Create kind cluster
+        # cluster_name is pinned to "kind" — the `kind load` step below
+        # side-loads the admin image into THAT cluster.
         uses: helm/kind-action@v1
         with:
           version: v0.24.0
           cluster_name: kind
 
-      - name: Build and load stub admin image for kind
-        run: |
-          # Build a minimal stub that serves HTTP on port 3001 so the admin
-          # Deployment's liveness/readiness probes (/health) pass in ct install.
-          # busybox httpd serves any request with 200 + directory listing.
-          # imagePullPolicy: Never in ci/ values ensures kind uses this pre-loaded
-          # image instead of trying to pull shipwright-admin from a registry.
-          mkdir -p /tmp/stub-admin/www
-          echo "ok" > /tmp/stub-admin/www/health
-          docker build -t shipwright-admin:0.1.0 -f - /tmp/stub-admin <<'EOF'
-          FROM busybox:1.36
-          COPY www/ /www/
-          EXPOSE 3001
-          CMD ["httpd", "-f", "-p", "3001", "-h", "/www"]
-          EOF
-          kind load docker-image shipwright-admin:0.1.0
+      # The admin Deployment runs the first-party `shipwright-admin` image, which
+      # is NOT published to any registry at PR time. Build the REAL image and
+      # side-load it into the kind cluster so `ct install` schedules the actual
+      # workload (not a stub): the admin pod boots, runs `prisma migrate deploy`
+      # against the bundled PostgreSQL subchart, and serves its DB-independent
+      # /health probe — so `ct install --wait` sees it reach Ready. The ci/ values
+      # set admin.image.pullPolicy: Never, so kind uses this loaded image and
+      # never reaches a registry. Tag 0.1.0 = Chart.yaml appVersion.
+      - name: Build admin image
+        run: docker build -t shipwright-admin:0.1.0 -f admin/Dockerfile .
+
+      - name: Load admin image into kind
+        run: kind load docker-image shipwright-admin:0.1.0 --name kind
 
       - name: ct install (per ci/ values variant, runs helm test)
         run: ct install --config ct.yaml --charts charts/shipwright
--- a/.github/workflows/helm.yml
+++ b/.github/workflows/helm.yml
@@ -98,6 +98,7 @@ jobs:
         uses: helm/kind-action@v1
         with:
           version: v0.24.0
+          cluster_name: kind
 
       - name: Build and load stub admin image for kind
         run: |
--- a/.github/workflows/helm.yml
+++ b/.github/workflows/helm.yml
@@ -101,8 +101,18 @@ jobs:
 
       - name: Build and load stub admin image for kind
         run: |
-          docker build -t shipwright-admin:0.1.0 - <<'EOF'
-          FROM scratch
+          # Build a minimal stub that serves HTTP on port 3001 so the admin
+          # Deployment's liveness/readiness probes (/health) pass in ct install.
+          # busybox httpd serves any request with 200 + directory listing.
+          # imagePullPolicy: Never in ci/ values ensures kind uses this pre-loaded
+          # image instead of trying to pull shipwright-admin from a registry.
+          mkdir -p /tmp/stub-admin/www
+          echo "ok" > /tmp/stub-admin/www/health
+          docker build -t shipwright-admin:0.1.0 -f - /tmp/stub-admin <<'EOF'
+          FROM busybox:1.36
+          COPY www/ /www/
+          EXPOSE 3001
+          CMD ["httpd", "-f", "-p", "3001", "-h", "/www"]
           EOF
           kind load docker-image shipwright-admin:0.1.0
 
--- a/.github/workflows/helm.yml
+++ b/.github/workflows/helm.yml
@@ -99,5 +99,12 @@ jobs:
         with:
           version: v0.24.0
 
+      - name: Build and load stub admin image for kind
+        run: |
+          docker build -t shipwright-admin:0.1.0 - <<'EOF'
+          FROM scratch
+          EOF
+          kind load docker-image shipwright-admin:0.1.0
+
       - name: ct install (per ci/ values variant, runs helm test)
         run: ct install --config ct.yaml --charts charts/shipwright
--- a/charts/shipwright/ci/minikube-values.yaml
+++ b/charts/shipwright/ci/minikube-values.yaml
@@ -5,7 +5,8 @@ networking:
   type: ClusterIP
 
 auth:
-  mode: none
+  # Dev auth (ADMIN_DEV_AUTH=true), no OAuth — fine for a throwaway local cluster.
+  mode: open
 
 agent:
   provisioning:
--- a/charts/shipwright/ci/minikube-values.yaml
+++ b/charts/shipwright/ci/minikube-values.yaml
@@ -8,6 +8,10 @@ auth:
   # Dev auth (ADMIN_DEV_AUTH=true), no OAuth — fine for a throwaway local cluster.
   mode: open
 
+admin:
+  image:
+    pullPolicy: Never
+
 agent:
   provisioning:
     persistence:
--- a/charts/shipwright/templates/admin-deployment.yaml
+++ b/charts/shipwright/templates/admin-deployment.yaml
@@ -22,6 +22,12 @@ spec:
         - name: {{ . }}
         {{- end }}
       {{- end }}
+      {{- if .Values.postgresql.enabled }}
+      initContainers:
+        - name: wait-for-postgres
+          image: busybox:1.36
+          command: ['sh', '-c', 'until nc -z {{ include "shipwright.postgresql.fullname" . }} 5432; do echo "waiting for postgresql..."; sleep 2; done']
+      {{- end }}
       containers:
         - name: admin
           image: "{{ with .Values.global.imageRegistry }}{{ . }}/{{ end }}{{ .Values.admin.image.repository }}:{{ .Values.admin.image.tag | default .Chart.AppVersion }}"
--- a/charts/shipwright/templates/admin-deployment.yaml
+++ b/charts/shipwright/templates/admin-deployment.yaml
@@ -1,0 +1,100 @@
+{{- if .Values.admin.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "shipwright.admin.fullname" . }}
+  labels:
+    {{- include "shipwright.admin.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.admin.replicas | default 1 }}
+  selector:
+    matchLabels:
+      {{- include "shipwright.admin.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "shipwright.admin.selectorLabels" . | nindent 8 }}
+    spec:
+      serviceAccountName: {{ include "shipwright.admin.serviceAccountName" . }}
+      {{- with .Values.global.imagePullSecrets }}
+      imagePullSecrets:
+        {{- range . }}
+        - name: {{ . }}
+        {{- end }}
+      {{- end }}
+      containers:
+        - name: admin
+          image: "{{ with .Values.global.imageRegistry }}{{ . }}/{{ end }}{{ .Values.admin.image.repository }}:{{ .Values.admin.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.admin.image.pullPolicy | default .Values.imagePullPolicy }}
+          ports:
+            - name: http
+              containerPort: 3001
+              protocol: TCP
+          livenessProbe:
+            httpGet:
+              path: /health
+              port: 3001
+            initialDelaySeconds: 10
+            periodSeconds: 15
+            timeoutSeconds: 3
+            failureThreshold: 3
+          readinessProbe:
+            httpGet:
+              path: /health
+              port: 3001
+            initialDelaySeconds: 5
+            periodSeconds: 10
+            timeoutSeconds: 3
+            failureThreshold: 3
+          env:
+            # Bind the admin service to the chart-standard port 3001 (the app
+            # defaults to 3000 via PORT; the chart standardizes on 3001).
+            - name: PORT
+              value: "3001"
+            {{- if .Values.postgresql.enabled }}
+            # Assembled in the admin Secret so the Postgres password is never in
+            # plaintext Deployment env.
+            - name: DATABASE_URL_SHIPWRIGHT_ADMIN
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "shipwright.admin.secretName" . }}
+                  key: DATABASE_URL_SHIPWRIGHT_ADMIN
+            {{- end }}
+            - name: SHIPWRIGHT_SESSION_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "shipwright.admin.secretName" . }}
+                  key: SHIPWRIGHT_SESSION_SECRET
+            - name: SHIPWRIGHT_ENCRYPTION_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "shipwright.admin.secretName" . }}
+                  key: SHIPWRIGHT_ENCRYPTION_KEY
+            {{- if eq .Values.auth.mode "open" }}
+            # auth.mode=open — dev auth, NO OAuth. Insecure: do not expose
+            # publicly. NODE_ENV is intentionally NOT set to production so the
+            # dev-auth path stays enabled.
+            - name: ADMIN_DEV_AUTH
+              value: "true"
+            {{- else if eq .Values.auth.mode "google" }}
+            # auth.mode=google — real Google OAuth. Production hardening on.
+            - name: NODE_ENV
+              value: "production"
+            - name: GOOGLE_CLIENT_ID
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "shipwright.admin.secretName" . }}
+                  key: GOOGLE_CLIENT_ID
+            - name: GOOGLE_CLIENT_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "shipwright.admin.secretName" . }}
+                  key: GOOGLE_CLIENT_SECRET
+            - name: SHIPWRIGHT_ADMIN_ALLOWED_EMAILS
+              value: {{ .Values.auth.google.allowedEmails | quote }}
+            {{- end }}
+          {{- with .Values.admin.resources }}
+          resources:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+{{- end }}
--- a/charts/shipwright/templates/admin-deployment.yaml
+++ b/charts/shipwright/templates/admin-deployment.yaml
@@ -70,10 +70,10 @@ spec:
                 secretKeyRef:
                   name: {{ include "shipwright.admin.secretName" . }}
                   key: SHIPWRIGHT_ENCRYPTION_KEY
-            {{- if eq .Values.auth.mode "open" }}
-            # auth.mode=open — dev auth, NO OAuth. Insecure: do not expose
-            # publicly. NODE_ENV is intentionally NOT set to production so the
-            # dev-auth path stays enabled.
+            {{- if or (eq .Values.auth.mode "open") (eq .Values.auth.mode "none") }}
+            # auth.mode=open (or deprecated alias "none") — dev auth, NO OAuth.
+            # Insecure: do not expose publicly. NODE_ENV is intentionally NOT set
+            # to production so the dev-auth path stays enabled.
             - name: ADMIN_DEV_AUTH
               value: "true"
             {{- else if eq .Values.auth.mode "google" }}
--- a/charts/shipwright/templates/admin-secret.yaml
+++ b/charts/shipwright/templates/admin-secret.yaml
@@ -1,0 +1,68 @@
+{{- if .Values.admin.enabled }}
+{{- /*
+  Chart-managed admin Secret.
+
+  Holds the admin session/encryption keys, and — when the bundled PostgreSQL
+  subchart is enabled — the fully-assembled DATABASE_URL_SHIPWRIGHT_ADMIN so the
+  Postgres password never appears in plaintext Deployment env.
+
+  Persistence idiom (survives `helm upgrade`): look up the already-installed
+  Secret and reuse its (already base64-encoded) values; only generate fresh
+  random material on first install. NOTE: `helm template` and helm-unittest do
+  NOT execute `lookup`, so they always take the generate branch — that is
+  expected. On a live install/upgrade the lookup finds the prior Secret and the
+  generated values stay stable.
+*/}}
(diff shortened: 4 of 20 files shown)


charts/shipwright/templates/admin-deployment.yaml
```diff
@@ -0,0 +1,100 @@
...
+      serviceAccountName: {{ include "shipwright.admin.serviceAccountName" . }}
+      {{- with .Values.global.imagePullSecrets }}
+      imagePullSecrets:
+        {{- range . }}
+        - name: {{ . }}
+        {{- end }}
+      {{- end }}
+      containers:
+        - name: admin
+          image: "{{ with .Values.global.imageRegistry }}{{ . }}/{{ end }}{{ .Values.admin.image.repository }}:{{ .Values.admin.image.tag | default .Chart.AppVersion }}"
```
dodizzle, line 33: This image line is the root cause of the `ct install` CI failure. `shipwright-admin:0.1.0` doesn't exist on Docker Hub — the kind cluster can't pull it. The helm.yml `ct install` step has no `kind load docker-image` step. Fix: either pre-load a stub image (`docker build -t shipwright-admin:0.1.0 . && kind load docker-image shipwright-admin:0.1.0`) before `ct install`, or add `imagePullPolicy: Never` to all ci/*.yaml variants so kind uses a pre-loaded image without hitting the registry.

charts/shipwright/values.schema.json
```diff
@@ -80,7 +80,16 @@
       "properties": {
         "mode": {
           "type": "string",
-          "enum": ["none", "session", "bearer"]
+          "enum": ["open", "google"]
```
dodizzle: Breaking change: the base branch has `enum: ["none", "session", "bearer"]` here with `none` as the default. Changing to `["open", "google"]` means any existing install running `helm upgrade` with default values (auth.mode: none) will get a schema validation error. Add a migration note to the CHANGELOG: `helm upgrade --set auth.mode=open`. Or keep `none` in the enum and map it to `open` behavior in the templates to avoid a hard break.

charts/shipwright/templates/admin-secret.yaml
```diff
@@ -0,0 +1,68 @@
...
+      1. .Values.postgresql.auth.password when set, else
+      2. the live <release>-postgresql Secret (Bitnami auto-generated), else
+      3. "" on `helm template`/unittest (no cluster to look up) — the live
+         install/upgrade resolves the real password from the subchart Secret.
+  */}}
+  {{- $pgHost := include "shipwright.postgresql.fullname" . }}
+  {{- $pgUser := .Values.postgresql.auth.username }}
+  {{- $pgDb := .Values.postgresql.auth.database }}
+  {{- $pgPassword := .Values.postgresql.auth.password }}
+  {{- if not $pgPassword }}
```
dodizzle, line 54: When `postgresql.auth.existingSecret` is set (the production path documented in values.yaml and the README), the Bitnami subchart reads the postgres password from that user-supplied secret. But this lookup still targets `<release>-postgresql` (the auto-generated name). That secret may not exist when `existingSecret` is set, producing an empty password in the assembled DATABASE_URL. Add a branch: `{{- if .Values.postgresql.auth.existingSecret }}` and look up the password from that secret name with the correct Bitnami key (`postgres-password` or `password` depending on existingSecret structure).

dodizzle (commented):
COMMENT — Two issues to fix before merge.

**1. ct install fails — admin image not pre-loaded into kind cluster** (`charts/shipwright/templates/admin-deployment.yaml:27`)
The `ct install` step in helm.yml has no image pre-loading for the kind cluster. `shipwright-admin:0.1.0` doesn't exist on Docker Hub — CI logs confirm `ImagePullBackOff` on the admin pod for all three ci/ values variants. This PR is the first to add admin workload templates; before it, ct install only needed the postgresql subchart image. Fix: either pre-load a stub image via `kind load docker-image` before the `ct install` step, or add `imagePullPolicy: Never` to the ci/ values variants and pre-load a scratch image.

**2. Breaking auth.mode enum change breaks helm upgrade for existing installs** (`charts/shipwright/values.schema.json:83`)
The base branch schema enforced `auth.mode: enum: ["none", "session", "bearer"]` with `none` as the default. Any existing installation will fail `helm upgrade` because the previously-valid `none` value is no longer in the schema. Fix: add a migration note to the CHANGELOG specifying `helm upgrade --set auth.mode=open`, or keep `none` in the enum as a handled alias.

**Also:** `postgresql.auth.existingSecret` is not handled in `admin-secret.yaml`'s DATABASE_URL assembly — when `existingSecret` is set, the lookup still targets `<release>-postgresql` (the auto-generated name), producing an empty or wrong password. Fix: branch on `.Values.postgresql.auth.existingSecret` and look up from that secret name instead.

Strengths: lookup idiom for session/encryption secrets is correctly implemented; loud NOTES.txt warning for `auth.mode=open`; helm-unittest coverage lands with the code.

charts/shipwright/ci/minikube-values.yaml
```diff
@@ -5,7 +5,8 @@ networking:
   type: ClusterIP
 
 auth:
-  mode: none
+  # Dev auth (ADMIN_DEV_AUTH=true), no OAuth — fine for a throwaway local cluster.
+  mode: open
```
dmcaulay, line 9: All three ci/ variants (minikube, eks, gke) leave `admin.enabled` at the chart default (`true`). The kind cluster in `ct install` attempts to pull `shipwright-admin:0.1.0` from Docker Hub — the image doesn't exist publicly — causing `ErrImagePull` and blocking the Helm CI gate.

Simplest fix: add `admin: enabled: false` to each ci/ file to skip the admin Deployment during ct install until the image is published. Apply the same to eks-values.yaml and gke-values.yaml.

charts/shipwright/values.schema.json
```diff
@@ -80,7 +80,16 @@
...
+        },
+        "google": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "clientId": { "type": "string" },
+            "clientSecret": { "type": "string" },
+            "allowedEmails": { "type": "string" }
```
dmcaulay, line 91: No `minLength` constraint and not in a `required` array. With `auth.mode: google` and `allowedEmails: ""`, helm lint passes cleanly. If the admin app treats an empty `SHIPWRIGHT_ADMIN_ALLOWED_EMAILS` as 'no restriction', the email allow-list is silently bypassed — any authenticated Google user gets in.

Suggested fix: `"allowedEmails": { "type": "string", "minLength": 1 }` and an `if/then` conditional making `clientId`/`clientSecret`/`allowedEmails` required when `mode = "google"`.

charts/shipwright/templates/admin-secret.yaml
```diff
@@ -0,0 +1,68 @@
...
+  {{- $pgSecret := (lookup "v1" "Secret" .Release.Namespace $pgHost) }}
+  {{- if and $pgSecret $pgSecret.data (index $pgSecret.data "password") }}
```
dmcaulay: The Bitnami postgresql subchart v16.x stores the application-user password under the key `postgresql-password`, not `password`. This lookup will always return nil when `postgresql.auth.password` is not set in values (the auto-generate default), leaving `$pgPassword` empty. The assembled `DATABASE_URL_SHIPWRIGHT_ADMIN` becomes `postgresql://shipwright:@host:5432/db` — the admin service can't authenticate.

This is invisible in CI because all three ci/ variants set explicit passwords. Fix both this line and line 57: change `"password"` → `"postgresql-password"`.

dmcaulay (commented):
COMMENT — Three issues before merge.

**1. ct install (kind) gate failing — ErrImagePull for `shipwright-admin:0.1.0` (critical)**
All three ci/ variants leave `admin.enabled: true` (the default). The kind cluster tries to pull `docker.io/library/shipwright-admin:0.1.0`, which doesn't exist publicly — `ErrImagePull` / `ImagePullBackOff`, context deadline exceeded. Simplest fix: add `admin:\n  enabled: false` to all three ci/ values files to skip the admin Deployment until a real image is published.

**2. Wrong Bitnami PostgreSQL secret key 'password' → 'postgresql-password' (important)**
`admin-secret.yaml:56-57` looks up `index $pgSecret.data "password"`. Bitnami postgresql subchart v16.x stores the app-user password under `postgresql-password`, not `password`. When `postgresql.auth.password` is empty (the auto-generate default), `$pgPassword` stays empty and `DATABASE_URL_SHIPWRIGHT_ADMIN` is assembled as `postgresql://shipwright:@host:5432/db`. Invisible in CI because all three ci/ variants set explicit passwords. Fix: change both lookup lines to `"postgresql-password"`.

**3. Empty `allowedEmails` accepted in google auth mode (important)**
`allowedEmails` has no `minLength` and is not `required` when `auth.mode=google`. A deployer who omits it passes helm lint cleanly, but if the admin service treats an empty env var as 'allow all authenticated users' the email allow-list is silently bypassed. Fix: add `"minLength": 1` to the schema property.

**Strengths:** Lookup-then-generate idiom is correct (no double-encoding); DATABASE_URL in the Secret is the right design; helm-unittest coverage is thorough.

.github/workflows/helm.yml
```diff
@@ -99,5 +99,22 @@ jobs:
...
+          # image instead of trying to pull shipwright-admin from a registry.
+          mkdir -p /tmp/stub-admin/www
+          echo "ok" > /tmp/stub-admin/www/health
+          docker build -t shipwright-admin:0.1.0 -f - /tmp/stub-admin <<'EOF'
+          FROM busybox:1.36
+          COPY www/ /www/
+          EXPOSE 3001
+          CMD ["httpd", "-f", "-p", "3001", "-h", "/www"]
+          EOF
+          kind load docker-image shipwright-admin:0.1.0
```
dodizzle: `kind load docker-image` defaults to cluster name `"kind"`, but `helm/kind-action@v1` with no `cluster_name` input creates the cluster as `"chart-testing"` (the chart-testing tool default). CI confirms: `ERROR: no nodes found for cluster "kind"` — the image builds successfully but is never loaded, so ct install still fails on ImagePullBackOff.

Fix either:
```yaml
# Option A — create the cluster with kind's own default name:
- name: Create kind cluster
  uses: helm/kind-action@v1
  with:
    version: v0.24.0
    cluster_name: kind
```
or
```bash
# Option B — target the existing cluster explicitly:
kind load docker-image shipwright-admin:0.1.0 --name chart-testing
```
Option A is simpler and avoids hard-coding the cluster name in two places.

dodizzle (commented):
COMMENT — One critical issue remains before merge.

**Prior findings resolution:**
- ✅ **auth.mode backward compat**: `"none"` re-added to schema enum and aliased to `"open"` behavior in the template. Existing upgrades from `auth.mode: none` pass schema validation cleanly.
- ✅ **existingSecret DATABASE_URL**: Branch added in `admin-secret.yaml` — looks up from the user-supplied secret using key `"password"`, which is correct for Bitnami postgresql v16.
- ❌ **ct install (ImagePullBackOff)**: Still failing — see inline comment.

**What's clean in this update:** The plugin-layer `model`/`complexity` additions are well-designed — soft-only warnings (non-throwing), bounds-checked `complexity` guard (1–5), comprehensive test coverage across json, github, and jira adapters with round-trip serialization verified.

One line away from a clean CI gate.

charts/shipwright/values.schema.json
```diff
@@ -80,7 +80,16 @@
...
-          "enum": ["none", "session", "bearer"]
+          "enum": ["open", "google", "none"]
+        },
+        "google": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "clientId": { "type": "string" },
+            "clientSecret": { "type": "string" },
+            "allowedEmails": { "type": "string", "minLength": 1 }
```
dodizzle: This `minLength: 1` constraint breaks both CI jobs. The default `values.yaml` ships with `auth.google.allowedEmails: ""` (an empty-string placeholder used when `auth.mode` is not `google`), which now fails `helm lint` unconditionally — Helm's schema validator has no awareness of the sibling `mode` field and `if/then` conditionals aren't supported.

CI error:
```
[ERROR] values.yaml: - at '/auth/google/allowedEmails': minLength: got 0, want 1
```

Fix: remove `minLength: 1` from the schema and enforce it at template render time:
```yaml
{{- if and (eq .Values.auth.mode "google") (empty .Values.auth.google.allowedEmails) }}
{{- fail "auth.google.allowedEmails is required when auth.mode=google" }}
{{- end }}
```
Place this near the top of `admin-secret.yaml` or `admin-deployment.yaml`. It fires only when `mode=google`, so the default empty-string placeholder stays valid for `mode=open` installs.

dodizzle (commented):
COMMENT — One critical issue introduced in this commit.

**Prior finding resolved:**
- ✅ **cluster_name: kind**: The `cluster_name: kind` addition to `helm/kind-action@v1` correctly fixes the cluster name mismatch. `kind load docker-image` now targets the right cluster. That path is unblocked.

**New issue introduced:**
- ❌ **minLength: 1 on allowedEmails breaks helm lint against default values** — see inline comment.

One template-level `fail` guard away from a clean CI gate.

dmcaulay (approved):
APPROVE — all three prior findings resolved.

**1. ct install ErrImagePull** ✅ Fixed: busybox:1.36 stub pre-loaded into kind via `kind load docker-image`; `cluster_name: kind` ensures the cluster name matches; `imagePullPolicy: Never` in all ci/ variants skips registry pull. The stub serves `/health` → 200 via `httpd -f -p 3001 -h /www` with `/www/health` present.

**2. Bitnami PostgreSQL secret key** ✅ Fixed (`622ab4fd`): auto-generated-secret path now looks up `"postgresql-password"` instead of `"password"`, matching Bitnami postgresql v16.x subchart output.

**3. allowedEmails empty in google mode** ✅ Fixed (`622ab4fd`): `if/then` JSON Schema conditional enforces `minLength: 1` on `clientId`, `clientSecret`, and `allowedEmails` only when `auth.mode=google`; open mode accepts empty values. Two helm-unittest cases verify both the rejection and the open-mode acceptance.

Code is clean. Merging pending CI green on this commit.

dmcaulay (approved):
APPROVE — all three prior findings resolved.

**1. ct install ErrImagePull** ✅ Fixed: busybox:1.36 stub pre-loaded into kind via `kind load docker-image`; `cluster_name: kind` ensures the cluster name matches; `imagePullPolicy: Never` in all ci/ variants skips registry pull. The stub serves `/health` → 200 via `httpd -f -p 3001 -h /www` with `/www/health` present.

**2. Bitnami PostgreSQL secret key** ✅ Fixed (`622ab4fd`): auto-generated-secret path now looks up `"postgresql-password"` instead of `"password"`, matching Bitnami postgresql v16.x subchart output.

**3. allowedEmails empty in google mode** ✅ Fixed (`622ab4fd`): `if/then` JSON Schema conditional enforces `minLength: 1` on `clientId`, `clientSecret`, and `allowedEmails` only when `auth.mode=google`; open mode accepts empty values. Two helm-unittest cases verify both the rejection and the open-mode acceptance.

Code is clean. Merging pending CI green on this commit.

dodizzle (commented):
APPROVE — All prior findings resolved. CI passes all gates (helm lint, helm unittest, kubeconform, ct install) on commit 622ab4f. No new issues.

**Prior finding resolved:**
- ✅ **allowedEmails minLength regression**: Moved from unconditional property definition into a proper JSON Schema `if/then` conditional that fires only when `auth.mode=google`. Helm 3.18.6 (pinned in CI) supports this; `helm lint` and `ct install` both pass. The two new schema tests in `values_schema_test.yaml` provide regression coverage.
- ✅ **Bitnami auto-generated secret key**: `postgresql-password` is the correct key for Bitnami chart 16.7.27 auto-generated secrets. The `existingSecret` path correctly retains `password` (user-defined key).

All findings across all 4 review rounds are now addressed. Ready to merge.

.github/workflows/helm.yml
```diff
@@ -95,9 +95,26 @@ jobs:
...
+        # side-loads the admin image into THAT cluster.
         uses: helm/kind-action@v1
         with:
           version: v0.24.0
+          cluster_name: kind
+
+      # The admin Deployment runs the first-party `shipwright-admin` image, which
+      # is NOT published to any registry at PR time. Build the REAL image and
+      # side-load it into the kind cluster so `ct install` schedules the actual
+      # workload (not a stub): the admin pod boots, runs `prisma migrate deploy`
```
dodizzle, line 108: **PostgreSQL startup race**: the comment says the admin pod "runs `prisma migrate deploy`" — which is accurate, but `main.ts` does this synchronously at boot before serving `/health`. If the Bitnami PostgreSQL pod hasn't finished its own init when the admin pod is scheduled, the migration throws a connection error, `process.exit(1)` fires, and the pod enters CrashLoopBackOff.

Kubernetes exponential backoff will eventually let the admin pod retry after PostgreSQL is up, so `ct install --wait` may still pass — but it is timing-dependent and will flake on slow runners. The stub approach avoided this entirely because it never connected to PostgreSQL.

The chart has no initContainer gating on PostgreSQL readiness. Consider adding one:
```yaml
initContainers:
  - name: wait-for-postgres
    image: busybox:1.36
    command: ['sh', '-c', 'until nc -z $PG_HOST 5432; do sleep 2; done']
```
Or document that CrashLoopBackOff→Ready on first deployment is expected behavior for this release stage.

dodizzle (commented):
COMMENT — One important issue to verify before merge.

**Prior findings from 622ab4f rounds resolved:**
- ✅ ct install ErrImagePull: busybox stub is replaced by real image build — correct progression
- ✅ Bitnami PostgreSQL secret key: fixed in 622ab4f, unchanged here
- ✅ allowedEmails minLength: fixed via if/then conditional in 622ab4f, unchanged here

**New issue introduced by building the real image:**
- ⚠️ **Admin pod races PostgreSQL at startup with no initContainer gate** (`.github/workflows/helm.yml:108`) — see inline comment.

The build invocation is correct (`-f admin/Dockerfile .` with repo-root context), `--name kind` correctly targets the pinned cluster, and splitting build/load into two named steps is cleaner. If CI Helm e2e passes green on this commit, the startup race is dormant for this run — but it remains a latent flakiness risk on slow runners.

dodizzle (commented):
COMMENT — One important issue before merge.

**Prior finding resolved:**
- ✅ **Admin pod startup race (no initContainer)**: `wait-for-postgres` initContainer added using `busybox:1.36` + `nc -z` poll loop; gated on `postgresql.enabled` so external-Postgres installs are unaffected. CI green.

**New issue introduced:**
- ⚠️ **Missing helm-unittest test for initContainer conditional** — see inline comment. CLAUDE.md requires tests to land with the code; `admin_workload_test.yaml` covers `postgresql.enabled` for DATABASE_URL wiring but not for the initContainer presence/absence.

charts/shipwright/templates/admin-deployment.yaml
```diff
@@ -0,0 +1,106 @@
...
+    spec:
+      serviceAccountName: {{ include "shipwright.admin.serviceAccountName" . }}
+      {{- with .Values.global.imagePullSecrets }}
+      imagePullSecrets:
+        {{- range . }}
+        - name: {{ . }}
+        {{- end }}
+      {{- end }}
+      {{- if .Values.postgresql.enabled }}
+      initContainers:
```
dodizzle, line 26: The initContainer fix is correct — `nc -z` poll loop, pinned busybox tag, gated on `postgresql.enabled`. CI passes.

Missing helm-unittest coverage: `admin_workload_test.yaml` already tests `postgresql.enabled=true/false` for DATABASE_URL wiring (lines 90-120) but doesn't assert whether `spec.template.spec.initContainers` is present or absent under the same flag. CLAUDE.md requires tests to land with the code at the correct layer.

Suggested additions to `admin_workload_test.yaml`:
```yaml
- it: renders wait-for-postgres initContainer when postgresql is enabled
  template: templates/admin-deployment.yaml
  set:
    postgresql.enabled: true
  asserts:
    - equal:
        path: spec.template.spec.initContainers[0].name
        value: wait-for-postgres
    - equal:
        path: spec.template.spec.initContainers[0].image
        value: busybox:1.36

- it: omits initContainers when postgresql is disabled
  template: templates/admin-deployment.yaml
  set:
    postgresql.enabled: false
  asserts:
    - notExists:
        path: spec.template.spec.initContainers
```

charts/shipwright/templates/admin-deployment.yaml
```diff
@@ -0,0 +1,106 @@
...
+            - name: SHIPWRIGHT_ENCRYPTION_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "shipwright.admin.secretName" . }}
+                  key: SHIPWRIGHT_ENCRYPTION_KEY
+            {{- if or (eq .Values.auth.mode "open") (eq .Values.auth.mode "none") }}
+            # auth.mode=open (or deprecated alias "none") — dev auth, NO OAuth.
+            # Insecure: do not expose publicly. NODE_ENV is intentionally NOT set
+            # to production so the dev-auth path stays enabled.
+            - name: ADMIN_DEV_AUTH
```
dmcaulay, line 83: `auth.mode: open` sets `ADMIN_DEV_AUTH=true` but omits `NODE_ENV`. The image bakes `ENV NODE_ENV=production` (`admin/Dockerfile:43`), so `dev-auth-guard.ts` sees `NODE_ENV=production` and returns `false` — dev-auth is silently blocked even with `ADMIN_DEV_AUTH=true`. Add an explicit override here:
```yaml
- name: NODE_ENV
  value: "development"
```

charts/shipwright/tests/admin_workload_test.yaml
```diff
@@ -0,0 +1,167 @@
...
+      value: admin
+      skipEmptyTemplates: true
+    templates:
+      - templates/admin-deployment.yaml
+      - templates/admin-service.yaml
+      - templates/admin-serviceaccount.yaml
+      - templates/admin-secret.yaml
+    asserts:
+      - hasDocuments:
+          count: 0
```
dmcaulay, line 167: The `wait-for-postgres` initContainer (added in `27216c43`) has two cases — enabled and disabled — but neither is tested here. CLAUDE.md requires tests to land with code in the same PR. Add:
```yaml
- it: includes wait-for-postgres initContainer when postgresql is enabled
  template: templates/admin-deployment.yaml
  set:
    postgresql.enabled: true
  asserts:
    - equal:
        path: spec.template.spec.initContainers[0].name
        value: wait-for-postgres

- it: omits initContainers when postgresql is disabled
  template: templates/admin-deployment.yaml
  set:
    postgresql.enabled: false
  asserts:
    - isEmpty:
        path: spec.template.spec.initContainers
```

dmcaulay (commented):
COMMENT — Two new issues introduced since the last APPROVE (622ab4fd).

**1. `NODE_ENV=production` baked in image silently disables dev-auth in `auth.mode: open`** (`charts/shipwright/templates/admin-deployment.yaml:79-84`)
`admin/Dockerfile:43` sets `ENV NODE_ENV=production`. `dev-auth-guard.ts:25` returns `false` whenever `NODE_ENV === "production"`, regardless of `ADMIN_DEV_AUTH`. The chart's `open` mode sets `ADMIN_DEV_AUTH=true` but omits `NODE_ENV` (the comment on line 81 states "NODE_ENV is intentionally NOT set to production") — but the baked-in env wins. Deployers using `auth.mode: open` get dev-auth silently blocked. CI passes because `ct install` gates on `/health` (unauthenticated), not the dev-login route. Fix: add `- name: NODE_ENV\n  value: "development"` in the open-mode env block to explicitly override the Dockerfile default.

**2. `wait-for-postgres` initContainer missing helm-unittest coverage** (`charts/shipwright/tests/admin_workload_test.yaml`)
The new initContainer (added in `27216c43`) under `{{- if .Values.postgresql.enabled }}` has two cases but `admin_workload_test.yaml` has no assertions for either. CLAUDE.md: "tests land with code, same PR." Add: (a) assert `spec.template.spec.initContainers[0].name == "wait-for-postgres"` when `postgresql.enabled: true`; (b) assert no initContainers when `postgresql.enabled: false`.

**Prior findings from 622ab4fd all resolved** — no regression on Bitnami key, allowedEmails, or ct install.

Strengths: real image build via `admin/Dockerfile` in CI is the right progression from the stub; `kind load docker-image --name kind` correctly targets the pinned cluster; the initContainer's `postgresql.enabled` gate correctly mirrors the DATABASE_URL gate.

PR #348, opened 2026-06-14
feat(charts): template metrics Deployment/Service in postgres mode + e2e image-load (HD-3.2)

dodizzle:
## Summary
- Template the **metrics** service in postgres mode: `metrics-deployment.yaml` (image from values, `/health` probes on :3460), `metrics-service.yaml` (dashboard), `metrics-serviceaccount.yaml`, `metrics-secret.yaml` (assembles `METRICS_DATABASE_URL` + persisted `SHIPWRIGHT_SESSION_SECRET`). All gated on `metrics.enabled` (default true) — `metrics.enabled=false` omits every metrics resource.
- **Separate metrics DB**: `METRICS_DATABASE_URL` targets `shipwright_metrics` (NOT the admin DB), created via a postgres `initdb` script. This is deliberate — sharing the admin DB makes its schema non-empty and breaks admin's `prisma migrate deploy` (Prisma **P3005**), reproduced and fixed here.
- Extended the helm-e2e CI to build + kind-load the `shipwright-metrics` image (mirrors HD-3.1's admin image-load) so `ct install` actually runs the metrics pod.
- Chart bumped **0.3.0 → 0.4.0** (CHANGELOG + artifacthub annotation). No application code changed.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Metrics Deployment renders with METRICS_DATABASE_URL → bundled postgres metrics DB + Service exposing dashboard; `metrics.enabled=false` omits all metrics resources | MET |
| 2 | helm-unit golden specs (enabled/disabled, DB wiring) + kubeconform on rendered resources; no tests retired | MET |

## Test Plan
- `helm lint` → 0 failed; `helm unittest` → 9 suites / 66 tests (25 new metrics); `kubeconform` → 16/16 valid
- `helm template` default → metrics Deployment/Service/SA/Secret present; `--set metrics.enabled=false` → 0 metrics resources
- **Real kind e2e**: built+loaded admin+metrics images, `ct install` → postgres + admin + metrics pods all `Ready`; both `shipwright_admin` + `shipwright_metrics` DBs created; metrics in POSTGRES mode, `/health` 200, `/dashboard` 302→login
- `ct lint --check-version-increment` → PASS (0.4.0); actionlint clean; check-strings + gitleaks clean; ci.yml/release.yml untouched
- [x] Pre-ship checks passing

### DIFF
--- a/charts/shipwright/templates/metrics-deployment.yaml
+++ b/charts/shipwright/templates/metrics-deployment.yaml
@@ -1,0 +1,86 @@
+{{- if .Values.metrics.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "shipwright.metrics.fullname" . }}
+  labels:
+    {{- include "shipwright.metrics.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.metrics.replicas | default 1 }}
+  selector:
+    matchLabels:
+      {{- include "shipwright.metrics.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "shipwright.metrics.selectorLabels" . | nindent 8 }}
+    spec:
+      serviceAccountName: {{ include "shipwright.metrics.serviceAccountName" . }}
+      {{- with .Values.global.imagePullSecrets }}
+      imagePullSecrets:
+        {{- range . }}
+        - name: {{ . }}
+        {{- end }}
+      {{- end }}
+      {{- if .Values.postgresql.enabled }}
+      initContainers:
+        - name: wait-for-postgres
+          image: busybox:1.36
+          command: ['sh', '-c', 'until nc -z {{ include "shipwright.postgresql.fullname" . }} 5432; do echo "waiting for postgresql..."; sleep 2; done']
+      {{- end }}
+      containers:
+        - name: metrics
+          image: "{{ with .Values.global.imageRegistry }}{{ . }}/{{ end }}{{ .Values.metrics.image.repository }}:{{ .Values.metrics.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.metrics.image.pullPolicy | default .Values.imagePullPolicy }}
+          ports:
+            - name: http
+              containerPort: 3460
+              protocol: TCP
+          livenessProbe:
+            httpGet:
+              path: /health
+              port: 3460
+            initialDelaySeconds: 10
+            periodSeconds: 15
+            timeoutSeconds: 3
+            failureThreshold: 3
+          readinessProbe:
+            httpGet:
+              path: /health
+              port: 3460
+            initialDelaySeconds: 5
+            periodSeconds: 10
+            timeoutSeconds: 3
+            failureThreshold: 3
+          env:
+            # Bind the metrics dashboard to the chart-standard port 3460.
+            - name: METRICS_API_PORT
+              value: "3460"
+            # Force the read backend out of offline/fixtures mode. With a postgres
+            # METRICS_DATABASE_URL present, the provider selects "postgres" mode
+            # (any METRICS_OFFLINE value other than "true" is ignored).
+            - name: METRICS_OFFLINE
+              value: "false"
+            # Signs the dashboard session JWT. Generated into the metrics Secret
+            # on first install and preserved across `helm upgrade`.
+            - name: SHIPWRIGHT_SESSION_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "shipwright.metrics.secretName" . }}
+                  key: SHIPWRIGHT_SESSION_SECRET
+            {{- if .Values.postgresql.enabled }}
+            # Assembled in the metrics Secret so the Postgres password is never in
+            # plaintext Deployment env. Targets the bundled PostgreSQL in postgres
+            # mode; the metrics provider bootstraps its own `events` table on boot
+            # (no separate migration job needed) and the dashboard serves /dashboard.
+            - name: METRICS_DATABASE_URL
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "shipwright.metrics.secretName" . }}
+                  key: METRICS_DATABASE_URL
+            {{- end }}
+          {{- with .Values.metrics.resources }}
+          resources:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+{{- end }}
--- a/charts/shipwright/templates/metrics-deployment.yaml
+++ b/charts/shipwright/templates/metrics-deployment.yaml
@@ -56,11 +56,6 @@ spec:
             # Bind the metrics dashboard to the chart-standard port 3460.
             - name: METRICS_API_PORT
               value: "3460"
-            # Force the read backend out of offline/fixtures mode. With a postgres
-            # METRICS_DATABASE_URL present, the provider selects "postgres" mode
-            # (any METRICS_OFFLINE value other than "true" is ignored).
-            - name: METRICS_OFFLINE
-              value: "false"
             # Signs the dashboard session JWT. Generated into the metrics Secret
             # on first install and preserved across `helm upgrade`.
             - name: SHIPWRIGHT_SESSION_SECRET
--- a/charts/shipwright/tests/metrics_workload_test.yaml
+++ b/charts/shipwright/tests/metrics_workload_test.yaml
@@ -1,0 +1,197 @@
+suite: metrics Deployment / Service / ServiceAccount workloads
+# HD-3.2 adds the metrics workload templates (mirrors the admin pattern). These
+# assert the structural facts: image (tag defaults to appVersion), container port
+# 3460, /health liveness + readiness probes, ServiceAccount wiring, Service shape,
+# the postgres-mode env (METRICS_API_PORT, METRICS_OFFLINE=false), and the
+# METRICS_DATABASE_URL wiring to the bundled PostgreSQL subchart (assembled in the
+# metrics Secret).
+tests:
+  - it: renders the metrics Deployment with the appVersion-defaulted image and port 3460
+    template: templates/metrics-deployment.yaml
+    asserts:
+      - isKind:
+          of: Deployment
+      - equal:
+          path: spec.template.spec.containers[0].image
+          value: shipwright-metrics:0.1.0
+      - equal:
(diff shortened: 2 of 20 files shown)


charts/shipwright/values.yaml
```diff
@@ -159,3 +191,17 @@ postgresql:
...
+    # (which would break admin's prisma migrate baseline — Prisma P3005). Runs once
+    # as the postgres superuser at initdb time; idempotent via the IF NOT EXISTS guard.
+    # If you override metrics.database.name, update the database name here to match.
+    initdb:
+      scripts:
+        # `.sql` initdb scripts are executed by the Bitnami entrypoint as the
+        # postgres superuser over the local socket (no password needed), against
+        # the default database, right after the app user/database are created.
+        00-create-metrics-db.sql: |
+          SELECT 'CREATE DATABASE shipwright_metrics OWNER shipwright'
```
dodizzle: **Important (conf 80, bug on override path): initdb DB name can diverge from METRICS_DATABASE_URL.** This script hardcodes `shipwright_metrics`, but the `shipwright.metrics.databaseName` helper (used to build `METRICS_DATABASE_URL`) reads `metrics.database.name`. If a user overrides `metrics.database.name`, the Deployment connects to the custom DB but initdb only created `shipwright_metrics` → fresh install fails to connect, and CI won't catch it (CI uses the default). Fix: template the init script so the created DB name tracks the helper — e.g. move it to a parent-chart ConfigMap rendered with `{{ include "shipwright.metrics.databaseName" . }}` and wire `postgresql.primary.initdb.scriptsConfigMap`, keeping `CREATE DATABASE ... WHERE NOT EXISTS` idempotent. (Default case is correct today; this only closes the override gap.)

dodizzle (commented):
**Verdict: COMMENT** (self-review). Strong PR — complete metrics.enabled gating, session-secret persistence mirrors admin, DB password kept out of plaintext env, non-vacuous golden specs, and a real 2-pod kind e2e. One important item before merge: the postgres initdb script hardcodes the metrics DB name while METRICS_DATABASE_URL honors the metrics.database.name override — overriding the name silently targets a DB that initdb never created (default works; only the override path breaks). See inline.

dodizzle (commented):
**Verdict: COMMENT** — One item before merge: in `.github/workflows/helm.yml`, the `helm-e2e` job pins Helm to `v3.16.4` (line 90) while `helm-checks` uses `v3.18.6`, violating the workflow comment on the `helm-checks` step that says to keep the two pins in step. helm-unittest isn't installed in `helm-e2e` so the `platformHooks` concern doesn't apply there, but if any `ct install` or chart template behavior changes between 3.16 and 3.18, the check gate and the install gate test different Helm semantics. One-line fix: change `helm-e2e`'s `azure/setup-helm@v4` version to `v3.18.6`.

Otherwise this is a solid PR. The initdb ConfigMap fix is the right architectural solution — using `shipwright.metrics.databaseName` in both the ConfigMap SQL and `metrics-secret.yaml` guarantees the created database and `METRICS_DATABASE_URL` stay in lockstep across any `metrics.database.name` override. The `postgresql.enabled`-only gate on the ConfigMap is the correct call given the Bitnami subchart `tpl` context constraint. The helm-unittest coverage for the override-safe behavior is thorough.

charts/shipwright/values.yaml
dmcaulay: This hardcodes `'shipwright_metrics'` but `METRICS_DATABASE_URL` targets `{{ include "shipwright.metrics.databaseName" . }}` (resolved from `metrics.database.name`). When a user sets `metrics.database.name: custom_name`, the URL points at `custom_name` but initdb only creates `shipwright_metrics` — pod crashes on connect. The fix requires templating the initdb SQL. Options: (1) use `postgresql.primary.initdb.scriptsConfigMap` pointing at a Helm-rendered ConfigMap that calls the `shipwright.metrics.databaseName` helper, or (2) remove `metrics.database.name` from the public values surface entirely until the initdb hook can be templated. The values.yaml comment acknowledges this but provides no enforcement.

charts/shipwright/templates/metrics-deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+            timeoutSeconds: 3
+            failureThreshold: 3
+          env:
+            # Bind the metrics dashboard to the chart-standard port 3460.
+            - name: METRICS_API_PORT
+              value: "3460"
+            # Force the read backend out of offline/fixtures mode. With a postgres
+            # METRICS_DATABASE_URL present, the provider selects "postgres" mode
+            # (any METRICS_OFFLINE value other than "true" is ignored).
+            - name: METRICS_OFFLINE
```
dmcaulay: `METRICS_OFFLINE=false` is a no-op. The provider (`select-provider.ts`) only checks `METRICS_OFFLINE === "true"` to enter fixtures mode — any other value, including `"false"`, is ignored. Postgres mode is selected automatically when `METRICS_DATABASE_URL` is present. Remove this env entry; it adds misleading noise for operators inspecting pod env.

dmcaulay (commented):
COMMENT — 1 important issue before merge: the `postgresql.primary.initdb.scripts` SQL hardcodes `'shipwright_metrics'` instead of using the resolved `metrics.database.name` value. When a user overrides `metrics.database.name`, `METRICS_DATABASE_URL` targets the custom DB but initdb only creates `shipwright_metrics` — the metrics pod crashes on connect. Fix: template the initdb SQL so the database name matches `{{ include "shipwright.metrics.databaseName" . }}`, or remove `metrics.database.name` from the public surface until an initdb hook is in place. Also 1 suggestion: `METRICS_OFFLINE=false` is a no-op env entry — the provider only checks for `=== "true"`.

dmcaulay (approved):
APPROVE — the initdb hardcoding finding from the previous review is fully resolved. The new `metrics-postgres-initdb-configmap.yaml` renders the CREATE DATABASE SQL via the `shipwright.metrics.databaseName` helper (the same one that assembles `METRICS_DATABASE_URL` in the metrics Secret), and is wired into the Bitnami subchart via `postgresql.primary.initdb.scriptsConfigMap` with a tpl-evaluated release-name string. Overriding `metrics.database.name` now correctly updates both the database created at initdb time and the URL the metrics pod connects to. Test coverage is thorough: default name, custom name, disabled-postgres gate, and absence-of-hardcoded-default assertions. No new issues in the fix commits.

charts/shipwright/tests/metrics_workload_test.yaml
```diff
@@ -0,0 +1,197 @@
...
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: METRICS_API_PORT
+            value: "3460"
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: METRICS_OFFLINE
```
dodizzle: `METRICS_OFFLINE` was removed from `metrics-deployment.yaml` in `d5dc5c6` (no-op in postgres mode), but this `contains` assertion still expects it. Remove lines 47-50 (the `METRICS_OFFLINE` entry) and update the test name on line 37 — "forces postgres (non-offline) mode via env" no longer applies since postgres mode is selected by `METRICS_DATABASE_URL` presence, not by this env var.

dodizzle (commented):
**Verdict: COMMENT** (re-review after 3 new commits)

**Prior finding addressed**: Helm version mismatch resolved — `helm-e2e` now pins `v3.18.6` matching `helm-checks`.

**New issue**: The `METRICS_OFFLINE=false` env var removed from `metrics-deployment.yaml` in `d5dc5c6` is still asserted in `metrics_workload_test.yaml:48`. The `contains` check on `spec.template.spec.containers[0].env` will fail `helm unittest charts/shipwright` and block CI. Remove the `METRICS_OFFLINE` assertion (lines 47-50) and update the test name and file header comment to drop the offline-mode reference.

charts/shipwright/tests/metrics_workload_test.yaml
```diff
@@ -0,0 +1,197 @@
...
+  - it: binds the metrics port and forces postgres (non-offline) mode via env
+    template: templates/metrics-deployment.yaml
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: METRICS_API_PORT
+            value: "3460"
+      - contains:
+          path: spec.template.spec.containers[0].env
```
dmcaulay: This assertion is now stale. `METRICS_OFFLINE=false` was removed from `metrics-deployment.yaml` in this commit (correctly — it was a no-op), but these 4 lines still assert it must be present, breaking CI.

Remove the `METRICS_OFFLINE` `contains` block (lines 46–49) and update the test name on line 37 to drop "forces postgres (non-offline) mode via env" — postgres mode is now implicit via `METRICS_DATABASE_URL` presence, not an explicit env flag.

dmcaulay (commented):
COMMENT — CI broken by stale test assertion after removing `METRICS_OFFLINE=false`.

The template removal is correct: `select-provider.ts` only reads `METRICS_OFFLINE === "true"` to enter fixtures mode — `"false"` is genuinely a no-op, and removing it cleans up operator noise. The Helm pin alignment (`v3.16.4` → `v3.18.6` in `helm-e2e`) is also the right call.

But `charts/shipwright/tests/metrics_workload_test.yaml` line 46–49 still asserts `env contains {name: METRICS_OFFLINE, value: "false"}` — CI confirms the `contains` fail. One fix needed before merge: remove those 4 lines from the test spec (and update the test name to drop "forces postgres (non-offline) mode via env").

dmcaulay (approved):
APPROVE — all findings resolved. The stale `METRICS_OFFLINE=false` test assertion is removed (lines 46-49 from `metrics_workload_test.yaml`, test name updated). Helm pin alignment (`helm-e2e` v3.18.6) and initdb ConfigMap fix were confirmed in prior review passes. No new issues in this commit. Clean to merge.

dodizzle (commented):
**Verdict: APPROVE** (re-review after dfe691f)

**Prior finding addressed**: Stale `METRICS_OFFLINE=false` assertion removed from `metrics_workload_test.yaml` — the `contains` block is gone, the suite header comment is updated, and the `METRICS_API_PORT` assertion is intact. Both prior findings (Helm version pin + stale test assertion) are now fully resolved.

No new issues. All three ci/ values variants carry `metrics.image.pullPolicy: Never`; the kind image-load steps are fully plumbed. Clean to merge pending Helm CI green.

PR #350, opened 2026-06-14
feat(admin): add KubernetesAgentProvisioner + NoopAgentProvisioner (HD-1.3)

dodizzle:
## Summary
- Add `admin/src/agent-provisioner.ts`: an `AgentProvisioner` interface with two implementations.
  - **`KubernetesAgentProvisioner`**: `provision(agentId)` mints a scoped per-agent token (real `AgentTokenService`), then creates the per-agent **Secret** (carrying the token) and **Deployment** via the injected `KubernetesClient` — Secret-before-Deployment, retry-safe (409 ConflictError tolerated → already-provisioned), with **rollback** (if Deployment creation fails after the Secret was freshly created, the Secret is best-effort deleted so a retry starts clean). `deprovision(agentId)` deletes Deployment then Secret, tolerating already-absent (404).
  - **`NoopAgentProvisioner`**: no-ops, so the admin service runs unchanged when k8s provisioning is disabled.
- Composes HD-1.1 (`KubernetesClient`/`RecordedKubernetesClient`), HD-1.2 (`buildAgent*Manifest`, `sanitizeAgentName`), and the existing `AgentTokenService`.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | provision() mints token, creates Secret then Deployment, retry-safe; deprovision() removes both, tolerates absent | MET |
| 2 | NoopAgentProvisioner no-ops the interface | MET |
| 3 | INTEGRATION tests (injected fake KubernetesClient + real AgentTokenService vs test Postgres): provision creates Secret+Deployment with minted token, deprovision deletes, failure/rollback path; none retired | MET |

## Test Plan
- 10 integration tests in `agent-provisioner.integration.test.ts` (provision Secret+Deployment carry the validated minted token; ordering; retry-safety; idempotent pre-existing; failure→Secret rollback; deprovision delete + absent-tolerant). 3 Noop tests run without a DB; 7 DB-backed tests guard on `DATABASE_URL_ADMIN_TEST` (same `describe.skip` pattern as the other admin integration tests) → run in CI's Postgres. Verified 10/0 locally against a real test DB.
- Biome lint clean; `@shipwright/admin` typecheck 0; check-strings + gitleaks clean.

## Known limitation (flagging for review)
The HD-1.2 manifest builders emit an `ownerReference` (for GC-cascade of agents on admin uninstall), but `KubernetesClient.createDeployment/createSecret` accept the simpler `DeploymentSpec`/`SecretSpec`, which don't carry ownerReferences — so the ownerRef is **not currently transmitted to the cluster**. Wiring it through would require extending the `KubernetesClient` create surface (HD-1.1 module), which is out of scope for this task. Flagging as a follow-up; HD-1.3's ACs (token + Secret/Deployment presence, retry, rollback) are fully met.

### DIFF
--- a/admin/src/agent-provisioner.integration.test.ts
+++ b/admin/src/agent-provisioner.integration.test.ts
@@ -178,6 +178,26 @@ describeOrSkip("KubernetesAgentProvisioner (integration)", () => {
     ).rejects.toThrow();
   });
 
+  it("provision() rolls back the freshly-minted Secret when the Deployment already exists (409)", async () => {
+    const agentId = await createAgent();
+    const name = sanitizeAgentName(agentId);
+    // Seed ONLY the Deployment — the Secret does NOT pre-exist. This call mints
+    // a token and creates a new Secret, then hits a 409 on the Deployment. The
+    // orphaned Secret (carrying a fresh, unused token) must be rolled back.
+    const k8s = emptyClient();
+    await k8s.createDeployment(NAMESPACE, {
+      name,
+      image: `${CONFIG.image}:${CONFIG.imageTag}`,
+    });
+    const provisioner = new KubernetesAgentProvisioner(k8s, tokens, CONFIG);
+
+    // (a) Idempotent success — does NOT throw despite the Deployment conflict.
+    await expect(provisioner.provision(agentId)).resolves.toBeDefined();
+
+    // (b) The Secret created during THIS call was deleted, not left behind.
+    await expect(k8s.getSecret(NAMESPACE, `${name}-token`)).rejects.toThrow();
+  });
+
   it("provision() succeeds idempotently when both resources already exist", async () => {
     const agentId = await createAgent();
     const name = sanitizeAgentName(agentId);
--- a/admin/src/agent-provisioner.integration.test.ts
+++ b/admin/src/agent-provisioner.integration.test.ts
@@ -1,0 +1,256 @@
+/**
+ * admin/src/agent-provisioner.integration.test.ts
+ * Integration tests for KubernetesAgentProvisioner.
+ *
+ * Composes the REAL AgentTokenService against a real Postgres test DB with an
+ * injected in-memory KubernetesClient (RecordedKubernetesClient) standing in
+ * for the cluster. No globals are overridden — the fake is injected via the
+ * constructor.
+ *
+ * Requires DATABASE_URL_ADMIN_TEST to be set; skips otherwise (CI provides it).
+ */
+
+import { beforeEach, describe, expect, it } from "bun:test";
+import { PrismaClient } from "../prisma/client/index.js";
+import {
+  buildAgentSecretManifest,
+  sanitizeAgentName,
+} from "./agent-manifest.ts";
+import {
+  type KubernetesAgentProvisionerConfig,
+  KubernetesAgentProvisioner,
+  NoopAgentProvisioner,
+} from "./agent-provisioner.ts";
+import { AgentTokenService } from "./agent-tokens.ts";
+import { ConflictError } from "./errors.ts";
+import {
+  type DeploymentSpec,
+  type KubernetesClient,
+  type KubernetesSecret,
+  RecordedKubernetesClient,
+  type SecretSpec,
+} from "./kubernetes-client.ts";
+
+const TEST_DB = process.env.DATABASE_URL_ADMIN_TEST;
+const describeOrSkip = TEST_DB ? describe : describe.skip;
+
+const NAMESPACE = "shipwright";
+
+const CONFIG: KubernetesAgentProvisionerConfig = {
+  namespace: NAMESPACE,
+  image: "ghcr.io/app-vitals/shipwright-agent",
+  imageTag: "v1.2.3",
+  apiUrl: "http://shipwright-admin.shipwright.svc:3001",
+  adminDeploymentName: "shipwright-admin",
+  adminDeploymentUid: "11112222-3333-4444-5555-666677778888",
+};
+
+function makePrisma(): PrismaClient {
+  return new PrismaClient({
+    datasources: { db: { url: TEST_DB as string } },
+  });
+}
+
+function emptyClient(): RecordedKubernetesClient {
+  return new RecordedKubernetesClient({ deployments: {}, secrets: {} });
+}
+
+/** Decode the token value the Secret carries under the default "token" key. */
+function decodeSecretToken(secret: KubernetesSecret): string {
+  return Buffer.from(secret.data.token, "base64").toString("utf-8");
+}
+
+describeOrSkip("KubernetesAgentProvisioner (integration)", () => {
+  let prisma: PrismaClient;
+  let tokens: AgentTokenService;
+
+  beforeEach(async () => {
+    prisma = makePrisma();
+    await prisma.agentToken.deleteMany();
+    await prisma.agentCronJob.deleteMany();
+    await prisma.agentTool.deleteMany();
+    await prisma.agentEnv.deleteMany();
+    await prisma.agent.deleteMany();
+    tokens = new AgentTokenService(prisma);
+  });
+
+  async function createAgent(name = "Test Agent"): Promise<string> {
+    const agent = await prisma.agent.create({ data: { name } });
+    return agent.id;
+  }
+
+  // ─── provision ──────────────────────────────────────────────────────────────
+
+  it("provision() mints a token, then creates the Secret and Deployment", async () => {
+    const agentId = await createAgent();
+    const k8s = emptyClient();
+    const provisioner = new KubernetesAgentProvisioner(k8s, tokens, CONFIG);
+
+    const result = await provisioner.provision(agentId);
+
+    const name = sanitizeAgentName(agentId);
+    expect(result.resourceName).toBe(name);
+    expect(result.secretName).toBe(`${name}-token`);
+    expect(result.deploymentName).toBe(name);
+    expect(result.rawToken).toBeDefined();
+
+    // Secret carries the minted token, and the token validates via the service.
+    const secret = await k8s.getSecret(NAMESPACE, `${name}-token`);
+    const carried = decodeSecretToken(secret);
+    expect(result.rawToken).toBeDefined();
+    expect(carried).toBe(result.rawToken as string);
+    const validated = await tokens.validate(carried);
+    expect(validated?.agentId).toBe(agentId);
+
+    // Deployment exists and points at the right image.
+    const dep = await k8s.getDeployment(NAMESPACE, name);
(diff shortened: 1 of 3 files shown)


admin/src/agent-provisioner.ts
```diff
@@ -0,0 +1,273 @@
...
+    // 3. Deployment. If creation fails AFTER we created the Secret in THIS call,
+    //    roll the Secret back (best-effort) so a retry starts clean and never
+    //    leaks a half-provisioned state.
+    try {
+      await this.k8s.createDeployment(
+        this.config.namespace,
+        this.deploymentSpec(agentId, resourceName, secretName),
+      );
+    } catch (err) {
+      if (isConflict(err)) {
```
dodizzle, line 181: **Important (conf 80, rollback gap): leaked Secret when Deployment returns 409 after Secret was created this call.** If createSecret succeeded (secretCreated=true) and createDeployment then throws ConflictError, this branch returns early as 'already provisioned' WITHOUT rolling back the Secret just created — leaving a Secret with a freshly minted token the pre-existing Deployment isn't using (orphaned token + leaked Secret). Trigger: Secret manually deleted while Deployment kept running, then re-provision. Fix: in this isConflict branch,  before returning.

admin/src/agent-provisioner.ts
```diff
@@ -0,0 +1,273 @@
...
+    await this.deleteDeploymentBestEffort(resourceName);
+    await this.deleteSecretBestEffort(secretName);
+  }
+
+  // ─── Spec derivation (via the pure manifest builders) ─────────────────────
+
+  private secretSpec(secretName: string, rawToken: string): SecretSpec {
+    // Build the canonical manifest (owner refs, encoding) for consistency, then
+    // hand the KubernetesClient the plain stringData it create-encodes itself.
+    buildAgentSecretManifest({
```
dodizzle: **Important (conf 85, dead code): buildAgentSecretManifest() return value discarded.** This call's result is unused — the SecretSpec at line 203 is constructed independently. Either derive the secret body from the manifest builder's output (so the builder is actually load-bearing) or remove the dead call and add a one-line comment on why SecretSpec is built inline. As-is, a future refactor will silently drop it.

dodizzle (commented):
**Verdict: COMMENT** (self-review). Solid provisioner — retry-safety, ordered creation, deprovision 404-tolerance, and the 'only roll back what THIS call created' invariant (for the non-409 path) are all correct and tested. Three items before merge (see inline + below). The disclosed ownerRef-not-transmitted limitation is acceptable as a documented follow-up.

**Finding 3 (test, conf 77):** no integration test covers the Secret-leaked-on-Deployment-409 path (AC3 requires the failure/rollback path). Add: seed the Deployment (not the Secret) as pre-existing, call provision(), assert no throw AND the Secret created this call is deleted (not left behind).

admin/src/agent-provisioner.ts
```diff
@@ -0,0 +1,273 @@
...
+    const resourceName = this.resourceName(agentId);
+    const secretName = this.secretNameFor(resourceName);
+    const result: ProvisionResult = {
+      resourceName,
+      secretName,
+      deploymentName: resourceName,
+    };
+
+    // 1. Token — mint a fresh scoped token for this agent.
+    const { rawToken } = await this.tokens.create(agentId, "k8s-provision");
```
dodizzle: Token is minted unconditionally on every call. On retry when the Secret already exists (409 path below), this token is orphaned in the DB — the cluster Secret still holds the original. `result.rawToken` will point to a credential that doesn't match the agent's actual Secret.

Options: (1) attempt `getSecret` first and skip minting if the Secret already exists, returning `rawToken: undefined` to signal the prior credential is in effect; (2) or mint but explicitly document that callers must not trust `rawToken` when the Secret was pre-existing.

admin/src/agent-provisioner.ts
dodizzle: `buildAgentSecretManifest()` is called here but the return value is discarded. The function is pure — this call has no observable effect and may mislead maintainers into thinking ownerReferences are being applied to the cluster Secret (they aren't, since `SecretSpec` has no `ownerReferences` field). Remove the call and add a comment noting that ownerReferences are intentionally omitted from the Secret spec.

dodizzle (commented):
COMMENT — One issue to address before merge.

**Token minted on every retry (`agent-provisioner.ts:141`):** `provision()` unconditionally calls `this.tokens.create()` on every invocation. On a retry where the Secret already exists (409 path), the freshly-minted token is orphaned in the DB — it's never written to the cluster Secret. The returned `rawToken` doesn't match what the cluster Secret holds. On each retry a new unreachable token row accumulates. Fix: check whether the Secret already exists before minting (e.g. attempt `getSecret` first), or set `rawToken: undefined` in the conflict path to signal callers that the original credential is already in effect.

**Suggestion (non-blocking):** `secretSpec()` at line 195 calls `buildAgentSecretManifest()` and discards the return — dead code. Remove the call and add a comment noting ownerReferences are intentionally omitted because `SecretSpec` doesn't support them.

admin/src/agent-provisioner.ts
```diff
@@ -0,0 +1,273 @@
...
+      namespace: this.config.namespace,
+      token: rawToken,
+      tokenKey: this.tokenKey,
+      adminDeploymentName: this.config.adminDeploymentName,
+      adminDeploymentUid: this.config.adminDeploymentUid,
+    });
+    return { name: secretName, stringData: { [this.tokenKey]: rawToken } };
+  }
+
+  private deploymentSpec(
```
dmcaulay, line 232: The `deploymentSpec()` method builds the full manifest via `buildAgentDeploymentManifest()` (which wires `SHIPWRIGHT_AGENT_API_KEY` via `secretKeyRef`) but then constructs its own `env` with only plain string literals — `DeploymentSpec.env` is `Record<string, string>`, so there's no way to express `valueFrom`/`secretKeyRef`. The agent container launched from this Deployment will have no token env var and cannot authenticate.

The module header explicitly states "the Deployment references the Secret by name (`secretKeyRef`)" but the code doesn't implement this. Either extend `DeploymentSpec`/`deploymentBody` to support `valueFrom` entries, or pass the token as a plain `SHIPWRIGHT_TOKEN` value in `env`.

admin/src/agent-provisioner.ts
dmcaulay: `tokens.create()` is called unconditionally before the Secret/Deployment conflict checks. On a retry (both resources already exist), a new token is minted, the Secret returns 409 immediately, and the new token is never stored or revoked — leaking an orphaned `AgentToken` row per retry call.

Consider moving `tokens.create()` after the Secret conflict check, or revoking the newly-minted token in the conflict path.

admin/src/agent-provisioner.ts
```diff
@@ -0,0 +1,273 @@
...
+    await this.deleteSecretBestEffort(secretName);
+  }
+
+  // ─── Spec derivation (via the pure manifest builders) ─────────────────────
+
+  private secretSpec(secretName: string, rawToken: string): SecretSpec {
+    // Build the canonical manifest (owner refs, encoding) for consistency, then
+    // hand the KubernetesClient the plain stringData it create-encodes itself.
+    buildAgentSecretManifest({
+      name: secretName,
```
dmcaulay, line 218: `buildAgentSecretManifest()` is called here but its return value is discarded — this is dead code. The method then constructs a bare `{ name, stringData }` manually. The ownerReferences and other fields the builder emits are silently dropped. Remove this call (or actually use the returned manifest's fields).

admin/src/agent-provisioner.integration.test.ts
```diff
@@ -0,0 +1,256 @@
...
+/** Decode the token value the Secret carries under the default "token" key. */
+function decodeSecretToken(secret: KubernetesSecret): string {
+  return Buffer.from(secret.data.token, "base64").toString("utf-8");
+}
+
+describeOrSkip("KubernetesAgentProvisioner (integration)", () => {
+  let prisma: PrismaClient;
+  let tokens: AgentTokenService;
+
+  beforeEach(async () => {
```
dmcaulay, line 67: `beforeEach` creates a new `PrismaClient` on every test but there's no `afterEach` calling `prisma.$disconnect()`. Each test opens a connection pool without releasing it. Add `afterEach(async () => { await prisma.$disconnect(); })` to avoid connection pool exhaustion under a full CI suite run.

dmcaulay (commented):
COMMENT — 3 issues before merge:

1. **[Critical] Agent token secretKeyRef dropped** (`admin/src/agent-provisioner.ts:206`): `deploymentSpec()` calls `buildAgentDeploymentManifest()` (which builds the token secretKeyRef env var) but then substitutes a hardcoded `env` block with only `SHIPWRIGHT_AGENT_ID` and `SHIPWRIGHT_API_URL` as plain string values — `DeploymentSpec.env` is `Record<string, string>`, no `valueFrom` support. The agent container will have no token and cannot authenticate to the admin API. The module header says "the Deployment references the Secret by name (secretKeyRef)" but the code doesn't implement this.

2. **[Important] Orphaned tokens on retry** (`admin/src/agent-provisioner.ts:141`): `tokens.create()` is called unconditionally before any ConflictError check. On a retry (both resources already exist), a new DB token is minted but the Secret returns 409 immediately, so the token is never stored and never revoked. Each `provision()` call on an already-provisioned agent leaks one `AgentToken` row.

3. **[Important] Dead call in secretSpec()** (`admin/src/agent-provisioner.ts:195`): `buildAgentSecretManifest()` is called but its return value is immediately discarded. The comment says "for consistency" but no consistency is achieved — a bare `{ name, stringData }` is returned instead. Remove the dead call.

dodizzle (commented):
**APPROVE** — Re-review after 3 follow-up commits.

All findings from the prior review are addressed:
- **Token orphaning on retry**: `provision()` now calls `getSecret()` before minting — token is only created when the Secret is absent, so idempotent re-invocation neither leaks DB rows nor returns a mismatched `rawToken`.
- **secretKeyRef dropped**: `deploymentSpec()` now passes `container.env` from the manifest builder directly via the new `envVars: KubernetesEnvVar[]` field on `DeploymentSpec`. The `deploymentBody()` merges plain `env` and `envVars` so the `SHIPWRIGHT_AGENT_API_KEY` secretKeyRef reaches the container spec as intended.
- **Dead manifest call in secretSpec()**: `secretSpec()` now actually uses `buildAgentSecretManifest()`'s return value — decodes the base64 `data` back to `stringData` so the builder is the single source of truth.
- **409-rollback test**: Line 185 in the integration suite covers the previously-untested path (Deployment pre-exists, Secret newly minted → Secret rolled back).

Provisioner is correct, retry-safe, rollback-tested, and the agent container receives its token via secretKeyRef as documented.

PR #351, opened 2026-06-14
feat(charts): add Ingress networking + helm test connection hook (HD-3.3)

dodizzle:
## Summary
- Add `templates/ingress.yaml` — a `networking.k8s.io/v1` Ingress rendered ONLY when `networking.type == "ingress"` (new option alongside ClusterIP/NodePort/LoadBalancer; default stays ClusterIP so Ingress is off by default). Configurable `ingressClassName` (nginx/alb), `host`, and `annotations`. Routes `/` → admin Service and `/dashboard` → metrics Service (the /dashboard path is omitted when `metrics.enabled=false`); `pathType: Prefix`, named `http` ports.
- Add `templates/tests/test-connection.yaml` — a `helm.sh/hook: test` Pod (pinned `curlimages/curl:8.11.1`) that curls admin `/health` (must 200) and, when metrics enabled, metrics `/dashboard` (200/302); `set -eu` + explicit non-zero exit so `helm test` fails loudly.
- `values.yaml`/`values.schema.json`: `networking.ingress.{className,host,annotations}` + `"ingress"` in the type enum.
- Chart bumped **0.5.0 → 0.6.0** (CHANGELOG + artifacthub annotation).

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | networking.type=ingress renders Ingress w/ className/host + correct admin + /dashboard backends/paths; no Ingress for other types | MET |
| 2 | helm test runs the connection Pod, succeeds against a live install (ct install e2e) | MET |
| 3 | helm-unit golden specs (className/host/paths, conditional rendering) + kubeconform; test-connection is the helm-e2e smoke; none retired | MET |

## Test Plan
- `helm lint` 0 failed; `helm unittest` → 11 suites / 86 tests (11 new ingress specs); `kubeconform` 18/18 valid (Ingress validates as networking.k8s.io/v1)
- Render checks: `--set networking.type=ingress` → Ingress (className=nginx, host=shipwright.local, /→admin, /dashboard→metrics); default ClusterIP → no Ingress; `metrics.enabled=false` → no /dashboard path
- **Live kind e2e `helm test` PASS**: connection Pod logs `admin /health -> 200`, `metrics /dashboard -> 302`, `OK: connection checks passed`; `kubectl get ingress` shows the nginx/shipwright.local rule
- `ct lint --check-version-increment` PASS (0.6.0); check-strings + gitleaks clean; ci.yml/helm.yml/release.yml untouched
- [x] Pre-ship checks passing

### DIFF
--- a/charts/shipwright/templates/ingress.yaml
+++ b/charts/shipwright/templates/ingress.yaml
@@ -1,0 +1,42 @@
+{{- if eq .Values.networking.type "ingress" }}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ include "shipwright.fullname" . }}
+  labels:
+    {{- include "shipwright.labels" . | nindent 4 }}
+  {{- with .Values.networking.ingress.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- with .Values.networking.ingress.className }}
+  ingressClassName: {{ . }}
+  {{- end }}
+  rules:
+    - host: {{ .Values.networking.ingress.host | quote }}
+      http:
+        paths:
+          {{- if .Values.metrics.enabled }}
+          # Metrics dashboard. Declared BEFORE the admin "/" catch-all so the
+          # more-specific /dashboard prefix wins (ingress controllers match the
+          # longest prefix, but ordering keeps intent explicit). Omitted entirely
+          # when metrics is disabled.
+          - path: /dashboard
+            pathType: Prefix
+            backend:
+              service:
+                name: {{ include "shipwright.metrics.fullname" . }}
+                port:
+                  name: http
+          {{- end }}
+          # Admin UI/API at the root. Everything not matched by a more-specific
+          # path above is routed to the admin service.
+          - path: /
+            pathType: Prefix
+            backend:
+              service:
+                name: {{ include "shipwright.admin.fullname" . }}
+                port:
+                  name: http
+{{- end }}
--- a/charts/shipwright/values.schema.json
+++ b/charts/shipwright/values.schema.json
@@ -28,7 +28,16 @@
       "properties": {
         "type": {
           "type": "string",
-          "enum": ["ClusterIP", "NodePort", "LoadBalancer"]
+          "enum": ["ClusterIP", "NodePort", "LoadBalancer", "ingress"]
+        },
+        "ingress": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "className": { "type": "string" },
+            "host": { "type": "string" },
+            "annotations": { "type": "object" }
+          }
         }
       },
       "required": ["type"]
--- a/charts/shipwright/CHANGELOG.md
+++ b/charts/shipwright/CHANGELOG.md
@@ -10,6 +10,25 @@ independent of `appVersion`. CI enforces this with
 `ct lint --check-version-increment`. Each release here must mirror the
 `artifacthub.io/changes` annotation in `Chart.yaml`.
 
+## [0.6.0]
+
+### Added
+
+- Ingress networking. Setting `networking.type=ingress` renders a
+  `networking.k8s.io/v1` `Ingress` (`templates/ingress.yaml`) with a configurable
+  `ingressClassName` (`networking.ingress.className`, default `nginx`), `host`
+  (`networking.ingress.host`, default `shipwright.local`), and controller-specific
+  `annotations` (`networking.ingress.annotations`, default `{}`). Rules route the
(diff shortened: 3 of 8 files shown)


dodizzle (commented):
APPROVE — Ingress + helm test hook, all 3 ACs met. Conditional render is correct (single eq guard, no stray output; absence verified for ClusterIP+LoadBalancer); path ordering /dashboard-before-/ is defensive + commented; /dashboard gated on metrics.enabled; backends use the correct fullname helpers + http port; className/annotations render conditionally. test-connection hook is production-grade (pinned public image, set -eu, restartPolicy Never, explicit exit 1, --retry-connrefused) and passed live helm test in the kind e2e. schema enum + version pin + artifacthub annotation all consistent at 0.6.0. CI fully green. Non-blocking suggestion: add an explicit NodePort no-Ingress test case (logic is correct, just not independently covered). Ready to merge.

charts/shipwright/values.schema.json
```diff
@@ -28,7 +28,16 @@
       "properties": {
         "type": {
           "type": "string",
-          "enum": ["ClusterIP", "NodePort", "LoadBalancer"]
+          "enum": ["ClusterIP", "NodePort", "LoadBalancer", "ingress"]
```
dodizzle, line 31: `"ingress"` is lowercase while its peers (`ClusterIP`, `NodePort`, `LoadBalancer`) are PascalCase — inconsistent with Kubernetes naming conventions and a usability trap. An operator will naturally try `networking.type: Ingress` and receive a schema validation error with no guidance. Change to `"Ingress"` here and update the template conditional (`eq .Values.networking.type "ingress"`) and unit test `set:` blocks to match.

dodizzle (commented):
COMMENT — Two issues before merge.

**Finding 1 (schema, conf 78):** `networking.type` enum uses lowercase `"ingress"` while its three peers (`ClusterIP`, `NodePort`, `LoadBalancer`) are PascalCase Kubernetes-conventional values. An operator who tries `networking.type: Ingress` (the natural spelling) gets a cryptic schema validation error with no hint. Fix: change the enum value to `"Ingress"` and update the template conditional (`eq .Values.networking.type "ingress"`) and all unit test `set:` blocks to match.

**Finding 2 (CI coverage, conf 77):** All three `ci/` values variants use `ClusterIP` or `LoadBalancer` — the `networking.type=ingress` path is validated by `helm-unittest` and `kubeconform` but is never rendered into the kind cluster during `ct install`. A future regression in the Ingress template would pass lint and kubeconform but escape the e2e job. Either add a `ci/ingress-values.yaml` variant (with the NGINX ingress controller wired into kind) or add a comment in `ci/` noting the deferral and capturing it as a follow-up task.

charts/shipwright/templates/ingress.yaml
```diff
@@ -0,0 +1,42 @@
+{{- if eq .Values.networking.type "ingress" }}
```
dmcaulay, line 1: When `admin.enabled=false` and `networking.type=ingress`, this renders an Ingress with a `/` backend pointing at a Service that doesn't exist. The metrics path is correctly guarded by `{{- if .Values.metrics.enabled }}`; the outer condition needs the same for admin:

```diff
-{{- if eq .Values.networking.type "ingress" }}
+{{- if and (eq .Values.networking.type "ingress") .Values.admin.enabled }}
```

In the edge case where both `admin.enabled=false` and `metrics.enabled=false`, this also produces an Ingress with zero path rules — invalid K8s.

dmcaulay (commented):

1. **Dangling admin backend when `admin.enabled=false`** (`ingress.yaml:1`): The outer guard is `{{- if eq .Values.networking.type "ingress" }}` but the admin `/` backend path renders unconditionally without checking `admin.enabled`. When `admin.enabled=false`, the Ingress references a non-existent Service. In the extreme case (`admin.enabled=false` AND `metrics.enabled=false`), the Ingress renders with zero path rules — an invalid K8s object. Fix: change the guard to `{{- if and (eq .Values.networking.type "ingress") .Values.admin.enabled }}`.

2. **No `ct install` e2e variant for `networking.type=ingress`** (`charts/shipwright/ci/` missing file): AC#2 claims `helm test` was run in a live kind install, but `ct install` only picks up files under `ci/` — none sets `networking.type: ingress`. The Ingress template and test-connection hook are exercised only by helm-unittest, not a live cluster install. Add `ci/ingress-values.yaml` with `networking.type: ingress` and wire the NGINX addon into `helm.yml` before `ct install`.

Non-blocking: the `values.yaml` comment for `networking.ingress.host` copy-pastes the `className` description (says "IngressClass to use: nginx/alb" for the `host` field). Easy one-liner fix.

PR #352, opened 2026-06-14
feat(admin): wire AgentProvisioner into POST /agents + add DELETE /agents/:id (HD-1.4)

dodizzle:
## Summary
- Wire the `AgentProvisioner` into the admin agents API (completes the HD-1.x provisioning vertical: client → manifests → provisioner → **API**):
  - `AdminDeps` now carries `provisioner`. `POST /agents` calls `provisioner.provision(agent.id)` after creating the row, still returns **201** with the record. If provision throws, the agent row is rolled back (no half-created agent) and the error maps to 500.
  - New **admin-only `DELETE /agents/:id`**: returns **204**, cascades child deletes, calls `provisioner.deprovision`; **404** unknown id, **401** missing auth, **403** non-admin scope.
- `main.ts`: `buildProvisioner()` selects `KubernetesAgentProvisioner` when `SHIPWRIGHT_K8S_PROVISIONING=enabled` (reads agent image/tag, namespace, internal URL, replicas, admin Deployment name/uid from env), else `NoopAgentProvisioner`. Default (unset) preserves today's behavior exactly.
- New env vars documented in `docs/configuration.md` (`check-config-docs` green).
- No admin-ui change (it has no agent create/delete flow — the create wizard is Slack onboarding, unrelated).

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | POST /agents provisions (real) or no-ops (Noop), still 201; DELETE /agents/:id admin-only, 204, cascades, deprovisions | MET |
| 2 | main.ts selects real vs Noop from SHIPWRIGHT_K8S_PROVISIONING, injects; default preserves behavior; env vars in configuration.md | MET |
| 3 | SMOKE tests (app.request): POST provision-invoked + admin-only + Noop path; DELETE 204/deprovision/401/403/404; existing tests kept | MET |

## Test Plan
- `bun test agents-api.smoke.test.ts api.smoke.test.ts` → 65/0 (RecordingProvisioner fake injected via deps; provision-invoked, Noop-path 201, provision-throws→500+rollback; DELETE 204+deprovision via cookie & admin bearer, 404 unknown, 401 missing, 403 per-agent bearer)
- Full `bun test` → no new failures vs the 2 known-pre-existing `task_store.test.ts` baseline
- `task check-config-docs` → all 39 env vars documented; typecheck/Biome clean; check-strings + gitleaks clean

## Breaking-change check
Safe to deploy standalone under rolling deploy: `POST /agents` gated by `SHIPWRIGHT_K8S_PROVISIONING` (Noop default → identical 201 + no side effects, verified by Noop-path test); `DELETE /agents/:id` is purely additive (previously 404'd). No existing request/response shapes changed.

### DIFF
--- a/admin/src/agents-api.smoke.test.ts
+++ b/admin/src/agents-api.smoke.test.ts
@@ -948,6 +948,57 @@ describe("admin API — delete agent", () => {
     expect(provisioner.deprovisioned).toEqual([AGENT_ID]);
   });
 
+  it("DELETE /agents/:id returns 500 and preserves the row when deprovision throws", async () => {
+    const cookie = await makeSessionCookie();
+    const deleted: string[] = [];
+    // Provisioner whose deprovision() rejects with a non-NotFound error. The
+    // throw must propagate (→ 500) BEFORE the agent row is deleted, leaving the
+    // row intact so the delete is retry-safe.
+    const provisioner: AgentProvisioner = {
+      async provision(agentId: string): Promise<ProvisionResult> {
+        return {
+          resourceName: agentId,
+          secretName: `${agentId}-token`,
+          deploymentName: agentId,
+        };
+      },
+      async deprovision(): Promise<void> {
+        throw new Error("k8s API timeout");
+      },
+    };
+    const base = makeMockDeps();
+    const deps: AdminDeps = {
+      ...base,
+      provisioner,
+      prisma: {
+        agent: {
+          findUnique: async (args: { where: { id: string } }) =>
+            args.where.id === AGENT_ID
+              ? { id: AGENT_ID, name: "Existing Agent" }
+              : null,
+          delete: async (args: { where: { id: string } }) => {
+            deleted.push(args.where.id);
+            return {
+              id: args.where.id,
+              name: "Existing Agent",
+              slackId: null,
+              createdAt: new Date("2024-01-01"),
+              updatedAt: new Date("2024-01-01"),
+            };
+          },
+        },
+      } as unknown as AdminDeps["prisma"],
+    };
+    const app = createAdminApp(deps);
+    const res = await app.request(`/agents/${AGENT_ID}`, {
+      method: "DELETE",
+      headers: { Cookie: `admin_session=${cookie}` },
+    });
+    expect(res.status).toBe(500);
+    // The deprovision error surfaced before the row was deleted — row preserved.
+    expect(deleted).toEqual([]);
+  });
+
   it("DELETE /agents/:id unknown id → 404 (no deprovision)", async () => {
     const cookie = await makeSessionCookie();
     const provisioner = new RecordingProvisioner();
--- a/admin/src/agents-api.smoke.test.ts
+++ b/admin/src/agents-api.smoke.test.ts
@@ -8,9 +8,37 @@
 
 import { beforeAll, describe, expect, it } from "bun:test";
 import { sign } from "hono/jwt";
+import type { AgentProvisioner, ProvisionResult } from "./agent-provisioner.ts";
 import type { AgentTokenService } from "./agent-tokens.ts";
 import { createAdminApp, parseAdminApiKeys } from "./agents-api.ts";
 import type { AdminDeps } from "./agents-api.ts";
+
+// ─── Recording fake provisioner ───────────────────────────────────────────────
+
+/**
+ * In-memory AgentProvisioner double that records every provision/deprovision
+ * call. Injected via deps — no mock.module, no global overrides.
+ */
+class RecordingProvisioner implements AgentProvisioner {
+  readonly provisioned: string[] = [];
+  readonly deprovisioned: string[] = [];
+
+  constructor(private readonly onProvision?: (agentId: string) => void) {}
+
+  async provision(agentId: string): Promise<ProvisionResult> {
+    this.onProvision?.(agentId);
+    this.provisioned.push(agentId);
+    return {
+      resourceName: agentId,
+      secretName: `${agentId}-token`,
+      deploymentName: agentId,
+    };
+  }
+
+  async deprovision(agentId: string): Promise<void> {
+    this.deprovisioned.push(agentId);
+  }
+}
 
 // ─── Constants ────────────────────────────────────────────────────────────────
 
@@ -241,8 +269,20 @@ function makeMockDeps(): AdminDeps {
           createdAt: new Date("2024-01-01"),
           updatedAt: new Date("2024-01-01"),
         }),
+        findUnique: async (args: { where: { id: string } }) =>
+          args.where.id === AGENT_ID
+            ? { id: AGENT_ID, name: "Existing Agent" }
+            : null,
+        delete: async (args: { where: { id: string } }) => ({
+          id: args.where.id,
+          name: "Existing Agent",
+          slackId: null,
+          createdAt: new Date("2024-01-01"),
+          updatedAt: new Date("2024-01-01"),
+        }),
       },
     } as unknown as AdminDeps["prisma"],
+    provisioner: new RecordingProvisioner(),
     sessionSecret: SESSION_SECRET,
   };
 }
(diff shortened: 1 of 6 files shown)


dodizzle (commented):
**Verdict: COMMENT** (self-review). Implementation is correct and well-tested: POST rollback re-throws after deleting the row (no 201-after-failed-provision, cleanup failure doesn't mask the original error); DELETE auth is genuinely admin-only (middleware re-applied + isAdmin check, full matrix tested: session/admin-bearer/per-agent-bearer→403/no-auth→401/404); cascade deletes backed by onDelete:Cascade on all 5 child relations; deprovision-before-DB-delete ordering correct; Noop path byte-identical. **One AC3 gap before merge:** no smoke test for DELETE when deprovision() throws (non-404). The behavior is correct (5xx, agent row preserved, retry-safe since deprovision is idempotent) but untested — a future refactor that swallowed the error or deleted the row before deprovisioning would go undetected. Add: a RecordingProvisioner that throws on deprovision + a prisma.agent.delete spy → assert res 500 AND delete never called (parallels the existing POST-rollback test). Then re-review.

dodizzle (commented):
APPROVE — re-review of ed2792e. The single prior finding is resolved: a smoke test now covers DELETE /agents/:id when deprovision() throws (asserts 500 AND the agent row is preserved / prisma.agent.delete never called), parallel to the POST-rollback test. Implementation was already correct (rollback re-throws, admin-only auth matrix, onDelete:Cascade children, deprovision-before-delete ordering, Noop default byte-identical). CI green, 0 unresolved threads. Ready to merge.

admin/src/agents-api.ts
```diff
@@ -147,6 +180,46 @@ export function createAdminApp(deps: AdminDeps): Hono<AdminAuthEnv> {
...
+      });
+      if (!existing) {
+        throw new NotFoundError(`agent ${agentId} not found`);
+      }
+
+      // Tear down the workload first, then delete the row. Child rows
+      // (AgentEnv / AgentCronJob / AgentTool / AgentToken / AgentPlugin) cascade
+      // via `onDelete: Cascade` in the Prisma schema. deprovision() tolerates an
+      // already-absent workload, so a retried delete is a no-op.
+      await provisioner.deprovision(agentId);
```
dodizzle, line 216: Ordering issue: `deprovision` before `prisma.agent.delete` creates a zombie-row failure mode. If `deprovision` succeeds but the DB delete throws, the K8s workload is gone but the agent row persists — the caller gets a 500 and the agent appears live in the admin UI with no backing workload. No compensation path exists.

The POST handler rolls back the row when provision fails; the DELETE path needs the inverse: delete the DB row first, then call `deprovision` with log-but-don't-propagate for any failure. The row is already gone, so the agent is effectively deleted; ownerReferences in the Deployment manifest allow K8s GC to clean orphaned resources if deprovision is skipped.

admin/src/agents-api.smoke.test.ts
```diff
@@ -806,6 +916,73 @@ describe("admin API — create agent", () => {
   });
 });
 
+// ─── Delete agent smoke tests ─────────────────────────────────────────────────
+
+describe("admin API — delete agent", () => {
```
dodizzle, line 921: The DELETE suite is missing a `deprovision-throws` test. The current cases cover 204, 404, 401, 403 — but not the 500 path when `deprovision` throws. Worth adding a test that injects a failing provisioner, asserts 500, and confirms `prisma.agent.delete` was not called (once the ordering is fixed). The `RecordingProvisioner` constructor already accepts a callback — a similar `onDeprovision` hook would cover this without duplicating the class.

dodizzle (commented):
COMMENT — One ordering issue in the DELETE handler: `deprovision` is called before `prisma.agent.delete`, which means a successful deprovision followed by a DB failure leaves a zombie row (agent appears live in the DB with no K8s workload). The POST handler explicitly rolls back on provision failure; the DELETE path needs the same care — delete the row first, then deprovision with log-and-ignore on failure. Everything else is solid: POST rollback is correct and well-tested, RecordingProvisioner is a clean test double, and the Noop default preserves existing behavior. One test gap: a `deprovision-throws` path for DELETE.

admin/src/main.ts
```diff
@@ -79,6 +85,50 @@ async function runMigrations(): Promise<void> {
...
+function buildProvisioner(
+  env: NodeJS.ProcessEnv,
+  agentTokenService: AgentTokenService,
+): AgentProvisioner {
+  if (env.SHIPWRIGHT_K8S_PROVISIONING !== "enabled") {
+    return new NoopAgentProvisioner();
+  }
+
+  const namespace = env.SHIPWRIGHT_K8S_NAMESPACE ?? "default";
+  const image = env.SHIPWRIGHT_AGENT_IMAGE ?? "";
```
dmcaulay, line 109: `SHIPWRIGHT_AGENT_IMAGE` defaults to `""` here, producing `":latest"` as the container image spec when the var is unset — Kubernetes will reject it. The admin service starts cleanly but the first `POST /agents` with provisioning enabled fails at the K8s API call after a DB row has been created (then rolled back). Same risk for `SHIPWRIGHT_ADMIN_DEPLOYMENT_NAME` / `SHIPWRIGHT_ADMIN_DEPLOYMENT_UID` on the lines below.

Fail fast at startup instead:
```ts
if (!image) throw new Error("SHIPWRIGHT_AGENT_IMAGE is required when SHIPWRIGHT_K8S_PROVISIONING=enabled");
if (!apiUrl) throw new Error("SHIPWRIGHT_API_URL is required when SHIPWRIGHT_K8S_PROVISIONING=enabled");
```

docs/configuration.md
```diff
@@ -139,6 +139,20 @@ Provide either the GitHub App vars (recommended) or `GH_TOKEN` (PAT). App auth i
...
+| `SHIPWRIGHT_K8S_NAMESPACE` | `string` | `default` | Namespace the per-agent Secret + Deployment are created in. Only read when provisioning is enabled. |
+| `SHIPWRIGHT_AGENT_IMAGE` | `string` | — | Agent container image (without tag) used for the provisioned Deployment. Only read when provisioning is enabled. |
+| `SHIPWRIGHT_AGENT_IMAGE_TAG` | `string` | `latest` | Image tag joined as `image:tag` for the provisioned Deployment. Only read when provisioning is enabled. |
+| `SHIPWRIGHT_ADMIN_DEPLOYMENT_NAME` | `string` | — | Name of the admin Deployment, used as the `ownerReference` target so per-agent resources are garbage-collected with the admin Deployment. Only read when provisioning is enabled. |
+| `SHIPWRIGHT_ADMIN_DEPLOYMENT_UID` | `string` | — | UID of the admin Deployment, paired with `SHIPWRIGHT_ADMIN_DEPLOYMENT_NAME` for the `ownerReference`. Only read when provisioning is enabled. |
+| `SHIPWRIGHT_AGENT_REPLICAS` | `number` | `1` | Replica count for the provisioned agent Deployment. Only read when provisioning is enabled. |
```
dmcaulay, line 154: `buildProvisioner()` reads `SHIPWRIGHT_API_URL` (`main.ts:111`) and injects it into every provisioned agent Deployment as the agent's home URL. This var isn't listed in the 'Agent provisioning' table above — an operator enabling `SHIPWRIGHT_K8S_PROVISIONING=enabled` would not know to set it in the admin service, producing agents that can never bootstrap.

Add a row:
| `SHIPWRIGHT_API_URL` | `string` | — | In-cluster URL of the admin service, injected into each provisioned agent Deployment. Must be set when provisioning is enabled (the agent runtime also reads this var at bootstrap). |

dmcaulay (commented):
COMMENT — two issues in the newly introduced K8s provisioning path. Noop default and all existing behavior unaffected.

**Finding 1:** `buildProvisioner()` defaults `SHIPWRIGHT_AGENT_IMAGE` to `""` when the env var is absent, producing `":latest"` as the container image spec — invalid, K8s will reject it. Same pattern for `SHIPWRIGHT_ADMIN_DEPLOYMENT_NAME` / `SHIPWRIGHT_ADMIN_DEPLOYMENT_UID`. No startup guard: the service starts cleanly but the first `POST /agents` with K8s provisioning enabled fails at the K8s API call after the DB row has been created and then rolled back. Add fail-fast validation after the `SHIPWRIGHT_K8S_PROVISIONING=enabled` branch.

**Finding 2:** `buildProvisioner()` reads `SHIPWRIGHT_API_URL` (main.ts:111) to inject into each spawned agent Deployment, but the new 'Agent provisioning' section in configuration.md doesn't list it. Operators reading that section won't know to set it in the admin service — agents will bootstrap with an empty home URL and can never connect.

Strengths: `RecordingProvisioner` is a clean DI double (no `mock.module()`, no globals), test matrix is thorough (9 smoke cases covering the full auth × outcome space), and the Noop feature-flag design makes this safe to merge and deploy standalone.

PR #353, opened 2026-06-14
feat(charts): add Gateway API networking + cert-manager Certificate option (HD-4.1)

dodizzle:
## Summary
- Add a `networking.type=gateway` path (Gateway API): `templates/gateway.yaml` (`gateway.networking.k8s.io/v1` Gateway, configurable `gatewayClassName` default `gke-l7-global-external-managed`, HTTP :80 listener + HTTPS :443 when cert-manager enabled) and `templates/httproute.yaml` (admin `/` + metrics `/dashboard` HTTPRoutes via parentRefs). Mutually exclusive with Ingress.
- Add `templates/certificate.yaml` — a `cert-manager.io/v1` Certificate rendered only when `tls.certManager.enabled`, wired to the Gateway's HTTPS listener via the cert Secret; disabled → plain HTTP.
- `values.yaml`/`schema`: `networking.gateway.{gatewayClassName,host}`, `tls.certManager.{enabled,issuerRef}`, `"gateway"` in the type enum. Defaults Minikube-friendly (ClusterIP, certManager off).
- ci/gke-values demonstrates `gke-l7-global-external-managed`; the **full** `type=gateway`+certManager config lives in `examples/values-gke-gateway.yaml` (NOT a ct variant — keeps the kind e2e installable, since kind has no Gateway/cert-manager CRDs).
- Chart bumped **0.6.0 → 0.7.0**.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | type=gateway renders Gateway + HTTPRoutes (admin + /dashboard); no Ingress (and vice versa) | MET |
| 2 | tls.certManager.enabled=true renders a Certificate wired to gateway/route; disabled → none; gke values demonstrate gke-l7-global-external-managed | MET |
| 3 | helm-unit specs (Gateway/HTTPRoute, cert on/off, mutual exclusivity) + kubeconform validates the CRDs; none retired | MET |

## Test Plan
- `helm lint` 0 failed; `helm unittest` → 12 suites / 106 tests (24 new gateway specs + 5 schema)
- **kubeconform actually validated the CRDs** (not skipped): `Gateway is valid`, `HTTPRoute …admin/…metrics is valid`, `Certificate is valid` (datreeio CRDs-catalog `-schema-location` already resolves these groups — no workflow change needed); default render 17/17 valid
- Render/mutual-exclusivity: type=gateway → Gateway+2 routes(+Certificate when cert on), no Ingress; type=ingress → Ingress only; default ClusterIP → none; certManager toggles the Certificate
- **kind e2e**: all 3 ci/ variants `ct install` + `helm test` pass; ct never tries to create Gateway/Certificate (ci values keep type kind-installable)
- `ct lint --check-version-increment` PASS (0.7.0); check-strings + gitleaks clean; ci.yml/helm.yml/release.yml untouched
- [x] Pre-ship checks passing

### DIFF
--- a/charts/shipwright/templates/certificate.yaml
+++ b/charts/shipwright/templates/certificate.yaml
@@ -1,8 +1,11 @@
-{{- if .Values.tls.certManager.enabled }}
-# cert-manager Certificate, rendered ONLY when tls.certManager.enabled=true.
-# Disabled → no Certificate (plain HTTP). cert-manager writes the issued cert
-# into the Secret named here ("<fullname>-tls"), which the Gateway's HTTPS
-# listener references via certificateRefs (templates/gateway.yaml).
+{{- if and .Values.tls.certManager.enabled (eq .Values.networking.type "gateway") }}
+# cert-manager Certificate, rendered ONLY when tls.certManager.enabled=true AND
+# networking.type=gateway. The guard prevents accidental rendering when certManager
+# is enabled on a non-gateway setup (networking.type=ingress or ClusterIP), which
+# would produce a Certificate referencing networking.gateway.host (not the ingress
+# host) whose Secret is never mounted or used. Disabled → no Certificate (plain HTTP).
+# cert-manager writes the issued cert into the Secret named here ("<fullname>-tls"),
+# which the Gateway's HTTPS listener references via certificateRefs (templates/gateway.yaml).
 apiVersion: cert-manager.io/v1
 kind: Certificate
 metadata:
--- a/charts/shipwright/templates/certificate.yaml
+++ b/charts/shipwright/templates/certificate.yaml
@@ -1,0 +1,21 @@
+{{- if .Values.tls.certManager.enabled }}
+# cert-manager Certificate, rendered ONLY when tls.certManager.enabled=true.
+# Disabled → no Certificate (plain HTTP). cert-manager writes the issued cert
+# into the Secret named here ("<fullname>-tls"), which the Gateway's HTTPS
+# listener references via certificateRefs (templates/gateway.yaml).
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: {{ include "shipwright.fullname" . }}-tls
+  labels:
+    {{- include "shipwright.labels" . | nindent 4 }}
+spec:
+  # Secret cert-manager populates with the issued certificate + key.
+  secretName: {{ include "shipwright.fullname" . }}-tls
+  dnsNames:
+    - {{ .Values.networking.gateway.host | quote }}
+  issuerRef:
+    name: {{ .Values.tls.certManager.issuerRef.name | quote }}
+    kind: {{ .Values.tls.certManager.issuerRef.kind }}
+    group: cert-manager.io
+{{- end }}
--- a/charts/shipwright/templates/gateway.yaml
+++ b/charts/shipwright/templates/gateway.yaml
@@ -1,0 +1,41 @@
+{{- if eq .Values.networking.type "gateway" }}
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
+metadata:
+  name: {{ include "shipwright.fullname" . }}
+  labels:
+    {{- include "shipwright.labels" . | nindent 4 }}
+  {{- with .Values.networking.gateway.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  gatewayClassName: {{ .Values.networking.gateway.gatewayClassName | quote }}
+  listeners:
+    # Plain HTTP listener on :80, always present so the gateway answers for the
+    # configured host even without TLS (Minikube/dev = plain HTTP).
+    - name: http
+      protocol: HTTP
+      port: 80
+      hostname: {{ .Values.networking.gateway.host | quote }}
+      allowedRoutes:
+        namespaces:
+          from: Same
+    {{- if .Values.tls.certManager.enabled }}
+    # HTTPS listener on :443, rendered only when cert-manager issues a cert. The
+    # terminate-mode TLS references the Certificate's Secret (templates/certificate.yaml),
+    # whose name tracks the release fullname.
+    - name: https
+      protocol: HTTPS
+      port: 443
+      hostname: {{ .Values.networking.gateway.host | quote }}
+      tls:
+        mode: Terminate
+        certificateRefs:
+          - kind: Secret
+            name: {{ include "shipwright.fullname" . }}-tls
+      allowedRoutes:
+        namespaces:
+          from: Same
+    {{- end }}
+{{- end }}
--- a/charts/shipwright/tests/gateway_test.yaml
+++ b/charts/shipwright/tests/gateway_test.yaml
@@ -318,8 +318,8 @@ tests:
       tls.certManager.issuerRef.name: letsencrypt
       networking.type: ingress
     asserts:
-      - hasDocuments:
-          count: 0
+      - failedTemplate:
+          errorPattern: "/networking/type"
 
   - it: renders NO Certificate when certManager is enabled with ClusterIP networking
     template: templates/certificate.yaml
@@ -327,8 +327,8 @@ tests:
       tls.certManager.enabled: true
       tls.certManager.issuerRef.name: letsencrypt
     asserts:
-      - hasDocuments:
-          count: 0
+      - failedTemplate:
+          errorPattern: "/networking/type"
 
   # ---- Mutual exclusivity with Ingress ----
   - it: networking.type=gateway renders NO Ingress
--- a/charts/shipwright/tests/gateway_test.yaml
+++ b/charts/shipwright/tests/gateway_test.yaml
@@ -1,10 +1,11 @@
 suite: Gateway API networking (Gateway + HTTPRoute) + cert-manager Certificate
 # HD-4.1 adds templates/gateway.yaml + templates/httproute.yaml (rendered only
 # when networking.type=gateway) and templates/certificate.yaml (rendered only
-# when tls.certManager.enabled). These assert the conditional rendering, the
-# gatewayClassName/host/backends, the cert-manager wiring, and mutual exclusivity
-# with the Ingress (networking.type=ingress renders Ingress and NO Gateway/route;
-# networking.type=gateway renders Gateway+route and NO Ingress).
+# when tls.certManager.enabled AND networking.type=gateway). These assert the
+# conditional rendering, the gatewayClassName/host/backends, the cert-manager
+# wiring, and mutual exclusivity with the Ingress (networking.type=ingress renders
+# Ingress and NO Gateway/route; networking.type=gateway renders Gateway+route and
+# NO Ingress).
 tests:
   # ---- Gateway rendering ----
   - it: renders a gateway.networking.k8s.io/v1 Gateway when networking.type=gateway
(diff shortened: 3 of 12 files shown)


charts/shipwright/templates/certificate.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if .Values.tls.certManager.enabled }}
```
dodizzle: Missing `networking.type=gateway` guard. A user who sets `tls.certManager.enabled=true` with `networking.type=ingress` or `ClusterIP` will get a Certificate rendered here — but it references `networking.gateway.host` (not the ingress host), so the Certificate requests a cert for the wrong hostname and its Secret is never used. Change to:

```
{{- if and .Values.tls.certManager.enabled (eq .Values.networking.type "gateway") }}
```

This matches the guard pattern already used in `gateway.yaml`.

charts/shipwright/templates/gateway.yaml
```diff
@@ -0,0 +1,41 @@
...
+  {{- with .Values.networking.gateway.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  gatewayClassName: {{ .Values.networking.gateway.gatewayClassName | quote }}
+  listeners:
+    # Plain HTTP listener on :80, always present so the gateway answers for the
+    # configured host even without TLS (Minikube/dev = plain HTTP).
+    - name: http
```
dodizzle, line 17: When `tls.certManager.enabled=true`, both HTTP (:80) and HTTPS (:443) listeners are active, and the HTTPRoutes in `httproute.yaml` attach by Gateway name without a `sectionName` — so they're admitted on both ports simultaneously. There's no HTTP→HTTPS redirect. On GKE with `gke-l7-global-external-managed`, the standard expectation is that HTTP redirects to HTTPS when TLS is configured. Either add a `RequestRedirect` HTTPRoute on `sectionName: http`, or add a comment to `values.yaml` and the GKE example making the intentional plaintext-alongside-TLS behavior explicit.

dodizzle (commented):
COMMENT — Two issues to address before merge.

**1. Certificate template missing `networking.type=gateway` guard** (`certificate.yaml:1`): The template is gated only on `tls.certManager.enabled`, so enabling certManager on a non-gateway setup (`networking.type=ingress` or `ClusterIP`) silently renders a Certificate pointing at `networking.gateway.host` rather than the ingress host. The Certificate requests a cert for the wrong hostname and its Secret is never referenced by anything. Fix: `{{- if and .Values.tls.certManager.enabled (eq .Values.networking.type "gateway") }}`.

**2. HTTP listener serves alongside HTTPS with no redirect** (`gateway.yaml:17`): When cert-manager is enabled, the Gateway renders both HTTP (:80) and HTTPS (:443) listeners; the HTTPRoutes attach by Gateway name without `sectionName`, admitting plaintext traffic on both ports simultaneously. Add a `RequestRedirect` filter HTTPRoute on the HTTP listener, or document the intended plaintext-alongside-TLS behavior explicitly.

CI green (all 4 checks passed). Helm unit suite comprehensive — 24 new gateway specs covering rendering, mutual exclusivity, and cert toggling.

charts/shipwright/templates/certificate.yaml
dmcaulay: The guard here should also check `networking.type == "gateway"`. With the current guard, setting `tls.certManager.enabled=true` on a non-gateway install (e.g. `networking.type=ingress`) renders a Certificate for `networking.gateway.host` (default `shipwright.local`) — a different hostname than the Ingress host — while no Gateway HTTPS listener exists to reference the issued Secret. The cert is orphaned and covers the wrong hostname.

```yaml
{{- if and .Values.tls.certManager.enabled (eq .Values.networking.type "gateway") }}
```

charts/shipwright/values.schema.json
```diff
@@ -38,10 +38,58 @@
...
+          "properties": {
+            "gatewayClassName": { "type": "string" },
+            "host": { "type": "string" },
+            "annotations": { "type": "object" }
+          }
         }
       },
       "required": ["type"]
     },
+    "tls": {
```
dmcaulay, line 54: No cross-field constraint prevents `tls.certManager.enabled=true` with `networking.type != "gateway"`. The broken combination passes `helm lint` silently. The `auth.mode=google` pattern in this same schema shows how to add an `if/then` guard — worth applying here so operators get a validation error before a bad render reaches the cluster.

dmcaulay (commented):

1. `certificate.yaml` gates only on `tls.certManager.enabled`, not `networking.type == "gateway"`. A user enabling cert-manager with `networking.type=ingress` (or any non-gateway type) gets a Certificate rendered for `networking.gateway.host` (default `shipwright.local`) — a different hostname than `networking.ingress.host` — while no Gateway HTTPS listener exists to reference the issued Secret. Fix: `{{- if and .Values.tls.certManager.enabled (eq .Values.networking.type "gateway") }}`.

2. `values.schema.json` has no cross-field constraint tying `tls.certManager.enabled=true` to `networking.type=gateway` — the broken combination passes `helm lint` silently. The `auth.mode=google` pattern already shows how to do this.

charts/shipwright/values.schema.json
dodizzle, line 54: Schema missing cross-field constraint: `tls.certManager.enabled=true` with `networking.type=ingress` or `ClusterIP` silently passes `helm lint` — the template guard handles it correctly (renders nothing), but the operator gets no feedback that their config is a no-op. Suggest adding a top-level if/then on the root schema object (the `auth` block shows the exact pattern):

```json
{
  "if": {
    "properties": {
      "tls": {
        "properties": {
          "certManager": {
            "properties": { "enabled": { "const": true } },
            "required": ["enabled"]
          }
        }
      }
    }
  },
  "then": {
    "properties": {
      "networking": {
        "properties": { "type": { "const": "gateway" } }
      }
    }
  }
}
```

And a matching test in `values_schema_test.yaml`. Not blocking merge — template behaviour is correct — but this prevents a confusing operator experience.

dodizzle (commented):
COMMENT — Re-review of #353 (af35c43). Both prior blocking findings are addressed: `certificate.yaml` now correctly guards on `and .Values.tls.certManager.enabled (eq .Values.networking.type "gateway")`, and `httproute.yaml` adds a sectionName-pinned HTTP→HTTPS redirect when cert-manager is enabled, with backend routes bound to the HTTPS listener only. One schema quality gap remains: `values.schema.json` has no cross-field if/then constraint requiring `networking.type=gateway` when `tls.certManager.enabled=true`, so a misconfigured combination silently passes `helm lint` — the template renders correctly (nothing), but the operator gets no feedback that their configuration is a no-op. Suggest adding the constraint following the `auth.mode=google` pattern and a test in `values_schema_test.yaml`. Not blocking, but worth a follow-up.

charts/shipwright/tests/values_schema_test.yaml
```diff
@@ -63,3 +63,56 @@ tests:
...
+          errorPattern: "name"
+
+  - it: accepts certManager disabled with an empty issuerRef.name (constraint is enabled-only)
+    set:
+      tls.certManager.enabled: false
+      tls.certManager.issuerRef.name: ""
+    asserts:
+      - notFailedTemplate: {}
+
+  - it: rejects certManager enabled with networking.type=ingress (cross-field guard)
```
dodizzle, line 103: The cross-field guard is only tested against `networking.type=ingress` as the rejection case. The default type is `ClusterIP` — the most likely operator mistake (enabling certManager without switching to gateway). Add a case with no `networking.type` set (so it defaults to `ClusterIP`) that asserts `failedTemplate: errorPattern: "/networking/type"`. That ensures the constraint catches the most common misconfiguration, not just the ingress path.

charts/shipwright/tests/values_schema_test.yaml
```diff
@@ -63,3 +63,56 @@ tests:
...
+    set:
+      networking.type: ingress
+      tls.certManager.enabled: true
+      tls.certManager.issuerRef.name: letsencrypt
+    asserts:
+      - failedTemplate:
+          errorPattern: "networking.type"
```
dodizzle: `"networking.type"` uses `.` as a regex wildcard — it incidentally matches `"/networking/type"` in the error output, but for the wrong reason. Line 41 in this same file uses `"/networking/type"` (JSON Pointer notation). Change to `errorPattern: "/networking/type"` for consistency and to avoid a latent false-pass if the error format ever changes.

dodizzle (commented):
COMMENT — Re-review of #353 (bb5f4a3). The schema constraint is correctly implemented: the root-level `if/then` in `values.schema.json` now rejects `tls.certManager.enabled=true` when `networking.type` is not `"gateway"`, following the same pattern as the `auth.mode=google` constraint. Two test quality gaps remain.

**1. Rejection test missing for default `networking.type=ClusterIP` with certManager enabled** (`values_schema_test.yaml:103`): Only `ingress` is tested as the rejected case. The default `ClusterIP` is the most realistic operator mistake — enabling certManager without switching to gateway. Add a test with no `networking.type` set (defaulting to `ClusterIP`) + `certManager.enabled: true`, asserting `failedTemplate: errorPattern: "/networking/type"`.

**2. `errorPattern` uses regex dot instead of JSON Pointer slash** (`values_schema_test.yaml:110`): `errorPattern: "networking.type"` uses `.` as a regex wildcard matching the `/` in the actual error path `"/networking/type"`. Works accidentally, but is inconsistent with line 41 which uses `"/networking/type"` correctly. Change to `errorPattern: "/networking/type"`.

Neither is blocking, but the ClusterIP test in particular ensures the constraint catches the most common misconfiguration.

dmcaulay (approved):
APPROVE — Re-review of #353 (bb5f4a3). All prior blocking findings resolved: `certificate.yaml` now correctly guards on `and .Values.tls.certManager.enabled (eq .Values.networking.type "gateway")`, the HTTP→HTTPS redirect route is pinned via `sectionName: http` with backend routes bound to `sectionName: https`, and the cross-field schema constraint requiring `networking.type=gateway` when `tls.certManager.enabled=true` is in place and test-covered. No new issues. Clean to merge.

dmcaulay (approved):
APPROVE — Re-review of #353 (0ff53e37). Test fix is correct and complete: `failedTemplate` is the right assertion for schema-rejected renders (schema errors abort template rendering, they don't produce 0 documents), the `errorPattern: "/networking/type"` matches the JSON Pointer path the go-jsonschema evaluator emits, and the new ClusterIP default case fills the one remaining gap (certManager enabled with no `networking.type` override). All prior blocking findings remain resolved. Clear to merge.

charts/shipwright/tests/gateway_test.yaml
```diff
@@ -0,0 +1,356 @@
...
+          value: Issuer
+
+  - it: renders NO Certificate when certManager is disabled (default — plain HTTP)
+    template: templates/certificate.yaml
+    asserts:
+      - hasDocuments:
+          count: 0
+
+  - it: renders NO Certificate when certManager is enabled but networking.type is not gateway
+    template: templates/certificate.yaml
```
dodizzle, line 315: `template: templates/certificate.yaml` here is incompatible with `failedTemplate` for schema-level errors. Helm-unittest only surfaces chart-wide JSON Schema validation errors when no `template:` is specified — with a specific template file, the schema rejection happens before rendering and the `failedTemplate` assertion never fires, causing this test to error instead of pass. The working counterparts in `values_schema_test.yaml` have no `template:` field. Fix: remove this `template:` line (and the same on line 325), or delete both test cases since `values_schema_test.yaml` now covers the same schema rejections.

dodizzle (commented):
COMMENT — Re-review of #353 (ef04c36). Both prior findings are addressed: ClusterIP rejection test added to `values_schema_test.yaml`, `errorPattern` corrected to JSON Pointer notation. One new issue: the two `gateway_test.yaml` test cases were changed from `hasDocuments: count: 0` to `failedTemplate` but still specify `template: templates/certificate.yaml` — helm-unittest can't catch chart-wide schema errors in a template-scoped test, so both are failing in CI. Fix: remove the `template:` line from those two cases, or delete them (covered by `values_schema_test.yaml`).

PR #354, opened 2026-06-14
feat(charts): agent-provisioning RBAC + admin env contract (HD-4.2)

dodizzle:
## Summary
- Add agent-provisioning RBAC + the admin env contract, all gated on `agent.provisioning.enabled` (default false → renders nothing, admin stays Noop):
  - `templates/agent-provisioning-rbac.yaml`: a **namespace-scoped `Role`** (NOT ClusterRole) granting **create/get/list/delete** on `apps`/Deployments and core/Secrets — least-privilege, exactly the provisioner's verbs — plus a `RoleBinding` to the admin ServiceAccount.
  - `templates/agent-serviceaccount.yaml`: the agent ServiceAccount that provisioned pods run as.
  - `admin-deployment.yaml`: injects the provisioner env contract **matching `admin/src/main.ts buildProvisioner` exactly** when enabled — `SHIPWRIGHT_K8S_PROVISIONING=enabled`, `SHIPWRIGHT_K8S_NAMESPACE` via the **downward API** (`fieldRef: metadata.namespace`), agent image/tag, replicas, in-cluster `SHIPWRIGHT_API_URL`, admin Deployment name. (`SHIPWRIGHT_ADMIN_DEPLOYMENT_UID` is omitted unless overridden — the parent Deployment's UID isn't downward-API-exposable; ownerRef propagation is a tracked follow-up.)
- `agent.provisioning.{enabled,image,replicas,pvc,serviceAccount}` values + schema. Chart bumped **0.6.0 → 0.8.0** (0.7.0 is HD-4.1/#353, landing concurrently).

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | enabled → Role(deployments+secrets verbs)+RoleBinding+agent SA + provisioner env (incl. namespace downward API) into admin Deployment; disabled → none, admin Noop | MET |
| 2 | RBAC namespace-scoped (Role) + least-privilege for exactly the provisioner's verbs | MET |
| 3 | helm-unit specs (RBAC rules + admin env enabled/disabled) + kubeconform; none retired | MET |

## Test Plan
- `helm lint` 0 failed; `helm unittest` → 12 suites / 96 tests (11 new RBAC specs)
- `kubeconform` → enabled render 20/20 valid (Role/RoleBinding/SA); disabled 17/17
- Render: enabled → RBAC+SA + `SHIPWRIGHT_K8S_PROVISIONING` + namespace `fieldRef`; disabled → none, no provisioning env
- **kind e2e**: all 3 ci variants `ct install` + `helm test` pass — **including minikube with provisioning RBAC ENABLED** (Role/RoleBinding applied on the live cluster)
- `ct lint --check-version-increment` PASS (0.8.0 > 0.6.0); check-strings + gitleaks clean; workflows untouched
- [x] Pre-ship checks passing

### DIFF
--- a/charts/shipwright/templates/agent-provisioning-rbac.yaml
+++ b/charts/shipwright/templates/agent-provisioning-rbac.yaml
@@ -1,0 +1,41 @@
+{{- if .Values.agent.provisioning.enabled }}
+# Agent-provisioning RBAC (HD-4.2). Gated on agent.provisioning.enabled (default
+# false → this file renders nothing and the admin service stays in Noop mode).
+#
+# Namespace-scoped Role (NOT a ClusterRole): the admin service provisions agent
+# Deployments + Secrets only within its own release namespace, so a Role bound by
+# a RoleBinding is the least-privilege fit — no cluster-wide grant. The verb set
+# is exactly what the provisioner (HD-1.3 KubernetesAgentProvisioner / HD-1.4
+# wiring) exercises: create + get + list + delete on Deployments (apps) and
+# Secrets (core).
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "shipwright.admin.fullname" . }}-agent-provisioner
+  labels:
+    {{- include "shipwright.admin.labels" . | nindent 4 }}
+rules:
+  - apiGroups: ["apps"]
+    resources: ["deployments"]
+    verbs: ["create", "get", "list", "delete"]
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["create", "get", "list", "delete"]
+---
+# RoleBinding granting the admin ServiceAccount the Role above, scoped to the
+# release namespace.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ include "shipwright.admin.fullname" . }}-agent-provisioner
+  labels:
+    {{- include "shipwright.admin.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ include "shipwright.admin.fullname" . }}-agent-provisioner
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "shipwright.admin.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }}
+{{- end }}
--- a/charts/shipwright/templates/agent-provisioning-rbac.yaml
+++ b/charts/shipwright/templates/agent-provisioning-rbac.yaml
@@ -6,7 +6,7 @@
 # Deployments + Secrets only within its own release namespace, so a Role bound by
 # a RoleBinding is the least-privilege fit — no cluster-wide grant. The verb set
 # is exactly what the provisioner (HD-1.3 KubernetesAgentProvisioner / HD-1.4
-# wiring) exercises: create + get + list + delete on Deployments (apps) and
+# wiring) exercises: create + get + delete on Deployments (apps) and
 # Secrets (core).
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
@@ -17,10 +17,10 @@ metadata:
 rules:
   - apiGroups: ["apps"]
     resources: ["deployments"]
-    verbs: ["create", "get", "list", "delete"]
+    verbs: ["create", "get", "delete"]
   - apiGroups: [""]
     resources: ["secrets"]
-    verbs: ["create", "get", "list", "delete"]
+    verbs: ["create", "get", "delete"]
 ---
 # RoleBinding granting the admin ServiceAccount the Role above, scoped to the
 # release namespace.
--- a/charts/shipwright/tests/agent_provisioning_rbac_test.yaml
+++ b/charts/shipwright/tests/agent_provisioning_rbac_test.yaml
@@ -8,7 +8,7 @@ suite: agent-provisioning RBAC + admin env contract (HD-4.2)
 #       carries NO provisioning env (admin stays in Noop mode)
 tests:
   # ── (a) enabled → RBAC + agent SA ────────────────────────────────────────────
-  - it: renders a namespace-scoped Role with exactly create/get/list/delete on Deployments and Secrets
+  - it: renders a namespace-scoped Role with exactly create/get/delete on Deployments and Secrets
     template: templates/agent-provisioning-rbac.yaml
     documentIndex: 0
     set:
@@ -29,7 +29,7 @@ tests:
           value: ["deployments"]
       - equal:
           path: rules[0].verbs
-          value: ["create", "get", "list", "delete"]
+          value: ["create", "get", "delete"]
       - equal:
           path: rules[1].apiGroups
           value: [""]
@@ -38,7 +38,7 @@ tests:
           value: ["secrets"]
       - equal:
           path: rules[1].verbs
-          value: ["create", "get", "list", "delete"]
+          value: ["create", "get", "delete"]
 
   - it: binds the Role to the admin ServiceAccount via a namespace-scoped RoleBinding
     template: templates/agent-provisioning-rbac.yaml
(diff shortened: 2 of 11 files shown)


charts/shipwright/templates/agent-provisioning-rbac.yaml
```diff
@@ -0,0 +1,41 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "shipwright.admin.fullname" . }}-agent-provisioner
+  labels:
+    {{- include "shipwright.admin.labels" . | nindent 4 }}
+rules:
+  - apiGroups: ["apps"]
+    resources: ["deployments"]
+    verbs: ["create", "get", "list", "delete"]
```
dodizzle: **Important (conf 82, least-privilege/security): drop the `list` verb.** KubernetesClient (kubernetes-client.ts) exposes only create/get/delete on Deployments and Secrets — there is no `list` method anywhere, so this grant is unused today. AC2 requires least-privilege for *exactly* the verbs the provisioner uses, and the PR/CHANGELOG claim exactly that. `list` on **Secrets** is a real over-grant: a compromised admin pod could enumerate every Secret in the namespace, not just the ones it created. Change both rules to `verbs: ["create", "get", "delete"]` (and fix the comment + CHANGELOG). Add `list` in the future PR that actually needs it. (The HD-4.2 task description listed `list`, but AC2's least-privilege rule + the actual code take precedence.)

dodizzle (commented):
**Verdict: COMMENT** (self-review). Env contract matches main.ts buildProvisioner exactly (verified), RoleBinding correctly subjects the ADMIN SA, gating is atomic, schema defaults off. Two items before merge. **Finding 2 (test, conf 78):** the disabled-absence test only notContains 2 of 7 provisioning env vars (SHIPWRIGHT_K8S_PROVISIONING + namespace) — AC3 wants the negative coverage complete; add notContains for the other 5 (SHIPWRIGHT_AGENT_IMAGE, _IMAGE_TAG, _REPLICAS, SHIPWRIGHT_API_URL, SHIPWRIGHT_ADMIN_DEPLOYMENT_NAME). See inline for finding 1.

charts/shipwright/values.yaml
```diff
@@ -139,6 +139,44 @@ agent:
...
+    # -- Optional admin Deployment UID for ownerRef propagation. The downward API
+    # does NOT expose the parent Deployment's UID to its pods, so it can only be
+    # supplied explicitly. Leave empty to OMIT the env var (acceptable until
+    # ownerRef propagation lands as a follow-up).
+    adminDeploymentUid: ""
+    # -- PVC settings for provisioned agent workspaces, surfaced for the agent
+    # manifest builder. NOTE: buildProvisioner does not read these as env today,
+    # so they are NOT injected into the admin Deployment — they exist for the
+    # manifest builder's future use.
+    pvc:
```
dodizzle, line 176: `pvc` (new, lines 176–179, not injected yet) and `persistence` (lines 181–188, pre-existing agent home PVC) both live under `agent.provisioning` with overlapping field shapes (`size`, `storageClass`). These describe different things but a user enabling provisioning will see both with no guidance on which governs what. Either add a clear disambiguating comment above each (`# For provisioned agent workspace PVCs` vs `# For the agent container's own home directory PVC`), or move this new block under a distinct key like `agent.provisioning.workspace` so the two storage surfaces don't look interchangeable.

charts/shipwright/tests/agent_provisioning_rbac_test.yaml
```diff
@@ -0,0 +1,192 @@
...
+          content:
+            name: SHIPWRIGHT_AGENT_REPLICAS
+            value: "1"
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: SHIPWRIGHT_API_URL
+            value: http://r-shipwright-admin:3001
+
+  - it: omits SHIPWRIGHT_ADMIN_DEPLOYMENT_UID unless an override is provided, and injects it when set
```
dodizzle: This test only asserts `SHIPWRIGHT_ADMIN_DEPLOYMENT_UID` IS present when `adminDeploymentUid` is set. Add a companion `notContains` case: `agent.provisioning.enabled: true` with no `adminDeploymentUid` override, asserting the var is absent. The `{{- with }}` guard in `admin-deployment.yaml` needs a negative test to protect against accidental removal.

dodizzle (commented):
COMMENT — Two issues before merge.

**1. Duplicate `pvc` and `persistence` sub-sections under `agent.provisioning`** (`values.yaml:176`): The block now has both `pvc` (new, lines 176–179, not injected yet) and `persistence` (pre-existing, lines 181–188, the agent container's home directory PVC). Both expose `size` and `storageClass` with no structural distinction between provisioned-workspace PVCs and the agent's own home PVC. A user enabling provisioning will see both with no guidance on which governs what. Distinguish them with clear comments or rename the new `pvc` block to something like `workspace` to make the separation explicit.

**2. Test missing negative assertion for `SHIPWRIGHT_ADMIN_DEPLOYMENT_UID` omit path** (`agent_provisioning_rbac_test.yaml:149`): The test only asserts the var IS present when `adminDeploymentUid` is set. No test confirms it's absent when unset — if the `{{- with }}` guard in `admin-deployment.yaml` were removed, no test would catch it. Add a `notContains` assertion for the default (unset) case.

CI green (all 4 checks). RBAC verb set correctly matches KubernetesClient operations — no missing verbs, no surplus grants.

charts/shipwright/tests/agent_provisioning_rbac_test.yaml
dmcaulay: This test title promises "omits SHIPWRIGHT_ADMIN_DEPLOYMENT_UID unless an override is provided" but the asserts block only covers the positive (UID-injected) case. The `{{- with .Values.agent.provisioning.adminDeploymentUid }}` guard in the template — the boundary that prevents accidentally leaking or fabricating a UID — has no negative test.

Add a second test (or a second asserts block without `adminDeploymentUid` set) that asserts:
```yaml
- notContains:
    path: spec.template.spec.containers[0].env
    content:
      name: SHIPWRIGHT_ADMIN_DEPLOYMENT_UID
```

charts/shipwright/templates/agent-provisioning-rbac.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["create", "get", "list", "delete"]
```
dmcaulay: The provisioner's stated semantics are create-and-delete only — it creates a Secret when provisioning an agent and deletes it by name when deprovisioning. `list` is not needed for that flow; `get` is sufficient for named lookups. `list` on Secrets grants enumeration of every Secret in the release namespace, including session secrets, encryption keys, and the PostgreSQL password created by the Bitnami subchart — it exceeds the least-privilege claim in the acceptance criteria.

Suggestion: `verbs: ["create", "get", "delete"]`

dmcaulay (commented):

1. **Missing negative UID-omission assertion** (`agent_provisioning_rbac_test.yaml:149`): The test title promises "omits SHIPWRIGHT_ADMIN_DEPLOYMENT_UID unless an override is provided" but only has a `contains` assert for the positive case. The `{{- with }}` guard in the template is left untested for the default-empty path — if it were removed, no test would catch it. Add a `notContains` assert for `SHIPWRIGHT_ADMIN_DEPLOYMENT_UID` when `adminDeploymentUid` is unset.

2. **`list` on Secrets exceeds least-privilege** (`agent-provisioning-rbac.yaml:23`): The provisioner's stated semantics are create-and-delete: create a secret when provisioning an agent, delete it when deprovisioning, look it up by name (`get`). `list` is not needed for this flow and allows the admin SA to enumerate every Secret in the namespace — including session secrets, encryption keys, and the PostgreSQL password. Remove `list` from the Secrets rule: `verbs: ["create", "get", "delete"]`.

dodizzle (commented):
APPROVE — converged and clean. RBAC is least-privilege (Role create/get/delete on deployments+secrets — the unused `list` verb was removed; namespace-scoped Role, RoleBinding subjects the admin SA), the admin env contract matches main.ts buildProvisioner exactly with namespace via downward API, and gating is atomic (disabled → no RBAC/SA/env, admin Noop). My findings (list verb + disabled-env test) and the follow-up review findings (pvc/persistence comment disambiguation + UID negative test) are all addressed; 0 unresolved threads; CI + Helm 2-pod e2e (RBAC enabled in a ci variant) + Chart Release + pr-title-lint all green. Chart 0.8.0. Ready to merge.

PR #357, opened 2026-06-14
feat(plan-session): add complexity scoring and model tier assignment

dodizzle:
## Summary

- Adds **Complexity** (1–5) and **Model** (haiku/sonnet/opus) fields to the Step 5 task breakdown field list in `plan-session.md`
- Adds a **heuristics table** mapping complexity score to signal patterns and model tier, plus tie-breaking rules and bundle inheritance rule
- Updates both **Path A and Path B** Step 6 JSON templates to include `"complexity"` and `"model"` fields in generated task JSON

## Complexity Scoring Heuristic

| Score | Signal | Model |
|-------|--------|-------|
| 1 | Single file, config/copy change, no logic | `haiku` |
| 2 | 1–2 files, straightforward logic, unit tests only | `haiku` |
| 3 | 2–5 files, standard feature, integration tests | `sonnet` |
| 4 | 5+ files, cross-layer, new patterns | `sonnet` |
| 5 | Architectural, new abstraction, migration, perf-sensitive | `opus` |

## Test plan

- [ ] 553 existing tests pass — no new tests needed (pure markdown edit)
- [ ] Biome format issues in `create-task-store.unit.test.ts` and `check-deploy.ts` are pre-existing on `main` (not introduced by this PR)
- [ ] `plan-session` invocations now emit `complexity` and `model` in task JSON, enabling downstream routing in `dev-task` (DM-1.3)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/plugins/shipwright/commands/plan-session.md
+++ b/plugins/shipwright/commands/plan-session.md
@@ -187,7 +187,7 @@ Assign a complexity score (1–5) and model tier to every task:
 **Tie-breaking rules:**
 - New abstraction required (interface, base class, shared module) → bump up one tier
 - Pure modification of existing code (no new patterns) → stay at current tier
-- When uncertain, prefer the lower tier — the executor escalates automatically on BLOCKED
+- When uncertain, prefer the lower tier — the planner can escalate in a follow-up plan revision if execution is blocked
 
 **Bundle inheritance:** When tasks share a branch, all tasks in the bundle inherit the highest model tier among them. A haiku-tier task bundled with a sonnet-tier task runs at sonnet.
 
--- a/plugins/shipwright/commands/plan-session.md
+++ b/plugins/shipwright/commands/plan-session.md
@@ -169,6 +169,27 @@ For each task:
 - **Branch**: `feat/{id-lowered-dashes}-{first-3-words-kebab}` — or a shared branch name for bundled tasks (see below)
 - **Layer**: API | Frontend | Database | Shared | Background | CLI
 - **Hours**: rough estimate (1-8h; break tasks larger than 8h)
+- **Complexity**: integer 1–5 — use the scoring table below
+- **Model**: `haiku` | `sonnet` | `opus` — derived from complexity score (see table)
+
+### Complexity and Model Scoring
+
+Assign a complexity score (1–5) and model tier to every task:
+
+| Score | Signal | Model |
+|-------|--------|-------|
+| 1 | Single file, config/copy change, no logic, unit tests only | `haiku` |
+| 2 | 1–2 files, straightforward logic, unit tests only | `haiku` |
+| 3 | 2–5 files, standard feature, integration tests | `sonnet` |
+| 4 | 5+ files, cross-layer, new patterns, integration + smoke tests | `sonnet` |
+| 5 | Architectural, cross-layer, new abstraction, migration, or perf-sensitive | `opus` |
+
+**Tie-breaking rules:**
+- New abstraction required (interface, base class, shared module) → bump up one tier
+- Pure modification of existing code (no new patterns) → stay at current tier
+- When uncertain, prefer the lower tier — the executor escalates automatically on BLOCKED
+
+**Bundle inheritance:** When tasks share a branch, all tasks in the bundle inherit the highest model tier among them. A haiku-tier task bundled with a sonnet-tier task runs at sonnet.
 
 ### Bundles
 
@@ -265,7 +286,9 @@ Write the new tasks to a temp file `/tmp/new-tasks-{session}.json` as a JSON arr
     "status": "pending",
     "pr": null,
     "addedAt": "{ISO timestamp}",
-    "hours": 2
+    "hours": 2,
+    "complexity": {complexity},
+    "model": "{model}"
   }
 ]
 ```
@@ -374,7 +397,9 @@ Write the tasks to `/tmp/new-tasks-{session}.json`. Set `source` to `"gh:{owner}
     "status": "pending",
     "pr": null,
     "addedAt": "{ISO timestamp}",
-    "hours": 2
+    "hours": 2,
+    "complexity": {complexity},
+    "model": "{model}"
   }
 ]
 ```
--- a/plugins/shipwright/commands/plan-session.md
+++ b/plugins/shipwright/commands/plan-session.md
@@ -242,6 +242,16 @@ Present the task list and dependency map as a first pass. The engineer reviews a
 
 Once the task breakdown is approved, write each task to the task store via `task_store.ts`.
 
+### Bundle Model Inheritance (Pre-Write)
+
+Before constructing any JSON, apply bundle inheritance to the full task list:
+
+1. Group planned tasks by `branch`.
+2. For each group with more than one task (a bundle), find the highest model tier present: `opus` > `sonnet` > `haiku`.
+3. Set every task in that bundle to the highest tier before writing.
+
+A task on its own branch is unaffected. This ensures a `haiku`-scored task bundled with a `sonnet`-scored task is written as `model: "sonnet"`.
+
 The code path depends on `taskStore` in `SHIPWRIGHT_CONFIG`. Read it the same way Phase-0 does:
 
 ```bash


dodizzle (commented):
## Review findings

### Bug: Bundle inheritance rule is stated but not enforced in Step 6 write path

**File:** `plugins/shipwright/commands/plan-session.md` (around the Step 6 task JSON templates)

**Severity:** Important

The "Bundle inheritance" rule says all tasks sharing a branch should inherit the highest model tier among them. But Step 6 writes each task individually from a flat JSON array — there is no recalculation step before the `append` call. A complexity-1 (`haiku`) task bundled with a complexity-3 (`sonnet`) task would be written as `model='haiku'`, violating the stated rule.

**Suggested fix:** Add an explicit pre-write step in Step 6:

> Before writing tasks, group them by `branch`. For each bundle group, set `model` to the highest tier among all tasks in that group (opus > sonnet > haiku) before constructing the JSON.

---

### Test readiness: No test for plan-session.md complexity/model schema content

**Severity:** Important (mitigated)

No test asserts the new `complexity` and `model` fields exist in the Step 6 JSON templates. A future refactor could silently drop them.

**Mitigated by:** The task acceptance criteria explicitly call this out as a known trade-off ("no automated test — markdown prompt file; observable via a manual plan-session run"). Acknowledged and accepted.

dodizzle:
**Patch applied** ().

Added a **Bundle Model Inheritance (Pre-Write)** section at the top of Step 6, before the branch gate. Groups tasks by `branch`, finds the highest tier among bundle-mates, and elevates all tasks in the group before any JSON is written. Applies uniformly to both Path A (JSON) and Path B (GitHub Issues).

Review finding resolved. Independent review of the fix: APPROVE — correctly addresses the prior finding, no new issues introduced.

plugins/shipwright/commands/plan-session.md
```diff
@@ -169,6 +169,27 @@ For each task:
...
+| 1 | Single file, config/copy change, no logic, unit tests only | `haiku` |
+| 2 | 1–2 files, straightforward logic, unit tests only | `haiku` |
+| 3 | 2–5 files, standard feature, integration tests | `sonnet` |
+| 4 | 5+ files, cross-layer, new patterns, integration + smoke tests | `sonnet` |
+| 5 | Architectural, cross-layer, new abstraction, migration, or perf-sensitive | `opus` |
+
+**Tie-breaking rules:**
+- New abstraction required (interface, base class, shared module) → bump up one tier
+- Pure modification of existing code (no new patterns) → stay at current tier
+- When uncertain, prefer the lower tier — the executor escalates automatically on BLOCKED
```
dmcaulay: This tie-breaking rule claims the executor "escalates automatically on BLOCKED" to a higher model tier, but `dev-task.md`'s BLOCKED path instructs the executor to escalate to the user — it does not auto-switch model tiers. Until DM-1.3 implements model routing, this overstates the system's current capability and could lead planners to under-assign tiers.

Suggested fix: `When uncertain, prefer the lower tier — the planner can escalate in a follow-up plan revision if execution is blocked.`

dmcaulay (commented):
COMMENT — one documentation accuracy issue: the "escalates automatically on BLOCKED" tie-breaking rule references model auto-escalation that `dev-task.md` does not currently implement (its BLOCKED path escalates to the user, not to a higher model tier). Until DM-1.3 lands, planners may under-assign tiers relying on a safety net that doesn't exist. Otherwise the complexity heuristics table, bundle inheritance logic, and pre-write enforcement are well-structured. The TypeScript schema already handles these fields cleanly.

dmcaulay (approved):
APPROVE — all prior findings resolved. The tie-breaking rule now correctly states the planner (not the system) escalates on BLOCKED, and bundle model inheritance is enforced pre-write with a clear algorithm. No new issues. Clean documentation that closes the plan → execute data contract gap.

plugins/shipwright/commands/plan-session.md
```diff
@@ -169,6 +169,27 @@ For each task:
...
+
+### Complexity and Model Scoring
+
+Assign a complexity score (1–5) and model tier to every task:
+
+| Score | Signal | Model |
+|-------|--------|-------|
+| 1 | Single file, config/copy change, no logic, unit tests only | `haiku` |
+| 2 | 1–2 files, straightforward logic, unit tests only | `haiku` |
+| 3 | 2–5 files, standard feature, integration tests | `sonnet` |
```
dodizzle, line 183: Boundary overlap: Score 3 here is "2–5 files" and Score 4 below is "5+ files" — at exactly 5 files both rows match. Suggest changing this to "2–4 files" so the file-count boundary is exclusive.

plugins/shipwright/commands/plan-session.md
```diff
@@ -169,6 +169,27 @@ For each task:
...
+| 4 | 5+ files, cross-layer, new patterns, integration + smoke tests | `sonnet` |
+| 5 | Architectural, cross-layer, new abstraction, migration, or perf-sensitive | `opus` |
+
+**Tie-breaking rules:**
+- New abstraction required (interface, base class, shared module) → bump up one tier
```
dodizzle, line 188: This bump rule has no ceiling behavior: a task already scored 5 / `opus` with a new abstraction has nowhere to bump. Add an explicit "if already at opus, stay" clause so the planner can't emit an out-of-range tier.

plugins/shipwright/commands/plan-session.md
```diff
@@ -169,6 +169,27 @@ For each task:
...
+- Pure modification of existing code (no new patterns) → stay at current tier
+- When uncertain, prefer the lower tier — the planner can escalate in a follow-up plan revision if execution is blocked
+
+**Bundle inheritance:** When tasks share a branch, all tasks in the bundle inherit the highest model tier among them. A haiku-tier task bundled with a sonnet-tier task runs at sonnet.
```
dodizzle, line 192: Bundle inheritance is also defined as a numbered procedure in Step 6 ("Bundle Model Inheritance (Pre-Write)"). To avoid a planner applying it here and skipping the Step 6 re-check, consider replacing this paragraph with a pointer to Step 6 as the authoritative definition.

dodizzle (commented):
Verdict: COMMENT (self-review — recorded as COMMENT since GitHub rejects self-APPROVE).

The complexity/model scoring heuristic is a clean, well-scoped addition and reuses the established haiku/sonnet/opus tiers. Three small text-level ambiguities would cause inconsistent planner output in edge cases and are worth tightening before this becomes the routing source of truth:

1. **File-count boundary overlap (Score 3 vs 4)** — Score 3 says "2–5 files" and Score 4 says "5+ files". At exactly 5 files both rows match. Harmless today (both map to sonnet) but sets a non-deterministic precedent. Suggest Score 3 → "2–4 files".
2. **Tie-break bump has no ceiling** — "New abstraction required → bump up one tier" has no defined behavior when the task is already Score 5 / opus (opus is the ceiling). Add an explicit "already at opus → stay" clause.
3. **Bundle inheritance defined twice** — once as prose in the scoring section and again as the Step 6 "Bundle Model Inheritance (Pre-Write)" procedure. Mark the first mention as a pointer to Step 6 (the authoritative location) so a planner doesn't apply it once and skip the re-check, or vice versa.

None are blocking correctness bugs; all are small wording fixes.

plugins/shipwright/commands/plan-session.md
dodizzle, line 183: Score 3 says "2–5 files" and Score 4 says "5+ files" — both include exactly 5 files. Both map to `sonnet` so there's no routing impact, but it's a non-deterministic boundary. Suggest changing Score 3 to "2–4 files".

dodizzle (commented):
Verdict: APPROVE (self-review — recorded as COMMENT since GitHub rejects self-APPROVE).

Clean, well-scoped documentation addition. The pre-write bundle inheritance algorithm is unambiguous, the tie-breaking rule correctly scopes BLOCKED escalation to the planner (not the system), and both JSON templates are updated consistently. dmcaulay approved the head commit after all blocking findings were resolved.

One minor suggestion: Score 3 says "2–5 files" and Score 4 says "5+ files" — both include exactly 5 files. Harmless since both rows map to `sonnet`, but the overlap creates a non-deterministic precedent. Suggest changing Score 3 to "2–4 files" to make the boundary exclusive.

PR #364, opened 2026-06-14
feat(charts): add PostHog provider + admin env surface to metrics deployment (SWD-1.3)

dodizzle:
## Summary

- Adds `metrics.provider` block in `values.yaml` with 5 configurable env surfaces: PostHog credentials (via `existingSecret` refs), `METRICS_ADMIN_URL`, `METRICS_BASE_PATH`, `METRICS_REQUIRE_OWNER_ROLE`, and the metrics↔admin internal API key
- Updates `metrics-deployment.yaml` to conditionally inject all 5 env vars when configured; existing `SHIPWRIGHT_SESSION_SECRET` and bundled-PG `METRICS_DATABASE_URL` paths are untouched
- Adds 10 new `helm-unittest` specs asserting each env renders in PostHog mode and is absent in the default path; all 13 previously-passing test suites remain green

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | `values.yaml` `metrics.provider` block with PostHog keys (secret refs), adminUrl, basePath, requireOwnerRole, internalKey; defaults preserve SQLite fallback | ✅ MET |
| 2 | `metrics-deployment.yaml` injects env when configured; `SHIPWRIGHT_SESSION_SECRET` + `METRICS_DATABASE_URL` paths unchanged | ✅ MET |
| 3 | `helm-unittest` specs: PostHog-mode renders when configured, absent in default; no existing tests retired (additive) | ✅ MET |
| 4 | Safe to deploy standalone (additive, gated) | ✅ MET |

## Test Plan

- [x] `helm unittest charts/shipwright -f 'tests/metrics_workload_test.yaml'` — 23 tests pass
- [x] `helm lint charts/shipwright` — 0 failures
- [x] New tests cover each env in both on (configured) and off (default) paths
- [x] Pre-existing suite failures (gateway, postgresql, values_schema) confirmed pre-existing on main

### DIFF
--- a/charts/shipwright/tests/values_schema_test.yaml
+++ b/charts/shipwright/tests/values_schema_test.yaml
@@ -124,3 +124,54 @@ tests:
       tls.certManager.issuerRef.name: letsencrypt
     asserts:
       - notFailedTemplate: {}
+
+  - it: rejects posthog with personalApiKeyRef set but existingSecret empty (conditional minLength)
+    set:
+      metrics.provider.posthog.personalApiKeyRef: my-key-ref
+      metrics.provider.posthog.existingSecret: ""
+    asserts:
+      - failedTemplate:
+          errorPattern: "existingSecret"
+
+  - it: rejects posthog with projectIdRef set but existingSecret empty (conditional minLength)
+    set:
+      metrics.provider.posthog.projectIdRef: my-project-ref
+      metrics.provider.posthog.existingSecret: ""
+    asserts:
+      - failedTemplate:
+          errorPattern: "existingSecret"
+
+  - it: accepts posthog with both ref fields and a non-empty existingSecret
+    set:
+      metrics.provider.posthog.personalApiKeyRef: my-key-ref
+      metrics.provider.posthog.projectIdRef: my-project-ref
+      metrics.provider.posthog.existingSecret: my-secret
+    asserts:
+      - notFailedTemplate: {}
+
+  - it: accepts posthog with no ref fields and an empty existingSecret (guard is ref-triggered)
+    set:
+      metrics.provider.posthog.existingSecret: ""
+    asserts:
+      - notFailedTemplate: {}
+
+  - it: rejects internalKey with key set but existingSecret empty (conditional minLength)
+    set:
+      metrics.provider.internalKey.key: my-key
+      metrics.provider.internalKey.existingSecret: ""
+    asserts:
+      - failedTemplate:
+          errorPattern: "existingSecret"
+
+  - it: accepts internalKey with key set and a non-empty existingSecret
+    set:
+      metrics.provider.internalKey.key: my-key
+      metrics.provider.internalKey.existingSecret: my-secret
+    asserts:
+      - notFailedTemplate: {}
+
+  - it: accepts internalKey with no key and an empty existingSecret (guard is key-triggered)
+    set:
+      metrics.provider.internalKey.existingSecret: ""
+    asserts:
+      - notFailedTemplate: {}
--- a/charts/shipwright/tests/values_schema_test.yaml
+++ b/charts/shipwright/tests/values_schema_test.yaml
@@ -125,21 +125,19 @@ tests:
     asserts:
       - notFailedTemplate: {}
 
-  - it: rejects posthog with personalApiKeyRef set but existingSecret empty (conditional minLength)
+  - it: accepts posthog with personalApiKeyRef set and existingSecret empty (guard is existingSecret-triggered)
     set:
       metrics.provider.posthog.personalApiKeyRef: my-key-ref
       metrics.provider.posthog.existingSecret: ""
     asserts:
-      - failedTemplate:
-          errorPattern: "existingSecret"
+      - notFailedTemplate: {}
 
-  - it: rejects posthog with projectIdRef set but existingSecret empty (conditional minLength)
+  - it: accepts posthog with projectIdRef set and existingSecret empty (guard is existingSecret-triggered)
     set:
       metrics.provider.posthog.projectIdRef: my-project-ref
       metrics.provider.posthog.existingSecret: ""
     asserts:
-      - failedTemplate:
-          errorPattern: "existingSecret"
+      - notFailedTemplate: {}
 
   - it: accepts posthog with both ref fields and a non-empty existingSecret
     set:
@@ -149,19 +147,26 @@ tests:
     asserts:
       - notFailedTemplate: {}
 
-  - it: accepts posthog with no ref fields and an empty existingSecret (guard is ref-triggered)
+  - it: accepts posthog with no ref fields and an empty existingSecret (guard never triggers)
     set:
       metrics.provider.posthog.existingSecret: ""
     asserts:
       - notFailedTemplate: {}
 
-  - it: rejects internalKey with key set but existingSecret empty (conditional minLength)
+  - it: rejects posthog with existingSecret set but personalApiKeyRef empty (conditional minLength)
+    set:
+      metrics.provider.posthog.existingSecret: my-secret
+      metrics.provider.posthog.personalApiKeyRef: ""
+    asserts:
+      - failedTemplate:
+          errorPattern: "personalApiKeyRef"
+
+  - it: accepts internalKey with key set and existingSecret empty (guard is existingSecret-triggered)
     set:
       metrics.provider.internalKey.key: my-key
       metrics.provider.internalKey.existingSecret: ""
     asserts:
-      - failedTemplate:
-          errorPattern: "existingSecret"
+      - notFailedTemplate: {}
 
   - it: accepts internalKey with key set and a non-empty existingSecret
     set:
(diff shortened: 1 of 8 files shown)


charts/shipwright/values.schema.json
```diff
@@ -112,6 +112,32 @@
               "properties": {
                 "name": { "type": "string" }
               }
+            },
+            "provider": {
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "posthog": {
```
dmcaulay, line 120: Silent misconfiguration gap: `personalApiKeyRef` and `projectIdRef` can be set while `existingSecret` remains `""`. This passes `helm lint` but the template gate (`{{- if .Values.metrics.provider.posthog.existingSecret }}`) fires on the secret name — so no PostHog env vars are injected and the service silently stays in SQLite mode with no error.

The `auth.google` block in this same file demonstrates the fix — add an `if/then` requiring `existingSecret` to be non-empty (`minLength: 1`) when either ref field is populated:
```json
"if": {
  "anyOf": [
    {"properties": {"personalApiKeyRef": {"minLength": 1}}, "required": ["personalApiKeyRef"]},
    {"properties": {"projectIdRef": {"minLength": 1}}, "required": ["projectIdRef"]}
  ]
},
"then": {
  "properties": {"existingSecret": {"type": "string", "minLength": 1}},
  "required": ["existingSecret"]
}
```

charts/shipwright/values.schema.json
```diff
@@ -112,6 +112,32 @@
...
+                  "properties": {
+                    "existingSecret":   { "type": "string" },
+                    "personalApiKeyRef": { "type": "string" },
+                    "projectIdRef":      { "type": "string" }
+                  }
+                },
+                "adminUrl":          { "type": "string" },
+                "basePath":          { "type": "string" },
+                "requireOwnerRole":  { "type": "boolean" },
+                "internalKey": {
```
dmcaulay, line 143: Same silent misconfiguration pattern: `key` can be set while `existingSecret` is `""`. The template gates `METRICS_INTERNAL_API_KEY` injection on `internalKey.existingSecret`, so setting only `key` produces no injection and no helm error. This is particularly risky when `requireOwnerRole: true` depends on the internal key for admin role lookups.

Add an `if/then`: if `key` is non-empty, require `existingSecret` with `minLength: 1`.

dmcaulay (commented):
COMMENT — two schema gaps allow silent misconfiguration before merge: (1) `provider.posthog.personalApiKeyRef`/`projectIdRef` can be set with `existingSecret: ""` — passes `helm lint` but injects no PostHog env vars at runtime; (2) same issue with `provider.internalKey.key` without `existingSecret` — silently breaks METRICS_INTERNAL_API_KEY injection. The `auth.google` `if/then` pattern already in this file is the fix for both. Everything else is clean: env var names match the metrics service source, all injections are correctly gated, default path is unchanged, and the 10 new helm-unittest specs are thorough.

charts/shipwright/values.schema.json
dmcaulay, line 120: Silent misconfiguration gap: `personalApiKeyRef` and `projectIdRef` can be set while `existingSecret` remains `""`. This passes `helm lint` but the template gate (`{{- if .Values.metrics.provider.posthog.existingSecret }}`) fires on the secret name — so no PostHog env vars are injected and the service silently stays in SQLite mode with no error.

The `auth.google` block in this same file demonstrates the fix — add an `if/then` requiring `existingSecret` to be non-empty (`minLength: 1`) when either ref field is populated:
```json
"if": {
  "anyOf": [
    {"properties": {"personalApiKeyRef": {"minLength": 1}}, "required": ["personalApiKeyRef"]},
    {"properties": {"projectIdRef": {"minLength": 1}}, "required": ["projectIdRef"]}
  ]
},
"then": {
  "properties": {"existingSecret": {"type": "string", "minLength": 1}},
  "required": ["existingSecret"]
}
```

charts/shipwright/values.schema.json
dmcaulay, line 143: Same silent misconfiguration pattern: `key` can be set while `existingSecret` is `""`. The template gates `METRICS_INTERNAL_API_KEY` injection on `internalKey.existingSecret`, so setting only `key` produces no injection and no helm error. This is particularly risky when `requireOwnerRole: true` depends on the internal key for admin role lookups.

Add an `if/then`: if `key` is non-empty, require `existingSecret` with `minLength: 1`.

dmcaulay (commented):
COMMENT — version bump commit (f5deecf) is clean, but the two schema gaps flagged in the prior review remain unaddressed: (1) `provider.posthog.personalApiKeyRef`/`projectIdRef` can be set while `existingSecret` is empty — passes helm lint but no PostHog env vars are injected at runtime; (2) `provider.internalKey.key` settable without `internalKey.existingSecret` — same silent no-op. No new issues introduced by the version bump. Add `if/then` guards to both schema objects (the `auth.google` pattern already in this file is the model).

dodizzle (commented):
COMMENT (self-review — recorded as COMMENT since GitHub rejects self-APPROVE via the API).

The deployment template logic is correct and the helm-unittest coverage is thorough, but the `if/then` schema guard added in `b843107` has inverted logic that breaks a plain default `helm install`. The CI `ct install` failure confirms it: `at '/metrics/provider': validation failed`.

**The issue:** `values.yaml` defaults both `personalApiKeyRef` and `projectIdRef` to non-empty strings (`POSTHOG_PERSONAL_API_KEY`, `POSTHOG_PROJECT_ID`). The schema's `if` fires when either ref field has `minLength: 1` — which is always true in the default state. The `then` requires `existingSecret` to have `minLength: 1`, but it defaults to `""`. Schema validation rejects a plain install.

**The fix:** Reverse the direction. Trigger on `existingSecret` being non-empty, then require the ref fields. This preserves the "if you configure the secret, you must name the keys" constraint without firing on the default unconfigured state.

Example for `posthog`:
```json
"if": {
  "properties": { "existingSecret": { "type": "string", "minLength": 1 } },
  "required": ["existingSecret"]
},
"then": {
  "properties": {
    "personalApiKeyRef": { "type": "string", "minLength": 1 },
    "projectIdRef": { "type": "string", "minLength": 1 }
  },
  "required": ["personalApiKeyRef", "projectIdRef"]
}
```
Apply the same reversal to `internalKey` (`if existingSecret non-empty, then key must be non-empty`).

charts/shipwright/values.schema.json
```diff
@@ -112,6 +112,50 @@
...
+              "properties": {
+                "posthog": {
+                  "type": "object",
+                  "additionalProperties": false,
+                  "properties": {
+                    "existingSecret":   { "type": "string" },
+                    "personalApiKeyRef": { "type": "string" },
+                    "projectIdRef":      { "type": "string" }
+                  },
+                  "if": {
```
dodizzle, line 128: The `if/then` direction is inverted. Both `personalApiKeyRef` and `projectIdRef` have non-empty defaults in `values.yaml` (`POSTHOG_PERSONAL_API_KEY` / `POSTHOG_PROJECT_ID`), so this `if` fires unconditionally in a default install — the `then` then requires `existingSecret` to have `minLength: 1`, but it defaults to `""`. Result: plain `helm install` fails with `at '/metrics/provider': validation failed`. Confirmed by the CI `ct install` failure.

Fix: trigger on `existingSecret` being non-empty, require refs to be non-empty in `then`:
```json
"if": {
  "properties": { "existingSecret": { "type": "string", "minLength": 1 } },
  "required": ["existingSecret"]
},
"then": {
  "properties": {
    "personalApiKeyRef": { "type": "string", "minLength": 1 },
    "projectIdRef": { "type": "string", "minLength": 1 }
  },
  "required": ["personalApiKeyRef", "projectIdRef"]
}
```

charts/shipwright/tests/values_schema_test.yaml
```diff
@@ -124,3 +124,54 @@ tests:
       tls.certManager.issuerRef.name: letsencrypt
     asserts:
       - notFailedTemplate: {}
+
+  - it: rejects posthog with personalApiKeyRef set but existingSecret empty (conditional minLength)
```
dmcaulay, line 128: These 3 `failedTemplate` tests (this block + the two that follow) were added in b8431075 alongside the schema guards, but the guards were reverted in b44bca8. With no if/then guards in `values.schema.json`, helm-unittest will render these configs successfully — the `failedTemplate` assertions will fail CI.

Fix: remove all 3 tests, or convert them to `notFailedTemplate` with an updated description documenting that partial config passes schema validation silently (an acceptable trade-off given the helm validator constraint). The test file currently has 3 tests expecting `errorPattern: "existingSecret"` that will never match.

dmcaulay (commented):
COMMENT — schema guard investigation is thorough (b8431075 added if/then guards, b44bca8 reverted them after discovering the helm validator applies `minLength` from the `if` condition as a base constraint — a legitimate blocker). One fix needed before merge: 3 `failedTemplate` tests in `values_schema_test.yaml` (added in b8431075) were not cleaned up when the guards were reverted in b44bca8. With no guards in the schema, these tests assert a failure that won't happen — they will fail Helm CI. Everything else is clean: template injection logic is correct, default path is unchanged, and the 10 feature tests are thorough.

dmcaulay (approved):
APPROVE — all findings resolved. The inverted schema guard approach (if existingSecret is non-empty, then require ref fields) cleanly handles the helm validator constraint: empty existingSecret (default) keeps the if-condition false so no constraint is applied to refs; non-empty existingSecret triggers the then-constraint ensuring refs are populated. Tests are updated to match — failedTemplate when existingSecret is set but refs are empty, notFailedTemplate when refs are set but existingSecret is empty. Template injection, schema, and defaults are all consistent.

PR #375, opened 2026-06-15
fix(github-adapter): ensure session milestone exists before gh issue create

### LINKED ISSUE #374: GitHub task-store adapter requires a session milestone it never creates
## Bug: GitHub task-store adapter requires a session milestone it never creates

The GitHub backend's insert/create path (`scripts/adapters/github.ts`) is asymmetric about the two things it derives from `task.session`:

- The **session label** is ensured idempotently right before create:
  ```js
  await this.runGh(["label","create",`session:${task.session}`,"--repo",this.repoFlag,"--force"]);
  ```
- The **session milestone** is passed straight to `gh issue create` with no ensure:
  ```js
  createArgs.push("--milestone", task.session);   // fails if the milestone doesn't exist
  ```

`gh issue create --milestone <x>` errors if `<x>` doesn't exist, so the **first `append` of any new GitHub-backed session dies on task 1 and creates zero child issues**.

`setup()` doesn't help: it ensures all `status:*` labels with `--force`, but milestones are per-session and only known at append time, so there's nothing for setup to pre-create. The fix belongs in the insert path.

### Repro
1. Fresh session, GitHub backend (`~/.shipwright.json` → `taskStore: github`).
2. `task_store.ts append --file tasks.json` where tasks carry a `session` with no matching milestone.
3. → `could not add to milestone '<session>': '<session>' not found`; `{"inserted":0}`.

Observed live while standing up session `vitals-agent-image` (parent #367); had to hand-create milestone #23 and re-run `append` (idempotent by id → then `{"inserted":6}`).

### Fix
Mirror the label-ensure for milestones. `gh` has no `milestone create --force`, so find-or-create via REST, guarded once per distinct session in a batch:
```js
const ms = JSON.parse(await this.runGh(["api", `repos/${owner}/${repo}/milestones`, "--paginate"]));
if (!ms.some(m => m.title === session))
  await this.runGh(["api","--method","POST",`repos/${owner}/${repo}/milestones`,"-f",`title=${session}`]);
```
Add a unit test asserting `ensureMilestone` POSTs only when absent.

### Secondary (optional, same area)
- **Failure surfacing:** the failed append printed the full task JSON + raw `gh` command + error, with no clear "aborted, 0 inserted" line — easy to misread as success. Add a clear abort message.
- **Non-atomic batch:** `append` isn't transactional; a mid-batch failure can leave partially-created issues. The milestone fix removes this specific trigger, but worth noting.

dmcaulay:
## Summary
- Add `ensureMilestone(session)` to `GitHubTaskStore` — finds-or-creates the session milestone via REST before calling `gh issue create --milestone`
- Guard with a `Set<string>` in the batch loop so the milestone API is only hit once per distinct session, not once per task
- Update 3 existing fake-gh scripts in unit tests to handle the new `gh api` call

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `ensureMilestone` called before `--milestone` in createArgs | MET |
| 2 | POST only when milestone is absent (idempotent) | MET |
| 3 | Unit test: no POST when milestone exists | MET |
| 4 | Unit test: POST when milestone is missing | MET |
| 5 | All existing tests still pass | MET |

## Test Plan
- [x] `bun test plugins/shipwright/scripts/adapters/github.unit.test.ts` — 74/74 pass
- [x] `bun run --filter='*' typecheck` — all exit 0
- [x] Coverage: 94.55% on github.ts (target: 90%)

Fixes #374

### DIFF
--- a/plugins/shipwright/scripts/adapters/github.ts
+++ b/plugins/shipwright/scripts/adapters/github.ts
@@ -208,6 +208,27 @@ export class GitHubTaskStore implements TaskStore {
     }
   }
 
+  private async ensureMilestone(session: string): Promise<void> {
+    const g = this.config.github;
+    if (!g) return;
+    const raw = await this.runGh([
+      "api",
+      `repos/${g.owner}/${g.repo}/milestones`,
+      "--paginate",
+    ]);
+    const existing = JSON.parse(raw) as Array<{ title: string }>;
+    if (!existing.some((m) => m.title === session)) {
+      await this.runGh([
+        "api",
+        "--method",
+        "POST",
+        `repos/${g.owner}/${g.repo}/milestones`,
+        "-f",
+        `title=${session}`,
+      ]);
+    }
+  }
+
   private async fetchAllIssues(): Promise<GhIssue[]> {
     const raw = await this.runGh([
       "issue",
@@ -314,6 +335,7 @@ export class GitHubTaskStore implements TaskStore {
     let inserted = 0;
     const updated = 0;
     const currentUser = await this.resolveCurrentGhUser();
+    const ensuredMilestones = new Set<string>();
 
     for (const task of tasks) {
       if (!task.id) {
@@ -340,6 +362,10 @@ export class GitHubTaskStore implements TaskStore {
           this.repoFlag,
           "--force",
         ]);
+        if (!ensuredMilestones.has(task.session)) {
+          await this.ensureMilestone(task.session);
+          ensuredMilestones.add(task.session);
+        }
       }
 
       const createArgs = [
--- a/plugins/shipwright/scripts/adapters/github.ts
+++ b/plugins/shipwright/scripts/adapters/github.ts
@@ -213,8 +213,7 @@ export class GitHubTaskStore implements TaskStore {
     if (!g) return;
     const raw = await this.runGh([
       "api",
-      `repos/${g.owner}/${g.repo}/milestones`,
-      "--paginate",
+      `repos/${g.owner}/${g.repo}/milestones?per_page=100`,
     ]);
     const existing = JSON.parse(raw) as Array<{ title: string }>;
     if (!existing.some((m) => m.title === session)) {
--- a/plugins/shipwright/scripts/adapters/github.unit.test.ts
+++ b/plugins/shipwright/scripts/adapters/github.unit.test.ts
@@ -614,6 +614,12 @@ if (args[0] === "issue" && args[1] === "list") {
   process.exit(0);
 }
 
+// Handle milestones API calls (ensureMilestone)
+if (args[0] === "api") {
+  console.log("[]");
+  process.exit(0);
+}
+
 // Capture issue create calls
 if (args[0] === "issue" && args[1] === "create") {
   const count = parseInt(readFileSync(${JSON.stringify(createCountFile)}, "utf8")) + 1;
@@ -1443,6 +1449,11 @@ if (args[0] === "label" && args[1] === "create") {
   process.exit(0);
 }
 
+if (args[0] === "api") {
+  console.log("[]");
+  process.exit(0);
+}
+
 if (args[0] === "issue" && args[1] === "create") {
   writeFileSync(${JSON.stringify(argsFile)}, JSON.stringify(args));
   console.log("https://github.com/test-owner/test-repo/issues/77");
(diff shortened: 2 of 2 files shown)


plugins/shipwright/scripts/adapters/github.ts
```diff
@@ -208,6 +208,27 @@ export class GitHubTaskStore implements TaskStore {
...
 
+  private async ensureMilestone(session: string): Promise<void> {
+    const g = this.config.github;
+    if (!g) return;
+    const raw = await this.runGh([
+      "api",
+      `repos/${g.owner}/${g.repo}/milestones`,
+      "--paginate",
+    ]);
+    const existing = JSON.parse(raw) as Array<{ title: string }>;
```
dodizzle, line 218: `gh api --paginate` concatenates pages as `[{...}][{...}]` — not valid JSON. `JSON.parse(raw)` throws a `SyntaxError` on any repo with more than 30 milestones. Drop `--paginate` (milestones rarely reach 30), or switch to `--jq '.[].title'` and check string inclusion directly to avoid `JSON.parse` entirely.

plugins/shipwright/scripts/adapters/github.unit.test.ts
```diff
@@ -614,6 +614,12 @@ if (args[0] === "issue" && args[1] === "list") {
   process.exit(0);
 }
 
+// Handle milestones API calls (ensureMilestone)
+if (args[0] === "api") {
+  console.log("[]");
```
dodizzle, line 625: This broad catch-all intercepts `gh api user --jq .login` from `resolveCurrentGhUser`, returning `"[]"` instead of a user login. `currentUser` becomes `"[]"` and `--assignee []` is silently passed to `gh issue create`. Narrow this to `args[1].includes("milestones")`, or add an explicit `args[1] === "user"` branch before this fallback that returns a real dummy login.

plugins/shipwright/scripts/adapters/github.unit.test.ts
```diff
@@ -1443,6 +1449,11 @@ if (args[0] === "label" && args[1] === "create") {
   process.exit(0);
 }
 
+if (args[0] === "api") {
+  console.log("[]");
```
dodizzle, line 1466: Same issue as the stub at line 619 — this broad `gh api` catch-all will also intercept `resolveCurrentGhUser`. Narrow to `args[1].includes("milestones")` here too.

dodizzle (commented):
COMMENT — Two issues to address before merging: (1) `--paginate` + `JSON.parse` in `ensureMilestone` produces invalid JSON on repos with more than 30 milestones, throwing a `SyntaxError` from `append()`; (2) the broad `gh api` catch-all in two test stubs silently intercepts `resolveCurrentGhUser`, setting `currentUser` to `"[]"` and masking assignee regressions.

dmcaulay (commented):
APPROVE — Clean fix. `ensureMilestone()` correctly finds-or-creates the milestone before `gh issue create --milestone`, and the `ensuredMilestones` Set ensures it's only hit once per distinct session in a batch. Three new test cases (absent, exists, deduplication) all cover the right behavior. One minor note for the future: the milestones `GET` uses default `state=open`, so a closed session milestone wouldn't be found and a re-create attempt would 422 — but that scenario is rare and would fail loudly, not silently, so no action needed here.

dmcaulay (commented):
APPROVE — Both findings addressed. (1) `--paginate` replaced with `?per_page=100` — no more multi-page JSON concatenation, `JSON.parse` gets a single valid array. (2) Broad `args[0] === "api"` catch-alls in the two test stubs narrowed to `args[1].includes("milestones")`, with an explicit `args[1] === "user"` handler returning `"test-user"` — `resolveCurrentGhUser` is now tested rather than silently intercepted. 74/74 tests pass.

dodizzle (approved):
APPROVE — Both prior findings addressed. (1) `--paginate` replaced with `?per_page=100` — single valid JSON array. (2) Test stubs narrowed: explicit `api user` handler returns `"test-user"`, milestones catch-all now only fires for milestone endpoints. Three new test cases (absent milestone, exists, batch-dedup) cover the right behavior. `fetch_all_tags: true` in the three build workflows is a correct shallow-clone fix. Clean to merge.

PR #381, opened 2026-06-16
fix(admin): parse plugin records as canonical claude install spec

dmcaulay:
## Problem

The agent config bundle derives `{marketplace, plugin}` from a stored plugin `name`, and the harness installs `` `${plugin}@${marketplace}` `` (`agent/src/setup.ts`). The old derivation (`admin/src/api.ts`) assumed an npm-scoped convention:

```js
marketplace: name.startsWith("@") ? name.split("/")[0].slice(1) : "shipwright",
plugin: name,   // ← kept the full scoped string
```

This is unreachable for any real plugin. Tracing every possible stored `name`:

| stored `name` | derived marketplace | derived plugin | install spec | valid? |
|---|---|---|---|---|
| `@scope/name` | `scope` | `@scope/name` | `@scope/name@scope` | ❌ `claude plugin install` rejects scoped ids |
| `name` | `shipwright` | `name` | `name@shipwright` | ❌ can't target another marketplace |

**No stored value could ever produce the correct `<plugin>@<marketplace>`** (e.g. `time-tracker@acme`).

## Fix

Store `name` as the exact string you'd pass to `claude plugin install` — the canonical spec — and parse on the first `@`:

- `"<plugin>@<marketplace>"` → `{ plugin, marketplace }`
- bare `"<plugin>"` → `{ plugin, marketplace: "shipwright" }`

Round-trips through the harness unchanged, and matches what operators actually type.

## Changes

- `admin/src/api.ts` — split on first `@`; bare name defaults to the shipwright marketplace
- `admin/src/api.smoke.test.ts` — assert canonical parsing, including a marketplace that itself contains `/` (`my-plugin@org/my-marketplace`)
- `admin/src/agents-api.{ts,smoke.test.ts}`, `admin/src/agent-plugins.ts` — align CRUD fixtures, comments, and JSDoc to the canonical convention
- `plugins/shipwright/skills/agent-admin/SKILL.md` — document the `name` format, add a marketplace-scoped install example, warn against the npm-scoped form

## Compatibility

No agent currently has a plugin record (verified against the deployed admin API across all agents), so there is no stored data to migrate.

## Verification

- `bunx biome lint .` — clean
- `bun test admin/` — 409 pass / 0 fail
- `task typecheck` — clean

### DIFF
--- a/admin/src/agent-plugins.ts
+++ b/admin/src/agent-plugins.ts
@@ -2,7 +2,10 @@
  * agent/src/agent-plugins.ts
  * AgentPluginService — CRUD for Claude Code plugins per agent.
  *
- * Each plugin is identified by its package name (e.g. "@shipwright/plugin").
+ * Each plugin is identified by its canonical Claude install spec — the string
+ * passed to `claude plugin install`: "<plugin>@<marketplace>" (e.g.
+ * "shipwright@shipwright"), or a bare "<plugin>" that defaults to the bundled
+ * "shipwright" marketplace.
  * The unique constraint on [agentId, name] prevents duplicates; add() uses
  * upsert so re-adding an existing plugin updates its version and re-enables it.
  */
--- a/admin/src/agents-api.smoke.test.ts
+++ b/admin/src/agents-api.smoke.test.ts
@@ -239,7 +239,7 @@ function makeMockDeps(): AdminDeps {
         {
           id: PLUGIN_ID,
           agentId: AGENT_ID,
-          name: "@shipwright/plugin",
+          name: "shipwright@shipwright",
           version: "1.0.0",
           enabled: true,
           createdAt: new Date("2024-01-01"),
@@ -249,7 +249,7 @@ function makeMockDeps(): AdminDeps {
       add: async () => ({
         id: PLUGIN_ID,
         agentId: AGENT_ID,
-        name: "@shipwright/plugin",
+        name: "shipwright@shipwright",
         version: "1.0.0",
         enabled: true,
         createdAt: new Date("2024-01-01"),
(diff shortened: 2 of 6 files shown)


dodizzle (commented):
Two issues before merge.

**1. `ensureMilestone` pagination cap** (`plugins/shipwright/scripts/adapters/github.ts:216`) — `milestones?per_page=100` doesn't paginate. If the repo accumulates >100 milestones, `existing.some(m => m.title === session)` won't find the existing one and a duplicate is silently created. GitHub allows duplicate milestone titles, so this fails quietly. Fix: use `gh api --paginate repos/{owner}/{repo}/milestones` or follow the `Link` header.

**2. PR scope vs title** — Title says `fix(admin):` but the diff also adds `discoverBakedMarketplaces` in `agent/src/setup.ts`, `ensureMilestone` in `plugins/shipwright/scripts/adapters/github.ts`, and `fetch_all_tags: true` across three CI workflows — none mentioned in the PR body. Retitle to `fix:` (or broader) and add a sentence per bundled change.

Core spec-parsing fix (`indexOf('@')`) is correct and handles `org/my-marketplace` specs correctly. Baked marketplace design is clean. Test coverage is solid across unit/integration/smoke layers.

dodizzle (approved):
Correction to my previous review — I diffed against the wrong base and flagged changes in `github.ts` and `agent/setup.ts` that are not in this PR. Apologies for the noise.

Actual diff is 6 files scoped entirely to `admin/`. The `indexOf('@')` fix is correct: handles bare names, marketplace-scoped specs, and marketplaces containing `/` (the `org/my-marketplace` smoke test covers this explicitly). No existing plugin records means no migration risk. SKILL.md update is clear and the npm-scoped-form warning is useful. LGTM.

PR #387, opened 2026-06-16
fix(task-store): don't strip the status label when re-applying the current status

dmcaulay:
## Summary

Fixes a latent bug in the GitHub task-store adapter that **orphans tasks** (makes them invisible to `query`/`update`), and hardens the `task-store` skill so the lifecycle is self-documenting and agents stop reaching for the backing store file.

## The bug

`GitHubTaskStore.update()` swaps status labels with a single call:
`gh issue edit --add-label status:<new> --remove-label status:<old>`. When **new === old** (re-applying the current status — e.g. updating a `pr_open` task to `pr_open` to set `pr`), gh adds and removes the *same* label in one invocation and nets to a **removal**. The issue is left with no `status:*` label; `fetchAllIssues` filters by that label, so the task is no longer enumerable → `query --id` returns `[]` and `update` throws `task not found`. Recovery required manually re-adding the label.

The prior guard only skipped `--remove-label` when there was *no* prior label (`_labelStatus === undefined`); it didn't handle `_labelStatus === newStatus`. Hit live while moving SWD-2.2 (vitals-os#1509) onto a `pr_open` record.

## Fix

Skip `--remove-label` when `_labelStatus === newStatus` (re-adding an already-present label is an idempotent no-op). Regression test added asserting a same-status update emits `--add-label` without a paired same-label `--remove-label`.

## Docs / skill

- **`references/task-schema.md`** — adds the canonical, backend-agnostic Task schema (every field + per-status required fields), restoring the previously-dangling reference link.
- **Self-sufficient lifecycle** — the skill now documents the label↔body model, the orphan gotcha + recovery, and every status transition, and is marked the canonical reference (no need to read `dev-task.md`).
- **"CLI is the only interface"** — reframed positively to steer agents off the backing store, *without* naming the store file (so it doesn't reinforce the file-reaching tendency it discourages). Dropped the "todos" concept (→ "task" / "task schema").

## Test Plan
- [x] `bun test github.unit.test.ts` → 75 pass (incl. new regression test)
- [x] `biome lint` + `tsc --noEmit` clean

Version bump handled by semantic-release (`fix:` → patch / 4.27.3).

### DIFF
--- a/plugins/shipwright/.claude-plugin/plugin.json
+++ b/plugins/shipwright/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "shipwright",
-  "version": "4.27.2",
+  "version": "4.27.3",
   "description": "Shipwright — conversational planning, queue-driven execution, policy-controlled code review, a five-phase test-readiness pipeline, and autonomous deploy (merge → canary → promote) for any software project",
   "author": {
     "name": "App Vitals",
--- a/plugins/shipwright/README.md
+++ b/plugins/shipwright/README.md
@@ -1,4 +1,4 @@
-# Shipwright v4.27.2
+# Shipwright v4.27.3
 
 A structured dev pipeline plugin for Claude Code. Plan sessions, execute tasks, run autonomous dev loops, perform multi-agent code reviews, and conduct integrated project research — for any software project.
 
--- a/plugins/shipwright/scripts/adapters/github.ts
+++ b/plugins/shipwright/scripts/adapters/github.ts
@@ -457,10 +457,19 @@ export class GitHubTaskStore implements TaskStore {
         "--add-label",
         newLabel,
       ];
-      // Only remove the old label if it was actually present on the issue.
-      // Attempting to remove a label that isn't there causes gh to error,
-      // which leaves the issue with no status label and invisible to fetchAllIssues.
-      if (targetTask._labelStatus !== undefined) {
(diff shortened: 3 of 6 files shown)


dmcaulay:
Added a second commit (`docs(task-store)`) extending scope slightly beyond the bug fix:

- **`references/task-schema.md`** — the canonical, backend-agnostic Task schema (every field + per-status required fields), restoring the previously-dangling reference link.
- **Reframed the skill around "the CLI is the only interface"** — the underlying persistence is private to the backend; agents must use `task_store.ts`, not reach for the store file. Stated *positively and without naming the file*, so the guidance doesn't reinforce the very file-reaching tendency it's meant to discourage.
- **Dropped the "todos" concept** from the agent-facing docs (→ "task"/"task schema").

Still a `fix:` PR overall (semantic-release → patch / 4.27.3).

dodizzle (approved):
APPROVE — Surgical fix for the task-orphaning bug where re-applying the current status caused `gh` to net `--add-label X --remove-label X` to a removal. Single guard clause addition with a targeted regression test. Documentation and schema reference are accurate and useful. No issues found.

dmcaulay (commented):
APPROVE — Fix is correct and minimal. The `_labelStatus !== newStatus` guard prevents `gh issue edit --add-label X --remove-label X` from netting to a removal (task orphan). Regression test validates the previously unguarded path. CI green. No blocking concerns.

PR #388, opened 2026-06-16
fix(admin): correct Metrics toolbar link default from /sw/dashboard to /dashboard

dodizzle:
## Summary

- The admin toolbar's **Metrics** link was using `/sw/dashboard` as its default — a path that doesn't exist. The metrics service serves at `/dashboard` on the same public host.
- Fixed the default in both `renderAdminToolbar()` (`admin/src/admin-ui-styles.ts`) and the shared `renderShipwrightToolbar()` (`lib/web/toolbar.ts`).
- Documented `METRICS_DASHBOARD_URL` in `docs/configuration.md` — same-host deployments need no change; set it to an absolute URL when the metrics service is on a different host/port (as in local dev via `task stack`, which already sets `http://localhost:3460/dashboard`).

## No new env var needed for production

The umbrella chart (`infra/helm/shipwright-deploy`) does **not** need a new value. The default `/dashboard` resolves correctly on the shared ingress host. The existing `METRICS_DASHBOARD_URL` override remains available if ever needed.

## Test plan

- [x] Existing `dev-tmux.unit.test.ts` (63 tests) passes — verifies local dev still gets the correct absolute URL override
- [x] No changes to metrics service asset paths (`/sw/dashboard/styles.css`, `/sw/dashboard/app.js`) — those are internal and correct

### DIFF
--- a/admin/src/admin-ui-styles.ts
+++ b/admin/src/admin-ui-styles.ts
@@ -280,6 +280,6 @@ export function renderAdminToolbar(userName: string, activePath = ""): string {
     userName,
     activePath,
     logoutAction: "/admin/logout",
-    metricsUrl: process.env.METRICS_DASHBOARD_URL ?? "/sw/dashboard",
+    metricsUrl: process.env.METRICS_DASHBOARD_URL ?? "/dashboard",
   });
 }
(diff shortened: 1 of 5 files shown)


dmcaulay (commented):
COMMENT — 2 stale comments referencing the old `/sw/dashboard` path were not updated:

1. **`lib/web/toolbar.ts:12`** — JSDoc on `ShipwrightToolbarOptions.metricsUrl` still reads `Defaults to "/sw/dashboard"`. The actual default was just changed to `/dashboard` on line 131 of the same file. Code and its own doc now contradict each other. Fix: `/** Full URL of the metrics dashboard. Defaults to "/dashboard". */`

2. **`scripts/dev-tmux.ts:179`** — Comment says "falls back to /sw/dashboard on the admin host (:3001) and 404s." The fallback is now `/dashboard`. The PR correctly updated the parallel comment in `dev-tmux.unit.test.ts:473` but missed the source file.

Core fix is correct — both call sites updated, CI green, docs entry is accurate.

dodizzle (commented):
APPROVE — correct, complete, and consistent. Both functional call sites updated, all stale comments/JSDoc addressed in follow-up, `METRICS_DASHBOARD_URL` documented in `docs/configuration.md`. CI green, no blocking concerns.

*(Self-review: event forced to COMMENT by GitHub API — verdict is APPROVE.)*

PR #391, opened 2026-06-16
feat(chart): admin.encryptionKeys, appBaseUrl/extraEnv, and GKE HealthCheckPolicy knobs

dodizzle:
## Summary

Three chart knobs that retire the post-deploy hacks added in **app-vitals/vitals-os#1523** after the 2026-06-16 prod outage. Each section below describes the gap, the vitals-os workaround that currently compensates, and the new chart value that replaces it.

---

### 1. `admin.encryptionKeys.existingSecret` — stable encryption/session keys across installs

**Gap:** The chart's `admin-secret.yaml` only reuse-or-generates `SHIPWRIGHT_ENCRYPTION_KEY` and `SHIPWRIGHT_SESSION_SECRET` — there is no way to inject pre-existing values. A fresh namespace install mints random keys that cannot decrypt agent env already stored encrypted-at-rest, causing uniform `invalid_auth` on every agent.

**vitals-os workaround (retire this):** The "Pin shipwright-admin encryption keys" step in `.github/workflows/deploy.yml` — fetches both keys from GCP Secret Manager, compares to live values in the `shipwright-admin` Secret, and patches + restarts the admin Deployment when they differ.

**New knob:**
```yaml
admin:
  encryptionKeys:
    existingSecret: shipwright-secrets          # pre-existing Secret name
    encryptionKeyRef: SHIPWRIGHT_ENCRYPTION_KEY # key within that Secret
    sessionSecretRef: SHIPWRIGHT_SESSION_SECRET  # key within that Secret
```
When set, the chart-managed Secret **omits** these keys entirely, and the Deployment sources them via `secretKeyRef` against the caller-managed Secret. Default is empty (existing behaviour: chart generates/reuses via lookup).

---

### 2. `admin.appBaseUrl` + `admin.extraEnv` — public base URL and env passthrough

**Gap:** The admin container has no way to receive `SHIPWRIGHT_ADMIN_APP_BASE_URL` or other arbitrary env vars from chart values. Without the public base URL, OAuth redirect URIs use `localhost:3001` and login fails behind the GKE Gateway.

**vitals-os workaround (retire this):** The "Set shipwright-admin public base URL" step in `.github/workflows/deploy.yml` — runs `kubectl set env` after every install to inject `SHIPWRIGHT_ADMIN_APP_BASE_URL` and then waits for rollout.

**New knobs:**
```yaml
admin:
  appBaseUrl: "https://shipwright.vitals-os.com"  # sets SHIPWRIGHT_ADMIN_APP_BASE_URL
  extraEnv: []                                      # arbitrary envVar objects
```
`extraEnv` is appended last in the container env list, so it can override earlier entries if needed.

---

### 3. `networking.gateway.healthCheckPolicy.enabled` — GKE HealthCheckPolicy for /health

**Gap:** The chart renders no `networking.gke.io/v1 HealthCheckPolicy`. The GKE Gateway controller default-probes `"/"` — both the admin and metrics services return 404 there — and marks the backends UNHEALTHY, returning 503 on the external host.

**vitals-os workaround (retire this):** `infra/helm/shipwright-deploy/templates/healthcheckpolicy.yaml` in the vitals-os umbrella chart — a separate template that creates HealthCheckPolicies for both services, targeting `/health`.

**New knob:**
```yaml
networking:
  type: gateway
  gateway:
    healthCheckPolicy:
      enabled: true   # default: false
```
When `networking.type=gateway` AND `enabled=true`, renders `networking.gke.io/v1 HealthCheckPolicy` resources for the admin Service (and metrics if `metrics.enabled=true`), probing `/health` with 15s interval / 5s timeout / 1-healthy / 2-unhealthy thresholds. Disabled by default so non-GKE Gateway installs are unaffected.

---

## Validation

- `helm lint` — clean (0 failures)
- `helm template` — verified all three new knobs render correctly
- New helm-unittest tests added: `admin_secret_test.yaml` (existingSecret omits keys), `admin_workload_test.yaml` (secretKeyRef, appBaseUrl, extraEnv), `healthcheckpolicy_test.yaml` (new file, 6 cases)

## Test plan

- [ ] `helm lint charts/shipwright` passes
- [ ] `helm unittest charts/shipwright` passes (once runner has the plugin)
- [ ] `helm template` with `admin.encryptionKeys.existingSecret` set: admin-secret has no SESSION/ENCRYPTION keys; Deployment uses the external secret refs
- [ ] `helm template` with `admin.appBaseUrl` set: `SHIPWRIGHT_ADMIN_APP_BASE_URL` appears in the Deployment env
- [ ] `helm template` with `networking.type=gateway` + `healthCheckPolicy.enabled=true`: two HealthCheckPolicies rendered with `/health` probe path
- [ ] After merging: update `app-vitals/vitals-os` to set these three knobs in `values-prod.yaml` and remove the three workaround steps from `deploy.yml`

### DIFF
--- a/charts/shipwright/CHANGELOG.md
+++ b/charts/shipwright/CHANGELOG.md
@@ -10,6 +10,15 @@ independent of `appVersion`. CI enforces this with
 `ct lint --check-version-increment`. Each release here must mirror the
 `artifacthub.io/changes` annotation in `Chart.yaml`.
 
+## [1.1.0]
+
+### Added
+
+- `admin.encryptionKeys` block: source `SHIPWRIGHT_ENCRYPTION_KEY` and `SHIPWRIGHT_SESSION_SECRET` from a pre-existing Secret instead of generating random values on a fresh namespace install. Set `admin.encryptionKeys.existingSecret` to the Secret name; `encryptionKeyRef` and `sessionSecretRef` select the keys within it (default to the env var names). When set, the chart-managed Secret omits these two keys entirely and the Deployment injects them via `secretKeyRef` against the caller-managed Secret. Default is empty — existing generate-on-install / reuse-on-upgrade behaviour is unchanged.
+- `admin.appBaseUrl`: sets `SHIPWRIGHT_ADMIN_APP_BASE_URL` in the admin container env when non-empty. Required when the admin service is behind a Gateway or Ingress so that OAuth redirect URIs reference the real public host rather than `localhost:3001`. Omitted from the env when left empty (default).
+- `admin.extraEnv`: list of Kubernetes `envVar` objects appended to the admin container env. Provides a generic passthrough for env vars not otherwise covered by chart values. Defaults to `[]` (no extra vars).
+- `networking.gateway.healthCheckPolicy.enabled`: when `true` and `networking.type=gateway`, renders a `networking.gke.io/v1 HealthCheckPolicy` for the admin Service and (when `metrics.enabled=true`) a second one for the metrics Service, both probing `/health` with 15 s interval / 5 s timeout / 1 healthy / 2 unhealthy thresholds. Without this the GKE Gateway controller default-probes `"/"` (both services → 404) and marks the backends UNHEALTHY, returning 503 on the external host. Disabled by default (`false`) so non-GKE Gateway installs are unaffected.
+
 ## [1.0.0]
 
 _First publicly published chart version. New features in this release: externalDatabase and cloudSqlProxy (SWD-1.x). Gateway API networking, cert-manager Certificate, agent-provisioning RBAC, and metrics.provider were shipped in 0.9.0/0.9.1._
--- a/charts/shipwright/Chart.yaml
+++ b/charts/shipwright/Chart.yaml
@@ -8,7 +8,7 @@ type: application
 # Chart version (SemVer). Bump on every chart change, independent of appVersion.
 # See CHANGELOG.md and the README "Versioning" section — CI enforces this bump
 # via `ct lint --check-version-increment`.
-version: 1.0.0
+version: 1.1.0
 # Version of the Shipwright application this chart deploys by default.
 appVersion: "0.1.0"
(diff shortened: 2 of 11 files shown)


dodizzle (commented):
APPROVE — Three chart knobs (encryptionKeys.existingSecret, appBaseUrl, healthCheckPolicy.enabled) are correctly implemented with mutually exclusive template branching, consistent schema/defaults, and comprehensive helm-unittest coverage. No blocking issues.

dodizzle (commented):
APPROVE — Three chart knobs (encryptionKeys.existingSecret, appBaseUrl, healthCheckPolicy.enabled) are correctly implemented with mutually exclusive template branching, consistent schema/defaults, and comprehensive helm-unittest coverage. No blocking issues.

dodizzle (commented):
APPROVE — Follow-on commit (`56a8d18`) updates `metadata_test.yaml` to assert chart version `1.1.0`, matching the `Chart.yaml` bump in this PR. Trivially correct; Helm CI is now passing. No new issues since the prior APPROVE.

dmcaulay (commented):
COMMENT — Implementation is correct and safe, all defaults are backward-compatible, and the `healthcheckpolicy.yaml` test coverage is thorough. One gap before merge:

**`values_schema_test.yaml` not updated for the new `encryptionKeys` if/then guard.** The schema adds: when `admin.encryptionKeys.existingSecret` is non-empty, both `encryptionKeyRef` and `sessionSecretRef` must be non-empty. Every other `if/then` constraint in `values.schema.json` has corresponding `values_schema_test.yaml` cases (posthog, internalKey, certManager, cloudSqlProxy). The `encryptionKeys` guard is the only new conditional with zero coverage in that file.

Suggested additions following the posthog pattern:
- Accept: `existingSecret` set + both refs non-empty
- Reject (`encryptionKeyRef`): `existingSecret` set + `encryptionKeyRef: ""`
- Reject (`sessionSecretRef`): `existingSecret` set + `sessionSecretRef: ""`

PR #414, opened 2026-06-17
fix(deploy): check any self-review for APPROVE, not just the first

dmcaulay:
## Summary

- The deploy scan used `| first` to get a self-review body, so PRs with multiple self-reviews (e.g. a COMMENT finding then an APPROVE) always evaluated the earliest review and silently skipped the PR
- Changed to `| any(startswith("APPROVE"))` — matches the intent already stated in the prose above the command
- Bumps plugin to 4.27.4

## Root cause

PR #1529 (vitals-os) had two self-reviews from the agent: the first was a `COMMENT —` finding, the second was `APPROVE — Both prior findings resolved`. The scan picked the COMMENT, failed the approval check, and skipped the PR at every deploy cron (5pm and 5:30pm PDT on Jun 16).

## Test plan
- [ ] Deploy cron picks up vitals-os #1529 after this ships
- [ ] PRs with a single APPROVE review continue to deploy normally
- [ ] PRs with only COMMENT reviews are still skipped

### DIFF
--- a/plugins/shipwright/.claude-plugin/plugin.json
+++ b/plugins/shipwright/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "shipwright",
-  "version": "4.27.3",
+  "version": "4.27.4",
   "description": "Shipwright — conversational planning, queue-driven execution, policy-controlled code review, a five-phase test-readiness pipeline, and autonomous deploy (merge → canary → promote) for any software project",
   "author": {
     "name": "App Vitals",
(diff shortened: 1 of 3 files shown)


dmcaulay (commented):
APPROVE — One-line fix: `| first` → `| any(startswith("APPROVE"))` in the deploy scan. The prose above the command already said "check if any review... has a body where trimStart().startsWith('APPROVE')" — the jq just didn't match. Confirmed the regression via session logs: vitals-os #1529 was skipped at 5pm and 5:30pm PDT on Jun 16 because a COMMENT finding review preceded the APPROVE review and `| first` returned the wrong one. Fix is correct and minimal.

PR #415, opened 2026-06-17
fix(patch): include workspace/repos in GitHub adapter resolveRepos

dodizzle:
## Summary

- `check-patch.ts` uses `resolveAllRepos` (check-helpers), which unions the primary repo with everything under `workspace/repos/`
- The patch and deploy skills call `task_store.ts repos` → `GitHubTaskStore.resolveRepos`, which returned **only** the primary repo from `shipwright.config.json`
- Since that return is non-empty, the `resolveReposForStore` fallback to `workspace/repos/` never fired
- Net effect: check-patch detects failing CI in a workspace repo (e.g. `bodhi`), triggers the cron, but the patch skill scans only the primary repo, finds nothing, and exits silently

Fix: `GitHubTaskStore.resolveRepos` now merges the primary repo with any repos scanned from `workspace/repos/` (matching the logic in `check-helpers.resolveAllRepos`). Primary repo is always first; duplicates are filtered.

## Test plan

- [ ] Existing 76 unit tests pass
- [ ] New test: `merges config repo with workspace repos, config repo first, no duplicates` — creates a fake git repo under a temp `SHIPWRIGHT_REPOS_DIR` and verifies both repos appear in the result

### DIFF
--- a/plugins/shipwright/scripts/adapters/github.ts
+++ b/plugins/shipwright/scripts/adapters/github.ts
@@ -20,6 +20,7 @@ import type {
   TaskStore,
   TaskStoreConfig,
 } from "../store.ts";
+import { resolveRepos } from "../check-helpers.ts";
 import { warnMissingFields } from "./validation.ts";
 
 // ─── Constants ────────────────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


dmcaulay (commented):
APPROVE — Fix is correct. GitHubTaskStore.resolveRepos() was only returning the primary config repo while check-patch/check-review were already scanning workspace/repos/ via resolveAllRepos — this closes that gap. The additional lint fix (delete → assignment) is semantically equivalent in Bun and keeps CI clean.

PR #416, opened 2026-06-17
feat: enforce single-backend selection in task store, add doctor coexistence warning

dodizzle:
## Summary
- Adds `doctorCheck()` to `create-task-store.ts` — emits backend/config status lines and warns when a non-empty `state/todos.json` coexists with a non-JSON backend (GitHub or Jira)
- Updates `cmdDoctor` in `task_store.ts` to delegate to `doctorCheck()` for non-JSON backends, removing inline duplicate logic
- Adds 7 new unit tests covering single-backend enforcement and the doctor coexistence warning

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | GitHub backend → `query`/`append`/`update` use only GitHub Issues; `todos.json` never read/written | ✅ MET |
| 2 | No backend configured → JSON backend used; no GitHub calls attempted | ✅ MET |
| 3 | `doctor` emits `[warn] config: todos.json exists and is non-empty while github backend is active` | ✅ MET |
| 4 | Existing JSON backend tests unaffected; new test asserts no `todos.json` reads with GitHub config | ✅ MET |

## Test Plan
- [x] `createTaskStore single-backend enforcement` — 3 tests: GitHub config → GitHubTaskStore; no config → JsonTaskStore; GitHub config + todos.json → todos.json untouched
- [x] `doctor coexistence warning` — 4 tests: warns/no-warn for empty/absent todos.json; no warn for JSON backend
- [x] All 19 tests in `create-task-store.unit.test.ts` pass
- [x] Lint clean (`bunx biome lint`)
- [x] Typecheck clean (`tsc --noEmit`)

Closes #407

Generated with [Claude Code](https://claude.com/claude-code)

PR #437, opened 2026-06-17
fix(check-deploy): match **APPROVE** markdown bold in self-review body check

dodizzle:
## Summary

- `hasSelfApproveReview` was calling `startsWith("APPROVE")` on the review body
- The review skill posts verdicts as `**APPROVE** — ...` (markdown bold), so the leading `**` caused the check to fail
- PR #416 was silently skipped every deploy tick because of this mismatch — preCheck exited 1, no session created, no deploy

## Fix

Strip leading `*` characters before the prefix check:
```ts
r.body.trimStart().replace(/^\*+/, "").startsWith("APPROVE")
```

Added a test case for the `**APPROVE**` body format in `check-deploy.test.ts`.

## Test plan

- [x] `bun test check-deploy.test.ts check-deploy.unit.test.ts` — 34 pass
- [x] New test: `exits 0 when self-review body uses markdown bold (**APPROVE**)`

### DIFF
--- a/plugins/shipwright/scripts/check-deploy.test.ts
+++ b/plugins/shipwright/scripts/check-deploy.test.ts
@@ -160,6 +160,31 @@ describe("check-deploy", () => {
     expect(result.exit).toBe(1);
   });
 
+  test("exits 0 when self-review body uses markdown bold (**APPROVE**)", async () => {
+    const pr = makeGhPr({
+      author: { login: "bodhi-agent" },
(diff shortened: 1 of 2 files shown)


dodizzle (commented):
**APPROVE** — Clean, targeted fix with a regression test. The `/^\*+/` strip correctly handles the `**APPROVE**` format the review skill posts. No blocking concerns.

PR #440, opened 2026-06-17
feat(chart): metrics.sessionSecret.existingSecret to share admin's session secret

dmcaulay:
## Problem

The metrics Deployment hardcodes `SHIPWRIGHT_SESSION_SECRET` to the chart-generated `<release>-metrics` Secret (lookup-or-`randAlphaNum`), with **no override**. When the admin sources its session secret from a caller-managed Secret via `admin.encryptionKeys.existingSecret` (the recommended prod posture — see the 1.1.0 encryption-key feature), metrics ends up holding a **different random**.

The metrics dashboard validates the admin-signed session JWT with `SHIPWRIGHT_SESSION_SECRET`. A mismatch means metrics rejects every admin-minted session → **401 on the dashboard's metrics view** when admin + metrics sit behind a shared Gateway. We hit exactly this in prod (admin on the canonical GCP secret `a10724…`, metrics on a chart-generated random `633c5b…`).

## Change

Add `metrics.sessionSecret.existingSecret` (+ `sessionSecretRef`), mirroring the existing `admin.encryptionKeys` idiom:

- **`templates/metrics-deployment.yaml`** — if/else: when set, source `SHIPWRIGHT_SESSION_SECRET` via `secretKeyRef` against the caller-managed Secret; else the chart-managed metrics Secret (unchanged default).
- **`templates/metrics-secret.yaml`** — gate the generate/lookup and the `data` key on the knob, so the chart doesn't write an unused random when the value is externally supplied.
- **`values.yaml`** — `metrics.sessionSecret` block (default empty).

Point it at the same Secret the admin uses and admin-minted dashboard JWTs validate at metrics.

## Compatibility

Purely additive. Default empty preserves the generate-on-install / reuse-on-upgrade behaviour exactly.

## Tests

- `helm unittest charts/shipwright`: **178 passed** (added workload cases for the external + custom-`sessionSecretRef` paths, and secret cases for key-omission + empty-data-when-postgres-disabled).
- `helm lint`: clean.
- `Chart.yaml` `1.2.0 → 1.3.0` (+ artifacthub annotation); `CHANGELOG.md` `[1.3.0]` (and backfilled the missing `[1.2.0]` entry whose only record was the annotation); README values row; metadata version-pin test updated.

### DIFF
--- a/charts/shipwright/templates/metrics-secret.yaml
+++ b/charts/shipwright/templates/metrics-secret.yaml
@@ -13,14 +13,21 @@
   fresh random material on first install. NOTE: `helm template` and helm-unittest
   do NOT execute `lookup`, so they always take the generate branch — that is
   expected.
+
+  When metrics.sessionSecret.existingSecret is set the generate/lookup path is
+  skipped and no session key is written here — the Deployment sources
+  SHIPWRIGHT_SESSION_SECRET via secretKeyRef against the caller-managed Secret
+  (point it at the admin's Secret so admin-minted dashboard JWTs validate).
 */}}
 {{- $secretName := include "shipwright.metrics.secretName" . }}
 {{- $existing := (lookup "v1" "Secret" .Release.Namespace $secretName) }}
 {{- $sessionSecret := "" }}
+{{- if not .Values.metrics.sessionSecret.existingSecret }}
 {{- if and $existing $existing.data (index $existing.data "SHIPWRIGHT_SESSION_SECRET") }}
 {{- $sessionSecret = index $existing.data "SHIPWRIGHT_SESSION_SECRET" }}
 {{- else }}
 {{- $sessionSecret = (randAlphaNum 32 | b64enc) }}
+{{- end }}
 {{- end }}
 apiVersion: v1
 kind: Secret
@@ -30,8 +37,12 @@ metadata:
     {{- include "shipwright.metrics.labels" . | nindent 4 }}
 type: Opaque
 data:
+{{- if not .Values.metrics.sessionSecret.existingSecret }}
   # Generated on first install, reused on upgrade via the lookup idiom above.
+  # Omitted when metrics.sessionSecret.existingSecret is set — the Deployment
+  # sources SHIPWRIGHT_SESSION_SECRET from the caller-managed Secret instead.
   SHIPWRIGHT_SESSION_SECRET: {{ $sessionSecret | quote }}
+{{- end }}
 {{- if .Values.postgresql.enabled }}
   {{- /*
     Assemble METRICS_DATABASE_URL here so the Postgres password is not rendered
--- a/charts/shipwright/CHANGELOG.md
+++ b/charts/shipwright/CHANGELOG.md
@@ -10,6 +10,18 @@ independent of `appVersion`. CI enforces this with
 `ct lint --check-version-increment`. Each release here must mirror the
 `artifacthub.io/changes` annotation in `Chart.yaml`.
 
+## [1.3.0]
+
+### Added
+
+- `metrics.sessionSecret.existingSecret`: source the metrics service's `SHIPWRIGHT_SESSION_SECRET` from a caller-managed Secret instead of the chart-generated random. Point it at the same Secret the admin uses (`admin.encryptionKeys.existingSecret`) so admin-minted dashboard session JWTs validate at the metrics service when both sit behind a shared Gateway — a mismatch returns 401 on the dashboard's metrics view. `sessionSecretRef` selects the key within that Secret (defaults to `SHIPWRIGHT_SESSION_SECRET`). When set, the chart-managed metrics Secret omits `SHIPWRIGHT_SESSION_SECRET` and the Deployment injects it via `secretKeyRef` against the caller-managed Secret. Default empty preserves the existing generate-on-install / reuse-on-upgrade behaviour — purely additive.
+
+## [1.2.0]
+
+### Added
+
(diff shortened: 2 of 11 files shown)


dodizzle (commented):
## Review

Good fix for the prod 401 — the template logic is correct and the helm-unittest coverage for the new paths is solid.

Two items before merge, both following the pattern every other `existingSecret` knob in this chart already satisfies:

### 1. `values.schema.json` missing `metrics.sessionSecret` definition

Every other `existingSecret` knob — `admin.encryptionKeys`, `metrics.provider.posthog`, `metrics.provider.internalKey`, `auth.google` — has a schema object with an `if`/`then` guard enforcing that the companion `*Ref` field is non-empty when `existingSecret` is set. This PR adds `metrics.sessionSecret.{existingSecret, sessionSecretRef}` to `values.yaml` and templates but makes no change to `values.schema.json`. Without a schema entry, a misconfigured install (e.g. `existingSecret` set but `sessionSecretRef` cleared to `""`) passes `helm install --validate` silently, and the pod injects the wrong key from the caller-managed Secret with no deployment-time error.

Add a `sessionSecret` object inside the `metrics` allOf second member's `properties`, mirroring the `internalKey` pattern:
```json
"sessionSecret": {
  "type": "object",
  "additionalProperties": false,
  "properties": {
    "existingSecret":   { "type": "string" },
    "sessionSecretRef": { "type": "string" }
  },
  "if": {
    "properties": { "existingSecret": { "type": "string", "minLength": 1 } },
    "required": ["existingSecret"]
  },
  "then": {
    "properties": { "sessionSecretRef": { "type": "string", "minLength": 1 } },
    "required": ["sessionSecretRef"]
  }
}
```

### 2. `values_schema_test.yaml` missing coverage for `metrics.sessionSecret`

The established pattern (exercised for `posthog`, `internalKey`, `encryptionKeys`, `auth.google`) requires at minimum two schema-behavior cases per `existingSecret` knob: one `notFailedTemplate` when `existingSecret` is non-empty and the ref is set, and one `failedTemplate` when `existingSecret` is non-empty but the ref is empty. These are absent for `metrics.sessionSecret`. CLAUDE.md is explicit: tests land with the code in the same PR.

```yaml
- it: accepts metrics.sessionSecret with existingSecret and sessionSecretRef set
  set:
    metrics.sessionSecret.existingSecret: my-secret
    metrics.sessionSecret.sessionSecretRef: SHIPWRIGHT_SESSION_SECRET
  asserts:
    - notFailedTemplate: {}

- it: rejects metrics.sessionSecret with existingSecret set but sessionSecretRef empty
  set:
    metrics.sessionSecret.existingSecret: my-secret
    metrics.sessionSecret.sessionSecretRef: ""
  asserts:
    - failedTemplate:
        errorPattern: "sessionSecretRef"
```

---

Everything else looks right — the if/else in the Deployment and Secret templates exactly mirrors the admin pattern, and the backfilled `[1.2.0]` CHANGELOG entry is good housekeeping.

dmcaulay:
Thanks @dodizzle — both addressed in `2a5d693` (rebased onto latest main):

1. **`values.schema.json`** — added the `metrics.sessionSecret` object (`additionalProperties: false`) with the `existingSecret`-triggered `if`/`then` guard requiring `sessionSecretRef` (`minLength: 1`), mirroring `internalKey`.
2. **`values_schema_test.yaml`** — added 4 schema-behavior cases mirroring the `internalKey` set, including the `failedTemplate` reject when `sessionSecretRef` is cleared (verified it fires on `errorPattern: sessionSecretRef`).

`helm unittest`: 182 passed · `helm lint`: clean.

dodizzle (approved):
APPROVE — Both prior findings are fully and correctly resolved:

1. `values.schema.json` — `metrics.sessionSecret` object added with `additionalProperties: false` and the `if`/`then` guard (consistent with `admin.encryptionKeys`, `posthog`, and `internalKey` guards already in the file).
2. `values_schema_test.yaml` — four schema-validation cases covering all permutations of the guard (accept: ref set + empty existingSecret; accept: both set; accept: no ref + empty existingSecret; reject: existingSecret set + ref empty).

Test coverage is thorough across all three fixtures (schema, workload wiring, secret key omission + empty-Secret edge case). Ready to merge.

dmcaulay (commented):
APPROVE — core logic correct, solves a real prod 401, follows the established chart idiom. One edge case to cover: when `existingSecret` is set and `postgresql.enabled=false`, the chart-managed metrics Secret renders with an empty `data:` block (not a crash, but a useless chart-owned resource with no test coverage).

_Note: Self-review — verdict recorded as COMMENT per GitHub API constraint._

charts/shipwright/templates/metrics-secret.yaml
```diff
@@ -30,8 +37,12 @@ metadata:
     {{- include "shipwright.metrics.labels" . | nindent 4 }}
 type: Opaque
 data:
+{{- if not .Values.metrics.sessionSecret.existingSecret }}
```
dmcaulay, line 40: When `metrics.sessionSecret.existingSecret` is set **and** `postgresql.enabled=false`, this `{{- if not ... }}` gate skips `SHIPWRIGHT_SESSION_SECRET`, and the outer postgres gate skips `METRICS_DATABASE_URL` — leaving the Secret with a bare `data:` block (no keys). Kubernetes accepts it but the resource is never used.

Consider suppressing the whole Secret in this case:
```yaml
{{- if or (not .Values.metrics.sessionSecret.existingSecret) .Values.postgresql.enabled }}
  # ... data block ...
{{- end }}
```
Or add a helm-unittest case asserting this combination produces `hasDocuments: count: 0` (suppressed) or explicitly documents the empty-Secret behavior as intentional.

PR #442, opened 2026-06-17
chore: remove arc-queue-probe from SYSTEM_CRONS

dodizzle:
## Summary

- Removes `arc-queue-probe` from `SYSTEM_CRONS` — it's a vitals-os-specific cron that monitors ARC runner queue times for that platform's infra
- Shipwright has no ARC runners or `arc-queue-probe` script, so it doesn't belong here
- Updates the unit test count (11 → 10) and removes the arc-queue-probe test block

### DIFF
--- a/admin/src/system-crons.ts
+++ b/admin/src/system-crons.ts
@@ -94,12 +94,4 @@ export const SYSTEM_CRONS: readonly SystemCron[] = [
     silent: true,
     enabled: false,
   },
-  {
-    name: "arc-queue-probe",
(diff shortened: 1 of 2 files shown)


dmcaulay (approved):
APPROVE — clean, complete removal with no stale references left behind. `arc-queue-probe` was always vitals-os-specific (the `arc-queue-probe.ts` script never existed in this repo), the test count/names and polling-cron filter exclusion are all updated correctly.

PR #456, opened 2026-06-17
feat(admin): add web delete agent danger-zone + POST route (ALP-B.2)

dmcaulay:
## Summary
- Adds a danger-zone section to the agent detail page (admin-only) with a "Delete agent" button and confirmation dialog
- Adds `POST /admin/agents/:id/delete` route that calls `provisioner.deprovision()` then `prisma.agent.delete()`; redirects to `/admin/agents?success=deleted` on success or back to the agent page with an error param on failure
- Wires `provisioner` into `AdminUIDeps` / `createAdminUIApp` factory (it was already constructed in `main.ts` but not passed to the UI layer)

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | renderAgentDetailPage() renders danger-zone section with Delete button visible only when isAdmin=true | ✅ MET |
| 2 | POST /admin/agents/:id/delete calls provisioner.deprovision() then deletes the agent row; redirects to /admin/agents on success | ✅ MET |
| 3 | Route is admin-gated: non-admin session returns 403 | ✅ MET |
| 4 | Smoke test covers both the admin success path (302 redirect) and the non-admin rejection (403) | ✅ MET |
| 5 | Safe to deploy standalone: yes | ✅ MET |

## Test Plan
- [x] `bun test admin/src/admin-ui.smoke.test.ts` — 58 tests pass (2 new: admin delete success + non-admin 403)
- [x] `bun test admin/src/` — 425 tests pass, 0 fail
- [x] `bun run --filter='*' typecheck` — all packages pass

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -547,10 +547,18 @@ export function renderAgentDetailPage(
         Deleting this agent permanently removes all its data (env vars, crons, tools, tokens, plugins, members)
         and terminates its pod. This action cannot be undone.
       </p>
-      <form method="POST" action="/admin/agents/${escapeHtml(agent.id)}/delete"
-            onsubmit="return confirm('Delete agent ${escapeHtml(agent.name)}? This cannot be undone.')">
+      <form id="delete-agent-form" method="POST" action="/admin/agents/${escapeHtml(agent.id)}/delete"
+            data-agent-name="${escapeHtml(agent.name)}">
         <button type="submit" class="btn btn-danger">Delete agent</button>
       </form>
+      <script>
+        document.getElementById('delete-agent-form').addEventListener('submit', function(e) {
+          var name = this.dataset.agentName;
+          if (!confirm('Delete agent ' + name + '? This cannot be undone.')) {
+            e.preventDefault();
+          }
+        });
+      </script>
     </div>`
         : ""
     }
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -538,6 +538,23 @@ export function renderAgentDetailPage(
 
     ${isAdmin ? membersSection : ""}
 
+    ${
+      isAdmin
+        ? `<!-- Danger Zone -->
+    <div class="card" style="border:1px solid #fca5a5">
+      <div class="card-title" style="color:#dc2626">Danger Zone</div>
+      <p style="font-size:13px;color:#6b7280;margin-bottom:16px">
+        Deleting this agent permanently removes all its data (env vars, crons, tools, tokens, plugins, members)
+        and terminates its pod. This action cannot be undone.
+      </p>
+      <form method="POST" action="/admin/agents/${escapeHtml(agent.id)}/delete"
+            onsubmit="return confirm('Delete agent ${escapeHtml(agent.name)}? This cannot be undone.')">
+        <button type="submit" class="btn btn-danger">Delete agent</button>
+      </form>
+    </div>`
+        : ""
+    }
+
   </div>
 </body>
 </html>`;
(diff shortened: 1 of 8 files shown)


admin/src/admin-ui-pages.ts
```diff
@@ -538,6 +538,23 @@ export function renderAgentDetailPage(
...
+      isAdmin
+        ? `<!-- Danger Zone -->
+    <div class="card" style="border:1px solid #fca5a5">
+      <div class="card-title" style="color:#dc2626">Danger Zone</div>
+      <p style="font-size:13px;color:#6b7280;margin-bottom:16px">
+        Deleting this agent permanently removes all its data (env vars, crons, tools, tokens, plugins, members)
+        and terminates its pod. This action cannot be undone.
+      </p>
+      <form method="POST" action="/admin/agents/${escapeHtml(agent.id)}/delete"
+            onsubmit="return confirm('Delete agent ${escapeHtml(agent.name)}? This cannot be undone.')">
```
dmcaulay: `escapeHtml` escapes `"` but not `'`. An agent name like `O'Brien` breaks this JS string; a crafted name like `foo') || alert(document.cookie) || ('` executes arbitrary JS in any admin's browser — stored XSS. Either escape `'` → `\'` (and `\` → `\\`) for the JS string context, or drop the inline `onsubmit` and use a `data-agent-name` attribute + a small `<script>` listener where `escapeHtml` is sufficient.

dmcaulay (commented):
COMMENT — one security issue before merge.

**Finding 1** (critical): Agent name not JS-escaped in inline `confirm()`. `escapeHtml` (lib/web/toolbar.ts) escapes `&`, `<`, `>`, `"` — but NOT `'`. The danger zone button uses `onsubmit="return confirm('Delete agent ${escapeHtml(agent.name)}?...')"`. A name with a single quote (`O'Brien`) produces malformed JS and silently disables the confirm guard. A crafted name like `foo') || alert(document.cookie) || ('` executes arbitrary JS in any admin's browser when they visit that agent's detail page — stored XSS in the admin surface. Fix: escape `'` → `\'` (and `\` → `\\`) in the JS string context, or move to a `data-agent-name` attribute + script block listener where `escapeHtml` is correct.

Strengths: route double-gated (render + 403 guard), deprovision-then-delete pattern consistent with the REST API, try/catch redirects cleanly on error, tests verify both spy closure IDs not just HTTP status.

_Note: Self-review — verdict recorded as COMMENT per GitHub API constraint, but this is a genuine COMMENT — one security bug needs fixing._

dodizzle (commented):
APPROVE — Critical XSS finding resolved. The confirm dialog now uses `data-agent-name` attribute + event listener pattern: the agent name is HTML-escaped into a safe attribute via `escapeHtml`, then read from `this.dataset.agentName` in the JS event handler — no JS string injection vector. Tests updated to assert the raw XSS payload cannot appear unescaped. CI green.

_Note: Self-review — verdict recorded as COMMENT per GitHub API constraint, but this is a clean APPROVE._

dmcaulay (commented):
APPROVE — XSS finding resolved. The danger zone form now uses `data-agent-name="${escapeHtml(agent.name)}"` (double-quoted HTML attribute, `escapeHtml` correct here) with a script-block event listener reading `this.dataset.agentName` — no JS string interpolation, no injection vector. The `O'Brien` test and non-admin absence test confirm both the fix and access control. No new issues.

_Note: Self-review — verdict recorded as COMMENT per GitHub API constraint, but this is a genuine APPROVE — prior security finding is resolved._

dmcaulay (commented):
APPROVE — XSS finding resolved. The danger zone form now uses `data-agent-name="${escapeHtml(agent.name)}"` (double-quoted HTML attribute, `escapeHtml` correct here) with a script-block event listener reading `this.dataset.agentName` — no JS string interpolation, no injection vector. The `O'Brien` test and non-admin absence test confirm both the fix and access control. No new issues.

_Note: Self-review — verdict recorded as COMMENT per GitHub API constraint, but this is a genuine APPROVE — prior security finding is resolved._

PR #501, opened 2026-06-18
fix(task-store): treat cancelled deps as satisfied; parse array fields in --set

dodizzle:
## Summary

Two bugs in the task-store adapter:

**Bug 1 — `cancelled` dep blocks downstream (`store.ts`)**
The `--ready` check treated `cancelled` as unsatisfied, so any task with a cancelled dependency would never unblock. A cancelled dep means the work is moot — the downstream should proceed. Fix: add `cancelled` to the terminal satisfied-status set alongside `merged`, `done`, `deploying`, `deployed`.

**Bug 2 — `--set dependencies=[]` serialises as a string (`task_store.ts`)**
`coerceValue` had no array-field handling, so `dependencies=[]` was written as the literal string `"[]"`. Fix: add `ARRAY_FIELDS = new Set(["dependencies", "acceptanceCriteria"])` and JSON-parse values for those fields. Widen `nonStatusFields` type to include `string[]`.

## Changes

- `store.ts` — add `cancelled` to satisfied-status block; update JSDoc (point numbering shifts)
- `store.unit.test.ts` — sync local pure helper; add `cancelled` test case
- `task_store.ts` — `ARRAY_FIELDS`, updated `coerceValue` signature + body, widened `nonStatusFields` type
- `task_store.test.ts` — two new coercion tests (`dependencies=[]`, `dependencies=["TS-0.1","TS-0.2"]`)

## Test plan

- [ ] `bun test plugins/shipwright/scripts/store.unit.test.ts` — all pass (includes new `cancelled` case)
- [ ] `bun test plugins/shipwright/scripts/task_store.test.ts` — all pass (includes two new array coercion cases)
- [ ] 139 tests, 0 failures confirmed locally


plugins/shipwright/scripts/store.unit.test.ts
```diff
@@ -386,6 +387,11 @@ describe("dep_satisfied semantics (pure — uses only Task type)", () => {
     expect(depSatisfied(dep, "feat/any")).toBe(true);
   });
 
+  test("dep with status=cancelled is satisfied (downstream unblocks)", () => {
```
dmcaulay, line 390: This tests the local `depSatisfied` helper (defined at line 29 of this test file), not the exported `resolveReadyTasks`. The actual fix to `store.ts` lines 244–253 has no direct test. Consider adding a case to the `resolveReadyTasks` describe block: task with a single dep at `status: "cancelled"` should appear in the ready list.

plugins/shipwright/scripts/task_store.ts
```diff
@@ -41,6 +42,13 @@ function coerceValue(key: string, rawValue: string): boolean | number | string {
     if (rawValue === "true") return true;
     if (rawValue === "false") return false;
   }
+  if (ARRAY_FIELDS.has(key)) {
+    try {
+      return JSON.parse(rawValue) as string[];
+    } catch {
+      return rawValue;
```
dmcaulay: Silent data corruption path: when `JSON.parse` fails (e.g. `--set dependencies=TS-1.1`), this returns `rawValue` (a plain string) which then gets written to a `string[]` field with no error. Also `JSON.parse(rawValue) as string[]` doesn't validate the parsed shape is actually an array — a valid JSON object like `{}` would pass. Consider `if (!Array.isArray(parsed)) throw new Error(...)` after parsing, and add a test for the invalid-JSON case.

dmcaulay (commented):
COMMENT — Two important gaps worth addressing: the primary `resolveReadyTasks` code path isn't directly tested for `cancelled`, and the `coerceValue` fallback can silently corrupt task data. Both are small fixes.

**Finding 1 — `resolveReadyTasks` not tested for `cancelled`**: The new test in `store.unit.test.ts` exercises the local `depSatisfied` helper (a pure function defined within the test file), not the exported `resolveReadyTasks`. The actual bug fix in `store.ts` lines 244–253 has no direct test. Suggest adding a case to the `resolveReadyTasks` describe block: a task whose only dep has `status: "cancelled"` should appear in the ready list.

**Finding 2 — silent data corruption in `coerceValue`**: When an `ARRAY_FIELDS` key receives invalid JSON (`--set dependencies=TS-1.1`), the catch returns `rawValue` (a string), which gets stored into a `string[]` field with no error. Also, `JSON.parse(rawValue) as string[]` doesn't validate the parsed shape is actually an array — a `{}` would pass. Suggest an `Array.isArray` guard after parsing and a test for the invalid-JSON path.

Otherwise solid — the logic is correct, the JSDoc is kept in sync, and the happy-path array tests are thorough.

plugins/shipwright/scripts/task_store.ts
```diff
@@ -41,6 +42,13 @@ function coerceValue(key: string, rawValue: string): boolean | number | string {
     if (rawValue === "true") return true;
     if (rawValue === "false") return false;
   }
+  if (ARRAY_FIELDS.has(key)) {
+    try {
+      return JSON.parse(rawValue) as string[];
+    } catch {
```
dodizzle, line 49: Silent failure: when `JSON.parse` throws, this returns the raw string — the same bug the PR was written to fix, but via the error path. User sees no warning and the string is written to a `string[]` field.

Suggest erroring instead:
```ts
} catch {
  process.stderr.write(`error: --set value for '${key}' must be a valid JSON array\n`);
  process.exit(1);
}
```

plugins/shipwright/scripts/task_store.ts
```diff
@@ -41,6 +42,13 @@ function coerceValue(key: string, rawValue: string): boolean | number | string {
     if (rawValue === "true") return true;
     if (rawValue === "false") return false;
   }
+  if (ARRAY_FIELDS.has(key)) {
+    try {
+      return JSON.parse(rawValue) as string[];
```
dodizzle: `JSON.parse` succeeds for any valid JSON — `42`, `{"a":1}`, etc. Add an array check before returning:
```ts
const parsed = JSON.parse(rawValue);
if (!Array.isArray(parsed)) {
  process.stderr.write(`error: --set value for '${key}' must be a JSON array\n`);
  process.exit(1);
}
return parsed as string[];
```

dodizzle (commented):
COMMENT — Two important issues in the `coerceValue` catch path: (1) bad JSON silently falls back to a string (same class of bug the PR was written to fix), and (2) valid JSON that isn't an array (e.g. `dependencies=42`) passes through unchecked. Both should error-exit with a clear message. Also a minor test gap: the new `cancelled` test exercises the local test-file helper, not `resolveReadyTasks` directly.

dodizzle (commented):
APPROVE — all prior findings addressed. coerceValue now error-exits on invalid/non-array JSON, resolveReadyTasks has a direct cancelled-dep test, and both error paths are covered by new tests. Clean to merge.

dmcaulay (commented):
APPROVE — both prior findings addressed: `resolveReadyTasks` now has a direct cancelled-dep test, and `coerceValue` error-exits on invalid/non-array JSON with four new tests covering both failure modes. Clean to merge.

PR #503, opened 2026-06-18
fix(agent): add channel message events and history scopes to Slack manifest

dodizzle:
## Summary

- Adds `message.channels` and `message.groups` to `bot_events` — without these Slack never delivers non-mention channel messages, so the `app.message` handler never fires
- Adds `channels:history` and `groups:history` to `DEFAULT_BOT_SCOPES` — required for both receiving those events and calling `conversations.replies` to load thread context on first mention

## Root cause

The `app.message` handler in `slack.ts` is wired up correctly and will respond to thread follow-ups once a session exists. But Slack only delivers channel messages to the bot when `message.channels` / `message.groups` are subscribed _and_ the corresponding `*:history` scopes are granted. Without them, the bot only receives `app_mention` events — explaining why a mention is always required.

## Test plan

- [x] Existing 42 manifest integration tests pass
- [x] New tests added for `channels:history`, `groups:history` default scopes and `message.channels`, `message.groups` bot events
- [ ] After merging, existing Slack apps need to be updated via the Slack Manifest API (`apps.manifest.update`) or reinstalled to pick up the new scopes — Slack will prompt for re-authorization on the new permissions


dmcaulay (commented):
APPROVE — clean, correct, and well-tested. Scope-count test is updated appropriately; each new scope and event has its own test. No blocking concerns.

dodizzle (commented):
APPROVE — correct, targeted fix with full test coverage. Adding `message.channels`/`message.groups` bot events and `channels:history`/`groups:history` scopes is the right fix for the non-firing `app.message` handler. Scope-count test correctly updated. No blocking concerns.

*(Self-review: verdict is APPROVE; posted as COMMENT per GitHub API constraint.)*

PR #506, opened 2026-06-18
fix(dev-task): reduce stale in_progress reset threshold to 45 minutes

dodizzle:
## Summary

- Reduces `STALE_THRESHOLD_MS` in `check-dev-task.ts` from 4 hours to 45 minutes
- Sessions time out after 30 minutes, so 4 hours meant stuck tasks sat for up to 4 runs before auto-recovery
- Updates test descriptions and fixed timestamps accordingly (all 15 tests pass)

## Test plan

- [x] `bun test plugins/shipwright/scripts/check-dev-task.test.ts` — 15/15 pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #519, opened 2026-06-18
fix(ci): scope build triggers to module paths instead of paths-ignore

dmcaulay:
## Summary

Replaces the broad `paths-ignore` (from #515) with positive `paths` filters so each build workflow only fires on changes relevant to that module.

| Workflow | Paths |
|---|---|
| `build-admin.yml` | `admin/**`, `lib/**`, root deps |
| `build-metrics.yml` | `metrics/**`, `lib/**`, root deps |
| `build-agent.yml` | `agent/**`, `plugins/**`, `.claude-plugin/**`, root deps |

`plugins/**` does not affect admin or metrics. `lib/**` is shared and affects admin + metrics (both use `COPY . .`). The agent Dockerfile explicitly copies `agent/`, `plugins/`, and `.claude-plugin/` — lib source is not in its build context.

## Why positive paths over paths-ignore

`paths-ignore` only excluded chart bumps and CHANGELOG — any other cross-module change (e.g. editing `admin/` code) would still rebuild all three images unnecessarily. Positive filters are explicit about what triggers each build.

### DIFF
--- a/.github/workflows/build-admin.yml
+++ b/.github/workflows/build-admin.yml
@@ -3,9 +3,12 @@ name: Build and Push Admin Image
 on:
   push:
     branches: [main]
-    paths-ignore:
-      - "charts/**"
-      - "CHANGELOG.md"
+    paths:
+      - "admin/**"
+      - "lib/**"
+      - "package.json"
+      - "bun.lock"
+      - ".github/workflows/build-admin.yml"
 
 permissions:
   contents: write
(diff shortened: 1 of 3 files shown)


dmcaulay (commented):
APPROVE — Path scoping is correct: admin/metrics include lib/** (shared types); agent includes plugins/** and .claude-plugin/** (plugin bundle). The build-agent dispatch step to vitals-os is guarded on CHART_DISPATCH_REPO and correctly fires shipwright-agent-released. The auto-bump-chart.yml token addition prevents push-back failures from GITHUB_TOKEN write limitations. check-dev-task threshold changes are pre-existing from PR #506 and will be no-ops on merge since main already has them.

PR #525, opened 2026-06-18
feat: parameterize k8s namespace + promote RBAC Role to ClusterRole

dmcaulay:
## Summary
- Promote RBAC `Role` + `RoleBinding` to `ClusterRole` + `ClusterRoleBinding` so the provisioner can reach any namespace in the cluster
- Add `agent.provisioning.namespace` Helm value to explicitly override the target namespace for provisioned agent resources (defaults to the pod's own release namespace via downward API)
- Update values schema and documentation to reflect the new field

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | RBAC Role promoted to ClusterRole in the Shipwright Helm chart | MET |
| 2 | Provisioner reads SHIPWRIGHT_K8S_NAMESPACE env var with fallback to release namespace | MET |
| 3 | Helm chart exposes the namespace config value | MET |
| 4 | Existing unit tests pass, new test covers namespace override | MET |
| 5 | TypeScript compiles clean | MET |

## Test Plan
- [x] `helm unittest charts/shipwright` — 195/195 tests pass (17 suites), including the new namespace override test
- [x] `bunx biome lint .` — clean (298 files)
- [x] Admin unit tests pass (63/63 pure unit tests)
- [x] New Helm test: `injects explicit SHIPWRIGHT_K8S_NAMESPACE value when agent.provisioning.namespace is set`

### DIFF
--- a/charts/shipwright/tests/agent_provisioning_rbac_test.yaml
+++ b/charts/shipwright/tests/agent_provisioning_rbac_test.yaml
@@ -182,7 +182,6 @@ tests:
             name: SHIPWRIGHT_ADMIN_DEPLOYMENT_UID
             value: "abc-123-uid"
 
-  # ── (c) disabled (default) → nothing rendered + admin Noop ──────────────────
   - it: injects explicit SHIPWRIGHT_K8S_NAMESPACE value when agent.provisioning.namespace is set
     template: templates/admin-deployment.yaml
     set:
@@ -204,6 +203,7 @@ tests:
               fieldRef:
                 fieldPath: metadata.namespace
 
+  # ── (c) disabled (default) → nothing rendered + admin Noop ──────────────────
   - it: renders no agent-provisioning RBAC by default (provisioning disabled)
     template: templates/agent-provisioning-rbac.yaml
     asserts:
--- a/charts/shipwright/tests/agent_provisioning_rbac_test.yaml
+++ b/charts/shipwright/tests/agent_provisioning_rbac_test.yaml
@@ -1,14 +1,15 @@
-suite: agent-provisioning RBAC + admin env contract (HD-4.2)
+suite: agent-provisioning RBAC + admin env contract (HD-4.2 / K8S-NS-1.1)
 # Asserts the agent-provisioning surface, gated on agent.provisioning.enabled:
-#   (a) enabled → namespace-scoped Role (exact resources + verbs) + RoleBinding to
-#       the admin SA + a separate agent ServiceAccount
+#   (a) enabled → cluster-scoped ClusterRole (exact resources + verbs) + ClusterRoleBinding
+#       to the admin SA + a separate agent ServiceAccount. ClusterRole enables cross-
+#       namespace provisioning (e.g. vitals-os agents with PVCs pinned to their namespace).
 #   (b) enabled → admin Deployment gets SHIPWRIGHT_K8S_PROVISIONING=enabled and the
-#       downward-API namespace env, matching admin/src/main.ts buildProvisioner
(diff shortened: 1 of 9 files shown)


dmcaulay (commented):
COMMENT — one important finding before merging:

**ClusterRole is now unconditional — escalates same-namespace installs unnecessarily** (`templates/agent-provisioning-rbac.yaml`): The Role→ClusterRole promotion applies to all installs with `agent.provisioning.enabled: true`, even when `agent.provisioning.namespace` is empty (same-namespace provisioning, the default). This silently escalates from a namespace-scoped boundary to cluster-wide Deployment/Secret/PVC access for all existing enabled installs. A more least-privilege approach: keep the namespace-scoped Role for the default (`namespace == ""`) case and only promote to ClusterRole when `namespace` is explicitly set. Requires a template conditional but preserves the tight security boundary for single-namespace deployments.

Also worth noting: on upgrade from a prior release with `agent.provisioning.enabled: true`, the old `Role` and `RoleBinding` will be orphaned — Helm won't delete them because the resource kind changed. Callers may want a `kubectl delete role/rolebinding` cleanup step.

Minor: the `values.yaml` comment says cross-namespace provisioning "requires ClusterRole RBAC" but ClusterRole is now unconditional. Update the comment to reflect that.

Strengths: clean conditional env injection in admin-deployment.yaml, well-structured new helm test, schema/values/docs updated consistently, downward API fallback preserved.

charts/shipwright/tests/agent_provisioning_rbac_test.yaml
```diff
@@ -17,7 +18,7 @@ tests:
       name: r
```
dmcaulay, line 22: This test sets `agent.provisioning.enabled: true` without `agent.provisioning.namespace`, so the conditional template (`{{- if .Values.agent.provisioning.namespace }}`) evaluates to false and renders a namespace-scoped `Role` — not a `ClusterRole`. The `isKind: ClusterRole` assertion therefore fails (confirmed by `Helm: failure` in CI). Fix: add `agent.provisioning.namespace: vitals-os` (or any non-empty value) to this test so it hits the ClusterRole branch. Also add two new tests (enabled=true, no namespace) asserting `isKind: Role` and `isKind: RoleBinding` to cover the same-namespace path.

dmcaulay (commented):
COMMENT — prior finding (unconditional ClusterRole) resolved correctly. However the patch introduced a test/template mismatch that breaks helm unittest in CI: the two existing RBAC tests assert ClusterRole/ClusterRoleBinding without setting agent.provisioning.namespace, but the template now renders Role/RoleBinding for empty namespace. Fix the stale test assertions and add Role/RoleBinding coverage before merging.

dmcaulay (commented):
APPROVE — helm unittest coverage fully addresses the test/template mismatch. Existing ClusterRole/ClusterRoleBinding tests updated with `agent.provisioning.namespace: vitals-os`; two new same-namespace Role/RoleBinding tests added for the default (empty namespace) path; a new SHIPWRIGHT_K8S_NAMESPACE injection test for admin-deployment.yaml adds explicit env coverage. All prior findings resolved.

dmcaulay (commented):
APPROVE — CI values variant (`ci/vitals-os-values.yaml`) correctly exercises the cross-namespace ClusterRole path via the declared `agent.provisioning.namespace` field. Template conditional and values.yaml declaration are consistent. Doc and comment updates accurately describe the conditional Role/ClusterRole behavior. Chart bumped to 1.5.11. All CI green. No new issues.

PR #529, opened 2026-06-19
fix(admin): consolidate Slack manifest into buildAgentManifest, fix sync-manifest

dmcaulay:
## Summary

- `sync-manifest` was calling `defaultAgentManifest()` — the provisioning-only manifest with `socket_mode_enabled: false`, no event subscriptions, and no `assistant_view` — overwriting the live agent config and breaking DMs, event delivery, and the Agents/Assistant experience
- `defaultAgentManifest()` was wrong from the start: `socket_mode_enabled` controls event delivery, not OAuth installation — agents should be fully configured from the moment they're provisioned
- Replaces both `defaultAgentManifest()` and the redundant `agent/src/slack-manifest.ts` with a single `buildAgentManifest(appName, redirectUri?)` in `slack-provisioning-client.ts`, used for both provisioning (with the Shipwright OAuth callback URL) and sync-manifest (no redirectUri)
- Adds `mpim:history` scope alongside `im:history` for consistency
- Deletes unused `agent/src/slack-manifest.ts` and its integration test

## Test plan

- [ ] Type-check passes (`bunx tsc --noEmit -p admin/tsconfig.json`)
- [ ] Lint clean (`bunx biome lint`)
- [ ] 480 unit tests pass, 0 fail
- [ ] Manually click "Sync Manifest" on a provisioned agent — verify socket mode, event subscriptions, and Agents/Assistant feature remain enabled afterward


admin/src/slack-provisioning-client.ts
```diff
@@ -96,48 +100,75 @@ export interface SlackProvisioningClient {
...
- * Scopes allow the agent to post messages, read channel history, and send DMs.
+ * Builds the Slack app manifest for a Shipwright agent.
+ *
+ * Used for both initial provisioning (pass redirectUri for the OAuth callback)
+ * and subsequent manifest syncs (omit redirectUri to keep localhost default).
+ * Socket Mode and the full event/assistant config are always applied — there
+ * is no separate "provisioning-only" manifest.
  */
-export function defaultAgentManifest(
+export function buildAgentManifest(
```
dodizzle, line 113: No unit test covers `buildAgentManifest`'s output. The smoke tests verify route behavior (redirects, error cases) but nothing asserts that `socket_mode_enabled === true`, `event_subscriptions.bot_events` includes `message.im`, or `assistant_view` is present — the exact fields that were broken before this fix.

Suggested: add `slack-provisioning-client.unit.test.ts` with assertions on the manifest shape for both the no-arg case (sync) and the with-redirectUri case (provision). Without this, a future regression on manifest content would pass all tests.

admin/src/slack-provisioning-client.ts
```diff
@@ -96,48 +100,75 @@ export interface SlackProvisioningClient {
...
-          "im:read",
           "im:write",
-          "mpim:history",
+          "reactions:read",
+          "reactions:write",
           "users:read",
         ],
       },
-      redirect_urls: [redirectUri],
+      redirect_urls: [redirectUri ?? "http://localhost:3460"],
```
dodizzle: The `redirectUri ?? "http://localhost:3460"` fallback means sync-manifest calls will include `redirect_urls: ["http://localhost:3460"]` in the Slack manifest update, overwriting any production redirect URL the app might have. Slack's `apps.manifest.update` accepts omitting `redirect_urls` to leave the existing value unchanged — consider omitting the field entirely when `redirectUri` is not provided to make sync-only calls non-destructive on that field.

dodizzle (commented):
COMMENT — correct fix for a production-breaking bug (sync-manifest was using the provisioning-only manifest with `socket_mode_enabled: false`). The consolidation to `buildAgentManifest` is clean. One important gap: no unit test asserts the manifest output shape, so the core behavioral fix is untested. See inline comments.

admin/src/slack-provisioning-client.ts
```diff
@@ -96,48 +100,75 @@ export interface SlackProvisioningClient {
...
+// ─── Agent manifest ───────────────────────────────────────────────────────────
 
 /**
- * Default Slack app manifest for a Shipwright agent.
- * Scopes allow the agent to post messages, read channel history, and send DMs.
+ * Builds the Slack app manifest for a Shipwright agent.
+ *
+ * Used for both initial provisioning (pass redirectUri for the OAuth callback)
+ * and subsequent manifest syncs (omit redirectUri to keep localhost default).
+ * Socket Mode and the full event/assistant config are always applied — there
```
dmcaulay, line 110: This function has no unit tests, and CLAUDE.md requires tests in the same PR. The parallel function in `agent/src/slack-manifest.ts` has 30+ tests covering scope correctness, `socket_mode_enabled`, `assistant_view`, and `redirectUri` fallback behavior — that file is the pattern to follow. Suggest adding `admin/src/slack-provisioning-client.unit.test.ts`.

dmcaulay, line 110: `buildAgentManifest` here is a near-duplicate of `buildManifest` in `agent/src/slack-manifest.ts` (same scopes, events, description template, assistant_view shape). Worth extracting to a shared `lib/` utility so future manifest shape changes — new scope, new event, new assistant config — only need to land in one place.

dmcaulay (commented):
COMMENT — The bug fix is correct and critical: sync-manifest was silently overwriting the operational Slack config with the provisioning-only manifest (socket mode off, no events) on every sync. The consolidated `buildAgentManifest` approach is clean. Two issues before merging:

1. **No unit tests** for `buildAgentManifest` — CLAUDE.md requires tests in the same PR, and the parallel `agent/src/slack-manifest.ts` function has 30+ tests showing the expected coverage shape.
2. **Near-duplicate logic** in `admin/` vs `agent/` — both packages now define the same manifest shape independently; any future scope/event change needs two edits.

admin/src/slack-provisioning-client.ts
```diff
@@ -96,48 +100,76 @@ export interface SlackProvisioningClient {
...
+ * is no separate "provisioning-only" manifest.
  */
-export function defaultAgentManifest(
+export function buildAgentManifest(
```
dodizzle, line 113: **Still unaddressed: no unit test for `buildAgentManifest` output shape.**

The `80ada0b` commit deleted `agent/src/slack-manifest.integration.test.ts` (288 tests) without adding any replacement coverage for this function. The deleted tests covered exactly the behavioral properties that matter here: `socket_mode_enabled === true`, `event_subscriptions.bot_events` contains `"message.im"`, `oauth_config.scopes.bot` includes `"assistant:write"`, and `features.assistant_view` is present.

Add `admin/src/slack-provisioning-client.unit.test.ts` with at minimum:
```ts
const m = buildAgentManifest("MyAgent");
expect(m.settings.socket_mode_enabled).toBe(true);
expect(m.settings.event_subscriptions.bot_events).toContain("message.im");
expect(m.oauth_config.scopes.bot).toContain("assistant:write");
expect(m.features.assistant_view).toBeDefined();
// redirectUri path:
const m2 = buildAgentManifest("MyAgent", "https://example.com/cb");
expect(m2.oauth_config.redirect_urls[0]).toBe("https://example.com/cb");
```

dodizzle (commented):
COMMENT — prior finding still open: `buildAgentManifest` (the production fix for the socket-mode-killing bug) has zero behavioral test coverage. The `80ada0b` commit deleted the 288-line `agent/src/slack-manifest.integration.test.ts` that proved the manifest shape was correct, without adding any replacement tests for the new canonical builder. `mpim:history` addition in `6710fe1` is clean and consistent. No new issues. Holding COMMENT until a `slack-provisioning-client.unit.test.ts` lands.

dmcaulay (commented):
APPROVE — both prior findings addressed:

1. ✅ **Unit tests added** — `slack-provisioning-client.unit.test.ts` covers socket_mode, events, assistant_view, scopes, and redirect_urls behavior for both sync and provisioning paths.
2. ✅ **Duplicate removed** — `agent/src/slack-manifest.ts` deleted (was never imported by production callers).

The `redirect_urls` omission on sync (via conditional spread) is the right call — Slack's manifest update API leaves existing values intact when the field is absent. `mpim:history` restored. Clean.

dodizzle (approved):
APPROVE — both prior findings fully addressed. `slack-provisioning-client.unit.test.ts` now covers socket_mode_enabled, all 7 bot events, assistant_view, required scopes, and redirect_urls behavior (absent for sync, set for provisioning). The conditional spread correctly omits `redirect_urls` when no `redirectUri` is provided, making sync-manifest non-destructive on the live redirect config. CI green. Clean to merge.

PR #532, opened 2026-06-19
fix(agent): skip @mention in threads where bot already has a session

dmcaulay:
## Summary

- When a user @mentions the agent in an active thread, both `app_mention` and `message` events fire — causing two responses
- Fix: add an early return in the `app_mention` handler if `ev.thread_ts` exists and `getSessionFn(sessionKey)` returns a session
- The `message` handler already has the same awareness (line 377-378) — this makes `app_mention` consistent with it

## Test plan

- [ ] 136 existing Slack integration tests pass unchanged
- [ ] Manual: @mention bot in an active thread → single response
- [ ] Manual: @mention bot to start a new thread → still responds (no session yet)
- [ ] Manual: @mention bot in a DM → still responds (`ev.thread_ts` is undefined in DMs)

### DIFF
--- a/agent/src/slack.ts
+++ b/agent/src/slack.ts
@@ -500,6 +500,12 @@ export function createSlackApp(
     const sessionKey = getThreadKey(ev.channel, ev.thread_ts ?? ev.ts);
     const replyTs = ev.thread_ts ?? ev.ts;
 
+    // Drop @mention if bot is already participating in this thread —
+    // the message handler covers it and would double-respond otherwise.
+    if (ev.thread_ts && getSessionFn(sessionKey)) {
+      return;
+    }
+
     const setStatus = async (status: string) => {
       // biome-ignore lint/suspicious/noExplicitAny: Bolt types don't include Agents API yet
       await (client.assistant.threads as any)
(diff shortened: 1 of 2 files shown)


agent/src/slack.ts
```diff
@@ -500,6 +500,12 @@ export function createSlackApp(
     const sessionKey = getThreadKey(ev.channel, ev.thread_ts ?? ev.ts);
     const replyTs = ev.thread_ts ?? ev.ts;
 
+    // Drop @mention if bot is already participating in this thread —
+    // the message handler covers it and would double-respond otherwise.
+    if (ev.thread_ts && getSessionFn(sessionKey)) {
```
dmcaulay, line 505: The fix is correct, but no test covers this new early-return path. The pre-existing test ('subsequent @mention when session exists: no history fetch') only asserts `mockRepliesFn` was not called — that was true before this PR too. CLAUDE.md requires tests to land in the same PR. Add a test: simulate an `app_mention` event with `thread_ts` set and `getSessionFn` returning a session, then assert `mockRunClaude` was not called.

dmcaulay (commented):
COMMENT — Fix is correct and the logic handles all edge cases well. One issue: no test covers the new early-return contract. CLAUDE.md requires tests to ship in the same PR. A test asserting `mockRunClaude` is not called when `thread_ts` is set and a session exists would close this gap.

dmcaulay (commented):
APPROVE — prior finding addressed. The new test directly covers the early-return contract: `app_mention` with `thread_ts` set and an existing session → `mockRunClaude` not called. Logic is correct, CI is green, test coverage is complete.

PR #552, opened 2026-06-19
feat: add reinstall-after-sync flow (SMR-1.1)

dmcaulay:
## Summary
- Extends `sync-manifest` route to trigger a full Slack OAuth reinstall after updating the manifest, reusing the existing `provision/complete` flow
- `provision/start` now persists `SLACK_CLIENT_ID` and `SLACK_CLIENT_SECRET` to the agent env alongside `SLACK_APP_ID` so they're available for the reinstall path
- `provision/complete` detects reinstalls (when `SLACK_APP_TOKEN` is already set) and redirects to `?success=reinstalled` instead of rendering the xapp-token page

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | provision/start upserts SLACK_CLIENT_ID and SLACK_CLIENT_SECRET alongside SLACK_APP_ID | MET |
| 2 | Smoke: sync-manifest with creds in env → 302 to slack.com/oauth/v2/authorize + provision cookie | MET |
| 3 | Smoke: sync-manifest without SLACK_CLIENT_ID (legacy agent) → 302 to ?success=manifest_synced | MET |
| 4 | Smoke: provision/complete with SLACK_APP_TOKEN in env → 302 to ?success=reinstalled | MET |
| 5 | Smoke tests only; existing provision/complete tests remain green | MET |

## Test Plan
- [x] 75 smoke tests pass across `admin-ui.smoke.test.ts` and `provision-callback.smoke.test.ts`
- [x] Biome lint clean
- [x] All 4 packages typecheck clean

### DIFF
--- a/admin/src/admin-ui.smoke.test.ts
+++ b/admin/src/admin-ui.smoke.test.ts
@@ -1179,6 +1179,8 @@ describe("admin UI — provision start form", () => {
     expect(upsertedAgentId).toBe(AGENT_ID);
     expect(envVars.SLACK_APP_ID).toBe("A_HAPPY");
     expect(envVars.SLACK_SIGNING_SECRET).toBe("ssec_happy");
+    expect(envVars.SLACK_CLIENT_ID).toBe("cid_happy");
+    expect(envVars.SLACK_CLIENT_SECRET).toBe("csec_happy");
     expect(envVars.GH_TOKEN).toBe("ghp_my_token");
   });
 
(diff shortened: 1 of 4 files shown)


dodizzle (commented):
**COMMENT** — One finding: the bot scope list at `admin/src/admin-ui.ts:974` is a hardcoded duplicate of the scopes defined in `buildAgentManifest`. If any scope is added or removed from the manifest, the reinstall OAuth URL diverges silently — the bot token will be issued for a different scope set. Suggest exporting the scope list from slack-provisioning-client.ts and importing it here.

Everything else is clean: the signed JWT cookie CSRF pattern is consistent with the fresh-provision flow, all 5 acceptance criteria are tested, and legacy agents fall through unchanged.

admin/src/admin-ui.ts
```diff
@@ -932,14 +940,69 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
...
+        httpOnly: true,
+        maxAge: PROVISION_STATE_TTL_SECONDS,
+        sameSite: "Lax",
+        path: "/",
+        secure: appBaseUrl.startsWith("https://"),
+      });
+
+      // Build the Slack OAuth v2 authorize URL
+      const scopes = [
+        "app_mentions:read",
```
dodizzle: These scopes are a hardcoded duplicate of the bot scope list in `buildAgentManifest`. If a scope is added or removed from the manifest, the reinstall OAuth URL will diverge silently — the reinstalled bot token will have a different scope set than what the app declares.

Suggest exporting the canonical list from slack-provisioning-client.ts:
```ts
export const AGENT_BOT_SCOPES: string[] = [
  "app_mentions:read",
  // ...
];
```
Then import and use it here instead of repeating the array.

admin/src/admin-ui.ts
```diff
@@ -932,14 +940,69 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
...
+      setCookie(c, PROVISION_STATE_COOKIE, provisionToken, {
+        httpOnly: true,
+        maxAge: PROVISION_STATE_TTL_SECONDS,
+        sameSite: "Lax",
+        path: "/",
+        secure: appBaseUrl.startsWith("https://"),
+      });
+
+      // Build the Slack OAuth v2 authorize URL
+      const scopes = [
```
dmcaulay: These 14 bot scopes are a verbatim copy of the scope list in `buildAgentManifest()` in `slack-provisioning-client.ts`. In the fresh-provision path, scopes come from the manifest itself; here they're hardcoded. If a scope is added to the manifest later, reinstalls via sync-manifest will silently grant a different permission set than fresh installs — no CI signal, no test failure.

Consider exporting a `BOT_SCOPES` constant from `slack-provisioning-client.ts` and importing it here to keep both paths in sync.

admin/src/admin-ui.smoke.test.ts
```diff
@@ -1803,4 +1805,72 @@ describe("admin UI — manifest sync route", () => {
...
+        Cookie: `admin_session=${adminCookie}`,
+      },
+    });
+    expect(res.status).toBe(302);
+    const location = res.headers.get("location") ?? "";
+    expect(location).toContain("slack.com/oauth/v2/authorize");
+    expect(location).toContain("client_id=my-client-id");
+    // Provision state cookie should be set
+    const setCookieHeader = res.headers.get("Set-Cookie") ?? "";
+    expect(setCookieHeader).toContain("slack_provision_state=");
```
dmcaulay, line 1843: This only asserts the cookie is present, not that the JWT payload is correct. The analogous provision/start test (around line 1158) decodes the JWT and asserts `payload.agentId` and `payload.appId`. A bug encoding the wrong `agentId` or `clientId` into the sync-manifest reinstall cookie wouldn't be caught here — `makeProvisionStateCookie()` in `provision-callback.smoke.test.ts` constructs the cookie independently.

Consider decoding the JWT (`JSON.parse(Buffer.from(token.split('.')[1], 'base64url').toString())`) and asserting `payload.agentId === AGENT_ID` and `payload.clientId === 'my-client-id'`.

dmcaulay (commented):
COMMENT — 2 important findings: (1) hardcoded scope list in sync-manifest reinstall path will silently drift from buildAgentManifest if scopes change; (2) new reinstall smoke test doesn't verify JWT payload in the provision cookie. Both are straightforward fixes.

dmcaulay (commented):
APPROVE — Both prior findings are resolved: `AGENT_BOT_SCOPES` is now the single canonical scope list (imported from `slack-provisioning-client.ts` in both `buildAgentManifest` and the reinstall OAuth URL), and the reinstall smoke test now verifies `agentId` and `clientId` in the decoded JWT payload. One suggestion: the new JWT decode block skips `decodeURIComponent` before splitting (unlike the existing provision/start test at line 1165) — latent fragility but no current impact. Clean merge.

PR #554, opened 2026-06-19
fix(admin): pass redirectUri in sync-manifest handler

dmcaulay:
## Summary

- `sync-manifest` was calling `buildAgentManifest(agent.name)` without a `redirectUri`, which omits `redirect_urls` from the manifest update — reverting the Slack app's configured redirect URI to the localhost default
- When the reinstall OAuth flow then sends `https://shipwright.vitals-os.com/admin/provision/complete` as the redirect, Slack rejects it with a `redirect_uri did not match` error
- Fix: derive `redirectUri` from `appBaseUrl` (already in scope) and pass it to `buildAgentManifest`, matching what the provisioning flow already does at line 1092-1093

## Test plan

- [ ] All 194 admin-ui tests pass (run `bun test admin/src/admin-ui`)
- [ ] All 13 slack-provisioning tests pass (run `bun test admin/src/slack-provisioning`)
- [ ] Sync manifest on an already-provisioned agent and verify the reinstall OAuth URL redirects to `/admin/provision/complete` successfully

### DIFF
--- a/admin/src/admin-ui.ts
+++ b/admin/src/admin-ui.ts
@@ -34,11 +34,11 @@ import {
 import type { AgentCronJobService } from "./agent-cron-jobs.ts";
 import type { AgentEnvService } from "./agent-envs.ts";
 import type { AgentPluginService } from "./agent-plugins.ts";
+import type { AgentProvisioner } from "./agent-provisioner.ts";
 import type { AgentTokenService } from "./agent-tokens.ts";
 import type { AgentToolService } from "./agent-tools.ts";
 import { ForbiddenError, UnprocessableEntityError } from "./errors.ts";
 import type { GoogleAuthClient } from "./google-auth-client.ts";
-import type { AgentProvisioner } from "./agent-provisioner.ts";
 import type { AppManifest } from "./slack-provisioning-client.ts";
 import {
   AGENT_BOT_SCOPES,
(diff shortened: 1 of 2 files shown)


dodizzle (commented):
**COMMENT** — The functional fix is correct: deriving `redirectUri` from `appBaseUrl` and passing it to `buildAgentManifest` matches the provisioning flow and resolves the `redirect_uri did not match` error.

One item to address before merge:

**`admin/src/slack-provisioning-client.unit.test.ts:13`** — The describe block is labeled `"sync-manifest (no redirectUri)"` and the test at line 51 says `"omits redirect_urls when no redirectUri provided"`. Before this PR that label accurately described sync-manifest's behavior; after this fix, sync-manifest always passes a redirectUri, making the label misleading. Suggest renaming to `"buildAgentManifest without redirectUri (unit)"` so the test's scope is clear without implying sync-manifest omits the redirectUri.

dmcaulay (commented):
APPROVE — Fix correctly passes `redirectUri` to `buildAgentManifest` in the sync-manifest handler, aligning it with the provisioning flow and resolving the Slack `redirect_uri did not match` error. Formatter-only changes in the rest of the diff. Test describe-block rename addressed the prior review finding. No blocking issues.

dodizzle (approved):
**APPROVE** — Prior finding addressed: describe block renamed from `"sync-manifest (no redirectUri)"` to `"buildAgentManifest without redirectUri (unit)"`, which correctly scopes the test to the function in isolation rather than implying sync-manifest omits redirectUri. Functional fix is correct, CI is green. Ready to merge.

PR #557, opened 2026-06-20
fix(plugin): use version-aware resolution for plugin script lookup

dmcaulay:
## Summary

- `sort -V` on full paths sorted by publisher prefix (`vitals-os` > `shipwright` alphabetically), causing an older plugin version (4.26.8) to win over a newer one (4.28.0) when multiple marketplace publishers are installed
- Fix: pipe through `awk -F/ '{print $(NF-2), $0}'` to prepend the version segment before sorting, so the highest version always wins regardless of publisher
- Applied to all 22 occurrences across 10 command/skill files (`task_store.ts` × 21, `check-patch.ts` × 1)

## Test plan

- [ ] On an agent with both `vitals-os/shipwright` and `shipwright/shipwright` installed, `PLUGIN_SCRIPTS` resolves to the highest version from either publisher
- [ ] `task_store.ts append` with `acceptanceCriteria` stores a proper array (not a stringified array)
- [ ] All existing command flows (dev-task, review, deploy, patch, plan-session) unaffected

### DIFF
--- a/plugins/shipwright/.claude-plugin/plugin.json
+++ b/plugins/shipwright/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "shipwright",
-  "version": "4.28.1",
+  "version": "4.28.2",
   "description": "Shipwright — conversational planning, queue-driven execution, policy-controlled code review, a five-phase test-readiness pipeline, and autonomous deploy (merge → canary → promote) for any software project",
   "author": {
     "name": "App Vitals",
--- a/plugins/shipwright/README.md
+++ b/plugins/shipwright/README.md
@@ -1,4 +1,4 @@
-# Shipwright v4.28.1
+# Shipwright v4.28.2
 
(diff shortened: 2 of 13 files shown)


dodizzle (approved):
APPROVE — Correct fix for the publisher-prefix sort ordering bug. The `awk -F/ '{print $(NF-2), $0}'` approach extracts the version segment before sorting, ensuring the highest version wins regardless of publisher name. All 22 occurrences updated consistently, CI green, no blocking concerns.

dmcaulay (commented):
APPROVE — Fix is correct, complete, and directly matches the documented root cause (LEARNINGS.md: `sort -V` on full paths sorts by publisher prefix, not version). The `awk -F/ '{print $(NF-2), $0}'` approach is position-independent and applied consistently across all 22 occurrences. No regressions. The `-new` typo fix in `plan-session.md` is a bonus. No blocking concerns.

PR #560, opened 2026-06-20
fix(plugin): validate branch field and document required task fields

dmcaulay:
## Summary

- **`dev-task.md`**: After picking a task in Step 1, validate that `branch` is set. If missing, post a clear error with the fix command and stop — prevents the silent worktree failure that caused tasks to stay `pending` indefinitely while the cron kept silently completing.
- **`task-store/SKILL.md`**: Document the required fields for `append` in a table, with `branch` called out explicitly (and the convention for generating its value).

## Background

A task created without a `branch` field caused `dev-task` to attempt `git worktree add ... -b ""`, failing silently. The cron's preCheck kept finding the task (correct), but the session always exited `[silent]` without progressing the task. This adds an early-exit guard with a clear, actionable error message.

## Test plan
- [ ] Create a task without `branch` and run `/shipwright:dev-task` — should post the error message and stop
- [ ] Create a task with `branch` set — should proceed normally through worktree creation

### DIFF
--- a/plugins/shipwright/commands/dev-task.md
+++ b/plugins/shipwright/commands/dev-task.md
@@ -44,6 +44,16 @@ The command returns `pending` shipwright tasks whose dependencies are all satisf
 
 If the output is an empty JSON array (`[]`), respond `[silent]` and stop.
 
+**Validate required fields.** If the selected task has no `branch` field (null, undefined, or empty string), do not proceed. Post:
+
+```
+⚠ Task {id} has no branch field set. Set it with:
+  bun "$PLUGIN_SCRIPTS/task_store.ts" update --id {id} --set branch=feat/{id-lowercase}
+Then re-run /shipwright:dev-task.
+```
+
+Respond `[silent]` and stop. A missing branch cannot be recovered from at runtime — worktree creation will fail silently if attempted.
+
 Print:
 ```
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
--- a/plugins/shipwright/commands/dev-task.md
+++ b/plugins/shipwright/commands/dev-task.md
@@ -52,7 +52,13 @@ If the output is an empty JSON array (`[]`), respond `[silent]` and stop.
 Then re-run /shipwright:dev-task.
 ```
 
-Respond `[silent]` and stop. A missing branch cannot be recovered from at runtime — worktree creation will fail silently if attempted.
+Before stopping, mark the task blocked so the cron does not keep re-queuing it:
+
+```bash
+bun "$PLUGIN_SCRIPTS/task_store.ts" update --id {id} --set status=blocked
+```
+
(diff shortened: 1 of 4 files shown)


dodizzle (approved):
APPROVE — correct fix for the silent cron spin cycle caused by tasks with no `branch` field.

The early-exit guard is well-placed and the error message is actionable (exact `task_store.ts update` command with the conventional naming pattern). `[silent]` on validation failure is the right choice — config errors shouldn't spam Slack.

One suggestion: in `task-store/SKILL.md`, `assignee` is grouped with fields that cause silent failures when absent, but a missing `assignee` is often intentional (any agent picks up the task). Worth separating "required" from "recommended" in a follow-up, but not a blocker.

plugins/shipwright/commands/dev-task.md
```diff
@@ -44,6 +44,16 @@ The command returns `pending` shipwright tasks whose dependencies are all satisf
...
 
+**Validate required fields.** If the selected task has no `branch` field (null, undefined, or empty string), do not proceed. Post:
+
+```
+⚠ Task {id} has no branch field set. Set it with:
+  bun "$PLUGIN_SCRIPTS/task_store.ts" update --id {id} --set branch=feat/{id-lowercase}
+Then re-run /shipwright:dev-task.
+```
+
+Respond `[silent]` and stop. A missing branch cannot be recovered from at runtime — worktree creation will fail silently if attempted.
```
dmcaulay: **`[silent]` suppresses the warning this block is trying to deliver.** In the harness, `[silent]` suppresses the entire outgoing reply — so the operator never sees the warning message. The task stays `pending` (not `blocked`), and the cron picks it up on the next cycle and silently exits the same way — the exact silent retry loop the PR is fixing, just triggered here instead of at worktree creation.

Two fixes needed:
1. Replace `Respond [silent] and stop.` with `Post the message above and stop.` so the warning is delivered.
2. Before stopping, mark the task blocked so the cron does not keep re-queuing it:
   ```
   bun "$PLUGIN_SCRIPTS/task_store.ts" update --id {id} --set status=blocked
   ```

dmcaulay (commented):
COMMENT — one critical issue: the validation guard introduces a `Post:` + `[silent]` contradiction that swallows the operator warning and leaves the task stuck as `pending` indefinitely — the same silent-loop failure mode the PR is fixing. The SKILL.md required-fields table is clean and accurate.

dodizzle (approved):
APPROVE — dmcaulay's critical finding fully resolved.

**Prior findings resolution:**
- ✅ `[silent]` contradiction fixed: the validation block now posts the warning visibly and marks the task `blocked` before stopping, preventing the cron re-queue loop the PR is fixing.

Ordering is correct — the `status=blocked` update runs before the post, so the task is protected even if the post call fails. No new issues. CI green.

dmcaulay (commented):
APPROVE — prior finding fully resolved: the validation guard now posts the branch warning visibly and marks the task blocked before stopping, eliminating the silent retry loop. No new issues.

PR #561, opened 2026-06-20
feat(admin): add PVC name template support to K8s provisioner

dmcaulay:
## Summary

- Adds `SHIPWRIGHT_AGENT_PVC_NAME_TEMPLATE` env var to the admin service — when set, the K8s provisioner substitutes `{name}` with the agent's human-readable slug rather than its sanitized agent ID when naming PVCs
- Threads `opts.slug` through `provision(agentId, { slug })` → `pvcNameFor(resourceName, slug)` → `config.pvcName(resourceName, slug)` so the template has both the resource name and the slug available
- Passes `agent.name` as `opts.slug` when `agents-api.ts` calls `provisioner.provision()`, and wires the template into `buildProvisioner()` in `main.ts`
- Exposes `agent.provisioning.pvcNameTemplate` in the Helm chart and flows it to `SHIPWRIGHT_AGENT_PVC_NAME_TEMPLATE` on the admin Deployment

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Default behaviour unchanged: no template set → PVC named `{sanitizedAgentId}-home` | ✅ MET |
| 2 | With template `vitals-os-agent-{name}-home` set, provisioning okwow references `vitals-os-agent-okwow-home` | ✅ MET |
| 3 | Reconcile respects the template when re-provisioning missing Deployments | ✅ MET |
| 4 | Unit tests cover both default and templated PVC name paths | ✅ MET |

## Test Plan

- [x] 5 new unit tests in `admin/src/agent-provisioner.unit.test.ts` — all passing
- [x] Biome check clean on all changed files
- [x] Default PVC naming unchanged (no template set)
- [x] Templated PVC naming with slug (`vitals-os-agent-okwow-home`)
- [x] Reconcile uses template (via `provision()` path)

### DIFF
--- a/admin/src/agent-provisioner.ts
+++ b/admin/src/agent-provisioner.ts
@@ -284,6 +284,15 @@ export class KubernetesAgentProvisioner implements AgentProvisioner {
     for (const [resourceName, agent] of expectedNames) {
       if (!k8sNameSet.has(resourceName)) {
         // provision() is idempotent — it handles ConflictError on Secret/Deployment.
+        if (this.config.pvcName && !agent.slug) {
+          // When a pvcName template is active, the PVC name is derived from the slug.
+          // If slug is absent, reconcile falls back to resourceName — which may not
+          // match the slug-based PVC created by the original provision() call, risking
+          // a mount to an empty volume. Callers should populate slug in the agents array.
+          console.warn(
+            `[reconcile] pvcName template is set but agent ${agent.id} has no slug; PVC will be named from resourceName (${resourceName}) — verify this matches the PVC created at provision time`,
+          );
+        }
         try {
           await this.provision(agent.id, { slug: agent.slug });
           recreated.push(agent.id);
--- a/admin/src/agent-provisioner.ts
+++ b/admin/src/agent-provisioner.ts
@@ -71,7 +71,10 @@ export interface ReconcileResult {
 
 export interface AgentProvisioner {
   /** Mint a token and create the agent's Secret + Deployment. Idempotent. */
-  provision(agentId: string): Promise<ProvisionResult>;
+  provision(
+    agentId: string,
+    opts?: { slug?: string },
+  ): Promise<ProvisionResult>;
   /** Delete the agent's Deployment + Secret. Tolerates already-absent. */
   deprovision(agentId: string): Promise<void>;
   /**
@@ -98,10 +101,13 @@ export interface KubernetesAgentProvisionerConfig {
   /** Admin Deployment uid — ownerReference target for GC. */
   adminDeploymentUid: string;
   /**
-   * Build the PVC name for an agent from its sanitized resource name.
-   * Defaults to `<name>-home`.
+   * Build the PVC name for an agent from its sanitized resource name and
+   * optional human-readable slug (e.g. the agent's name). When a slug is
+   * provided (via `provision(agentId, { slug })`), it is passed as the second
+   * argument so callers can substitute `{name}` in a template. Defaults to
+   * `<resourceName>-home`.
    */
-  pvcName?: (resourceName: string) => string;
+  pvcName?: (resourceName: string, slug?: string) => string;
   /**
    * Build the Secret name from the sanitized resource name.
    * Defaults to `<name>-token`.
@@ -152,16 +158,19 @@ export class KubernetesAgentProvisioner implements AgentProvisioner {
       : `${resourceName}-token`;
   }
 
-  private pvcNameFor(resourceName: string): string {
+  private pvcNameFor(resourceName: string, slug?: string): string {
     return this.config.pvcName
-      ? this.config.pvcName(resourceName)
+      ? this.config.pvcName(resourceName, slug)
       : `${resourceName}-home`;
   }
 
-  async provision(agentId: string): Promise<ProvisionResult> {
+  async provision(
+    agentId: string,
+    opts?: { slug?: string },
+  ): Promise<ProvisionResult> {
     const resourceName = this.resourceName(agentId);
     const secretName = this.secretNameFor(resourceName);
-    const pvcName = this.pvcNameFor(resourceName);
+    const pvcName = this.pvcNameFor(resourceName, opts?.slug);
     const result: ProvisionResult = {
       resourceName,
       secretName,
@@ -172,10 +181,7 @@ export class KubernetesAgentProvisioner implements AgentProvisioner {
     //    PVC already exists from a prior provision; treat as idempotent success.
     //    On any subsequent failure, do NOT delete the PVC (data safety policy).
     try {
-      await this.k8s.createPvc(
-        this.config.namespace,
-        this.pvcSpec(pvcName),
-      );
+      await this.k8s.createPvc(this.config.namespace, this.pvcSpec(pvcName));
     } catch (err) {
       if (!isConflict(err)) throw err;
     }
@@ -218,7 +224,7 @@ export class KubernetesAgentProvisioner implements AgentProvisioner {
     try {
       await this.k8s.createDeployment(
         this.config.namespace,
-        this.deploymentSpec(agentId, resourceName, secretName),
+        this.deploymentSpec(agentId, resourceName, secretName, opts?.slug),
       );
     } catch (err) {
       if (isConflict(err)) {
@@ -334,6 +340,7 @@ export class KubernetesAgentProvisioner implements AgentProvisioner {
     agentId: string,
     resourceName: string,
     secretName: string,
+    slug?: string,
   ): DeploymentSpec {
     const manifest = buildAgentDeploymentManifest({
       agentId,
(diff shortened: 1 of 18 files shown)


dodizzle (commented):
COMMENT — 1 critical issue: `values.schema.json` is not updated.

**`pvcNameTemplate` missing from `values.schema.json`** — this is what's breaking Helm lint. `values.yaml` adds `agent.provisioning.pvcNameTemplate: ""` but the `provisioning` block in the schema has `additionalProperties: false`, so the new field is rejected:

```
[ERROR] values.yaml: - at '/agent': 'allOf' failed
  - at '/agent/provisioning': additional properties 'pvcNameTemplate' not allowed
```

Fix: add `"pvcNameTemplate": { "type": "string" }` to the `provisioning.properties` block in `charts/shipwright/values.schema.json` after `adminDeploymentUid`.

The code logic, tests, and documentation are all correct. One-line schema fix unblocks the merge.

dodizzle (commented):
**Re-review of `7a8fc5f6`** — prior finding addressed ✓, 2 new issues.

**Prior finding**: ✅ `pvcNameTemplate` added to `values.schema.json` — Helm lint now passes.

---

### 🔴 Critical: Chart.yaml version not bumped — `ct lint` CI gate fails

Chart version is still `1.5.23` but `main` is at `1.5.25` (two releases landed since this branch was cut). The `ct lint --check-version-increment` gate is failing:
```
✖︎ shipwright => (version: "1.5.23") > chart version not ok. Needs a version bump!
```
Fix: bump `charts/shipwright/Chart.yaml` → `version: 1.5.26` and update the `artifacthub.io/changes` annotation.

---

### 🟡 Important: `reconcile()` provisions without slug — PVC name mismatch risk

`reconcile()` calls `this.provision(agentId)` with no `opts`, so `slug` is always `undefined` when re-provisioning a missing Deployment. With an active template (e.g. `vitals-os-agent-{name}-home`), reconcile creates a PVC named after the sanitized agentId rather than the slug. If the slug-named PVC from the original provision already exists, the agent binds to a new empty volume — silent data loss for the agent home directory.

Options:
- Add a `slugForAgent?: (agentId: string) => Promise<string | undefined>` callback to `KubernetesAgentProvisionerConfig` so `reconcile()` can resolve the slug
- Or document this limitation explicitly with a `console.warn` when `pvcNameTemplate` is set and reconcile fires

---

Everything else is solid — the schema fix is confirmed, slug threading is consistent through `pvcNameFor`/`deploymentSpec`, and the 5-test suite covers all acceptance criteria cleanly.

dodizzle (commented):
**Re-review of `0dc95f8`** — prior findings addressed ✓, 1 new issue.

**Prior findings**:
- ✅ Chart version bumped to 1.5.26 — ct lint passing
- ✅ `reconcile()` slug mismatch — `agents-api.ts` now passes `agent.name` as slug; `console.warn` guards external callers

---

### 🟡 Important: New slug integration tests ignore the slug argument

All 4 new `pvcName` callbacks in the integration tests use `(name) => \`vitals-os-agent-${name}-home\`` — reading only the first arg (`resourceName`, UUID-derived) and ignoring the slug (second arg). Tests that assert slug-based PVC names like `vitals-os-agent-okwow-home` will fail against a real DB because the callback never produces that name. CI stays green only because `DATABASE_URL_ADMIN_TEST` is unset and `describeOrSkip` skips the whole block.

The production `main.ts` wiring is correct — fix the test callbacks to use the second arg:
```ts
pvcName: (_resourceName, slug) => `vitals-os-agent-${slug ?? _resourceName}-home`
```

Everything else is solid. Merge after this test fix.

admin/src/agent-provisioner.ts
```diff
@@ -153,13 +159,15 @@ export class KubernetesAgentProvisioner implements AgentProvisioner {
   }
 
   private pvcNameFor(resourceName: string, slug?: string): string {
-    if (this.config.pvcName) {
-      return this.config.pvcName(slug ? sanitizeAgentName(slug) : resourceName);
-    }
-    return `${resourceName}-home`;
+    return this.config.pvcName
+      ? this.config.pvcName(resourceName, slug)
```
dmcaulay: This passes `resourceName` (sanitized agentId) as the first arg and `slug` as the second. Every `pvcName` callback uses only one arg — `slug` is silently discarded. The original implementation (`this.config.pvcName(slug ? sanitizeAgentName(slug) : resourceName)`) correctly passed a single pre-resolved value. Revert to that pattern:
```ts
return this.config.pvcName
  ? this.config.pvcName(slug ? sanitizeAgentName(slug) : resourceName)
  : `${resourceName}-home`;
```
And revert `pvcName?` back to `(resourceName: string) => string` — the second-arg approach is a leaky abstraction.

admin/src/main.ts
```diff
@@ -147,8 +147,15 @@ function buildProvisioner(
...
     ...(Object.keys(voice).length > 0 ? { voice } : {}),
+    // When SHIPWRIGHT_AGENT_PVC_NAME_TEMPLATE is set (e.g. "vitals-os-agent-{name}-home"),
+    // substitute {name} with the slug (agent's human-readable name) when provided, or
+    // fall back to the sanitized resource name (agentId-derived). When unset, the
+    // provisioner uses the default `${resourceName}-home` naming.
     ...(pvcNameTemplate
-      ? { pvcName: (name: string) => pvcNameTemplate.replace("{name}", name) }
+      ? {
+          pvcName: (resourceName: string, slug?: string) =>
+            pvcNameTemplate.replace("{name}", slug ?? resourceName),
```
dmcaulay: `slug` is a raw agent name from the DB and can contain spaces, uppercase, etc. K8s PVC names must be RFC1123 (lowercase alphanumeric + hyphens, ≤63 chars). Use `sanitizeAgentName(slug)` here:
```ts
pvcNameTemplate.replace("{name}", slug ? sanitizeAgentName(slug) : resourceName)
```
Import `sanitizeAgentName` from `./agent-manifest.ts`. (This also becomes a no-op once `pvcNameFor` is fixed to a single-arg pattern — the fix belongs in `pvcNameFor`, not in `main.ts`.)

dmcaulay (commented):
COMMENT — 2 critical bugs before merge.

**Prior findings**: Both addressed ✓ — `pvcNameTemplate` added to `values.schema.json`, chart bumped to `1.5.26`.

---

### 🔴 Critical: `pvcNameFor` passes slug as second arg — it's never used by any callback

`pvcNameFor()` calls `this.config.pvcName(resourceName, slug)`. Every callback in the codebase — including all integration tests and `main.ts` — uses only the first arg: `(name) => pvcNameTemplate.replace('{name}', name)`. The slug is in position 2 and is discarded. The integration test at line 323 provisions with slug `"okwow"` expecting `vitals-os-agent-okwow-home`, but the callback receives `resourceName` (sanitized agentId), not `"okwow"` — that test would fail.

Fix: pass `slug ?? resourceName` as a single first argument and keep the config type as `(resourceName: string) => string`:
```ts
private pvcNameFor(resourceName: string, slug?: string): string {
  return this.config.pvcName
    ? this.config.pvcName(slug ?? resourceName)
    : `${resourceName}-home`;
}
```

---

### 🔴 Critical: `main.ts` substitutes raw unsanitized slug — invalid K8s PVC names

`pvcNameTemplate.replace("{name}", slug ?? resourceName)` inserts slug verbatim. Agent names are user-supplied DB strings (e.g. `"My Agent"`, `"OkWOW"`) with uppercase/spaces. A raw slug produces names like `vitals-os-agent-My Agent-home` which K8s rejects (RFC1123: lowercase alphanumeric and hyphens only, max 63 chars). The integration test at line 362 (`"provision() sanitizes an unsafe slug"`) tests for `sanitizeAgentName(slug)` but the implementation skips that step — the test would fail (and it's skipped in CI because `DATABASE_URL_ADMIN_TEST` is unset there).

Fix: `pvcNameTemplate.replace("{name}", sanitizeAgentName(slug) ?? resourceName)` — import `sanitizeAgentName` from `./agent-manifest.ts` in `main.ts`.

---

Note: the previous `pvcNameFor` implementation (before this PR) did exactly the right thing — `this.config.pvcName(slug ? sanitizeAgentName(slug) : resourceName)` — passing one sanitized value. The refactor lost both behaviors.

dodizzle (approved):
**Re-review of `2eb508a7`** — prior finding addressed ✓, no new issues.

**Prior finding**: ✅ Unit tests ignore slug arg — tests now use single-arg `(name) => ...` callbacks matching the corrected `pvcName` type. The callbacks receive the pre-sanitized name directly from `pvcNameFor()`, so assertions on slug-based PVC names are now correct.

The single-arg revert is the right design: `pvcNameFor()` owns the slug-vs-resourceName resolution and passes one clean value to the callback. This also fixes the raw-slug substitution bug in `main.ts` — the callback now always gets an RFC1123-safe string. All CI green.

APPROVE — clean and ready to merge.

dmcaulay (commented):
APPROVE — prior findings addressed ✓, no new issues.

**Prior findings**: Both resolved in `2eb508a`:
- ✅ `pvcName` reverted to single-arg — `pvcNameFor` now passes `slug ? sanitizeAgentName(slug) : resourceName` as the sole argument
- ✅ Slug sanitization moved into `pvcNameFor` — `main.ts` callback receives a pre-sanitized RFC1123-safe name
- ✅ Unit tests updated to single-arg pattern, all 5 passing in CI

Clean merge — no blocking concerns.

PR #587, opened 2026-06-23
fix(chart): add Cloud SQL Auth Proxy sidecar to task-store Deployment (TSS-4.2)

dmcaulay:
## Summary

- The task-store Deployment template was missing the Cloud SQL Auth Proxy sidecar, causing `P1001: Can't reach database server at 127.0.0.1:5432` on startup.
- Adds the same `{{- if .Values.cloudSqlProxy.enabled }}` sidecar block that admin and metrics use.
- Bumps chart to 1.6.1.

## Root cause

TSS-4.1 (chart 1.6.0) added the task-store templates but didn't include the proxy sidecar. The task-store tries to run `prisma migrate deploy` at startup via `127.0.0.1:5432`, which is only reachable when the Cloud SQL Auth Proxy sidecar is present.

## Test plan
- [ ] Deploy shipwright chart 1.6.1 to vitals-os-prod
- [ ] Confirm task-store pod comes up Ready (no `P1001` in logs)
- [ ] Confirm `GET /health` returns 200


dmcaulay (commented):
APPROVE — Clean fix that correctly mirrors the admin-deployment Cloud SQL proxy sidecar. CI green across all Helm checks.

*Review file: state/reviews/PR_REVIEW_587.md*

PR #589, opened 2026-06-23
fix(chart): add serviceAccountName to task-store Deployment for Workload Identity (TSS-4.2)

dmcaulay:
## Summary

- task-store pods were using the `default` ServiceAccount, which has no Workload Identity binding → Cloud SQL Auth Proxy got 403 `cloudsql.instances.get` denied
- Adds `serviceAccountName: {{ include "shipwright.taskStore.serviceAccountName" . }}` to the Deployment spec
- Adds `shipwright.taskStore.serviceAccountName` helper to `_helpers.tpl` (mirrors admin/metrics pattern)
- Adds `taskStore.serviceAccount` values block so callers can set `create: false, name: shipwright` to reuse an existing WI-annotated SA
- Bumps chart to 1.6.2

## Root cause

chart 1.6.1 added the Cloud SQL Auth Proxy sidecar (PR #587) but the pod still ran as `default` SA — the proxy authenticated via ADC which has no Cloud SQL IAM binding.

## Test plan
- [ ] Deploy with `taskStore.serviceAccount.create: false, name: shipwright` in vitals-os values-prod.yaml
- [ ] Confirm cloud-sql-proxy starts cleanly (no 403 in logs)
- [ ] Confirm task-store pod reaches Ready
- [ ] Confirm `GET /health` returns 200


charts/shipwright/values.yaml
```diff
@@ -492,6 +492,15 @@ taskStore:
     # Defaults to "shipwright-secrets" (the canonical cluster secret name).
     # The chart does NOT manage this Secret — create it before deploying.
     existingSecret: "shipwright-secrets"
+  # -- ServiceAccount for the task-store pod. Set create: false and name: <sa>
+  # to reuse an existing SA (e.g. the umbrella chart'''s Workload Identity SA).
+  serviceAccount:
+    # -- Whether a ServiceAccount is created for the task-store workload.
+    create: true
```
dodizzle, line 499: `create: true` is the default, but there is no `task-store-serviceaccount.yaml` template that creates the ServiceAccount resource. Admin, agent, and metrics each have a matching `*-serviceaccount.yaml`. With `taskStore.enabled: true` and `create: true`, the Deployment will reference a ServiceAccount that Helm never creates — pod startup fails with `ServiceAccount not found`. Add `charts/shipwright/templates/task-store-serviceaccount.yaml` gated on `{{- if and .Values.taskStore.enabled .Values.taskStore.serviceAccount.create }}`.

dodizzle (commented):
**COMMENT** — One issue before merge: `taskStore.serviceAccount.create` defaults to `true` in `values.yaml`, but there is no `task-store-serviceaccount.yaml` template to actually create the ServiceAccount resource. Admin, agent, and metrics each have a matching `*-serviceaccount.yaml`; task-store is missing its. With `taskStore.enabled: true` and the default `create: true`, the pod spec references a ServiceAccount name that Helm never materialises — the pod will crash with `ServiceAccount not found`. Production avoids this only because the test plan uses `create: false, name: shipwright`, but fresh installs that leave the default will be broken.

Fix: add `charts/shipwright/templates/task-store-serviceaccount.yaml` gated on `{{- if and .Values.taskStore.enabled .Values.taskStore.serviceAccount.create }}`, mirroring `admin-serviceaccount.yaml`.

Minor: both `Chart.yaml` and `values.yaml` are missing trailing newlines (`\ No newline at end of file` in the diff) — causes git noise.

Everything else is clean: helper template mirrors the established pattern exactly, values comment is clear, chart bump and changelog are correct.

*Review file: state/reviews/PR_REVIEW_589.md*

charts/shipwright/templates/task-store-serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if and .Values.taskStore.enabled .Values.taskStore.serviceAccount.create }}
```
dodizzle, line 1: Template is correctly structured and mirrors the admin pattern. Follow-up: consider adding three Helm unit tests to `task_store_workload_test.yaml` for this template (creates SA by default, skips when `create: false`) and for the `serviceAccountName` wiring on the Deployment — `admin_workload_test.yaml` lines 35-40 and 307-321 have the exact pattern to copy.

dodizzle (approved):
## Re-Review: APPROVE

Prior blocking issue (missing `task-store-serviceaccount.yaml`) is addressed by `b8c662e0` — the new template is gated on `taskStore.enabled && taskStore.serviceAccount.create` and mirrors the admin/metrics pattern exactly.

**Follow-up worth adding** (not blocking): `task_store_workload_test.yaml` has no tests for the new SA template or the `serviceAccountName` wiring on the Deployment. The admin suite covers exactly these three cases and they run in CI via `helm unittest`. Worth adding:
1. Assert `spec.template.spec.serviceAccountName` is set when `taskStore.enabled: true`
2. Assert `task-store-serviceaccount.yaml` renders a `ServiceAccount` when `create: true`
3. Assert no `ServiceAccount` document when `create: false`

**Minor**: `CHANGELOG.md` v1.6.2 entry only says "auto-bump triggered by metrics-v0.79.0" — doesn't describe the actual fix. The `Chart.yaml` annotation captures it correctly, but worth cleaning up.

Good fix — template is clean, helper mirrors established patterns, `create: false` escape hatch is properly documented.

*Review file: state/reviews/PR_REVIEW_589.md*

dmcaulay (commented):
APPROVE (self-review: event=COMMENT per GitHub policy) — chart changes are correct, complete, and consistently patterned after the admin/metrics/agent SA implementations. The `task-store-serviceaccount.yaml` template is correctly dual-gated on `taskStore.enabled + serviceAccount.create`, the helper mirrors the established pattern exactly, and the bring-your-own-SA path (`create: false`) is well-documented. No findings. The Helm `ct install` failure is a transient kind cluster infrastructure issue (configmap cache sync timeout) unrelated to these changes.

*Review file: state/reviews/PR_REVIEW_589.md*

PR #596, opened 2026-06-23
fix(agent): set safe.directory on every startup to survive PVC uid mismatch

dmcaulay:
## Summary

- PVC-mounted repos may be owned by root (uid 0) while the agent runs as uid 1000
- Git's `safe.directory` check blocks all git operations in that case
- The previous band-aid lived in `/home/bun/.gitconfig`, but the startup sequence's `git config --global` calls for credential helper / user.name / user.email rewrite that file on every pod restart, wiping the fix
- This adds `git config --global safe.directory '*'` unconditionally at the top of `setupGitHubAuth`, before any auth-path branching, so it survives every restart regardless of which auth method is configured

## Test plan

- [ ] All 6 `setup-github-auth.integration.test.ts` tests updated and passing
- [ ] Biome lint clean on changed files
- [ ] Deploy to an agent with a PVC, confirm git operations work after a pod restart without manual `.gitconfig` edits

### DIFF
--- a/agent/src/setup-github-auth.integration.test.ts
+++ b/agent/src/setup-github-auth.integration.test.ts
@@ -135,11 +135,17 @@ describe("setupGitHubAuth — App path", () => {
       ],
       { stdio: "inherit", env: expectedEnv },
     );
+    // safe.directory always set unconditionally
+    expect(spawnSync).toHaveBeenCalledWith(
+      "git",
+      ["config", "--global", "safe.directory", "*"],
+      { stdio: "inherit", env: expect.any(Object) },
+    );
     // No `gh auth setup-git` invocation on the App path anymore
     for (const call of spawnSync.mock.calls) {
       expect(call[0]).not.toBe("gh");
     }
-    expect(spawnSync).toHaveBeenCalledTimes(3);
+    expect(spawnSync).toHaveBeenCalledTimes(4);
     expect(startBackgroundRefresh).toHaveBeenCalledTimes(1);
   });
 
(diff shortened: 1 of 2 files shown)


dmcaulay (commented):
COMMENT — One existing test now silently verifies the wrong thing after the call-order change.

**`agent/src/setup-github-auth.integration.test.ts:152` — error-logging test asserts safe.directory failure, not credential helper failure**

The test "logs an error when git config credential helper registration exits non-zero" uses `makeSpawnSync(1)`, making every `spawnSync` call return status 1. After this PR, the safe.directory call fires first — `errorSpy.mock.calls[0]` captures the safe.directory error, not the credential helper failure the test is named for. The assertions `toContain("git config")` and `toContain("status 1")` pass coincidentally. The credential helper failure path is no longer verified.

Fix: check `mock.calls[1]?.[0]` for the credential helper error, and add a separate assertion on `mock.calls[0]?.[0]` for safe.directory — or use position-independent `toHaveBeenCalledWith` matchers keyed on the git arguments to decouple from call order.

**Strengths:** Unconditional placement before all auth branches is correct; `expect.any(Object)` on the env arg is appropriate since env hasn't been mutated at that point; `runOrWarn` is the right wrapper for a non-fatal startup step.

*Review file: state/reviews/PR_REVIEW_596.md*

dmcaulay (commented):
**APPROVE** (re-review R2) — R1 finding fully resolved.

`mock.calls[0]` now explicitly asserts the safe.directory error (`toContain("safe.directory")`); `mock.calls[1]` asserts the credential helper error (`toContain("credential")`). The two assertions check distinct string content that cannot cross-match, so the credential helper failure path is now genuinely verified. All four integration test scenarios (App happy-path call count, error-logging, PAT path, no-auth paths) are updated consistently. CI green.

*Review file: state/reviews/PR_REVIEW_596.md*

PR #633, opened 2026-06-23
feat: remove VITALS_OS_DISPATCH_TOKEN dispatch coupling from all workflows (SEV-1.4)

dodizzle:
## Summary
- Removes the `VITALS_OS_DISPATCH_TOKEN` secret and all `repository_dispatch` coupling from six CI workflow files
- Replaces dispatch token with `GITHUB_TOKEN` where needed for tagging/PR operations
- Removes downstream vitals-os notification steps from `build-agent.yml` and `chart-release.yml`

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | No VITALS_OS_DISPATCH_TOKEN, 'vitals-os', or downstream-dispatch reference remains | ✅ MET |
| 2 | All six workflows remain syntactically valid and build logic intact | ✅ MET |
| 3 | Build workflows no longer emit a repository_dispatch event | ✅ MET |
| 4 | No test layer required (CI YAML removals only) | ✅ MET |

## Test Plan
- [x] Zero grep hits for `VITALS_OS_DISPATCH_TOKEN`, `vitals-os`, `CHART_DISPATCH_REPO` across all `.github/workflows/*.yml`
- [x] All 10 workflow files structurally valid (name/on/jobs keys present)
- [x] Docker build/push and chart-release steps preserved; only dispatch steps removed
- [x] No tests added or retired — removed dispatch had no coverage and is intentionally gone

Generated with [Claude Code](https://claude.com/claude-code)

PR #642, opened 2026-06-23
fix(ci): restore dispatch pipeline broken by #633

dodizzle:
## Summary

- Removes the broken `create-github-app-token` step from all 4 build workflows — `RELEASE_APP_ID`/`RELEASE_APP_PRIVATE_KEY` were never configured as repo secrets, causing immediate failure on every build
- Replaces with `VITALS_OS_DISPATCH_TOKEN` directly (already exists, a PAT — pushes tags that correctly trigger `auto-bump-chart`)
- Restores `Dispatch agent-released to downstream` step in `build-agent.yml` (drives vitals-os agent image build)
- Restores `Notify downstream of chart release` step in `chart-release.yml` (drives image bump + deploy)

Both dispatch steps were removed in #633 without a replacement, silently killing the downstream pipeline. The dispatch target remains configurable via `CHART_DISPATCH_REPO`.

## Test plan

- [ ] Push a change to `agent/` on main — `Build and Push Agent Image` should succeed, push a tag, and dispatch `shipwright-agent-released`
- [ ] Confirm `auto-bump-chart` fires on the new `agent-v*` tag
- [ ] After chart bump merges, confirm `chart-release.yml` publishes and dispatches `shipwright-chart-released` downstream

### DIFF
--- a/.github/workflows/build-agent.yml
+++ b/.github/workflows/build-agent.yml
@@ -28,7 +28,7 @@ jobs:
         id: tag
         uses: mathieudutour/github-tag-action@v6.2
         with:
-          github_token: ${{ secrets.VITALS_OS_DISPATCH_TOKEN }}
+          github_token: ${{ secrets.DISPATCH_TOKEN }}
           tag_prefix: "agent-v"
           default_bump: patch
           fetch_all_tags: true
@@ -55,7 +55,7 @@ jobs:
 
       - name: Dispatch agent-released to downstream
         env:
-          GH_TOKEN: ${{ secrets.VITALS_OS_DISPATCH_TOKEN }}
+          GH_TOKEN: ${{ secrets.DISPATCH_TOKEN }}
           NEW_TAG: ${{ steps.tag.outputs.new_tag }}
           DISPATCH_REPO: ${{ vars.CHART_DISPATCH_REPO }}
         run: |
--- a/.github/workflows/build-agent.yml
+++ b/.github/workflows/build-agent.yml
@@ -24,18 +24,11 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Generate app token for tag push
-        id: app-token
-        uses: actions/create-github-app-token@v1
-        with:
-          app-id: ${{ secrets.RELEASE_APP_ID }}
-          private-key: ${{ secrets.RELEASE_APP_PRIVATE_KEY }}
-
       - name: Bump version and push tag
         id: tag
         uses: mathieudutour/github-tag-action@v6.2
         with:
-          github_token: ${{ steps.app-token.outputs.token }}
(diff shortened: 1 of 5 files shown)


dmcaulay (commented):
COMMENT — The broken app-token step removal is correct and the fix unblocks the pipeline. Two items below.

**Strengths:** The `CHART_DISPATCH_REPO` guard (`[ -z "$DISPATCH_REPO" ] && exit 0`) makes dispatch opt-in and gracefully skips on unconfigured repos. Using `printf` with positional `%s` for JSON construction is the right approach. Restoring `continue-on-error: true` on the chart dispatch step so a notification failure can't block the chart release itself is intentional and good.

.github/workflows/chart-release.yml
```diff
@@ -61,6 +61,19 @@ jobs:
...
 
+      - name: Notify downstream of chart release
+        continue-on-error: true
+        env:
+          GH_TOKEN: ${{ secrets.DISPATCH_TOKEN }}
+          DISPATCH_REPO: ${{ vars.CHART_DISPATCH_REPO }}
+        run: |
+          [ -z "$DISPATCH_REPO" ] && echo "CHART_DISPATCH_REPO not configured, skipping" && exit 0
+          CHART_VERSION=$(grep '^version:' charts/shipwright/Chart.yaml | sed 's/^version: //')
+          [ -z "$CHART_VERSION" ] && echo "Error: could not parse chart version" && exit 1
```
dmcaulay, line 72: Silent failure: `exit 1` under `continue-on-error: true` is swallowed. If `Chart.yaml` is ever unreadable or the `grep/sed` produces empty output, the step emits an error line but the workflow step outcome is still "success (with warning)" — the downstream dispatch silently doesn't fire and the chart release goes unnoticed by downstream consumers.

Two options:
1. Move the chart version parse into a **separate required step** before this one (no `continue-on-error`), so a version parse failure is a hard error. Then this step only runs when a valid version is confirmed.
2. Accept silent-skip on version parse failure and change the `exit 1` to `exit 0` with a clear warning — consistent with the "best effort notification" framing.

Option 1 is safer if downstream consumers are important.

.github/workflows/build-agent.yml
```diff
@@ -59,3 +52,14 @@ jobs:
```
dmcaulay, line 58: Suggestion: the other three build workflows (`build-admin.yml`, `build-metrics.yml`, `build-task-store.yml`) get the token fix but no dispatch step. If downstream auto-bump or deploy workflows need to respond to those image releases too, the dispatch pattern here can be copied with appropriate `event_type` values (e.g. `shipwright-admin-released`). Fine to leave as-is if only agent releases drive downstream.

PR #644, opened 2026-06-23
feat(admin): wire task-store token into KubernetesAgentProvisioner (TSS-5.1)

dmcaulay:
## Summary
- Add `TaskStoreProvisioningClient` interface + `HttpTaskStoreProvisioningClient` (fetch-based) to mint/revoke task-store tokens via the task-store service API
- Update `KubernetesAgentProvisioner.provision()` to mint a per-agent task-store token and store it in the agent Secret (`task-store-token` key) alongside the existing agent API key
- Inject `SHIPWRIGHT_TASK_STORE_TOKEN` (secretKeyRef) and `SHIPWRIGHT_TASK_STORE_URL` (plain value) into provisioned agent Deployment env; rollback revokes the task-store token if Deployment creation fails

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | provision() calls POST /tokens and stores raw token in agent Secret | ✅ MET |
| 2 | Agent Deployments include SHIPWRIGHT_TASK_STORE_TOKEN and SHIPWRIGHT_TASK_STORE_URL | ✅ MET |
| 3 | Token revoked if Deployment creation fails after minting | ✅ MET |
| 4 | New agents can query task store without manual env patching | ✅ MET |
| 5 | Unit tests: token minted, revoked on rollback, env vars in manifest | ✅ MET |
| 6 | Integration test: SHIPWRIGHT_TASK_STORE_TOKEN present in Secret | ✅ MET |

## Test Plan
- [x] 6 new unit tests in `agent-provisioner.unit.test.ts` (task-store minting, rollback, no-op without config)
- [x] 2 new manifest unit tests in `agent-manifest.unit.test.ts` (env var injection with/without taskStoreUrl)
- [x] 1 new integration test verifying `task-store-token` key in provisioned Secret
- [x] Full admin test suite: 563 pass, 77 skip (integration skipped without DB), 0 fail
- [x] Biome lint clean on all changed files


dmcaulay (commented):
APPROVE (posted as COMMENT — self-review policy)

The provision path is solid: mint → Secret → Deployment with correct rollback (revoke + Secret delete) on Deployment failure, including the 409 conflict case. The reconcile gap I initially suspected is actually fine by design — when a Deployment is missing but the Secret already exists, `provision()` detects `secretAlreadyExists = true`, skips minting, and the new Deployment picks up the token from the existing Secret via `secretKeyRef`. The `main.ts` guard (`taskStoreUrl && taskStoreAdminToken`) ensures task-store env vars are never injected without a provisioned token. Test coverage for the mint/rollback/409 paths is comprehensive.

Three issues to address:

**1. deprovision() never revokes the task-store token (operational gap)**

`deprovision()` deletes the K8s Secret and Deployment but does not call `revokeToken()`. The token ID (`tsTokenId`) is a local variable in `provision()` — it is not stored in the Secret, the DB, or `ProvisionResult`. When `deprovision()` runs later, there is no token ID to revoke, so the task-store token row is never cleaned up. The Secret deletion makes the credential unusable, but orphaned token rows will accumulate in the task-store DB over time.

If the task-store service has a GC mechanism that handles this (e.g. expire-on-last-heartbeat), that contract should be documented here. Otherwise, persist the token ID — e.g. store it in the Secret under a stable key (`task-store-token-id`) and read it back in `deprovision()`.

**2. No test for reconcile() with taskStore configured**

`reconcile()` calls `provision()` for missing Deployments, which will mint tokens when `taskStore` is configured. No unit test exercises this path. Worth adding a minimal test: pre-seed no Deployment, call `reconcile()`, assert `mintToken` was called once and the created Deployment env includes `SHIPWRIGHT_TASK_STORE_URL`.

**3. NoopTaskStoreProvisioningClient is dead code**

`main.ts` uses `undefined` as the unconfigured fallback, not `NoopTaskStoreProvisioningClient`. The class's JSDoc says it's used "when ... not configured" but it's never imported in `main.ts`. Either wire it in as the fallback or remove it — as-is it's misleading.

Minor: the `SHIPWRIGHT_TASK_STORE_TOKEN` row in `docs/configuration.md` says "tokens are minted and stored but not yet used by agents" — more precisely, the token IS injected into the Deployment as an env var, it's just not yet read by the agent runtime. Worth clarifying so operators don't assume no injection happens.

admin/src/task-store-provisioning-client.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// ─── No-op implementation ─────────────────────────────────────────────────────
+
+/**
+ * No-op client used when SHIPWRIGHT_TASK_STORE_URL / SHIPWRIGHT_TASK_STORE_ADMIN_TOKEN
+ * are not configured. mintToken returns empty strings so callers can gate on
+ * the configured path without null-checking the return.
+ */
+export class NoopTaskStoreProvisioningClient
```
dodizzle, line 89: `NoopTaskStoreProvisioningClient` is never used — `main.ts` guards with `taskStoreUrl && taskStoreAdminToken ? new HttpTaskStoreProvisioningClient(...) : undefined` and passes `undefined` as the fallback. The class JSDoc says it's used "when not configured" but nothing imports it. Either wire it as the explicit fallback (instead of `undefined`) or delete it — as-is it's misleading.

dodizzle (commented):
**COMMENT** — TSS-5.1 is well-structured and the provision rollback path is solid. Three issues from @dmcaulay's prior review remain unaddressed at the current head.

**1. `deprovision()` never revokes the task-store token (operational gap)**

`deprovision()` deletes the K8s Deployment and Secret but does not call `revokeToken()`. The token ID (`tsTokenId`) is local to `provision()` and is never persisted to the Secret, DB, or `ProvisionResult`. Every deprovisioned agent leaves an orphaned token row in the task-store DB. Deleting the Secret makes the credential unusable but the row stays forever.

Fix: store `tsTokenId` as `stringData["task-store-token-id"]` in `secretSpec()`, read it back in `deprovision()`, and call `revokeTaskStoreTokenBestEffort` there. Or document that the task-store GC handles this via expiry (but that contract doesn't exist today).

**2. No test for `reconcile()` with `taskStore` configured**

`reconcile()` calls `provision()` for missing Deployments, which mints a token when `taskStore` is set. No test covers this path. Add a unit test: pre-seed no Deployment, call `reconcile()` with `taskStore: stubTaskStore().client` in config, assert `mintToken` was called once and the Deployment env includes `SHIPWRIGHT_TASK_STORE_URL`.

**3. `NoopTaskStoreProvisioningClient` is dead code** (inline on the file)

PR #645, opened 2026-06-23
fix(ci): hard-fail on chart version parse before dispatch

dodizzle:
## Summary

Addresses review finding from #642 (dmcaulay): `exit 1` inside `continue-on-error: true` was being swallowed — a failed version parse would silently skip the dispatch with no downstream notification.

Fix: split version parse into a separate required step (`Parse chart version`) with no `continue-on-error`. The notify step only runs with a confirmed version output. A bad `Chart.yaml` now fails the job hard instead of silently passing.

Leaves item 2 (dispatch for admin/metrics/task-store) open pending answer on whether downstream needs those events.


dodizzle (commented):
**APPROVE** (posted as COMMENT — self-review policy)

Two clean CI fixes:
1. `chart-release.yml`: version parse extracted into a required step — `continue-on-error: true` remains only on the dispatch notify step where it belongs, not on parse. A bad `Chart.yaml` now hard-fails the job.
2. `auto-bump-chart.yml`: `task-store-v*` added to trigger patterns — task-store releases now correctly fire chart bumps alongside admin/metrics/agent tags.

No issues.

PR #649, opened 2026-06-24
fix(chart): add task-store to GKE HealthCheckPolicy subchart

dodizzle:
## Summary

- Adds a `HealthCheckPolicy` resource for the `task-store` Service in `charts/shipwright/templates/healthcheckpolicy.yaml`
- Gated on `taskStore.enabled` (defaults `false`) — no impact on existing installs
- Task-store exposes `GET /health` (unauthenticated, returns 200) — same probe path as admin and metrics

## Test plan

- [ ] `helm template` with `taskStore.enabled=true` + gateway conditions renders 3 `HealthCheckPolicy` docs
- [ ] `helm template` with default values (taskStore off) still renders 2 docs — no regression
- [ ] Helm unittest suite updated with 3 new cases: all-enabled, metrics-off+taskStore-on, admin-only

### DIFF
--- a/charts/shipwright/Chart.yaml
+++ b/charts/shipwright/Chart.yaml
@@ -8,7 +8,7 @@ type: application
 # Chart version (SemVer). Bump on every chart change, independent of appVersion.
 # See CHANGELOG.md and the README "Versioning" section — CI enforces this bump
 # via `ct lint --check-version-increment`.
-version: 1.6.14
+version: 1.6.15
 # Version of the Shipwright application this chart deploys by default.
 appVersion: "0.1.0"
 home: https://shipwright-harness.com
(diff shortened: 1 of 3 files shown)


dmcaulay (approved):
APPROVE — correct, minimal, and well-tested infrastructure addition. Probe config (15s/5s/1/2 thresholds, HTTP /health) mirrors existing admin+metrics HealthCheckPolicies exactly. The `taskStore.enabled` gate (default false) means zero impact on existing installs. Three new Helm unittest cases cover all combinatorial paths. CI green.

dodizzle (commented):
APPROVE — correct, minimal infrastructure addition. Probe config (15s/5s/1/2 thresholds, HTTP `/health`) mirrors existing admin + metrics HealthCheckPolicies exactly. The `taskStore.enabled` gate (default false) means zero impact on existing installs. Four new Helm unittest cases cover all combinatorial paths. CI green. dmcaulay already approved at this head — no blocking issues found.

PR #650, opened 2026-06-24
docs(claude): how to diagnose admin 401s (auth middleware vs onError)

dmcaulay:
Adds a Debugging note to the admin/CLAUDE.md surface capturing a diagnostic that cost real hours: a `401 {"error":"Unauthorized"}` with **no `WWW-Authenticate` header** is *not* the auth middleware rejecting the token — `createAdminAuthMiddleware` always sets that header on its bearer-path 401s. A header-less 401 is surfaced via the app's `onError` from a downstream call inside a handler (typically a `KubernetesClient` request the K8s API rejected).

Tell-tale symptom: read-only routes (DB only) return 200 while a route that calls the K8s API (agent `reconcile` / `provision`) 401s with the **same** token — a calls-the-dependency correlation, not auth or routing. The note also records the common K8s-401 root cause (the pod's ServiceAccount recreated → new UID → bound token invalidated even with a far-future `exp`) and the fix (restart the pod; don't render the SA as a Helm hook, which recreates it on every `helm upgrade`).

Generic wording only — no deployment-specific details (public repo).

Pairs with #649 (task-store HealthCheckPolicy parity). The SA-as-hook and umbrella-chart specifics stay in the downstream consumer's docs — the OSS chart's own ServiceAccounts are normal resources, not hooks.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -145,6 +145,10 @@ The schema uses `provider = "postgresql"`. All database connection strings must 
 
 For the full configuration reference (all env vars, agent config, policy config), see [`docs/configuration.md`](./docs/configuration.md).
 
+## Debugging
+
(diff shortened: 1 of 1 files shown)


dmcaulay (commented):
APPROVE — clean, accurate docs addition. The WWW-Authenticate header signal is a precise diagnostic: auth middleware always sets it, onError from a downstream K8s API call doesn't. Public-repo safe wording; all CI green.

dodizzle (approved):
APPROVE — accurate, targeted diagnostic. The `WWW-Authenticate` header signal is precise: `createAdminAuthMiddleware` always sets it on bearer-path 401s; a header-less `{"error":"Unauthorized"}` is `onError` surfacing a downstream K8s API rejection, not the auth middleware. The correlation test (read-only routes 200 / K8s-calling routes 401 with the same token) is the fastest way to confirm it's a dependency problem, not auth or routing. SA delete+recreate → new UID → token invalidation chain is non-obvious and exactly right. Appropriately scoped for a public repo — no deployment-specific details. CI green. No issues found.

PR #651, opened 2026-06-24
feat(task-store): agent token scoping

dmcaulay:
## Summary

- Adds `agentId String?` to `TaskToken` — `null` = admin token, set = agent token scoped to that agent
- Agent tokens can only create, read, and update tasks where `assignee === agentId`; creates auto-set `assignee`
- Admin tokens retain unrestricted access to all tasks and token management
- Token management routes (`/tokens`) are now admin-only

## Changes

- **Schema + migration**: `agentId` column on `TaskToken`
- **`token-service`**: `create(label?, agentId?)` stores it; `validate()` returns `{ id, agentId }`
- **`auth`**: middleware exposes `agentId` on Hono context (`null` for admin)
- **`task routes`**: ownership enforced on all CRUD + action endpoints; `POST /tasks` and `POST /tasks/bulk` force `assignee = agentId` for agent tokens
- **`token routes`**: 403 for any agent token
- **`task-service`**: `listReady(agentId?)` filters results post-dependency-resolution (full graph still used for dep checking)
- **`errors`**: `ForbiddenError` (403) added

## Test plan

- [x] All existing smoke tests pass with updated fakes (`agentId: null` for admin tokens)
- [ ] Integration: verify agent token sees only its tasks, admin sees all, cross-agent PATCH returns 403
- [ ] Existing admin tokens (`agentId null` in DB) continue to work without migration

### DIFF
--- a/admin/src/agent-provisioner.ts
+++ b/admin/src/agent-provisioner.ts
@@ -226,6 +226,7 @@ export class KubernetesAgentProvisioner implements AgentProvisioner {
       if (this.config.taskStore) {
         const { id, rawToken } = await this.config.taskStore.mintToken(
           `agent:${agentId}`,
+          agentId,
         );
         tsTokenId = id;
         tsRawToken = rawToken;
--- a/task-store/src/api.smoke.test.ts
+++ b/task-store/src/api.smoke.test.ts
@@ -355,4 +355,41 @@ describe("task-store API (smoke)", () => {
     });
     expect(res.status).toBe(403);
   });
+
+  it("GET /tasks/:id returns 200 when agent token reads an unassigned task", async () => {
+    const app = makeApp({
+      tokenService: fakeAgentTokenService(),
+      // Task has no assignee — should be accessible to any agent token.
+      taskService: fakeTaskService({
+        getResult: makeTask({ id: "task-1", assignee: null }),
+      }),
+    });
+    const res = await app.request("/tasks/task-1", {
+      headers: { Authorization: `Bearer ${AGENT_TOKEN}` },
+    });
+    expect(res.status).toBe(200);
+  });
+
+  it("PATCH /tasks/:id with agent token pins assignee to the agent's ID", async () => {
+    const app = makeApp({
+      tokenService: fakeAgentTokenService(),
+      // Task is owned by agent-1 (the token's agent).
+      taskService: fakeTaskService({
+        getResult: makeTask({ id: "task-1", assignee: "agent-1" }),
+      }),
+    });
+    const res = await app.request("/tasks/task-1", {
+      method: "PATCH",
+      headers: {
+        Authorization: `Bearer ${AGENT_TOKEN}`,
+        "content-type": "application/json",
+      },
+      // Attempt to reassign to a different agent.
+      body: JSON.stringify({ assignee: "agent-2", status: "in_progress" }),
+    });
+    expect(res.status).toBe(200);
+    const body = (await res.json()) as Task;
+    // Assignee must be pinned back to agent-1, not agent-2.
+    expect(body.assignee).toBe("agent-1");
+  });
 });
--- a/task-store/src/api.smoke.test.ts
+++ b/task-store/src/api.smoke.test.ts
@@ -392,4 +392,58 @@ describe("task-store API (smoke)", () => {
     // Assignee must be pinned back to agent-1, not agent-2.
     expect(body.assignee).toBe("agent-1");
   });
+
+  it("GET /tasks?ready=true with agent token forwards agentId to listReady", async () => {
+    let capturedAgentId: string | undefined = undefined;
+
+    const spyTaskService: TaskServiceLike = {
+      ...fakeTaskService(),
+      async listReady(agentId?: string) {
+        capturedAgentId = agentId;
+        return [];
+      },
+    };
+
+    const app = makeApp({
+      tokenService: fakeAgentTokenService(),
+      taskService: spyTaskService,
+    });
+
+    const res = await app.request("/tasks?ready=true", {
+      headers: { Authorization: `Bearer ${AGENT_TOKEN}` },
+    });
+
+    expect(res.status).toBe(200);
+    // The token's agentId ("agent-1") must be forwarded to listReady.
+    expect(capturedAgentId).toBe("agent-1");
+  });
+
+  it("GET /tasks?status=in_progress with agent token ignores caller-supplied ?assignee", async () => {
+    let capturedAssignee: string | undefined = undefined;
+
+    const spyTaskService: TaskServiceLike = {
+      ...fakeTaskService(),
+      async list(opts) {
+        capturedAssignee = opts.assignee;
+        return [];
+      },
+    };
+
+    const app = makeApp({
+      tokenService: fakeAgentTokenService(),
+      taskService: spyTaskService,
+    });
+
+    // Caller tries to supply ?assignee=other-agent to peek at another agent's tasks.
+    const res = await app.request(
+      "/tasks?status=in_progress&assignee=other-agent",
+      {
+        headers: { Authorization: `Bearer ${AGENT_TOKEN}` },
+      },
+    );
+
+    expect(res.status).toBe(200);
+    // Token's agentId must win; caller-supplied assignee must be ignored.
+    expect(capturedAssignee).toBe("agent-1");
+  });
 });
--- a/task-store/src/api.smoke.test.ts
+++ b/task-store/src/api.smoke.test.ts
@@ -16,7 +16,7 @@ import { describe, expect, it } from "bun:test";
 import { createTaskStoreApp } from "./app.ts";
 import { ConflictError, NotFoundError } from "./errors.ts";
 import type { Task } from "./index.ts";
-import type { TaskServiceLike } from "./task-service.ts";
+import type { TaskListFilters, TaskServiceLike } from "./task-service.ts";
 import type { TokenServiceLike } from "./token-service.ts";
 
 // ─── Fakes ────────────────────────────────────────────────────────────────────
@@ -430,12 +430,12 @@ describe("task-store API (smoke)", () => {
   });
 
   it("GET /tasks?ready=true with agent token forwards agentId to listReady", async () => {
-    let capturedAgentId: string | undefined = undefined;
+    const capturedArgs: Array<string | undefined> = [];
 
     const spyTaskService: TaskServiceLike = {
       ...fakeTaskService(),
       async listReady(agentId?: string) {
-        capturedAgentId = agentId;
+        capturedArgs.push(agentId);
         return [];
       },
     };
@@ -451,16 +451,16 @@ describe("task-store API (smoke)", () => {
 
     expect(res.status).toBe(200);
     // The token's agentId ("agent-1") must be forwarded to listReady.
-    expect(capturedAgentId).toBe("agent-1");
+    expect(capturedArgs[0]).toBe("agent-1");
   });
 
   it("GET /tasks?status=in_progress with agent token ignores caller-supplied ?assignee", async () => {
-    let capturedAssignee: string | undefined = undefined;
+    const capturedAssignees: Array<string | undefined> = [];
 
     const spyTaskService: TaskServiceLike = {
       ...fakeTaskService(),
-      async list(opts) {
-        capturedAssignee = opts.assignee;
+      async list(opts?) {
+        capturedAssignees.push(opts?.assignee);
         return [];
       },
     };
@@ -480,6 +480,6 @@ describe("task-store API (smoke)", () => {
 
     expect(res.status).toBe(200);
     // Token's agentId must win; caller-supplied assignee must be ignored.
-    expect(capturedAssignee).toBe("agent-1");
+    expect(capturedAssignees[0]).toBe("agent-1");
   });
 });
--- a/task-store/src/api.smoke.test.ts
+++ b/task-store/src/api.smoke.test.ts
@@ -22,6 +22,7 @@ import type { TokenServiceLike } from "./token-service.ts";
 // ─── Fakes ────────────────────────────────────────────────────────────────────
 
 const VALID_TOKEN = "valid-token";
+const AGENT_TOKEN = "agent-token";
 
 function fakeTokenService(): TokenServiceLike {
   return {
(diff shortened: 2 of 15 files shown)


task-store/src/routes/tasks.ts
```diff
@@ -82,19 +115,32 @@ export function createTasksRoutes(taskService: TaskServiceLike): Hono {
...
     const task = await taskService.get(c.req.param("id"));
     if (!task) throw new NotFoundError("task not found");
+    if (agentId !== null && task.assignee !== agentId) {
+      throw new ForbiddenError("task belongs to a different agent");
+    }
     return c.json(task, 200);
   });
 
   // ─── Update ────────────────────────────────────────────────────────────────
   app.patch("/:id", async (c) => {
```
dmcaulay, line 141: `requireOwnership` verifies `task.assignee === agentId` at read time, but `body` is passed to `taskService.update()` without stripping `assignee`. An agent token that owns a task can patch `{ "assignee": "other-agent-id" }` or `{ "assignee": null }` to escape its scoping boundary.

Fix: add `if (agentId !== null) { delete body.assignee; }` before `taskService.update`.

task-store/src/api.smoke.test.ts
```diff
@@ -31,14 +31,15 @@ function fakeTokenService(): TokenServiceLike {
...
+          agentId: null,
           createdAt: new Date(),
           revokedAt: null,
         },
         rawToken: "raw",
       };
     },
     async validate(raw: string) {
-      return raw === VALID_TOKEN ? { id: "tok-1" } : null;
+      return raw === VALID_TOKEN ? { id: "tok-1", agentId: null } : null;
```
dmcaulay, line 43: `fakeTokenService()` always returns `agentId: null` (admin). The PR's core enforcement paths — forced `assignee` on `POST /tasks`, 403 on cross-agent writes, 403 on `POST /tokens` — have no smoke coverage.

Suggestion: add a second fake returning `{ id: "tok-2", agentId: "agent-1" }` and test: (a) `POST /tasks` response has `assignee: "agent-1"`, (b) `GET /tasks/:id` for a task owned by `"agent-2"` returns 403, (c) `POST /tokens` returns 403.

dmcaulay (commented):
COMMENT — two issues before merge:

**1. Agent token can escape ownership via PATCH body** (`task-store/src/routes/tasks.ts:141`): `requireOwnership` checks `task.assignee === agentId` at read time, but the raw body — including any caller-supplied `assignee` field — is passed directly to `taskService.update()`. An agent token that owns a task can patch it with `{ "assignee": "other-agent-id" }` or `{ "assignee": null }`, silently moving it outside its ownership scope. Fix: `if (agentId !== null) { delete body.assignee; }` before calling `taskService.update`.

**2. No smoke coverage for agent-scoping paths** (`task-store/src/api.smoke.test.ts:42`): `fakeTokenService()` hardcodes `agentId: null`. The core enforcement — forced `assignee` on POST /tasks, 403 on cross-agent writes, 403 on POST /tokens — has no smoke tests. These are exactly the route-enforcement checks that belong at the smoke layer and don't need a real DB.

Strengths: `requireOwnership` is consistently applied across all write routes; `listReady` correctly loads the full graph for dep resolution before filtering by `agentId`; SHA-256-only token storage is correct secret hygiene; atomic `$executeRaw` claim is the right pattern for a concurrent claim race.

task-store/src/routes/tasks.ts
```diff
@@ -38,19 +45,39 @@ async function readJson(c: {
...
-  const app = new Hono();
+/** Fetch a task and enforce agent ownership. Throws 404 or 403 as appropriate. */
+async function requireOwnership(
+  taskService: TaskServiceLike,
+  id: string,
+  agentId: string | null,
+) {
+  const task = await taskService.get(id);
+  if (!task) throw new NotFoundError("task not found");
+  if (agentId !== null && task.assignee !== agentId) {
```
dodizzle, line 56: **Unassigned tasks permanently blocked for agent tokens.** When `task.assignee === null`, this check always throws 403 for any agent token. Unassigned tasks already don't appear in `listReady` results for agent tokens, but tasks created without an assignee by an admin are now completely inaccessible for writes via agent tokens.

Consider: `if (agentId !== null && task.assignee !== null && task.assignee !== agentId)` to allow agent tokens to claim/operate on unassigned tasks. Or document explicitly that all tasks intended for agent consumption must have `assignee` set at creation time.

task-store/src/routes/tasks.ts
```diff
@@ -82,19 +115,32 @@ export function createTasksRoutes(taskService: TaskServiceLike): Hono {
...
+    if (agentId !== null && task.assignee !== agentId) {
+      throw new ForbiddenError("task belongs to a different agent");
+    }
     return c.json(task, 200);
   });
 
   // ─── Update ────────────────────────────────────────────────────────────────
   app.patch("/:id", async (c) => {
+    const agentId = c.get("agentId");
+    await requireOwnership(taskService, c.req.param("id"), agentId);
```
dodizzle, line 143: **Agent token can reassign its own task.** `requireOwnership` verifies the caller owns this task, but the raw `body` is then forwarded to `taskService.update` without stripping `assignee`. An agent token can send `{ "assignee": "other-agent" }` to reassign a task it owns, losing visibility of it.

Fix: after `requireOwnership`, add `if (agentId !== null) { body.assignee = agentId; }`.

task-store/src/api.smoke.test.ts
```diff
@@ -31,14 +31,15 @@ function fakeTokenService(): TokenServiceLike {
           id: "tok-1",
           token: "hash",
           label: label ?? null,
+          agentId: null,
           createdAt: new Date(),
           revokedAt: null,
         },
         rawToken: "raw",
       };
```
dodizzle, line 40: **No test coverage for agent token enforcement.** The fake `validate` only returns `agentId: null` (admin token). None of the new security paths are tested: 403 when agent token hits another agent's task, 403 on `/tokens`, forced assignee on `POST /tasks`, scoped results on `GET /tasks`.

Add a second fake token variant (`agentId: 'agent-x'`) and smoke cases for each enforcement path. CLAUDE.md requires tests to land in the same PR.

dodizzle (commented):
COMMENT — 1 critical + 2 important findings on the new agent-scoped token enforcement.

**Critical**: `requireOwnership` throws 403 when `task.assignee === null`, meaning unassigned tasks are permanently inaccessible to agent tokens for writes. Consider allowing writes when `task.assignee === null`: `if (agentId !== null && task.assignee !== null && task.assignee !== agentId)` — or document that all agent-accessible tasks must have `assignee` set at creation.

**Missing tests**: The smoke suite only tests admin tokens (`agentId: null`). The entire security boundary — agent 403s, scoped list filtering, forced assignee on create — is unexercised. CLAUDE.md requires tests in the same PR.

**PATCH assignee bypass**: After `requireOwnership` confirms ownership, the PATCH handler passes the raw body to `taskService.update` without stripping `assignee`. An agent token can reassign its own task to another agent. Add `if (agentId !== null) { body.assignee = agentId; }` after the ownership check.

admin/src/agent-provisioner.ts
```diff
@@ -226,6 +226,7 @@ export class KubernetesAgentProvisioner implements AgentProvisioner {
       if (this.config.taskStore) {
         const { id, rawToken } = await this.config.taskStore.mintToken(
           `agent:${agentId}`,
+          agentId,
```
dmcaulay, line 229: `agentId` is now forwarded to `mintToken` — correct — but the unit test stub captures only `label` in its `minted` array so this path has no assertion. Add `agentId` to the stub: `minted.push({ label, agentId, id })` and assert `expect(minted[0].agentId).toBe(AGENT_ID)` in the provision test.

dmcaulay (commented):
COMMENT (R2) — one issue before merge. Both round-1 findings (PATCH assignee bypass + smoke coverage) are addressed cleanly. The provisioner now correctly passes `agentId` to `mintToken`, but the unit test stub (`stubTaskStore`) captures only `label` in its `minted` array — the `agentId` passthrough from `ce9b3a0` has no unit assertion. If it were accidentally dropped, all provisioner tests would still pass.

Fix: extend `stubTaskStore.mintToken` to capture `agentId` and add `expect(minted[0].agentId).toBe(AGENT_ID)` to the provision test.

Strengths: both R1 findings fully resolved; 5 agent-scoped smoke tests covering forced assignee, cross-agent 403, admin-only token route, unassigned-task 403, and PATCH pin; unassigned-task blocking is intentional and explicitly documented; `listReady()` scopes post-graph-resolution so cross-agent dep chains remain intact.

task-store/src/api.smoke.test.ts
```diff
@@ -281,4 +311,85 @@ describe("task-store API (smoke)", () => {
...
+      },
+      // Attempt to reassign to a different agent.
+      body: JSON.stringify({ assignee: "agent-2", status: "in_progress" }),
+    });
+    expect(res.status).toBe(200);
+    const body = (await res.json()) as Task;
+    // Assignee must be pinned back to agent-1, not agent-2.
+    expect(body.assignee).toBe("agent-1");
+  });
 });
```
dodizzle, line 485: Two list-scoping paths are missing smoke coverage:

**1. `GET /tasks?ready=true` with AGENT_TOKEN** — `tasks.ts:72` calls `taskService.listReady(agentId ?? undefined)` but no test verifies the agentId is forwarded. The `fakeTaskService.listReady` stub ignores its argument so a regression here would be silent.

**2. `GET /tasks?status=in_progress` with AGENT_TOKEN** — `tasks.ts:80` overrides `assignee: agentId ?? c.req.query('assignee')`. No test verifies the token's agentId takes precedence over a caller-supplied `?assignee`.

Both can be added here using the existing `fakeAgentTokenService`. For `listReady`, the simplest approach is to make `fakeTaskService` accept an `onListReady` spy that captures the agentId arg.

dodizzle (commented):
COMMENT — prior critical findings addressed; two list-scoping smoke gaps remain.

**Fixed since last review:**
- PATCH `assignee` bypass: `body.assignee = agentId` now pins the value after `requireOwnership` ✅
- Smoke coverage added for per-task routes: forced assignee on create, 403 on cross-agent reads, 403 on POST /tokens, 403 on unassigned task, PATCH assignee pinning ✅
- Admin provisioner now passes `agentId` when minting agent tokens ✅

**Still missing:**
1. `GET /tasks?ready=true` with agent token — `listReady(agentId)` is the dev-task pickup path and has no smoke test verifying the agentId is forwarded.
2. `GET /tasks?status=in_progress` with agent token — the interrupted-task resume path relies on `assignee: agentId ?? c.req.query('assignee')`, also untested.

Both gaps are in the list-scoping code paths the dev-task cron runs on every cycle. See inline comment.

task-store/src/api.smoke.test.ts
```diff
@@ -281,4 +311,139 @@ describe("task-store API (smoke)", () => {
...
+      taskService: spyTaskService,
+    });
+
+    const res = await app.request("/tasks?ready=true", {
+      headers: { Authorization: `Bearer ${AGENT_TOKEN}` },
+    });
+
+    expect(res.status).toBe(200);
+    // The token's agentId ("agent-1") must be forwarded to listReady.
+    expect(capturedAgentId).toBe("agent-1");
```
dodizzle: TypeScript infers `capturedAgentId` as `undefined` here (from the `= undefined` initializer) and narrows the type before the async spy has a chance to assign it. `expect(capturedAgentId).toBe("agent-1")` hits TS2769 — no `toBe` overload accepts `string` when the arg type is narrowed to `undefined`. CI fails at typecheck.

Fix options:
```ts
// Option 1: assert non-undefined first
expect(capturedAgentId).not.toBeUndefined();
expect(capturedAgentId).toBe("agent-1");

// Option 2: initialize as string | null so TypeScript doesn't narrow to undefined
let capturedAgentId: string | null = null;
// ...
expect(capturedAgentId).toBe("agent-1");
```
Same fix needed at line 447 for `capturedAssignee`.

dodizzle (commented):
COMMENT (R3) — one issue before merge. Both R2 list-scoping tests are added and the logic is correct, but the spy variable type narrowing causes a typecheck failure that breaks CI.

**TypeScript type error — CI fails at typecheck** (`task-store/src/api.smoke.test.ts:418,447`): Both new spy tests declare `let capturedAgentId: string | undefined = undefined`. TypeScript narrows the type to `undefined` at the declaration site and doesn't widen it based on the async callback assignment. `expect(capturedAgentId).toBe("agent-1")` hits TS2769 — the only `toBe` overload that matches `undefined` is `toBe(expected: undefined)`, so `"agent-1"` doesn't fit. Same issue at line 447 for `capturedAssignee` plus TS18048 on the `opts` param in the spy closure.

Fix: add an assertion before the `.toBe()` call — `expect(capturedAgentId).not.toBeUndefined()` — or initialize with `let capturedAgentId: string | null = null` and assert against that type.

Strengths: the two missing smoke tests are the right approach — spy injection into `fakeTaskService` is clean. The `agentId ?? undefined` forwarding to `listReady` and the `agentId ??` override for `list()` are correctly exercised once this typecheck issue is resolved.

task-store/src/api.smoke.test.ts
```diff
@@ -281,4 +311,139 @@ describe("task-store API (smoke)", () => {
...
+    const res = await app.request(
+      "/tasks?status=in_progress&assignee=other-agent",
+      {
+        headers: { Authorization: `Bearer ${AGENT_TOKEN}` },
+      },
+    );
+
+    expect(res.status).toBe(200);
+    // Token's agentId must win; caller-supplied assignee must be ignored.
+    expect(capturedAssignee).toBe("agent-1");
```
dodizzle: Same narrowing issue as line 418 — `capturedAssignee` is typed `undefined` at this point, causing TS2769. Additionally, `opts` in the spy closure (`async list(opts)`) is typed as `TaskListFilters | undefined` because `list()` accepts optional params — accessing `opts.assignee` without a guard causes TS18048. Use `opts?.assignee` or narrow the fake's signature to `async list(opts: TaskListFilters)` (since in the smoke context it will always be called with an object).

task-store/src/api.smoke.test.ts
```diff
@@ -281,4 +313,173 @@ describe("task-store API (smoke)", () => {
...
+      taskService: spyTaskService,
+    });
+
+    const res = await app.request("/tasks?ready=true", {
+      headers: { Authorization: `Bearer ${AGENT_TOKEN}` },
+    });
+
+    expect(res.status).toBe(200);
+    // The token's agentId ("agent-1") must be forwarded to listReady.
+    expect(capturedAgentId).toBe("agent-1");
```
dodizzle: TS2769 still here — same narrowing issue as R3. `capturedAgentId` is typed `undefined` at this point; `toBe("agent-1")` doesn't fit. The result-based test added above (line 375) covers the same behavior. Simplest fix: delete this spy test and the one below it (lines 432–484). If you want to keep the spy approach, change `let capturedAgentId: string | undefined = undefined` to `let capturedAgentId: string | null = null` — TypeScript won't narrow `null` the same way.

dodizzle (commented):
COMMENT (R4) — CI still fails at typecheck on `8c6b872d`. The new commit added correct result-based tests and the provisioner `agentId` unit assertion, but the spy-based tests from `21525c00` at lines 432–484 are still present with the same TypeScript errors: TS2769 at lines 454 and 483, TS18048 at line 463. Either drop lines 432–484 (the result-based tests above already cover this behavior) or fix the spy variable types to `string | null` initialized to `null`. Either resolves the typecheck failure and unblocks merge.

PR #653, opened 2026-06-24
fix(plan-session): store plan on disk, drop phantom /tasks/repo endpoint

dmcaulay:
## Summary

- `plan-session` was calling `GET /tasks/repo` to discover owner/repo before creating a GitHub Issue to hold the plan — but that endpoint doesn't exist in the task store and the concept doesn't make sense for a shared store used across multiple agents and repos
- Switches to the same pattern as `/prd`: writes the plan to `planning/{session}/PLAN.md` on disk
- Tasks reference the plan via `source: "planning/{session}/PLAN.md"` and are posted directly to `/tasks/bulk`
- Removes the phantom `GET /tasks/repo` entry from the task-store skill's API reference table

## Test plan

- [ ] Run `/plan-session` — confirm plan is written to `planning/{session}/PLAN.md`
- [ ] Confirm tasks are posted to the task store with correct `source` field
- [ ] Confirm no attempt is made to call `/tasks/repo` or create a GitHub Issue

### DIFF
--- a/plugins/shipwright/.claude-plugin/plugin.json
+++ b/plugins/shipwright/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "shipwright",
-  "version": "4.29.0",
+  "version": "4.29.1",
   "description": "Shipwright — conversational planning, queue-driven execution, policy-controlled code review, a five-phase test-readiness pipeline, and autonomous deploy (merge → canary → promote) for any software project",
   "author": {
     "name": "App Vitals",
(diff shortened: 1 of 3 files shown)


dmcaulay (commented):
COMMENT — the core fix is correct (removes the phantom `/tasks/repo` endpoint and switches to the disk-storage pattern), but `plugins/shipwright/README.md` is now two versions behind.

**Finding**: `plugins/shipwright/README.md:1` still says `# Shipwright v4.28.2` while `plugin.json` is now `4.29.1` (two versions behind — was already stale before this PR). CLAUDE.md requires version sync across all three locations when bumping `plugin.json`. Update line 1 to `# Shipwright v4.29.1`.

PR #668, opened 2026-06-24
docs: clarify agent token scoping applies to all task-store endpoints

dodizzle:
## Summary
- Clarifies that agent bearer token scoping applies to **all** task-store API endpoints, not just `?ready=true`
- Adds a scoping note in the Authentication section so the contract is visible at first read
- Adds a scoping callout in the Full API reference table for quick reference

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Authentication section explains bearer token scopes all API operations | ✅ MET |
| 2 | `?ready=true` semantics section keeps per-token explanation | ✅ MET |
| 3 | Full API reference table has scoping note | ✅ MET |
| 4 | Existing examples and commands unchanged | ✅ MET |

## Test Plan
- [x] Docs-only change — no code modified
- [x] Diff reviewed: additions only, no existing content removed

### DIFF
--- a/plugins/shipwright/skills/task-store/SKILL.md
+++ b/plugins/shipwright/skills/task-store/SKILL.md
@@ -31,6 +31,8 @@ Both env vars are provisioned automatically by the agent harness:
 | `SHIPWRIGHT_TASK_STORE_URL` | Base URL of the task store service |
 | `SHIPWRIGHT_TASK_STORE_TOKEN` | Bearer token for this agent |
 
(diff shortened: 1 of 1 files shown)


dodizzle (commented):
APPROVE — Accurate, well-placed docs clarifying that agent bearer token scoping applies to all task-store endpoints, not just `?ready=true`. Verified against `task-store/src/routes/tasks.ts` — the implementation matches the documented contract. No issues.

PR #670, opened 2026-06-24
feat(task-store): state filter, pagination, and status dropdown fix

dodizzle:
## Summary

- **`GET /tasks` response** is now `{ tasks, total, limit, offset }` — consistent wrapper enabling pagination metadata
- **`?state=open|closed`** coarse filter maps to the terminal/active status sets already codified in `ready.ts`
- **`?limit` / `?offset`** pagination params, default limit 50
- **Admin UI** defaults to `state=open` (open tasks only on page load), adds an Open/Closed toggle in the page header, prev/next pagination controls, and fixes the missing `merged`/`deploying`/`deployed` statuses in the dropdown
- **Badge colors** extended: green for merged/done/deployed, red for blocked/cancelled

## Test plan

- [ ] `task-store`: integration tests cover new `{ tasks, total, limit, offset }` response shape, `?state=open`, `?state=closed`, `?limit=2&offset=0`, and `?limit=2&offset=2`
- [ ] `task-store`: smoke tests updated to match new `list()` return type
- [ ] `admin`: smoke tests updated to match new `fetchTaskStoreTasks` return shape
- [ ] Typecheck passes (only pre-existing Prisma client environment error)
- [ ] Biome lint/format clean

### DIFF
--- a/admin/src/admin-ui.ts
+++ b/admin/src/admin-ui.ts
@@ -1543,7 +1543,7 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
         c.var.userEmail,
         agentNames,
         { total, limit, page },
-        error ? { error } : undefined,
+        { ...(error ? { error } : {}), agentFilterActive: agentFilterIds !== null },
       ),
     );
   });
--- a/admin/src/admin-ui.ts
+++ b/admin/src/admin-ui.ts
@@ -187,7 +187,12 @@ export interface AdminUIDeps {
    * Fetch tasks from the task-store service. If absent, the tasks page renders
    * in degraded mode (empty table + yellow notice) rather than returning 500.
    */
-  fetchTaskStoreTasks?: (params: URLSearchParams) => Promise<TaskItem[]>;
+  fetchTaskStoreTasks?: (params: URLSearchParams) => Promise<{
+    tasks: TaskItem[];
+    total: number;
+    limit: number;
+    offset: number;
+  }>;
   /**
    * Fetch a single task by ID from the task-store service. If absent, the
    * detail route redirects back to the list.
@@ -1452,10 +1457,22 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
   app.get("/admin/tasks", requireAuth, async (c) => {
     if (!c.var.isAdmin) return new Response("Forbidden", { status: 403 });
     const status = c.req.query("status") ?? undefined;
+    const stateRaw = c.req.query("state");
+    // Default to "open" when neither state nor status is provided.
+    const state: "open" | "closed" | undefined =
+      stateRaw === "open" || stateRaw === "closed"
+        ? stateRaw
+        : status
+          ? undefined
+          : "open";
     const session = c.req.query("session") ?? undefined;
     const repo = c.req.query("repo") ?? undefined;
     const agent = c.req.query("agent") ?? undefined;
     const error = c.req.query("error") ?? undefined;
+    const pageRaw = c.req.query("page");
+    const page = pageRaw ? Math.max(1, Number.parseInt(pageRaw, 10)) : 1;
+    const limit = 50;
+    const offset = (page - 1) * limit;
 
     // When filtering by agent name, resolve matching IDs upfront so we can
     // filter tasks client-side (task store only supports a single assignee ID).
@@ -1468,6 +1485,7 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
     }
 
     let tasks: TaskItem[] = [];
+    let total = 0;
     let degraded = false;
 
     if (!fetchTaskStoreTasks) {
@@ -1475,10 +1493,17 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
     } else {
       const params = new URLSearchParams();
       if (status) params.set("status", status);
+      if (state) params.set("state", state);
       if (session) params.set("session", session);
       if (repo) params.set("repo", repo);
+      // Agent-name filtering is done client-side, so we fetch a larger slice
+      // when an agent filter is active to avoid under-counting across pages.
+      params.set("limit", agentFilterIds !== null ? "500" : String(limit));
+      params.set("offset", agentFilterIds !== null ? "0" : String(offset));
       try {
-        tasks = await fetchTaskStoreTasks(params);
+        const result = await fetchTaskStoreTasks(params);
+        tasks = result.tasks;
+        total = result.total;
       } catch {
         degraded = true;
       }
@@ -1491,6 +1516,8 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
           (t.assignee && ids.has(t.assignee)) ||
           (t.claimedBy && ids.has(t.claimedBy)),
       );
+      total = tasks.length;
+      tasks = tasks.slice(offset, offset + limit);
     }
 
     const agentIds = [
@@ -1511,10 +1538,11 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
     return html(
       renderTasksPage(
         tasks,
-        { status, session, repo, agent },
+        { status, state, session, repo, agent },
         degraded,
         c.var.userEmail,
         agentNames,
+        { total, limit, page },
         error ? { error } : undefined,
       ),
     );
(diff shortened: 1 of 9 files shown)


dmcaulay (commented):
COMMENT — CI is failing due to a shape mismatch between `createTaskStoreClient().query()` and the new `GET /tasks` API response.

**Root cause** (`plugins/shipwright/scripts/check-helpers.ts:346`):

```typescript
async query(params: URLSearchParams): Promise<Task[]> {
  const res = await fetch(`${baseUrl}/tasks?${params}`, { headers });
  if (!res.ok) throw new Error(...);
  return res.json() as Promise<Task[]>;  // ← BUG
}
```

This PR changes `GET /tasks` to return `{ tasks, total, limit, offset }` (the new `TaskListResult` shape in `task-store/src/task-service.ts`). But `createTaskStoreClient().query()` — also added in this PR — still casts the response as `Promise<Task[]>`.

Callers in `check-deploy.ts` (lines ~316, ~393) and `check-dev-task.ts` (lines ~94-98) do:
```typescript
const prOpenTasks = await taskStore.query(new URLSearchParams({ status: 'pr_open' }));
if (prOpenTasks.length === 0) return;          // undefined, not 0
for (const task of prOpenTasks) { ... }        // TypeError: not iterable
```

The response is now a plain object, so `.length` is `undefined` and `for...of` throws. This is why CI tests fail.

**Fix** — unwrap the array in `query()` so callers are unaffected:
```typescript
async query(params: URLSearchParams): Promise<Task[]> {
  const res = await fetch(`${baseUrl}/tasks?${params}`, { headers });
  if (!res.ok) throw new Error(...);
  const result = await res.json() as { tasks: Task[] };
  return result.tasks;
}
```
Or alternatively, update the return type to expose `TaskListResult` and update all callers to destructure `.tasks`.

task-store/src/routes/tasks.ts
```diff
@@ -73,17 +74,26 @@ export function createTasksRoutes(
...
       assignee: agentId ?? c.req.query("assignee"),
       claimedBy: c.req.query("claimedBy"),
       pr: prRaw !== undefined ? Number.parseInt(prRaw, 10) : undefined,
       branch: c.req.query("branch"),
+      limit: limitRaw !== undefined ? Number.parseInt(limitRaw, 10) : undefined,
+      offset:
+        offsetRaw !== undefined ? Number.parseInt(offsetRaw, 10) : undefined,
     });
-    return c.json(tasks, 200);
+    return c.json(result, 200);
```
dodizzle, line 96: This is a breaking API shape change: `GET /tasks` previously returned `Task[]`; it now returns `{ tasks, total, limit, offset }`. `check-helpers.ts` was updated in this PR, but any other caller that destructures the response as an array (`.length`, `for...of`) will silently get `undefined`/throw after deploy.

Suggestion: add a migration note to `docs/migration.md` documenting the `Task[]` → `{ tasks, total, limit, offset }` change so operators know what to update before rolling out the new task-store image.

task-store/src/task-service.ts
```diff
@@ -54,19 +84,32 @@ export class TaskService implements TaskServiceLike {
 
   // ─── Reads ─────────────────────────────────────────────────────────────────
 
-  async list(filters: TaskListFilters = {}): Promise<Task[]> {
+  async list(filters: TaskListFilters = {}): Promise<TaskListResult> {
     const where: Prisma.TaskWhereInput = {};
     if (filters.status) where.status = filters.status as Task["status"];
+    if (filters.state === "open") where.status = { in: [...OPEN_STATUSES] };
```
dodizzle: `state` silently overwrites `status` — line 89 sets `where.status = filters.status`, then line 90 immediately overwrites it when `state === 'open'`. A direct API call `GET /tasks?status=pending&state=open` returns all open-status tasks, not just pending ones, with no error.

The admin UI avoids this by clearing `state` when `status` is set, but the task-store API route passes both through, making the public surface inconsistent.

Suggestion: return 400 in `routes/tasks.ts` when both `?status` and `?state` are provided, or skip the `state` filter when `status` is also set and document the precedence.

admin/src/admin-ui.ts
```diff
@@ -1468,17 +1485,25 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
...
       degraded = true;
     } else {
       const params = new URLSearchParams();
       if (status) params.set("status", status);
+      if (state) params.set("state", state);
       if (session) params.set("session", session);
       if (repo) params.set("repo", repo);
+      // Agent-name filtering is done client-side, so we fetch a larger slice
+      // when an agent filter is active to avoid under-counting across pages.
+      params.set("limit", agentFilterIds !== null ? "500" : String(limit));
```
dodizzle, line 1501: When agent-name filtering is active, this fetches at most 500 tasks and filters/paginates client-side. If more than 500 tasks match upstream (by state/repo/session), the computed total and page content silently reflect only the first 500 — no UI indicator warns the user.

Suggestion: add a notice when `agentFilterIds !== null` (e.g. "Showing up to 500 results for agent name filter"), or implement agent-name filtering server-side in the task-store API.

dodizzle (commented):
COMMENT — 3 findings (1 critical, 2 important).

**Critical**: `GET /tasks` response shape changed from `Task[]` to `{ tasks, total, limit, offset }` — breaking change for any client that expected a bare array. `check-helpers.ts` was updated, but a migration note in `docs/migration.md` is needed so operators know to update agents before deploying.

**Important (1)**: `?status` and `?state` silently conflict in the service layer — `state` overwrites `status` when both are passed to `taskService.list()`. The admin UI avoids this, but the public API surface is inconsistent.

**Important (2)**: Agent-name pagination fetches at most 500 tasks with no UI indicator when the result is approximate.

Strengths: state/status precedence in the admin UI handler is correct; the check-helpers fix is clean; `$transaction` for count+findMany is a sound pagination pattern.

**Verdict**: COMMENT — migration note for the breaking shape change + a 400 or precedence rule for the status/state conflict before merge.

dodizzle (commented):
APPROVE — all 3 prior findings addressed in `8120b2e5`:

1. **Migration note** added to `docs/migration.md` — documents the `Task[]` → `{ tasks, total, limit, offset }` shape change with a "What to update" checklist.
2. **Status/state conflict** fixed — `task-service.ts` now uses `if/else if`, so `status` takes precedence over `state` when both are provided. Comment documents the precedence.
3. **Agent filter truncation notice** added — banner shown when agent-name filtering is active: "Showing up to 500 results — agent name filter is applied client-side and may not reflect all matching tasks."

CI green. No new issues.

**Verdict**: APPROVE

admin/src/admin-ui.ts
```diff
@@ -1452,10 +1457,22 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
...
+        ? stateRaw
+        : status
+          ? undefined
+          : "open";
     const session = c.req.query("session") ?? undefined;
     const repo = c.req.query("repo") ?? undefined;
     const agent = c.req.query("agent") ?? undefined;
     const error = c.req.query("error") ?? undefined;
+    const pageRaw = c.req.query("page");
+    const page = pageRaw ? Math.max(1, Number.parseInt(pageRaw, 10)) : 1;
```
dmcaulay: `Math.max(1, Number.parseInt(pageRaw, 10))` returns `NaN` when `pageRaw` is non-numeric — `Math.max(1, NaN) === NaN` in JavaScript. This makes `offset` NaN, which serializes as `"NaN"` in URLSearchParams and reaches Prisma's `skip`, likely causing a 500.

Fix: `Math.max(1, Number.parseInt(pageRaw, 10) || 1)` — the `|| 1` coerces NaN to 1. Apply the same fix to `limit` and `offset` parsing in `task-store/src/routes/tasks.ts`.

dmcaulay (commented):
COMMENT — All prior findings from round 2 are correctly addressed (shape unwrap in check-helpers.ts, migration note, status/state precedence, agent filter notice). One new issue introduced in this round:

**`?page=abc` sends `NaN` to Prisma** (`admin/src/admin-ui.ts:1473`): `Math.max(1, NaN) === NaN` in JS — the guard fails for non-numeric strings. `offset` becomes `NaN`, serialized as `"NaN"` in URLSearchParams → Prisma `skip: NaN` → likely 500. Same pattern in `routes/tasks.ts` for `?limit` and `?offset`. Fix: `Math.max(1, Number.parseInt(pageRaw, 10) || 1)`. Low-impact (admin-auth-gated) but a correctness bug.

dodizzle (commented):
APPROVE — dmcaulay's round-3 finding addressed in `9e7edb03`: `?page=abc` NaN guard fixed in `admin-ui.ts` (`|| 1` fallback) and `?limit`/`?offset` NaN guard fixed in `routes/tasks.ts` (`|| undefined` falls back to service default). All 4 findings across 3 review rounds are resolved. CI green.

**Verdict**: APPROVE

dmcaulay (approved):
APPROVE — NaN guard correctly applied in all three parseInt call sites. `|| 1` for `?page` in admin-ui.ts and `|| undefined` for `?limit`/`?offset` in routes/tasks.ts both fall through to sensible defaults. CI green. All prior findings resolved.

PR #679, opened 2026-06-24
fix(ci): hold concurrency slot until chart bump PR merges

dodizzle:
## Problem

`auto-bump-chart.yml` exits immediately after enabling auto-merge, releasing the concurrency slot while the bump PR is still open. The next queued run then:

1. Checks out `main` — still carries the pre-bump chart version
2. Computes the same new branch name (e.g. `chore/chart-v1.6.20`)
3. Finds the branch already exists → **silently skips**

This dropped the chart bump for every release after the first in any burst. Confirmed lost: `task-store-v0.5.0`, `task-store-v0.6.0`, `task-store-v0.7.0` — none appear in the CHANGELOG despite being tagged and pushed.

## Fix

Add a **"Wait for PR to merge"** step at the end of the job that polls until the PR reaches `MERGED` or `CLOSED` state (up to 10 minutes). This holds the concurrency slot until the version bump is actually on `main`, so each queued run reads the correct post-merge chart version and computes a fresh increment.

On timeout (very slow CI), the step warns but exits 0 — worst case is the same behaviour as before this fix, not a broken workflow.

## What still needs doing

The three lost task-store bumps (`v0.5.0`, `v0.6.0`, `v0.7.0`) won't self-heal — the tag push events are gone. Those chart bumps need to be triggered manually (re-push the tags or run the bump script directly).

---

### DIFF
--- a/.github/workflows/auto-bump-chart.yml
+++ b/.github/workflows/auto-bump-chart.yml
@@ -269,9 +269,9 @@ jobs:
       #
       # By blocking here until the merge lands, we guarantee the next run always
       # reads the post-merge chart version and computes a fresh increment.
-      # Timeout is 10 minutes; on expiry we warn but do not fail — the worst case
-      # is that a subsequent run may still encounter burst-collapse, which is the
-      # same behaviour as before this fix.
+      # Timeout is 10 minutes; on expiry we exit 1 so the failure surfaces in the
+      # Actions UI — a chart bump PR that never merged within the window needs
+      # investigation (e.g. a required check stalled or auto-merge was disabled).
       - name: Wait for PR to merge
         if: steps.idempotency.outputs.skip != 'true' && steps.commit-push.outputs.push_skipped != 'true'
         env:
@@ -288,4 +288,5 @@ jobs:
             echo "  attempt ${i}/60: state=${STATE}"
             sleep 10
           done
-          echo "WARNING: timed out after 10 minutes — next queued run may encounter burst-collapse."
+          echo "ERROR: timed out after 10 minutes waiting for ${PR_URL} to merge."
+          exit 1
--- a/.github/workflows/auto-bump-chart.yml
+++ b/.github/workflows/auto-bump-chart.yml
@@ -39,8 +39,10 @@ on:
 
 # Queue concurrent bumps rather than cancelling them — if multiple tags land
 # at nearly the same time, each should produce its own chart bump PR. Using
-# cancel-in-progress: false ensures they queue up and run sequentially so
-# each run sees the version committed by the previous one.
+# cancel-in-progress: false ensures they queue up and run sequentially. The
+# "Wait for PR to merge" step at the end of the job holds the slot until the
+# merge actually lands on main, so the next queued run always reads the
+# post-merge chart version and computes a fresh increment.
 concurrency:
   group: chart-bump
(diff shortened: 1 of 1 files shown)


dodizzle (commented):
APPROVE — clean fix for the burst-collapse race.

The polling loop is correct: `if:` guards ensure we only wait when this run actually opened a PR, `DISPATCH_TOKEN` has the right scope, `gh pr view` accepts full PR URLs, and CLOSED is handled alongside MERGED. Timeout exits 0 with a warning — no worse than the pre-fix behavior.

**Suggestion**: `auto-bump-chart.yml:281` — consider `exit 1` on timeout instead of `exit 0` with a warning if you want the timeout to surface as a visible failure in the Actions UI. Not a blocker — the current behavior is explicitly documented and reasonable.

No new issues. CI green. Clean to merge.

.github/workflows/auto-bump-chart.yml
```diff
@@ -262,3 +258,34 @@ jobs:
...
+      # new-version branch name, finds the branch already exists, and silently
+      # skips — dropping that release's chart bump entirely (burst-collapse).
+      #
+      # By blocking here until the merge lands, we guarantee the next run always
+      # reads the post-merge chart version and computes a fresh increment.
+      # Timeout is 10 minutes; on expiry we warn but do not fail — the worst case
+      # is that a subsequent run may still encounter burst-collapse, which is the
+      # same behaviour as before this fix.
+      - name: Wait for PR to merge
+        if: steps.idempotency.outputs.skip != 'true' && steps.commit-push.outputs.push_skipped != 'true'
```
dmcaulay: If `gh pr create` fails mid-race (e.g. a concurrent run opened the same PR between the ls-remote check and this run's push — the exact race documented above), `steps.pr.outputs.pr_url` is empty but this step still runs because the `if` condition doesn't check `steps.pr.outcome`. `gh pr view ""` fails silently via `2>/dev/null || echo "UNKNOWN"`, and the step holds the concurrency slot for the full 10 minutes emitting UNKNOWN every 10s.

Add `&& steps.pr.outcome == 'success'` to guard against this:
```yaml
if: steps.idempotency.outputs.skip != 'true' && steps.commit-push.outputs.push_skipped != 'true' && steps.pr.outcome == 'success'
```

dmcaulay (commented):
COMMENT — one edge case in the `if` condition: if `gh pr create` fails mid-race (the race the code documents), `steps.pr.outputs.pr_url` is empty and the Wait step runs anyway, calling `gh pr view ""` which fails silently, holding the concurrency slot for the full 10-minute timeout unnecessarily. Add `&& steps.pr.outcome == 'success'` to the condition to skip the wait when the PR open step didn't succeed.

The burst-collapse fix itself is correct and the timeout-exit-0 behavior is exactly right.

dodizzle (commented):
APPROVE — my suggestion from the first review is addressed: `exit 1` on timeout now surfaces the condition visibly in the Actions UI.

Re: dmcaulay's concern about `steps.pr.outcome == 'success'` — moot in practice. The "Open bump PR" step has no `continue-on-error: true`, so a failure there fails the job immediately and the Wait step never executes. No guard needed.

CI is in-progress on this commit; the prior failure was on the merge-from-main commit `16227b8f`, not the code change. Clean to merge once CI passes.

PR #681, opened 2026-06-24
feat(task-store): add blockedBy computed field to task API responses

dodizzle:
## Summary

- Adds `blockedBy: BlockedByEntry[]` computed field to all task API responses, explaining exactly why a task isn't ready
- `GET /tasks` list and `GET /tasks/:id` single-task endpoints both return `blockedBy`
- Exports `TaskWithBlockedBy`, `BlockedByEntry`, and updated `TaskListResult` from `@shipwright/task-store`

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | All tasks include a `blockedBy` array (empty if nothing blocks) | ✅ MET |
| 2 | HITL block as `{ type: "hitl" }` when `hitl=true` and `hitlNotifiedAt=null` | ✅ MET |
| 3 | Dep blocks as `{ type: "dependency", id, status }` per unsatisfied dep | ✅ MET |
| 4 | Tasks with all deps satisfied + no HITL have `blockedBy: []` | ✅ MET |
| 5 | `blockedBy` present on both list and single-task (`GET /tasks/:id`) responses | ✅ MET |
| 6 | List response shape is `{ tasks: TaskWithBlockedBy[], total, limit, offset }` | ✅ MET |
| 7 | Single-task response shape is `TaskWithBlockedBy` | ✅ MET |
| 8 | Existing `?state`, `?status`, `?ready`, `?limit`, `?offset` behavior unchanged | ✅ MET |
| 9 | TypeScript types `TaskWithBlockedBy`, `BlockedByEntry`, `TaskListResult` exported | ✅ MET |

## Test Plan

- [x] 18 unit tests for `computeBlockedBy()` pure helper — all pass
- [x] 9 smoke tests for API responses (`GET /tasks` and `GET /tasks/:id`) — all pass
- [x] `blocked-by.ts` line coverage: 94.59%
- [x] Typecheck passes (`tsc --noEmit`, excluding expected Prisma client errors)
- [x] Biome lint: no issues

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -37,6 +37,7 @@ Each `Task` in the response now includes a `blockedBy` array describing why the 
 
 `BlockedByEntry` is one of:
 - `{ type: "hitl" }` — the task has `hitl: true` and `hitlNotifiedAt` is null (awaiting human confirmation)
+- `{ type: "hitl", notified: true }` — notification was already sent (`hitlNotifiedAt` is set); the task is still excluded from `?ready=true` but no agent-actionable block remains
 - `{ type: "dependency"; id: string; status: string }` — a dependency task is not satisfied (see dependency satisfaction rules in `docs/task-store.md`)
 
 When `blockedBy` is empty, the task is ready to execute (assuming it has `status: pending`).
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -4,9 +4,9 @@ Durable notes for breaking changes and the steps needed to migrate across versio
 
 ---
 
-## `GET /tasks` response shape change
+## `GET /tasks` and `GET /tasks/:id` response shape change
 
-**Version**: next (feat/task-filters)
+**Version**: next (feat/task-filters, feat/ts-api-blocked-by)
 
 `GET /tasks` previously returned a bare `Task[]`. It now returns an envelope:
 
@@ -14,10 +14,38 @@ Durable notes for breaking changes and the steps needed to migrate across versio
 { "tasks": Task[], "total": number, "limit": number, "offset": number }
 ```
 
+Each `Task` in the response now includes a `blockedBy` array describing why the task is not yet ready:
+
+```json
+{
+  "tasks": [
+    {
+      "id": "task-1",
+      "status": "pending",
+      ...
+      "blockedBy": [
+        { "type": "hitl" },
+        { "type": "dependency", "id": "dep-1", "status": "in_progress" }
+      ]
+    }
+  ],
+  "total": 1,
+  "limit": 50,
+  "offset": 0
+}
+```
+
+`BlockedByEntry` is one of:
+- `{ type: "hitl" }` — the task has `hitl: true` and `hitlNotifiedAt` is null (awaiting human confirmation)
+- `{ type: "dependency"; id: string; status: string }` — a dependency task is not satisfied (see dependency satisfaction rules in `docs/task-store.md`)
+
+When `blockedBy` is empty, the task is ready to execute (assuming it has `status: pending`).
+
 **What to update:**
 - Any code that calls `GET /tasks` and expects an array must unwrap `.tasks` from the response.
 - `check-helpers.ts` in the plugin is updated in this PR. Custom scripts or agents calling `/tasks` directly need the same fix.
-- `GET /tasks?ready=true` is unchanged — it still returns `Task[]`.
+- Code that checks task readiness should use the `blockedBy` array instead of duplicating dependency logic.
+- `GET /tasks?ready=true` is unchanged — it still returns `Task[]` (filtered to ready tasks only, with `blockedBy` also present).
 
 ---
 
--- a/task-store/src/blocked-by.ts
+++ b/task-store/src/blocked-by.ts
@@ -20,9 +20,20 @@ import { CLOSED_STATUSES } from "./statuses.ts";
 
 // ─── Types ────────────────────────────────────────────────────────────────────
 
-/** A single reason why a task is not yet ready. */
+/**
+ * A single reason why a task is not yet ready.
+ *
+ * `{ type: "hitl" }` — notification has not yet been sent; task is awaiting
+ * the HITL gate to be opened.
+ *
+ * `{ type: "hitl"; notified: true }` — notification was already sent
+ * (`hitlNotifiedAt` is set) so the agent-actionable block is cleared, but the
+ * task is still excluded from `listReady` because `hitl=true` tasks are never
+ * considered ready. Consumers MUST NOT infer that `blockedBy: []` means
+ * "this task is ready" — it only means there are no agent-actionable blocks.
+ */
 export type BlockedByEntry =
-  | { type: "hitl" }
+  | { type: "hitl"; notified?: true }
   | { type: "dependency"; id: string; status: string };
 
 // ─── Terminal statuses ────────────────────────────────────────────────────────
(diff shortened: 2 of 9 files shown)


task-store/src/task-service.ts
```diff
@@ -133,8 +147,12 @@ export class TaskService implements TaskServiceLike {
     return ready;
   }
 
-  async get(id: string): Promise<Task | null> {
-    return this.prisma.task.findUnique({ where: { id } });
+  async get(id: string): Promise<TaskWithBlockedBy | null> {
+    const task = await this.prisma.task.findUnique({ where: { id } });
+    if (!task) return null;
+    // Load all tasks for dependency resolution.
+    const allTasks = await this.prisma.task.findMany();
```
dodizzle: This fetches the entire task table to compute `blockedBy` for a single task. Since `task.dependencies` is already available (it was just returned by `findUnique`), scope the all-tasks query to only the IDs that matter:

```ts
const allTasks = task.dependencies?.length
  ? await this.prisma.task.findMany({ where: { id: { in: task.dependencies } } })
  : [];
```

This caps the query at O(deps) instead of O(all tasks), which matters when `GET /tasks/:id` is called frequently.

task-store/src/blocked-by.ts
```diff
@@ -0,0 +1,90 @@
...
+// ─── Types ────────────────────────────────────────────────────────────────────
+
+/** A single reason why a task is not yet ready. */
+export type BlockedByEntry =
+  | { type: "hitl" }
+  | { type: "dependency"; id: string; status: string };
+
+// ─── Terminal statuses (mirrors task-service.ts CLOSED_STATUSES) ──────────────
+
+const TERMINAL_STATUSES = new Set([
```
dodizzle: `TERMINAL_STATUSES` is a manual copy of `CLOSED_STATUSES` from `task-service.ts` (which imports from this file, so a direct import would be circular). The comment acknowledges this — but the drift risk is real: a future PR that adds a new terminal status to `CLOSED_STATUSES` won't automatically update this set, and `computeBlockedBy` will silently over-report blocks.

Suggested fix: extract `CLOSED_STATUSES` (and optionally `OPEN_STATUSES`) to a shared `task-store/src/statuses.ts` that both files can import from. Eliminates the circularity and the drift.

dodizzle (commented):
COMMENT — two issues to address before merge: (1) `get()` fetches the entire task table to resolve deps for a single task — scope the query to `task.dependencies` IDs instead. (2) `TERMINAL_STATUSES` in `blocked-by.ts` is a hand-rolled copy of `CLOSED_STATUSES` in `task-service.ts` — extract to a shared constants module to eliminate silent drift.

Core logic is clean: `computeBlockedBy` is a well-structured pure function, test coverage is thorough (18 unit + 9 smoke), and the deliberate divergence from `ready.ts` on cross-branch `pr_open` handling is correctly documented.

Full review: state/reviews/PR_REVIEW_681.md

task-store/src/ready.ts
```diff
@@ -27,6 +27,9 @@ export interface ReadyTaskLike {
   dependencies?: string[];
   pr?: number | null;
   hitl?: boolean | null;
+  /** ISO timestamp set when HITL notification was sent; null while awaiting. */
```
dmcaulay, line 30: `assignee` is not referenced anywhere in `blocked-by.ts`, `ready.ts`, or any other consumer of `ReadyTaskLike`. Only `hitlNotifiedAt` is actually used by `computeBlockedBy()`. Dead interface surface widens the required shape for future fakes/adapters without benefit — remove it.

task-store/src/blocked-by.ts
```diff
@@ -0,0 +1,85 @@
...
+ * @param allTasks  The full task list used to resolve dependency IDs.
+ */
+export function computeBlockedBy(
+  task: ReadyTaskLike,
+  allTasks: ReadyTaskLike[],
+): BlockedByEntry[] {
+  const blocks: BlockedByEntry[] = [];
+
+  // HITL gate: blocked when hitl=true and the notification hasn't been sent yet.
+  if (task.hitl === true && task.hitlNotifiedAt == null) {
```
dmcaulay: When `hitl === true && hitlNotifiedAt != null` (notification already sent), this returns `[]` — but `resolveReadyTasks` still skips all `hitl=true` tasks unconditionally. A consumer reading `blockedBy: []` will incorrectly infer the task is unblocked. At minimum, add a note to `migration.md` that `blockedBy` reflects agent-actionable blocks only (not passive waits). Alternatively emit `{ type: "hitl", notified: true }` so consumers can distinguish the two states.

dmcaulay (commented):
COMMENT — prior findings (get() dep query scope, TERMINAL_STATUSES drift) are both correctly addressed. Two new issues in the fix commit: (1) `assignee` was added to `ReadyTaskLike` but is not used anywhere in blocked-by or ready logic — dead interface surface. (2) `blockedBy: []` for `hitl=true && hitlNotifiedAt!=null` tasks is semantically misleading — those tasks are still excluded from `listReady` but consumers can't tell from `blockedBy` alone.

Core feature is solid: `computeBlockedBy` is clean, the statuses extraction is correct, test coverage is thorough (18 unit + 9 smoke), and the get() scoping fix is well-implemented.

Full review: state/reviews/PR_REVIEW_681.md

dodizzle (commented):
COMMENT — all prior findings resolved cleanly. One remaining doc gap: `migration.md` lists two `BlockedByEntry` shapes but omits the third (`{ type: "hitl", notified: true }`) introduced in the fix commits. Everything else is merge-ready: statuses extracted, get() scoped, assignee removed, hitl notified-state emitted and tested across all layers.

Full review: state/reviews/PR_REVIEW_681.md

docs/migration.md
```diff
@@ -4,20 +4,48 @@ Durable notes for breaking changes and the steps needed to migrate across versio
...
+    }
+  ],
+  "total": 1,
+  "limit": 50,
+  "offset": 0
+}
+```
+
+`BlockedByEntry` is one of:
+- `{ type: "hitl" }` — the task has `hitl: true` and `hitlNotifiedAt` is null (awaiting human confirmation)
```
dodizzle, line 39: `BlockedByEntry` has a third observable shape introduced in the fix commits but not listed here: `{ type: "hitl", notified: true }` — emitted when `hitl=true && hitlNotifiedAt != null` (notification already sent; task is still excluded from `?ready=true` but no agent-actionable block remains). Add it as a third bullet so consumers have the full type contract.

dmcaulay (approved):
APPROVE — R2 findings (unused `assignee` on `ReadyTaskLike`, misleading `blockedBy: []` for notified-HITL tasks) are both correctly resolved. `f7dfb5c` drops `assignee` and emits `{ type: "hitl", notified: true }` when `hitlNotifiedAt` is set; `3dbdc46` documents the third `BlockedByEntry` variant in `migration.md`. The "empty = ready" doc claim is now accurate since both HITL states emit entries. No new issues. Merge-ready.

Full review: state/reviews/PR_REVIEW_681.md

dodizzle (commented):
APPROVE — R3 finding (migration.md missing `{ type: "hitl", notified: true }` BlockedByEntry variant) is correctly addressed in `3dbdc46`. All three rounds of findings are fully resolved. CI green. No new issues. Merge-ready.

Full review: state/reviews/PR_REVIEW_681.md

PR #683, opened 2026-06-24
feat(plan-session): flag release pipeline changes and net-new CI secrets

dodizzle:
## Summary

Two gaps in `plan-session` that caused silent breakage during release_prep, now closed:

**Breaking Change Scan (Step 2)** — new bullet:
- **Release/deployment pipeline**: the pipeline is itself an interface with external consumers. Changes to how artifacts are built, packaged, or published; where they land; what events or signals are emitted; or what triggers downstream workflows can silently break systems in other repos. Those consumers won't appear in a local grep and won't error loudly — they'll just stop running. Identify what depends on the current pipeline behavior before designing any task that modifies it.

**HITL Detection (Step 5.5)** — new scan step:
- **CI workflow secret scan**: when a task adds or modifies a CI workflow file, extract every `${{ secrets.* }}` reference and check whether each name already appears in other workflow files. Any net-new secret requires human provisioning — flag HITL and list the secret names in `## Human steps`.

Also bumps plugin to **4.29.2**.

## Test plan
- [x] No repo-specific names in added text
- [x] Existing scan bullets and keyword list unchanged
- [x] Only `plan-session.md` and `plugin.json` modified


dmcaulay (approved):
APPROVE — clean docs-only change. The new bullet correctly identifies that CI/workflow consumers live in other repos and won't appear in a local grep, which is exactly the failure mode that makes this class of breaking change dangerous. The 'substitutions that change behavior' addition to the closing sentence is accurate. No blocking concerns.

dmcaulay (approved):
APPROVE — HITL additions are clean. The structural CI secrets scan (extract all `${{ secrets.* }}` references from changed workflow files and flag any net-new name) is a stronger heuristic than keyword matching and the right direction. Plugin version bump to 4.29.2 is correct in `plugin.json`.

One non-blocking note: `plugins/shipwright/README.md` heading still reads `v4.29.0` while `plugin.json` is now at `4.29.2` — this pre-exists the PR but is worth cleaning up.

PR #684, opened 2026-06-24
fix(check-helpers): handle both Task[] and paginated envelope in query()

dmcaulay:
## Summary

- `createTaskStoreClient().query()` crashed with `TypeError: {} is not iterable` when iterating the result of `getInProgressTasks()` — the task-store API returns `{ tasks, total, limit, offset }` for `?status=...` queries but a bare `Task[]` for `?ready=true`
- Fix detects both shapes: unwraps `.tasks` from the envelope, falls through for bare arrays, throws on anything else
- Bumps plugin to 4.29.2
- Adds 4 unit tests to `check-helpers.unit.test.ts` covering both response shapes and the error path

## Root cause

The API comment in `routes/tasks.ts` line 16 already documented the split:
> `returns { tasks, total, limit, offset } — or Task[] when ?ready=true`

But `query()` in `check-helpers.ts` only handled one shape at a time (first assumed bare array in 4.29.1, then assumed envelope in main). Neither version handled both.

## Test plan

- [ ] `bun test plugins/shipwright/scripts/check-helpers.unit.test.ts` — 29 pass
- [ ] Install 4.29.2 on Bodhi, confirm `check-dev-task.ts` preCheck exits 1 (no work) or 0 (work) without crashing

### DIFF
--- a/plugins/shipwright/scripts/check-helpers.unit.test.ts
+++ b/plugins/shipwright/scripts/check-helpers.unit.test.ts
@@ -506,7 +506,12 @@ describe("createTaskStoreClient query()", () => {
     globalThis.fetch = async () =>
       ({
         ok: true,
-        json: async () => ({ tasks: [FAKE_TASK], total: 1, limit: 50, offset: 0 }),
+        json: async () => ({
+          tasks: [FAKE_TASK],
+          total: 1,
+          limit: 50,
+          offset: 0,
+        }),
       }) as Response;
 
     const client = createTaskStoreClient();
--- a/plugins/shipwright/scripts/check-helpers.unit.test.ts
+++ b/plugins/shipwright/scripts/check-helpers.unit.test.ts
@@ -463,6 +463,7 @@ describe("createTaskStoreClient query()", () => {
   };
 
   let savedEnv: { url?: string; token?: string };
+  let savedFetch: typeof globalThis.fetch;
 
   beforeEach(() => {
     savedEnv = {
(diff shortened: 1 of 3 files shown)


dodizzle (commented):
COMMENT — the `query()` fix is correct and the coverage is solid. One convention violation: `globalThis.fetch` is overridden in tests 2–4 without cleanup — `mock.restore()` in `afterEach` only restores `mock.module()` registrations, not `globalThis.fetch`. CLAUDE.md prohibits global overrides because Bun shares the test process. Save and restore the original fetch in `beforeEach`/`afterEach`, or inject a fetch function into `createTaskStoreClient` for testing.

Full review: state/reviews/PR_REVIEW_684.md

plugins/shipwright/scripts/check-helpers.unit.test.ts
```diff
@@ -448,3 +450,101 @@ describe("getCurrentUser", () => {
...
+    }
+    mock.restore();
+  });
+
+  test("handles bare Task[] response (returned by ?ready=true)", async () => {
+    mock.module("node:fetch", () => ({}));
+    globalThis.fetch = (async () =>
+      ({
+        ok: true,
+        json: async () => [FAKE_TASK],
```
dodizzle, line 500: `globalThis.fetch` is assigned here (and in tests 3–5) but never restored. `mock.restore()` in `afterEach` only cleans up `mock.module()` registrations, not `globalThis.fetch` — the last assigned value leaks into subsequent test suites. CLAUDE.md rule: no `global.*` overrides (Bun shares the test process).

Fix: `const originalFetch = globalThis.fetch` in `beforeEach`, `globalThis.fetch = originalFetch` in `afterEach`. Or inject a fetch parameter into `createTaskStoreClient` for clean testability.

dodizzle (approved):
APPROVE — R1 finding (globalThis.fetch leaks into sibling test suites per CLAUDE.md isolation rule) is correctly resolved in `d56acd3`: `savedFetch` is declared in the describe scope, captured in `beforeEach`, and restored in `afterEach` across all four tests. CI green. No new issues.

Full review: state/reviews/PR_REVIEW_684.md

PR #688, opened 2026-06-25
feat(plan-session): scan open cross-session tasks as potential deps

dmcaulay:
## Summary

- Step 1 of `plan-session` now queries the task store for tasks in non-terminal statuses (`pending`, `in_progress`, `blocked`, `pr_open`) from **prior sessions** before the orientation header
- If any exist, they are surfaced with ID, status, title, and session slug so the planner can wire them into the dependency map
- Updates the `Dependencies` bullet in Step 5 to explicitly allow cross-session task IDs
- Updates the visual graph template to show a `[PRIOR SESSIONS]` node when applicable

## Problem

Previously, `plan-session` only checked for existing tasks in the **current** session (to avoid duplicates). It was blind to tasks from earlier sessions still in flight — new tasks could have implicit dependencies the execution cron had no way to enforce.

## Test plan

- [ ] Run `plan-session` when the task store has open tasks from a prior session — verify the cross-session block appears before the orientation header
- [ ] Run `plan-session` with an empty store or all tasks terminal — verify the silent path (no extra output)
- [ ] Reference a cross-session task ID in a new task's `dependencies` — verify readiness logic blocks it until the prior task reaches terminal status


dodizzle (approved):
APPROVE — focused docs change that closes a real gap in cross-session dependency visibility. The four-status query pattern, jq merge+filter, and silent-on-empty path are all correct. Visual graph template and dependency description updated in lockstep. CI failure is an infra rate-limit flake unrelated to this change (single markdown file). One suggestion: `{session}` in the jq `--arg` at line 57 is inconsistent with `$SESSION` in the adjacent Step 1.3 curl — minor readability rough spot, not a blocker.

Full review: state/reviews/PR_REVIEW_688.md

plugins/shipwright/commands/plan-session.md
```diff
@@ -43,7 +43,28 @@ This is the engineering planning pass. The product spec (what and why) is alread
...
+   {
+     curl -sf -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
+       "$SHIPWRIGHT_TASK_STORE_URL/tasks?status=pending" | jq '.tasks // []'
+     curl -sf -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
+       "$SHIPWRIGHT_TASK_STORE_URL/tasks?status=in_progress" | jq '.tasks // []'
+     curl -sf -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
+       "$SHIPWRIGHT_TASK_STORE_URL/tasks?status=blocked" | jq '.tasks // []'
+     curl -sf -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
+       "$SHIPWRIGHT_TASK_STORE_URL/tasks?status=pr_open" | jq '.tasks // []'
+   } | jq -s --arg s "{session}" 'add // [] | map(select(.session != $s)) | unique_by(.id)'
```
dmcaulay: This four-status query misses `approved` — a task with a reviewed PR awaiting deploy is not terminal and could be a valid prerequisite from a prior session.

The task-store API already supports `?state=open` (maps to all five open statuses including `approved`). Suggest replacing the four curls with:
```bash
curl -sf -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
  "$SHIPWRIGHT_TASK_STORE_URL/tasks?state=open" | jq '.tasks // []'
```
Then pipe to `jq -s --arg s "$SESSION" 'add // [] | map(select(.session != $s)) | unique_by(.id)'`. One query, complete coverage.

dmcaulay (commented):
COMMENT — one finding to address before merging.

**`approved` status missing from cross-session scan**: The four-status query covers `pending`, `in_progress`, `blocked`, and `pr_open`, but the task store's `OPEN_STATUSES` also includes `approved`. A task in `approved` status has a reviewed PR awaiting deploy — it's not terminal and could be a valid dependency. The task-store API already supports `?state=open` which maps to all five open statuses; using that instead of four separate curls would be both correct and simpler.

Full review: state/reviews/PR_REVIEW_688_shipwright.md

dodizzle (approved):
APPROVE — both prior findings addressed. `?state=open` is correct and API-verified (maps to all five `OPEN_STATUSES` including `approved`, covered by an integration test), and simplifies the implementation from four separate curls to one. `$SESSION` consistency fix also landed. Clean.

Full review: state/reviews/PR_REVIEW_688.md

dmcaulay (commented):
APPROVE — prior finding addressed. The `?state=open` query covers all five open statuses (including `approved`), the jq filter correctly handles the paginated envelope and client-side session exclusion, and CI is green. Good to merge.

Full review: state/reviews/PR_REVIEW_688_shipwright.md

PR #692, opened 2026-06-25
feat(task-store): validate org/repo format on task and token writes

dmcaulay:
## Summary
- Adds `isOrgRepo(s)` validator in `task-store/src/validate.ts` — enforces `org/repo` format (exactly one slash, non-empty both sides)
- Applies format validation to `POST /tokens` (per-entry in `repos` array), `POST /tasks`, `POST /tasks/bulk` (per-task), and `PATCH /tasks/:id`
- Adds scope check for agent tokens: if `repo` is present it must be in the token's `repos` list; admin tokens bypass scope but still get format validation

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | POST /tokens with repos: ['vitals-os'] returns 400 | ✓ MET |
| 2 | POST /tokens with repos: ['app-vitals/vitals-os'] returns 201 | ✓ MET |
| 3 | POST /tasks with repo: 'vitals-os' returns 400 (any token) | ✓ MET |
| 4 | POST /tasks with repo: 'app-vitals/vitals-os' + agent scoped to 'app-vitals/shipwright' returns 400 | ✓ MET |
| 5 | POST /tasks with repo: 'app-vitals/vitals-os' + agent scoped to 'app-vitals/vitals-os' returns 201 | ✓ MET |
| 6 | POST /tasks without repo field + agent token returns 201 | ✓ MET |
| 7 | POST /tasks/bulk validates per-task; skips null/absent repo | ✓ MET |
| 8 | Admin tokens bypass scope check but still get format validation | ✓ MET |
| 9 | All cases covered as smoke tests | ✓ MET (18 smoke tests) |

## Test Plan
- [x] 18 new smoke tests in `validate.smoke.test.ts` covering all acceptance criteria
- [x] 77 tests pass, 0 fail
- [x] `bun run --filter='*' typecheck` — all packages clean

### DIFF
--- a/task-store/src/routes/tokens.ts
+++ b/task-store/src/routes/tokens.ts
@@ -13,8 +13,9 @@
 
 import { Hono } from "hono";
 import type { TaskStoreAuthEnv } from "../auth.ts";
-import { ForbiddenError, NotFoundError } from "../errors.ts";
+import { BadRequestError, ForbiddenError, NotFoundError } from "../errors.ts";
 import type { TokenServiceLike } from "../token-service.ts";
+import { isOrgRepo } from "../validate.ts";
 
 export function createTokensRoutes(
   tokenService: TokenServiceLike,
@@ -43,10 +44,21 @@ export function createTokensRoutes(
       const body = (await c.req.json()) as {
         label?: unknown;
         agentId?: unknown;
+        repos?: unknown;
       };
       if (typeof body.label === "string") label = body.label;
       if (typeof body.agentId === "string") agentId = body.agentId;
-    } catch {
+      if (Array.isArray(body.repos)) {
+        for (const entry of body.repos) {
+          if (typeof entry !== "string" || !isOrgRepo(entry)) {
+            throw new BadRequestError(
+              `invalid repos entry: "${entry}" — expected "org/repo" format`,
+            );
+          }
+        }
+      }
+    } catch (err) {
+      if (err instanceof BadRequestError) throw err;
       // No body / invalid JSON → create an unlabeled admin token.
     }
     const { token, rawToken } = await tokenService.create(label, agentId);
--- a/task-store/src/routes/tokens.ts
+++ b/task-store/src/routes/tokens.ts
@@ -13,9 +13,8 @@
 
 import { Hono } from "hono";
 import type { TaskStoreAuthEnv } from "../auth.ts";
-import { BadRequestError, ForbiddenError, NotFoundError } from "../errors.ts";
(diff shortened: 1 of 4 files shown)


task-store/src/validate.ts
```diff
@@ -0,0 +1,5 @@
+/** Returns true when `s` is a valid "org/repo" string with non-empty org and repo parts. */
+export function isOrgRepo(s: string): boolean {
+  const parts = s.split("/");
+  return parts.length === 2 && parts[0].length > 0 && parts[1].length > 0;
```
dmcaulay: `parts[0].length > 0` passes for whitespace-only strings like `' '`. A value like `' /repo'` or `'org/ '` would pass this check but reference a nonexistent repo. Add `.trim()` before each length check:
```typescript
return parts.length === 2 && parts[0].trim().length > 0 && parts[1].trim().length > 0;
```

dmcaulay (commented):
COMMENT — two findings to address before merging.

**`isOrgRepo` accepts whitespace-only segments**: `parts[0].length > 0` passes for `' '` (a single space), so values like `' /repo'` or `'org/ '` would slip through format validation but reference nonexistent repos. Fix: add `.trim()` before the length checks.

**`POST /tokens` validates `repos` but silently discards it**: The `repos` array is parsed and format-validated, but never passed to `tokenService.create()` and not persisted to `TaskToken`. A caller sending `repos: ['app-vitals/vitals-os']` expecting a scoped token receives an admin token with no repo restriction, indistinguishable from one created without `repos`. Either defer this field until persistence is implemented, or make the silent drop explicit in the response.

Full review: state/reviews/PR_REVIEW_692.md

task-store/src/routes/tokens.ts
```diff
@@ -43,10 +44,21 @@ export function createTokensRoutes(
       const body = (await c.req.json()) as {
         label?: unknown;
         agentId?: unknown;
+        repos?: unknown;
```
dmcaulay: `repos` is parsed and format-validated here, but never forwarded to `tokenService.create()` — there's no `repos` parameter and no `TaskToken.repos` column. The 201 response gives no indication the field was ignored. A caller expecting a repo-scoped token gets a full admin token instead. Either hold off validating `repos` until persistence lands, or explicitly document in the response that `repos` is not yet enforced (e.g., return `repos: null` in the token response body).

dmcaulay (commented):
APPROVE — both prior findings are resolved. `isOrgRepo` now trims whitespace before length checks. The `repos` field in `POST /tokens` has an explicit comment documenting deferred persistence. CI green.

PR #707, opened 2026-06-25
feat: render blockers and markdown on task detail view

dodizzle:
## Summary
- Add `blockedBy` field to `TaskItem` interface covering both `hitl` and `dependency` blocker types
- Render task description and acceptance criteria as markdown (headings, lists, code blocks, inline code, bold) with XSS-safe HTML escaping applied first
- Show a Blockers section on the task detail view when `blockedBy` is non-empty

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Detail view shows a blockers section when blockedBy is non-empty | ✅ MET |
| 2 | Each blocker entry shows its type and, for dep blockers, the dep ID and current status | ✅ MET |
| 3 | Description field is rendered as markdown (headings, lists, code blocks) | ✅ MET |
| 4 | Acceptance criteria items are rendered as markdown | ✅ MET |
| 5 | Plain text fields (title, id, status, etc.) are unaffected | ✅ MET |

## Test Plan
- [x] 129 new unit tests covering blockers section (show/hide, dep ID/status, HITL variants, XSS escaping) and markdown rendering (headings, code blocks, inline code, lists, bold, XSS)
- [x] Biome lint passes on changed files
- [x] 92.73% line coverage across admin package (above 90% threshold)


admin/src/admin-ui-pages.ts
```diff
@@ -109,7 +109,101 @@ export interface TaskItem {
...
+  out = out.replace(/```[\r\n]([\s\S]*?)[\r\n]```/g, (_m, code) => {
+    return `<pre><code>${code}</code></pre>`;
+  });
+  // Also handle ``` on same line as content: ```code```
+  out = out.replace(/```([^`\n]+)```/g, (_m, code) => {
+    return `<pre><code>${code}</code></pre>`;
+  });
+
+  // Step 3: process line-by-line for block-level elements
+  const lines = out.split("\n");
```
dodizzle, line 150: **Multi-line code block content leaks into the line processor.** The regex on line 127 replaces fenced blocks with `<pre><code>…</code></pre>` before this `split("\n")`. That means interior lines of the replaced block (which still contain their original content, including newlines) are handed to the loop below. A line starting with `- ` or `* ` inside a multi-line code block matches `/^[-*] /` and gets wrapped in `<li>…</li>`, breaking the `<pre>` block and producing malformed HTML.

Confirmed with: `\`\`\`\n- item one\n- item two\n\`\`\`` → `<pre><code>- item one\n<ul>\n<li>item two</code></pre></li>\n</ul>`

Fix: track whether the current output is inside a pre block (e.g. use a placeholder token approach — replace fenced blocks with `\x00CODE_BLOCK_N\x00` before the line loop, run the loop, then swap placeholders back), or collect the byte ranges consumed by the code-block regex and skip those lines in the loop.

dodizzle (commented):
COMMENT — 1 finding.

Multi-line code block content leaks into the line-level markdown processor: the code-block regex (line 127) replaces fenced blocks with `<pre><code>…</code></pre>` before `out.split("\n")`, so interior lines of the pre block are handed to the line loop. Any line inside a code block that starts with `- ` or `* ` matches the list-item test and gets wrapped in `<li>`, producing broken HTML.

**Self-review** — event set to COMMENT per GitHub API constraint.

*Review file*: state/reviews/PR_REVIEW_707.md

dodizzle (commented):
APPROVE — prior finding addressed.

The code-block interior line leak (finding #1 from the last review) is fixed: fenced blocks are now extracted into a `codeBlocks[]` array with Unicode PUA sentinel tokens before the line-by-line loop, so `- item` lines inside ` ``` ` blocks never reach the list-item matcher. The regression test at `renderTaskDetailPage — markdown / multi-line code block: interior lines starting with '- ' are not wrapped in <li>` confirms the fix. CI is green.

**Self-review** — event set to COMMENT per GitHub API constraint.

*Review file*: state/reviews/PR_REVIEW_707.md

dodizzle (commented):
APPROVE — biome lint compliance fix; no behavior changes, no new concerns.

The commit replaces `\x00` null-byte placeholder sentinels with Unicode PUA chars (U+E000/U+EFFF) to satisfy Biome's `noControlCharactersInRegex` rule, fixes `parseInt` → `Number.parseInt` (`useNumberNamespace`), and refactors placeholder detection to use a captured group from `PLACEHOLDER_RE.exec()` instead of a separate replace call. All three changes are purely mechanical — the placeholder mechanism introduced in `92b8e3b` is functionally unchanged and the prior code-block interior line leak finding remains addressed.

**Self-review** — event set to COMMENT per GitHub API constraint.

*Review file*: state/reviews/PR_REVIEW_707.md

PR #711, opened 2026-06-25
feat: add PullRequest model to task-store schema (PRI-1.1)

dmcaulay:
## Summary
- Add `PullRequest` model to `task-store/prisma/schema.prisma` with `PrState` and `PrReviewState` enums
- Additive migration creates the new table with `@@unique([repo, prNumber])` at DB level
- Integration tests verify all defaults and unique constraint rejection

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Migration is additive — no existing Task or TaskToken data affected | MET |
| 2 | @@unique([repo, prNumber]) enforced at DB level | MET |
| 3 | Defaults: state=open, reviewState=pending, patchCycles=0, staged=false | MET |
| 4 | Prisma client regenerated and TypeScript compiles cleanly | MET |
| 5 | Integration test covering defaults + unique constraint | MET |

## Test Plan
- Integration tests in `task-store/src/pull-request.integration.test.ts` gate on `DATABASE_URL_SHIPWRIGHT_TASK_STORE_TEST` (consistent with existing pattern)
- Full suite: 2791 pass, 121 skip, 0 fail
- Lint (Biome): 312 files, no issues
- Typecheck: all 5 packages clean


dodizzle (commented):
COMMENT — 1 finding (PRI-1.1).

The schema and migration are clean and fully additive. `@@unique([repo, prNumber])` is enforced at DB level, integration tests follow the established pattern, and all defaults are verified by the round-trip test.

**One blocking issue**: `PullRequest` types are not exported from the package index (`task-store/src/index.ts`). The file exports `Task`, `TaskToken`, and `TaskStatus` explicitly — but not `PullRequest`, `PrState`, or `PrReviewState`. Callers that depend on `@shipwright/task-store` (as the package comments encourage) won't be able to type PR records without reaching into the raw Prisma output path.

Fix: add to `task-store/src/index.ts`:
```ts
export type { PullRequest } from "../prisma/client/index.js";
export { PrState, PrReviewState } from "../prisma/client/index.js";
```

Two non-blocking suggestions:
- Schema header comment still says `// Owns: Task, TaskToken.` — should include `PullRequest`.
- `staged: Boolean` and `reviewState: PrReviewState` both appear to encode the "review written but not yet posted" state. Document which is authoritative before the API routes land.

*Review file*: state/reviews/PR_REVIEW_711.md

dmcaulay (commented):
COMMENT — 1 finding (PRI-1.1).

Migration is strictly additive, `@@unique([repo, prNumber])` is correctly enforced at DB level, all defaults are set, integration tests cover defaults + unique constraint, and all five acceptance criteria are met. Type exports are clean and complete.

**One issue**: `docs/architecture.md` row D was updated to mention `PullRequest` but the root `CLAUDE.md` (line 18) still says "Prisma schema defines `Task` and `TaskToken` models." CLAUDE.md is auto-loaded into every agent session — leaving it stale gives contributors and agents an inaccurate schema picture. Fix: change row D in `CLAUDE.md` to match `docs/architecture.md`: "Postgres-backed task queue, PR tracking, and scoped tokens. Prisma schema defines `Task`, `PullRequest`, and `TaskToken` models."

*Review file*: state/reviews/PR_REVIEW_711.md

dmcaulay (commented):
COMMENT — 3 findings (PRI-1.2 re-review, R2).

PRI-1.2 implements PullRequestService and /prs routes — the claim/heartbeat/complete/patch/release lifecycle is solid, atomic transaction logic is correct, and scope enforcement for agent tokens is applied consistently. Smoke and integration test coverage is thorough.

**Prior findings resolution**: dodizzle R1's two blocking items (missing type exports, schema header comment) are both addressed. Two items from R1 remain open:

1. **CLAUDE.md row D still stale** — flagged in both R1 reviews, still not updated. Row D in the root `CLAUDE.md` (line 18) reads "Postgres-backed task queue + scoped tokens. Prisma schema defines `Task` and `TaskToken` models" — no mention of `PullRequest`. Fix: update row D to "Postgres-backed task queue, PR tracking, and scoped tokens. Prisma schema defines `Task`, `PullRequest`, and `TaskToken` models."

2. **`staged` vs `reviewState` semantics undocumented** — still unclear which is authoritative for deploy-gating decisions after two review rounds. See inline comment.

**New issue in PRI-1.2**: Concurrent INSERT race in `claim()` surfaces as a 500 instead of 409 — see inline comment.

*Review file*: state/reviews/PR_REVIEW_711.md

task-store/src/pull-request-service.ts
```diff
@@ -0,0 +1,255 @@
...
+            claimedAt: now,
+            heartbeatAt: now,
+            ...(taskId !== undefined ? { taskId } : {}),
+          },
+        });
+        return { status: 200 as const, record };
+      }
+
+      // No existing record → create
+      const record = await tx.pullRequest.create({
```
dmcaulay: **Concurrent INSERT race returns 500, not 409**

When two agents race on an empty table, both execute `findUnique → null`, then both attempt `tx.pullRequest.create()`. The losing INSERT throws Prisma P2002 (unique constraint violation). `translateNotFound()` only maps P2025 — P2002 propagates unhandled and `onError` returns a 500. The integration test comment at line 118 says "The service maps P2002 → ConflictError on the transaction retry path" but this mapping doesn't exist.

Fix:
```ts
try {
  const record = await tx.pullRequest.create({ data: { ... } });
  return { status: 201 as const, record };
} catch (err: unknown) {
  if (
    typeof err === 'object' && err !== null &&
    'code' in err && (err as { code: string }).code === 'P2002'
  ) {
    throw new ConflictError(`pr ${repo}#${prNumber} is already claimed`);
  }
  throw err;
}
```

task-store/prisma/schema.prisma
```diff
@@ -92,6 +92,55 @@ model Task {
...
+// Tracks a GitHub PR through the Shipwright review → patch → deploy pipeline.
+// repo is in org/repo format (e.g. "app-vitals/shipwright").
+// @@unique([repo, prNumber]) ensures one record per GitHub PR.
+
+model PullRequest {
+  id          String        @id @default(cuid())
+  repo        String        // org/repo format
+  prNumber    Int
+  taskId      String?
+  staged      Boolean       @default(false)
```
dmcaulay: **`staged` vs `reviewState` — document which is authoritative for deploy-gating**

This was flagged in dodizzle R1 and remains unaddressed. Both fields could signal "review written but not yet posted". A few words clarifying their distinct roles would remove ambiguity for callers:

```prisma
// staged: true when the review body is written locally but not yet posted to GitHub.
//   Cleared by the review skill after a successful POST. Orthogonal to reviewState.
staged      Boolean       @default(false)
...
// reviewState: review pipeline lifecycle state.
//   posted = GitHub review submitted; staged may still be true until deploy skill clears it.
reviewState PrReviewState @default(pending)
```

task-store/src/pull-request-service.ts
dodizzle: Concurrent INSERT race: if two requests both read null from findUnique and both reach this create(), the second will throw a Prisma P2002 unique constraint violation. translateNotFound only catches P2025 — P2002 falls through as a 500, not a 409. The integration test comment says the service maps P2002 to ConflictError on the retry path but this mapping does not exist in the code. Fix: add P2002 handling alongside P2025 in translateNotFound and rethrow as ConflictError. Also wrap the create() call in a try/catch that uses the same helper.

dodizzle (commented):
COMMENT — 3 findings (re-review R3).

Prior findings resolved: Type exports (PrState, PrReviewState, PullRequest) are now exported from index.ts. CLAUDE.md row D is updated. staged vs reviewState semantics are now documented in the schema comment. The PullRequestService and /prs routes implementation (PRI-1.2) is solid — atomic transaction logic, Clock injection, scope enforcement for agent tokens, and thorough smoke and integration test coverage.

Three issues remain:

1. Concurrent INSERT race surfaces as 500 not 409 (carried from R3, not yet addressed). Two concurrent requests both reading null from findUnique will both attempt tx.pullRequest.create(). The second hits the @@unique([repo, prNumber]) constraint and Prisma throws P2002. translateNotFound only catches P2025 — P2002 falls through as a 500. The integration test comment says the service maps P2002 to ConflictError on the retry path but no such mapping exists in the code. Fix: catch P2002 in translateNotFound and rethrow as ConflictError.

2. PATCH /prs/:id accepts unrestricted field writes. The body is passed directly to prService.update() with no field allowlist. Any authenticated caller can set reviewState, state, staged, claimedBy, patchCycles, etc., bypassing the lifecycle endpoints that set dependent fields atomically. Fix: define an explicit allowlist (e.g. staged, commitSha, taskId, agentId) and reject other fields.

3. patching enum value is dead code. PrReviewState includes patching but patch() transitions to pending, not patching. No lifecycle endpoint ever sets this value. Either remove it from the enum and migration, or have patch() transition to patching.

Review file: state/reviews/PR_REVIEW_711.md

task-store/src/pull-request-service.ts
dmcaulay: Concurrent INSERT race: two agents on an empty table both hit `findUnique → null`, then both attempt this `create`. The loser gets a Prisma P2002 unique-constraint violation. `translateNotFound()` only maps P2025, so P2002 propagates to `onError` as a raw 500.

Fix:
```ts
try {
  const record = await tx.pullRequest.create({ data: { ... } });
  return { status: 201 as const, record };
} catch (err: unknown) {
  if (typeof err === 'object' && err !== null && 'code' in err &&
      (err as { code: string }).code === 'P2002') {
    throw new ConflictError(`pr ${repo}#${prNumber} is already claimed`);
  }
  throw err;
}
```

Also tighten the integration test at line 100–145: assert the rejected promise rejects with `ConflictError` rather than allowing both callers to succeed.

dmcaulay (commented):
COMMENT — 1 finding (PRI-1.1 re-review, R3).

Two documentation gaps from R2 are now resolved: CLAUDE.md row D correctly lists `Task`, `PullRequest`, and `TaskToken`, and the schema has a clear comment block explaining `staged` vs `reviewState` semantics. CI is green.

**One issue remains**: concurrent `claim()` calls on a fresh `(repo, prNumber)` pair surface as HTTP 500 instead of 409. Two agents racing on an empty table both read `null` from `findUnique`, then both attempt `tx.pullRequest.create()`. The losing INSERT throws a Prisma P2002 unique-constraint violation. `translateNotFound()` only maps P2025; P2002 propagates unhandled and the `onError` hook returns a raw 500. The integration test comment at line 115–119 describes the P2002 → ConflictError mapping as expected behavior, but the implementation doesn't deliver it and the assertion (`fulfilled.length >= 1`) is too lenient to catch the gap.

Fix: wrap the `create` call in a try/catch that maps P2002 to ConflictError, then tighten the integration test to assert the losing caller rejects with ConflictError.

*Review file*: state/reviews/PR_REVIEW_711.md

dodizzle (approved):
APPROVE — all three R3 findings addressed (R4 re-review).

All three prior blocking issues from R3 are cleanly resolved in `df31138`:

1. **P2002 concurrent INSERT race**: `claim()` now wraps `tx.pullRequest.create()` in a try/catch that maps Prisma P2002 to `ConflictError(409)`. The integration test is tightened to assert `fulfilled.length === 1`, `rejected.length === 1`, and that the loser gets `ConflictError` — not an overly-lenient `toBeGreaterThanOrEqual(1)` assertion.

2. **PATCH allowlist**: `PATCH_ALLOWED_FIELDS = ["staged", "commitSha", "taskId", "agentId"]` is enforced via a filter loop that silently drops non-allowlisted fields and returns 400 if nothing allowlisted was provided. `reviewState`, `state`, `claimedBy`, `patchCycles`, and other lifecycle-owned fields can no longer be written via PATCH.

3. **`patching` dead enum value**: Removed from `PrReviewState` in both `schema.prisma` and the migration SQL.

One non-blocking suggestion: a smoke test asserting that `PATCH /prs/:id` with `{ reviewState: 'posted' }` (non-allowlisted) returns 400 would prevent silent regression if the allowlist changes. Not holding up the merge.

*Review file*: state/reviews/PR_REVIEW_711.md

dmcaulay (commented):
APPROVE — R4 (PRI-1.1). All R3 blocking findings resolved: P2002 race in claim() now maps to ConflictError(409), PATCH /prs/:id enforces field allowlist [staged, commitSha, taskId, agentId], patching enum removed from PrReviewState. CI green, acceptance criteria met, test coverage thorough.

*Review file*: state/reviews/PR_REVIEW_711.md

PR #730, opened 2026-06-25
fix(admin): handle bare-array response from ready/blocked task endpoints

dodizzle:
## Summary

- `listReady()` and `listBlocked()` in the task-store return bare `Task[]`, not the `{ tasks, total }` wrapped shape
- The admin UI always destructured `.tasks`/`.total`, causing a 500 on every default page load (ready filter) and blocked filter
- Fix is admin-side only: normalize the raw result with `Array.isArray(rawResult) ? { tasks: rawResult, total: rawResult.length } : rawResult`
- Task-store contract for agents is untouched — they depend on the bare-array shape from `?ready=true`

**Caveat:** `total` on ready/blocked views reflects the page length, not a server-side count — pagination footer will be approximate on those views. Pre-existing behavior, not introduced by this fix.

## Test plan

- [ ] Visit the admin tasks page (default view = ready) — should no longer 500
- [ ] Switch to the blocked filter — should render without error
- [ ] Switch to in_progress / closed — still works as before

### DIFF
--- a/admin/src/admin-ui.smoke.test.ts
+++ b/admin/src/admin-ui.smoke.test.ts
@@ -2478,6 +2478,56 @@ describe("admin UI — tasks page", () => {
       "blocked",
     );
   });
+
+  it("GET /admin/tasks renders without 500 when task-store returns bare array for ready state", async () => {
+    const mockTasks = [
+      {
+        id: "task-1",
+        title: "Build auth module",
+        status: "pending",
+        session: "session-abc",
+        repo: "example-org/example-repo",
+        assignee: null,
+        claimedBy: null,
+      },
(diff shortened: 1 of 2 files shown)


dodizzle (commented):
## Review: fix(admin): handle bare-array response from ready/blocked task endpoints

Clean fix. The `Array.isArray` normalization in `admin-ui.ts` is minimal and surgical — one guard, no side-effects on any other path. Both new smoke tests land at the right layer with in-process `app.request()` calls. The `total`-as-page-length approximation on ready/blocked views is pre-existing and correctly noted in the PR body.

No blocking issues. ✓

PR #733, opened 2026-06-25
fix(task-store): unify list response shape — ready/blocked now return { tasks, total }

dodizzle:
## Summary

- All `GET /tasks` endpoints now return a consistent `{ tasks, total }` envelope
- Previously `?ready=true` and `?state=blocked` returned bare `Task[]`, causing the admin UI to 500 on every default page load (fixed in #730 with a normalization shim)
- This PR is the proper fix: the task-store itself always returns the wrapped shape, and all callers are cleaned up

## Changes

| File | Change |
|------|--------|
| `task-store/src/routes/tasks.ts` | Wrap `listReady()`/`listBlocked()` results in `{ tasks, total }` before responding |
| `admin/src/admin-ui.ts` | Remove `Array.isArray` normalization added in #730 (now dead code) |
| `plugins/shipwright/scripts/check-helpers.ts` | Remove bare-array branch from `query()` helper |
| `task-store/src/state-filter.smoke.test.ts` | Update ready/blocked assertions to expect envelope shape |
| `task-store/src/api.smoke.test.ts` | Same |
| `task-store/src/api.integration.test.ts` | Same |
| `plugins/shipwright/scripts/check-helpers.unit.test.ts` | Replace bare-array test with envelope test |
| `plugins/shipwright/commands/dev-task.md` | `jq .` → `jq '.tasks'` |
| `plugins/shipwright/commands/refresh-plan.md` | Same |
| `plugins/shipwright/skills/task-store/SKILL.md` | Update docs to reflect unified shape |

## Notes

- `listReady()` and `listBlocked()` on `TaskServiceLike` still return bare arrays internally — only the HTTP layer wraps. No service-level type changes needed.
- `total` on ready/blocked reflects the full result count (no server-side pagination on those endpoints).

## Test plan

- [ ] All task-store smoke tests pass
- [ ] check-helpers unit tests pass
- [ ] Admin tasks page loads on default (ready) view without error
- [ ] Admin blocked filter works


plugins/shipwright/scripts/check-helpers.ts
```diff
@@ -345,8 +345,6 @@ export function createTaskStoreClient(): {
       if (!res.ok)
         throw new Error(`task-store GET /tasks?${params} → ${res.status}`);
       const data = (await res.json()) as unknown;
-      // ?ready=true returns Task[]; all other queries return { tasks, total, limit, offset }
-      if (Array.isArray(data)) return data as Task[];
       if (
         data !== null &&
```
dmcaulay, line 351: Removing the `Array.isArray(data)` guard creates a one-way compatibility gap: the updated plugin throws `"Unexpected task-store response format"` if deployed to an agent while the task-store service still returns bare `Task[]` for `?ready=true`. This breaks `check-dev-task.ts`, `check-review.ts`, `check-patch.ts`, and `check-deploy.ts` during the plugin-before-service deployment window. The reverse (new service + old plugin) is safe.

Suggest keeping the guard as a temporary fallback:
```ts
// Temporary: accept legacy bare Task[] from older task-store instances
if (Array.isArray(data)) return data as Task[];
```
Remove in a follow-up once all deployed task-store instances are updated.

dmcaulay (commented):

**1. Rolling-deployment break in `check-helpers.ts`**: Removing the `Array.isArray(data)` guard means agents running the new plugin against an old task-store will throw on every `?ready=true` query, breaking all pre-check scripts. The reverse (new service + old plugin) is safe. Suggest retaining the fallback until all deployed task-stores are confirmed updated, or documenting the required deploy order.

**2. Stale JSDoc in `task-store/src/routes/tasks.ts:18`**: Line 18 still reads `returns { tasks, total, limit, offset } — or Task[] when ?ready=true` — this is the old shape. Should be updated to reflect the `{ tasks, total }` envelope.

Minor: `fetchTaskStorePr` is declared in `AdminUIDeps` and fully implemented but not injected in `admin/src/main.ts` — the PR section will never appear in production until wired. If intentional (follow-up PR), worth a TODO comment in `main.ts`.

Response shape fix itself is correct and well-tested. CI green.

dodizzle (commented):
APPROVE — All three findings from the prior review addressed: backward-compat `Array.isArray` fallback retained in `check-helpers.ts`, JSDoc corrected to reflect `{ tasks, total }`, and `fetchTaskStorePr` wired with a TODO. Shape is consistent, CI green, tests updated across all three layers. Ready to merge.

plugins/shipwright/scripts/check-helpers.ts
```diff
@@ -345,7 +345,7 @@ export function createTaskStoreClient(): {
       if (!res.ok)
         throw new Error(`task-store GET /tasks?${params} → ${res.status}`);
       const data = (await res.json()) as unknown;
-      // ?ready=true returns Task[]; all other queries return { tasks, total, limit, offset }
+      // Temporary: accept legacy bare Task[] from older task-store instances
       if (Array.isArray(data)) return data as Task[];
```
dmcaulay, line 349: The backward-compat fallback is correct and needed — but the unit test that covered the bare-array path was removed in commit `6f1f3c80` and not replaced. This branch now has zero test coverage, which defeats its purpose as a safety net. Add one test:
```ts
test('accepts legacy bare Task[] response from older task-store', async () => {
  globalThis.fetch = (async () => ({
    ok: true,
    json: async () => [FAKE_TASK],
  }) as Response) as unknown as typeof fetch;
  const client = createTaskStoreClient();
  const result = await client.query(new URLSearchParams({ ready: 'true' }));
  expect(result).toEqual([FAKE_TASK]);
});
```

dmcaulay (commented):
COMMENT — R1 findings all resolved: backward-compat `Array.isArray` fallback retained with an accurate comment, JSDoc corrected, `fetchTaskStorePr` TODO added in `main.ts`. One new issue: the legacy fallback added for rolling-deploy safety is now untested. The old bare-array test was removed (commit `6f1f3c80`) without a replacement covering the `Array.isArray(data)` branch. A future accidental deletion wouldn't be caught. See inline comment.

PR #734, opened 2026-06-25
feat: add SHIPWRIGHT_ADMIN_TZ timezone formatting to admin UI

dmcaulay:
## Summary
- Adds `timezone?: string` to `AdminUIDeps` (defaults to `"America/Los_Angeles"`) and threads it through all four admin UI render functions
- Replaces all `.toISOString().split("T")[0]` date formatting with `toLocaleDateString("en-US", { timeZone: timezone })` and adds `timeZone` to `toLocaleString` calls
- Reads `SHIPWRIGHT_ADMIN_TZ` env var in `admin/src/main.ts` and passes it to `createAdminUIApp`

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `AdminUIDeps.timezone` is optional; absent → factory defaults to `"America/Los_Angeles"` | ✅ MET |
| 2 | All 4 render functions accept `timezone: string` and use it for every date/datetime field | ✅ MET |
| 3 | `SHIPWRIGHT_ADMIN_TZ=America/New_York` causes timestamps to render in Eastern time | ✅ MET |
| 4 | Unit tests: all existing render-function call sites updated to pass `"UTC"`; one new test asserts a Pacific-formatted timestamp for a known UTC input | ✅ MET |
| 5 | `docs/configuration.md` admin table includes `SHIPWRIGHT_ADMIN_TZ` | ✅ MET |
| 6 | `task lint` and `task typecheck` pass | ✅ MET |

## Test Plan
- [x] 184 unit tests pass (admin/src/admin-ui-pages.unit.test.ts)
- [x] New test: `createdAt date uses the provided timezone` verifies UTC vs Pacific date rendering
- [x] New tests: `renderTaskDetailPage — timezone formatting` verifies Pacific vs UTC timestamp formatting
- [x] `bunx biome lint .` — no issues (320 files)
- [x] `bun run --filter="*" typecheck` — all 5 packages pass
- [x] Line coverage: 95.95% (above 90% threshold)


dodizzle (approved):
APPROVE — Clean timezone threading with no breaking changes. All four render functions updated with correct defaults, tests use UTC for determinism with midnight-crossing coverage. CI green, all acceptance criteria met.

admin/src/main.ts
```diff
@@ -318,6 +318,7 @@ async function startServer(): Promise<void> {
     slackClient,
     appBaseUrl,
     devAuthEnabled: isDevAuthAllowed(process.env),
+    timezone: process.env.SHIPWRIGHT_ADMIN_TZ,
```
dmcaulay: **Invalid SHIPWRIGHT_ADMIN_TZ crashes all admin pages with 500**

process.env.SHIPWRIGHT_ADMIN_TZ is passed without validation to toLocaleDateString()/toLocaleString() as the timeZone option in all four render functions. An invalid IANA value (typo, unsupported alias like `"US/Eastern"`, any garbage string) causes `RangeError: Invalid time zone specified` inside the render function — none of the call sites have try/catch, so the error propagates to a 500 on every admin page. The old .toISOString() path could not throw; this is a new failure mode.

Suggested fix at startup in main.ts (near this line):
```ts
const tz = process.env.SHIPWRIGHT_ADMIN_TZ;
if (tz) {
  try { Intl.DateTimeFormat(undefined, { timeZone: tz }); }
  catch { console.error('[admin] Invalid SHIPWRIGHT_ADMIN_TZ:', tz); process.exit(1); }
}
```

admin/src/admin-ui-pages.ts
```diff
@@ -1132,6 +1140,7 @@ export function renderTasksPage(
   },
   opts?: { error?: string; agentFilterActive?: boolean },
   suggestions?: { sessions?: string[]; repos?: string[]; agents?: string[] },
+  timezone = "America/Los_Angeles",
```
dmcaulay: **timezone parameter in renderTasksPage is dead code**

renderTasksPage now accepts timezone as a 9th parameter and the call site in admin-ui.ts passes timezone to it. But the tasks list table has no date columns (id, title, status, agent, session, repo, action button only) — the parameter is received, defaulted, and never used in any of the ~25 lines of row generation. This is misleading: it implies timezone formatting applies to the tasks list when it does not.

Either remove the parameter (and the corresponding argument in admin-ui.ts), or add a date column to the tasks list if that was the intended direction.

dmcaulay (commented):
## Review

Adds SHIPWRIGHT_ADMIN_TZ env var that threads an IANA timezone name through all four admin UI render functions, replacing UTC-only ISO date formatting with locale-aware toLocaleDateString/toLocaleString calls. Clean approach and solid test coverage — two issues before merge.

**COMMENT** — 2 important findings below.

dmcaulay (commented):
APPROVE — Both prior findings addressed cleanly. Startup validation uses `Intl.DateTimeFormat(undefined, { timeZone: adminTz })` — the same engine path that would have thrown at render time — with `process.exit(1)` on invalid TZ. Dead `timezone` parameter removed from `renderTasksPage` and all 13 test call sites cleaned up. All three affected pages (agents list, agent detail, task detail) now have timezone-aware date formatting. No new issues.

PR #750, opened 2026-06-25
feat: add token columns to AgentCronRun and PATCH complete endpoint

dmcaulay:
## Summary
- Add 6 nullable token-tracking columns to `AgentCronRun` (inputTokens, outputTokens, cacheReadTokens, cacheCreationTokens, costUsd, model)
- Add `PATCH /agents/:id/crons/:cronId/runs/:runId` endpoint to complete a cron run with outcome, error, and token data
- Enables the harness to CREATE a run at start and PATCH it at completion in a single round-trip

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | AgentCronRun schema has all 6 nullable token columns | MET |
| 2 | Migration applies cleanly | MET |
| 3 | PATCH endpoint accepts all specified fields, returns updated run | MET |
| 4 | PATCH returns 404 when runId not owned by agent | MET |
| 5 | Integration + smoke tests added, no existing tests retired | MET |

## Test Plan
- 4 new integration tests: patch updates token fields, patch updates error/outcome fields, patch 404 unknown runId, patch 404 wrong-agent runId
- 3 new smoke tests: PATCH 200 with token fields, PATCH 404 ownership check, PATCH 400 empty body
- [x] biome lint: no issues
- [x] typecheck: all workspaces clean
- [x] 682 smoke/unit tests passing


dodizzle (approved):
APPROVE — clean, well-tested addition of token-tracking columns and PATCH completion endpoint to AgentCronRun. Migration is fully backwards-compatible (nullable columns, no dropped fields). Ownership check in `patch()` is correct. Two minor suggestions: (1) `cronId` URL param is not validated against the run's actual cronId in the PATCH handler — agentId guard is the real security boundary and is correct, but consistency with list/create would validate cronId too; (2) token fields in `AgentCronRunSchema` are marked `optional()` but `serializeCronRun()` always emits them as `null` — consider dropping `optional()` to match the actual contract. Neither blocks merge.

admin/src/agents-api.ts
```diff
@@ -1057,6 +1086,40 @@ export function createAdminApp(deps: AdminDeps): OpenAPIHono<AdminAuthEnv> {
     );
   });
 
+  // PATCH /agents/:id/crons/:cronId/runs/:runId — update a cron run record
+  app.openapi(patchCronRunRoute, async (c) => {
+    const { id: agentId, runId } = c.req.valid("param");
```
dmcaulay: **cronId is dropped here — ownership check is incomplete.** The route URL contains `{cronId}` but it's never extracted from params. The service `patch()` only checks `run.agentId !== agentId`, so a request to `PATCH /agents/A/crons/C1/runs/runBelongingToC2` succeeds with 200 if C1 and C2 belong to the same agent. This is inconsistent with `create()` and `list()`, which both validate cronId explicitly.

Fix:
```ts
const { id: agentId, cronId, runId } = c.req.valid("param");
// then pass cronId to agentCronRunService.patch(runId, agentId, cronId, input)
```
And in `AgentCronRunService.patch()`:
```ts
if (run.cronId !== cronId) throw new NotFoundError(`cron run ${runId} not found`);
```

admin/src/openapi-schemas.ts
```diff
@@ -184,6 +184,12 @@ export const AgentCronRunSchema = z
       .openapi({ example: "pre-check returned false" }),
     outcome: z.string().nullable().openapi({ example: "success" }),
     error: z.string().nullable().openapi({ example: null }),
+    inputTokens: z.number().int().nullable().optional().openapi({ example: 1234 }),
```
dmcaulay: **Schema contract is broader than actual behavior.** These six token fields are marked `.nullable().optional()` — promising clients they may be absent (`undefined`) or null. But `serializeCronRun()` always emits them as `null` via `?? null`, never as `undefined`. Drop `.optional()` from all six to match what the serializer actually produces:
```ts
inputTokens: z.number().int().nullable().openapi({ example: 1234 }),
```

dmcaulay (commented):
COMMENT — two important issues with the PATCH endpoint's ownership validation.

**Finding 1 (Important):** The PATCH handler extracts only `{ id: agentId, runId }` from the params — `cronId` is silently dropped. A caller can PATCH a run belonging to cron C2 via the URL for cron C1 of the same agent and receive 200 instead of 404. This partially fails AC4 and is inconsistent with `create()` and `list()`, which both validate `cronId`. Fix: extract `cronId` in the handler and add `if (run.cronId !== cronId) throw new NotFoundError(...)` in `patch()`.

**Finding 2 (Important):** The smoke test for the 404 case injects a mock that throws `NotFoundError` unconditionally — it doesn't exercise whether the handler enforces cronId validation, because `cronId` is never passed to the service. Once the cronId check is added, update the mock and add a test case for the cross-cron scenario.

**Suggestion:** Token fields in `AgentCronRunSchema` are marked `.nullable().optional()` but `serializeCronRun()` always emits them as `null` (never `undefined`). Drop `.optional()` to match the actual contract.

admin/src/agents-api.ts
```diff
@@ -1057,6 +1112,40 @@ export function createAdminApp(deps: AdminDeps): OpenAPIHono<AdminAuthEnv> {
...
```
dmcaulay: Still extracting only `{ id: agentId, runId }` — `cronId` from the URL is silently dropped. Unchanged from R1.

Fix: `const { id: agentId, cronId, runId } = c.req.valid("param")` and add `if (run.cronId !== cronId) throw new NotFoundError(...)` in `AgentCronRunService.patch()` after the agentId check.

admin/prisma/migrations/20260627000000_add_agent_chat_token_usage_daily/migration.sql
```diff
@@ -0,0 +1,18 @@
...
+    "date" TEXT NOT NULL,
+    "inputTokens" INTEGER NOT NULL DEFAULT 0,
+    "outputTokens" INTEGER NOT NULL DEFAULT 0,
+    "cacheReadTokens" INTEGER NOT NULL DEFAULT 0,
+    "cacheCreationTokens" INTEGER NOT NULL DEFAULT 0,
+    "costUsd" DOUBLE PRECISION NOT NULL DEFAULT 0,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL,
+
+    CONSTRAINT "AgentChatTokenUsageDaily_pkey" PRIMARY KEY ("id")
```
dmcaulay, line 14: No FK from `agentId` to `Agent.id ON DELETE CASCADE`. Every other child table in this schema has cascade cleanup via a Prisma `@relation`. Without it, deleting an agent silently leaves all its `AgentChatTokenUsageDaily` rows as orphans — no cascade fires.

Fix: Add a Prisma `@relation` on `AgentChatTokenUsageDaily.agentId` referencing `Agent.id` with `onDelete: Cascade`, then regenerate the migration.

dmcaulay (commented):
COMMENT — R2 re-review. R1 Finding 1 (cronId silently dropped in PATCH cron-run handler) is still unresolved — the new commits do not touch the handler. The new AgentChatTokenUsageDaily migration introduces a data integrity gap: no FK from agentId to Agent.id means agent deletes leave orphan token rows. CI is green on HEAD.

**Finding 1 (unresolved from R1):** `PATCH /agents/:id/crons/:cronId/runs/:runId` handler extracts only `{ id: agentId, runId }` — `cronId` is silently dropped and never validated against the run's `cronId`. A caller can PATCH a run belonging to cron C2 via the URL for cron C1 of the same agent and receive 200 instead of 404. Partially fails AC4.

Fix: Extract cronId in the handler and add `if (run.cronId !== cronId) throw new NotFoundError(...)` in `patch()` after the agentId check.

**Finding 2 (new):** `AgentChatTokenUsageDaily` migration has no FK from `agentId` to `Agent.id ON DELETE CASCADE`. Every other child table (`AgentCronJob`, `AgentCronRun`, etc.) has cascade cleanup via Prisma relations. Without it, deleting an agent silently leaves all its daily token rows as orphans.

Fix: Add a Prisma `@relation` on `agentId` referencing `Agent` with `onDelete: Cascade`, regenerate the migration.

**Suggestion (unresolved from R1):** Token fields in `AgentCronRunSchema` are `.nullable().optional()` but `serializeCronRun()` always emits `null` (via `?? null`), never `undefined`. Drop `.optional()` from the six token fields to tighten the contract.

admin/src/agent-cron-runs.integration.test.ts
```diff
@@ -200,6 +200,88 @@ describeOrSkip("AgentCronRunService (integration)", () => {
...
+    const cronId = await createCron(cronJobService, agentId);
+    const startedAt = new Date("2026-01-15T10:00:00Z");
+    const completedAt = new Date("2026-01-15T10:05:00Z");
+
+    const run = await runService.create(cronId, agentId, {
+      startedAt,
+      skipped: false,
+    });
+
+    const updated = await runService.patch(run.id, agentId, {
```
dodizzle: This call now fails typecheck — `patch()` requires a 4th `cronId` argument. The `cronId` variable is in scope from `createCron()` above. Change to `runService.patch(run.id, agentId, cronId, { ... })`.

dodizzle (commented):
COMMENT — R3 re-review. All R2 findings addressed: cronId validation added to PATCH handler and service, FK cascade added to AgentChatTokenUsageDaily, `.optional()` dropped from token schema fields. One new blocker: the fix commit extended `patch()` to 4 params but did not update the 4 existing integration tests — CI is failing typecheck on `b9311619`.

**Finding (Critical):** `admin/src/agent-cron-runs.integration.test.ts` lines 216, 248, 262, 279 call `runService.patch(run.id, agentId, {...})` with 3 arguments. The signature now requires 4 (`runId, agentId, cronId, input`). Fix: pass the `cronId` that's already in scope as the 3rd argument in each call. One-line change per call site.

dodizzle (approved):
APPROVE (R4) — R3 finding fully addressed. `eb24c533` updates all 4 integration test `patch()` calls with the `cronId` 4th argument and adds the cross-cron ownership test case that was the gap. No new issues. CI running on this commit — prior failure was the 3-arg typecheck errors, which are now fixed. Clean to merge.

dmcaulay (commented):
APPROVE — R3 re-review. All prior findings resolved: cronId validation added to PATCH handler and service, FK cascade added to AgentChatTokenUsageDaily, .optional() dropped from token schema fields, integration test call sites updated to 4-arg signature, cross-cron test coverage added at both service (integration) and HTTP (smoke) layers. No new issues. Ready to merge once CI green on eb24c53.

PR #753, opened 2026-06-26
fix(admin): simplify PR list tabs — Open/Merged only

dodizzle:
## Summary

- Removes the **"In Review"** tab — it was filtering on `reviewState=in_review` which doesn't exist in the `PrReviewState` enum (`pending | in_progress | posted | approved`), causing Prisma to throw and the store-unavailable banner to appear only on that tab
- Removes the **"All"** tab
- Renames **"Closed" → "Merged"**, filtering on `state=merged` (the only terminal `PrState` value that's ever written — set by the deploy command; `closed` is a valid schema value but nothing writes it)
- Fixes `reviewStateBadgeClass` which referenced invalid enum values (`in_review`, `merged`, `changes_requested`)
- Updates filter form placeholders to reflect actual enum values
- `reviewState` remains rendered as a badge column in the table rows

## Test plan

- [ ] `/admin/prs` — no tab is highlighted (unfiltered view still works)
- [ ] `/admin/prs?state=open` — Open tab highlights, shows open PRs
- [ ] `/admin/prs?state=merged` — Merged tab highlights, shows merged PRs
- [ ] No more "PR store unavailable" banner on any tab
- [ ] reviewState badge column renders correctly for `pending`, `in_progress`, `posted`, `approved`

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -1701,9 +1701,8 @@ export function renderPrsPage(
   };
 
   const reviewStateBadgeClass = (s: string) => {
-    if (s === "in_review") return "badge-blue";
-    if (s === "approved" || s === "merged") return "badge-green";
-    if (s === "changes_requested") return "badge-red";
+    if (s === "in_progress" || s === "posted") return "badge-blue";
+    if (s === "approved") return "badge-green";
     return "badge-gray";
   };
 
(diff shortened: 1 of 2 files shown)


dodizzle (commented):
APPROVE — clean bug fix: removes invalid enum values from tab strip and badge function, replaces with correct `PrState`/`PrReviewState` values, tests updated. No blocking concerns. CI green on HEAD (6d0645d).

dmcaulay (commented):
COMMENT — R1 review. The crash fix (badge enum values + tab simplification) is correct, but removing the All tab leaves no tab active when visiting /admin/prs with no query params. `const activeState = filters.state` is undefined on unfiltered load; both Open and Merged tabs evaluate `activeState === tabState` as false. Suggest defaulting to `filters.state ?? 'open'` (with a matching route handler default) so the tab highlight and filter stay in sync. No other issues.

PR #754, opened 2026-06-26
docs: remote setup guide for running plan-session from a laptop

dodizzle:
## Summary

- Adds **`docs/remote-setup.md`** — step-by-step guide for running `/shipwright:plan-session` (or any HTTP-API-backed Shipwright skill) from a local Claude Code session and pushing tasks to the shared task store for a cloud agent to execute
- Fixes a misleading note in `task-store/SKILL.md` that said env vars are "provisioned automatically by the agent harness" (they are, for managed agents — but for remote sessions the user sets them manually)
- Adds the new doc to the `@docs` reference block in `CLAUDE.md`

## How it works

When a remote user uses a token scoped to a specific agent's ID, the task store server auto-assigns every task they create to that agent (enforced server-side in `routes/tasks.ts`). No `assignee` field needed in the task JSON — the token does the work.

## Test plan

- [ ] Set `SHIPWRIGHT_TASK_STORE_URL` and `SHIPWRIGHT_TASK_STORE_TOKEN` locally and run the verify curl from the guide — should return `{ total: 0 }` or higher
- [ ] Run `/shipwright:plan-session` from a local Claude Code session against a test repo — tasks should appear in the task store scoped to the target agent
- [ ] Confirm task-store SKILL.md note no longer misleads remote users

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -160,3 +160,4 @@ To load additional context into a session, add `@docs/filename.md` entries here 
 - **docs/deploy-kubernetes.md** — Kubernetes deployment guide: Minikube / GKE (Gateway API + cert-manager) / EKS (ALB), the agent runtime-provisioning RBAC model, and auth modes
 - **docs/test-readiness/test-system.md** — the authoritative test blueprint: layer matrix, boundary rules, per-component budgets, CI pipeline shape, and the full isolation contract
 - **docs/migration.md** — breaking changes and migration steps across versions (e.g. `AgentProvisioner.reconcile()` interface change)
+- **docs/remote-setup.md** — how to run Shipwright skills (e.g. `/plan-session`) from a laptop and push tasks to the shared task store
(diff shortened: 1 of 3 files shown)


dmcaulay:
Closing to re-approach through proper shipwright planning. Will reopen after plan-session.

dodizzle:
Worth noting for reviewers: this PR solves the mechanical access problem (env vars, token, URL) but `remote-setup.md` should also set expectations around planning discipline. Having task store access from a laptop doesn't mean skipping `plan-session` → queue → agent execution — that flow still applies. A local session posting tasks directly without a plan creates untracked, undependency-mapped work the agent picks up blindly.

Will add a "How this fits the workflow" section before merge, making clear the expected pattern: run `/shipwright:plan-session` locally → tasks queue → cloud agent executes. Not: use the task store API to post ad-hoc tasks from a laptop.

PR #761, opened 2026-06-26
fix(task-store): extend repo scope to all tasks in scope, not just unassigned

dmcaulay:
## Summary

- `list()`, `listBlocked()`, `distinct()` now show all tasks whose repo is in the agent's scope, regardless of assignee — fixing the case where Bodhi couldn't see Warchild's pending tasks in `app-vitals/shipwright`
- `listReady()` intentionally stays strict (unassigned only) so agents don't try to claim each other's work
- Write access unchanged — `requireOwnership` still requires self-assigned or unassigned for mutations

## Test plan

- [ ] Unit tests pass (`bun test task-store/src/task-service.unit.test.ts`)
- [ ] Bodhi can see Warchild's pending tasks via `GET /tasks?status=pending` after deploy

### DIFF
--- a/task-store/src/routes/tasks.ts
+++ b/task-store/src/routes/tasks.ts
@@ -109,7 +109,10 @@ export function createTasksRoutes(
 
     // ?state=blocked delegates to listBlocked().
     if (stateRaw === "blocked") {
-      const tasks = await taskService.listBlocked(agentId ?? undefined);
+      const tasks = await taskService.listBlocked(
+        agentId ?? undefined,
+        repos !== null ? repos : undefined,
+      );
       return c.json({ tasks, total: tasks.length }, 200);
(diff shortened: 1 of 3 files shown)


dmcaulay (commented):
APPROVE (self-review → COMMENT per policy)

Correctly extends read visibility to all tasks in scoped repos while preserving claim safety (`listReady()` unchanged). Logic is sound, tests cover all behavioral cases.

One suggestion: the JSDoc comment on `TaskListFilters.agentScope` (line 46 of `task-store/src/task-service.ts`) still reads "unassigned pool tasks whose repo is in the agent's scope" — this PR removes the `assignee: null` constraint, so now any task in a scoped repo is visible. The comment should be updated to match.

PR #764, opened 2026-06-26
feat(admin): tasks show-all default and Reset button

dmcaulay:
## Summary
- Remove implicit `state=ready` fallback from `GET /admin/tasks` so visiting with no params returns all tasks
- Update tab styling so no tab is highlighted when no state filter is active
- Add a Reset button next to the Filter button that clears all params back to `/admin/tasks`

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | GET /admin/tasks with no params returns all tasks, no tab highlighted | ✅ MET |
| 2 | Reset button appears next to Filter, links to /admin/tasks | ✅ MET |
| 3 | State tabs still work when explicit state= param is present | ✅ MET |
| 4 | HTML rendering only — no new backend/API tests needed | ✅ MET |

## Test Plan
- [x] Unit tests updated: `renderTasksPage — 4-state toggle` now asserts no active tab on unfiltered load
- [x] New unit test: Reset button present and links correctly
- [x] Smoke test updated: `GET /admin/tasks` with no params forwards no state to task-store
- [x] All 835 admin tests pass (0 failures)

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -1283,8 +1283,7 @@ export function renderTasksPage(
   const makePageUrl = (p: number) => {
     const params = new URLSearchParams();
     if (filters.status) params.set("status", filters.status);
-    else if (filters.state && filters.state !== "ready")
-      params.set("state", filters.state);
+    else if (filters.state) params.set("state", filters.state);
     if (filters.session) params.set("session", filters.session);
     if (filters.repo) params.set("repo", filters.repo);
     if (filters.agent) params.set("agent", filters.agent);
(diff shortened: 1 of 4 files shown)


dmcaulay (commented):
COMMENT (R1) — 1 important finding: the Ready tab link is broken by the new default semantics, plus a gap in AC3 test coverage.

**`admin/src/admin-ui-pages.ts` (makeStateParams)** — `makeStateParams("ready")` omits `state=` from the query string via `if (newState !== "ready") p.set("state", newState)`. This was intentional when `/admin/tasks` defaulted to ready, but the PR removes that default — `/admin/tasks` now means "show all". Result: clicking the Ready tab navigates to `/admin/tasks` with no state param, showing all tasks rather than filtering to ready. Clicking Ready is now equivalent to clicking Reset. Fix: remove the `!== "ready"` guard so all four tabs emit an explicit `state=` param: `p.set("state", newState)` unconditionally.

**`admin/src/admin-ui-pages.unit.test.ts`** — Acceptance criterion 3 says "State tabs still work correctly when an explicit state= param is present." Tests cover in_progress, blocked, and closed tabs but not the Ready tab. Once `makeStateParams` is fixed, add a test: `renderTasksPage([], { state: "ready" }, ...)` and assert `background:#6366f1` styling and href contains `state=ready`.

Strengths: backend change in `admin-ui.ts` is clean; Reset button as a plain `<a href="/admin/tasks">` is exactly right; smoke test correctly verifies the new show-all behavior.

dmcaulay (commented):
COMMENT (R2) — R1 findings fixed (makeStateParams guard removed, Ready tab test added). New finding: `makePageUrl` at line 1286 has the same `!== "ready"` guard that was just fixed in `makeStateParams`. Under the new show-all default, clicking page 2 on the Ready tab will silently drop the state filter and show all tasks instead of ready-only tasks.

**Fix**: `admin/src/admin-ui-pages.ts:1286` — change `else if (filters.state && filters.state !== "ready")` to `else if (filters.state)` so the ready state is preserved across pagination.

dmcaulay (commented):
APPROVE (R3) — All R2 findings resolved: `makePageUrl` guard removed so `state=ready` is preserved across pagination. Biome formatting fixed. CI green. No new issues.

PR #765, opened 2026-06-26
fix(site): add .nojekyll so GitHub Pages serves _astro assets

dodizzle:
## Summary
The live site at shipwrightharness.com rendered unstyled — GitHub Pages runs Jekyll by default, which strips `_`-prefixed paths, so the entire `_astro/` assets dir (CSS/JS) 404'd.

Adds an empty `site/public/.nojekyll`. Astro copies `public/*` to the `dist/` root, so the marker is published to `gh-pages` on every deploy and survives the deploy workflow's `clean: true`.

## Why source, not just gh-pages
A `.nojekyll` added directly to `gh-pages` would be wiped on the next `deploy-site.yml` run (clean-exclude only covers `index.yaml`/`*.tgz`). Putting it in `site/public/` makes it durable.

## Verification
- `bunx astro build` → `dist/.nojekyll` present.
- Live site already unblocked via a manual `.nojekyll` on gh-pages; this PR makes it permanent.
- Closes the last gap on SITE-6.2 (Pages/custom-domain/DNS).

### DIFF
--- a/charts/shipwright/CHANGELOG.md
+++ b/charts/shipwright/CHANGELOG.md
@@ -10,6 +10,18 @@ independent of `appVersion`. CI enforces this with
 `ct lint --check-version-increment`. Each release here must mirror the
 `artifacthub.io/changes` annotation in `Chart.yaml`.
 
+## [1.6.83] - 2026-06-26
+
+### Changed
+
+- auto-bump to chart v1.6.83 triggered by release tag `metrics-v0.89.0`
+
+## [1.6.82] - 2026-06-26
+
+### Changed
+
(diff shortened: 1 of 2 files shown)


dmcaulay (approved):
APPROVE — correct fix in the right place. `.nojekyll` in `site/public/` survives clean deploys; direct edit to `gh-pages` would be wiped on the next `deploy-site.yml` run.

dodizzle (commented):
APPROVE — single empty marker file `site/public/.nojekyll`. Correct fix: Astro copies `public/*` to `dist/` root, so the marker survives the build and tells GitHub Pages not to run Jekyll, allowing `_astro/` assets to be served. CI green.

PR #768, opened 2026-06-26
fix(dev-stack): make task stack:up fully functional locally (nav cross-links + task-store wiring)

dodizzle:
Two local-dev-only fixes so `task stack:up` is fully usable. **Both are scoped to the local stack** (`scripts/dev-tmux.ts` + a dev-only seed script) — no admin/metrics/task-store **service** code changes, so deployed stacks are unaffected by construction.

## 1. Dashboard nav cross-links to the admin console
**Problem:** on `:3460/dashboard`, the toolbar's Agents/Tasks/PRs links were relative → resolved to `:3460/admin/*` → 404.
**Fix:** the metrics dashboard now renders the shared toolbar with an `adminBaseUrl` (new env `METRICS_ADMIN_APP_URL`, default `""` = relative, so prod/ingress is unchanged). The local stack sets it to `http://localhost:3001`. Also fixes the misleading logs-pane banner so the admin console is the labeled front door.

## 2. Admin console wired to the task-store
**Problem:** the Tasks/PRs pages showed *"Task store unavailable — data shown may be stale or empty."* The admin builds its task-store client only when both `SHIPWRIGHT_TASK_STORE_URL` and `SHIPWRIGHT_TASK_STORE_ADMIN_TOKEN` are set, and the stack set neither → permanently degraded. (Not an empty-store or outage — a missing-client state.)
**Fix:**
- New `scripts/seed-task-store-token.ts` — idempotently upserts a dev admin token (`agentId = null`, hashed) into the task-store DB; run as a preflight after the task-store migrate.
- The admin pane sets `SHIPWRIGHT_TASK_STORE_URL=http://localhost:3002` + `SHIPWRIGHT_TASK_STORE_ADMIN_TOKEN=<dev token>` (inline prefix overrides any empty value the login shell exports).

## Tests (land with the code)
- `dashboard-page.unit.test.ts` / `dashboard-dev-auth.smoke.test.ts` — absolute vs relative admin links; factory passthrough.
- `dev-tmux.unit.test.ts` — metrics pane carries `METRICS_ADMIN_APP_URL`; admin pane carries both task-store vars; seed preflight runs after migrate; seeded token matches the admin pane's token (drift guard).
- `seed-task-store-token.unit.test.ts` — hashing (known SHA-256 vector), arg parsing, idempotent upsert shape.

## Verification
- `task lint` ✓, `task typecheck` ✓ (all 5 workspaces), full `bun test` = **2975 pass, 0 fail**.
- End-to-end: ran the seed against the live task-store DB → token row present with `agentId=null`, not revoked, and its stored hash equals an independent `shasum -a 256` of the raw token (so task-store `validate()` will accept it).

> `check-strings` failures in CI, if any, are from git-ignored `state/reviews/*` local files only — zero hits in tracked source.


dmcaulay (approved):
APPROVE — both sides of the local stack are now fully wired. `adminBaseUrl` propagates through all three render paths in `api.ts` with a production-safe empty-string default. Seed script is idempotent and local-dev-only. Token consistency test is a good drift guard. CI green, no issues.

dodizzle (commented):
APPROVE — both dev-stack fixes are clean and scoped to local-only surfaces. `adminBaseUrl` propagates through all three `renderDashboardPage` call sites with a production-safe empty-string default; tests cover absolute vs relative link behavior and the Metrics link's same-origin invariant. Seed script is gated by `import.meta.main`, uses an idempotent upsert (`update: {}`), and is never invoked in deployed stacks. Known SHA-256 vector test locks the hashing contract against drift from `task-store/src/token-service.ts`. CI green, all 2975 pass.

PR #771, opened 2026-06-26
feat: add admin proxy routes for task-store token CRUD

dodizzle:
## Summary
- Add `adminListTokens`, `adminCreateToken`, `adminRevokeToken` fetcher functions to `admin/src/main.ts` wired to the task-store admin token
- Add GET /admin/tokens, POST /admin/tokens, POST /admin/tokens/:id/revoke proxy routes in `admin/src/admin-ui.ts` with graceful degradation when task-store is unconfigured
- Add Tokens nav link in the admin sidebar toolbar

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | GET /admin/tokens fetches token list from task store and passes to view | ✅ MET |
| 2 | POST /admin/tokens proxies create request, passes rawToken to success view | ✅ MET |
| 3 | POST /admin/tokens/:id/revoke proxies DELETE to task store, redirects to list | ✅ MET |
| 4 | Smoke tests for all three proxy routes via Hono in-process driver with mocked task-store responses; no existing tests retired | ✅ MET |

## Test Plan
- 8 smoke tests in `admin/src/admin-tokens.smoke.test.ts` using `app.request()` with mocked dep injection — no real server or task-store needed
- Tests cover list, create, revoke; degraded mode (dep absent); 503 responses; 302 redirects; non-admin 403; unauthenticated 302
- All 8 tests passing; lint clean

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -2093,6 +2093,7 @@ export function renderTokensPage(
   activePath = "/admin/tokens",
   rawToken?: string,
   timezone?: string,
+  error?: string,
 ): string {
   const tz = timezone ?? "America/Los_Angeles";
   const fmt = (d: Date | string | null | undefined) => {
@@ -2103,6 +2104,10 @@ export function renderTokensPage(
 
   const degradedBanner = degraded
     ? `<div class="alert alert-warning">Token store unavailable — configure SHIPWRIGHT_TASK_STORE_URL and SHIPWRIGHT_TASK_STORE_ADMIN_TOKEN.</div>`
+    : "";
+
+  const errorBanner = error
+    ? `<div class="alert alert-danger" style="margin-bottom:16px">${escapeHtml(decodeURIComponent(error))}</div>`
     : "";
 
   const rawTokenBanner = rawToken
@@ -2123,7 +2128,7 @@ export function renderTokensPage(
           (t) => `<tr>
           <td style="font-size:12px;color:#6b7280;font-family:monospace">${escapeHtml(t.id)}</td>
           <td>${escapeHtml(t.label ?? "—")}</td>
-          <td style="font-size:12px;color:#6b7280;font-family:monospace">${escapeHtml(t.agentId ?? "admin")}</td>
+          <td style="font-size:12px;color:#6b7280;font-family:monospace">${escapeHtml(t.agentId ?? "(admin)")}</td>
           <td>${fmt(t.createdAt)}</td>
           <td>${t.revokedAt ? `<span style="color:#dc2626">Revoked ${fmt(t.revokedAt)}</span>` : '<span style="color:#16a34a">Active</span>'}</td>
           <td>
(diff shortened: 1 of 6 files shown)


admin/src/admin-ui.ts
```diff
@@ -1838,6 +1860,64 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
     return html(renderPrDetailPage(pr, c.var.userEmail, agentNames, timezone));
   });
 
+  // ─── Task-store token proxy routes ────────────────────────────────────────
+
```
dodizzle, line 1864: The GET handler doesn't read `c.req.query("error")`, so when `POST /admin/tokens` or `POST /admin/tokens/:id/revoke` redirect here with `?error=<msg>`, the error is never surfaced to the user. Add:
```ts
const error = c.req.query("error");
```
and pass it through to `renderTokensPage` (which needs a corresponding `error?: string` parameter) to render an alert banner. See the pattern at lines 400, 634, and 1614 for reference.

dodizzle (commented):
COMMENT — one important bug: `POST /admin/tokens` and `POST /admin/tokens/:id/revoke` redirect to `?error=<msg>` on failure, but `GET /admin/tokens` never reads the query param and `renderTokensPage` has no `error` parameter — the error is silently discarded. Every other error-redirect pattern in the codebase reads and forwards `?error` (lines 400, 634, 1614). Fix: read `c.req.query("error")` in the GET handler and pass it to `renderTokensPage`. Everything else is clean: inline render after create (never redirecting the raw token), graceful degradation, thorough smoke test coverage.

admin/src/admin-ui-pages.ts
```diff
@@ -2071,3 +2083,118 @@ export function renderProvisionCompletePage(
...
+    const date = d instanceof Date ? d : new Date(d);
+    return date.toLocaleString("en-US", { timeZone: tz, dateStyle: "medium", timeStyle: "short" });
+  };
+
+  const degradedBanner = degraded
+    ? `<div class="alert alert-warning">Token store unavailable — configure SHIPWRIGHT_TASK_STORE_URL and SHIPWRIGHT_TASK_STORE_ADMIN_TOKEN.</div>`
+    : "";
+
+  const errorBanner = error
+    ? `<div class="alert alert-danger" style="margin-bottom:16px">${escapeHtml(decodeURIComponent(error))}</div>`
```
dodizzle: `decodeURIComponent(error)` is unsafe. Hono's `c.req.query()` already returns percent-decoded values, so a second `decodeURIComponent` call will throw a `URIError` if the error message contains a literal `%` not followed by valid two-digit hex (e.g. a task-store error like "50% of requests failed"). Every other error banner in this file uses `escapeHtml(opts.error)` directly with no `decodeURIComponent` (lines 324, 626, 920, 1054, 1112, 1180, 2056). Change to:
```ts
? `<div class="alert alert-danger" style="margin-bottom:16px">${escapeHtml(error)}</div>`
```

dodizzle (commented):
COMMENT — prior finding addressed: error param is now wired to the GET handler and `renderTokensPage`. One new issue introduced: `decodeURIComponent(error)` in the error banner (line 2110 of `admin-ui-pages.ts`) double-decodes an already-decoded Hono query value. If any task-store error message contains a literal `%` not followed by valid hex (e.g. "50% capacity"), `decodeURIComponent` throws a `URIError` and the entire page crashes with a 500. Every other error-rendering site in this file uses `escapeHtml(opts.error)` with no `decodeURIComponent` (lines 324, 626, 920, 1054, 1112, 1180). Fix: `escapeHtml(error)` only.

dodizzle (commented):
APPROVE (R3) — R2 finding resolved: `decodeURIComponent` removed from the error banner in `renderTokensPage`. The fix is a one-line change (`escapeHtml(decodeURIComponent(error))` → `escapeHtml(error)`) that matches the pattern used at every other error-rendering site in `admin-ui-pages.ts`. Hono's `c.req.query()` already decodes the value; the extra call was the only remaining issue. No new issues. CI green.

PR #822, opened 2026-06-28
feat(metrics): add TaskStoreProvider — 16 query kinds via admin + task store APIs

dodizzle:
## Summary
- Adds `TaskStoreProvider` (`metrics/src/providers/task-store-provider.ts`) — a `MetricsProvider` that answers all 16 `MetricQuery` kinds from the task store (tasks + PRs) and admin token-aggregation endpoints, emitting `MetricTable`s whose `columns` and `table()` envelope are byte-identical to `sql-provider.ts` (api.ts parses positionally).
- Adds two slim read-only HTTP clients following the `accounts-client.ts` pattern: `HttpTaskStoreClient` (`/tasks`, `/prs`) and `HttpAdminMetricsClient` (cron-run + chat-daily token stats), each with an injectable interface + typed error.
- `HttpTaskStoreClient` applies **client-side date-range windowing** (the live `/tasks` and `/prs` routes ignore `from`/`to`) and **paginates on `offset`** so result sets are not truncated at the server default `limit` (50). The window helpers (`taskAnchor`, `prAnchor`, `inWindow`) are the single source of truth, imported by the recorded double so cassette and live paths filter identically.
- Token kinds combine two **disjoint** sources — cron-run tokens ("cron") and chat-daily tokens ("chat") — summed field-wise, so there is no double-counting.
- Wires `taskstore` mode into `select-provider.ts` (priority above postgres/sqlite) and `server.ts`, so the provider is live when `METRICS_TASK_STORE_URL` + `METRICS_ADMIN_URL` are configured.
- Adds `Recorded{TaskStore,AdminMetrics}Client` cassette doubles, a TDD-first integration test suite, and unit tests for the client's pagination + windowing.

## Known limitation (MME-4.2)
The two admin **stats endpoints** the `tokens*` kinds target are **not implemented server-side yet**:
- `GET /agents/:id/cron-runs/stats`
- `GET /agents/chat-tokens/daily/stats`

In `taskstore` mode the `tokens*` kinds therefore throw `AdminMetricsClientError` (404) until those endpoints land in MME-4.2; this is documented inline at the wiring site in `server.ts`. The task/PR-derived kinds (summary, summaryCycleTime, trends, features*, queue*) are fully live. The provider's token behavior is proven via the `RecordedAdminMetricsClient` cassettes.

Note: several `sql-provider` sub-metrics have no task-store source today (`avg_retries`, `avg_files_changed`, `avg_fix_cascade_depth`, `avg_review_findings`, the `simplify_avg_*` sub-scores). Those columns are kept and filled with `null` so the column shapes stay byte-identical; values await richer task records / the event stream.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Implements MetricsProvider — all 16 kinds return MetricTable without throwing (offline/fixtures path) | MET |
| 2 | Token queries aggregate cron + chat with no double-counting | MET |
| 3 | Date-range (from/to) filtering applied consistently across all kinds, including the live path | MET |
| 4 | Integration tests w/ recorded fixtures: summary, queueFunnel, tokensTotals, tokensByAgentByCron | MET |

## Test Plan
- `bun test metrics/` (full package) → 737 pass, 16 skip, 0 fail
- `bun test metrics/src/lib/task-store-client.unit.test.ts` → pagination past the server limit + client-side window filtering
- `bun run --filter='@shipwright/metrics' typecheck` → clean
- `bunx biome lint metrics/src` → clean


metrics/src/lib/task-store-client.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  async listPrs(params: {
+    from?: string;
+    to?: string;
+    reviewState?: string;
+  }): Promise<PrRecord[]> {
+    const res = await this.fetch<{ items: PrRecord[] }>(
+      `/prs${HttpTaskStoreClient.query(params)}`,
+    );
+    return res.items ?? [];
```
dodizzle: Wrong response envelope. The live task-store `/prs` endpoint returns `{ prs, total, limit, offset }` — there is no `items` key (verified against the running store). `res.items ?? []` will always yield `[]` against the real API, collapsing every PR-derived metric (`reviews_total`, `reviews_ship_it`, `tasks_approved`, `avg_review_findings`). The recorded double bypasses this HTTP client, so the integration tests stay green. `listTasks` already parses the `{ tasks }` envelope correctly — do the same here: `res.prs ?? []`.

metrics/src/providers/task-store-provider.ts
```diff
@@ -0,0 +1,653 @@
...
+    const blocked = tasks.filter((t) => t.status === BLOCKED_STATUS);
+
+    // CI gates: completed tasks that recorded a ciFixAttempts value.
+    const ciTasks = completed.filter((t) => num(t.ciFixAttempts) !== null);
+    const ciFirst = ciTasks.filter((t) => num(t.ciFixAttempts) === 0).length;
+
+    const complexityCount = (n: number) =>
+      completed.filter((t) => num(t.complexity) === n).length;
+
+    const shipIt = prs.filter((p) => p.reviewState === "SHIP IT").length;
```
dodizzle: `reviewState === "SHIP IT"` never matches a real task-store PR record. "SHIP IT" is a PostHog event `verdict` in sql-provider.ts; the task store's `reviewState` only takes `approved | posted | pending | in_progress` (verified live). So `reviews_ship_it`, `tasks_approved`, and the per-feature ship-it counts (also lines 386, 486, 501) will always be 0 against the real store. The fixtures invent `reviewState: "SHIP IT"`, which hides the mismatch. Map to `reviewState === "approved"` and update the cassettes to real values.

dodizzle (commented):
COMMENT — Well-tested provider with exact column parity to sql-provider, but the two new HTTP clients encode a task-store contract that doesn't match the live API: `listPrs` reads `res.items` (live `/prs` returns `{ prs }`), `reviewState === "SHIP IT"` never matches real reviewState values (`approved|posted|pending|in_progress`), and PR records have no `findings` field. The recorded doubles bypass the HTTP layer, so the green tests don't catch these. Nothing is wired in yet (deferred to MME-4.2), so not breaking today — but worth fixing before the live path lands. Full detail inline.

metrics/src/lib/task-store-client.ts
```diff
@@ -0,0 +1,139 @@
...
+  addedAt?: string | null;
+}
+
+/** Slim PR record — drives review (SHIP IT) derived metrics. */
+export interface PrRecord {
+  id?: string;
+  taskId?: string | null;
+  reviewState?: string | null;
+  /** Number of review findings, when recorded. */
+  findings?: number | null;
```
dodizzle: Live PR records carry no `findings` field (they have `reviewCycles`, `patchCycles`, etc., but no finding count). `avg_review_findings` (provider lines 401, 515) will always be null in production even though the fixtures supply it. Lower severity than the envelope/reviewState bugs since it degrades to null rather than zeroing a count — but either source it from a real recorded field or list `avg_review_findings` among the known-null columns in the PR body.

dodizzle (commented):
APPROVE (recorded as COMMENT — self-review API constraint). Re-review of 2e79389. All three prior findings are resolved correctly:

1. `listPrs` now parses the `{ prs }` envelope (`task-store-client.ts:137/140`) instead of the non-existent `items` key.
2. `reviewState` matching extracted to `const SHIP_IT_REVIEW_STATE = "approved"` (`task-store-provider.ts:47`), aligning with live values; fixtures updated from the invented "SHIP IT" to real `approved`/`posted`.
3. The phantom `PrRecord.findings` field is removed and `avg_review_findings` is emitted as `null` with documenting comments.

The fix commit is tightly scoped to the three findings with no new logic, and CI is green. Fixtures now reflect the live task-store contract.

metrics/src/lib/task-store-client.ts
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  private headers(): Record<string, string> {
+    return {
+      "Content-Type": "application/json",
+      Authorization: `Bearer ${this.apiKey}`,
+    };
+  }
+
+  private async fetch<T>(path: string): Promise<T> {
```
dodizzle, line 162: Two production gaps converge on this single-request seam now that select-provider wires taskstore mode live:

1. **Date range is a no-op.** `listTasks`/`listPrs` forward `from`/`to`, but the live task store ignores them — verified: `/tasks?to=2020-01-01T00:00:00Z` and `?from=2030-01-01T00:00:00Z` both return the full set. Nothing filters client-side, so every task/PR kind ignores the dashboard range and reports all-time data. `RecordedTaskStoreClient` (task-store-recorded.ts:30-42) implements window filtering, so the "custom range filters cassette rows" test passes while prod is unfiltered.

2. **Truncation.** One request, no pagination. `/prs` reports `total:64` but returns `limit:50` — 14 PRs dropped, growing with the store. The double returns the full array, masking it.

Fix: filter client-side in TaskStoreProvider (same anchors the double uses) and paginate on `offset` until `offset + returned >= total` (or pass a large `limit` and assert `total <= limit`).

dodizzle (commented):
Re-review (cycle 3). Commit 0994f6b wires TaskStoreProvider into select-provider.ts (taskstore mode, priority above postgres/sqlite) and server.ts — so the provider is now live in production, not deferred to MME-4.2 as the PR body states. That promotion exposes two production-incorrectness gaps that the recorded doubles mask:

1. Date-range filtering is a no-op against the live store. `from`/`to` are forwarded as query params, but the running task store ignores them (`/tasks?to=2020-01-01` and `?from=2030-01-01` both return the full set). The provider does no client-side windowing, so every task/PR kind reports all-time data regardless of the dashboard range. The recorded double DOES filter by window, so the range test passes while production is unfiltered.

2. Single-page fetch truncates at the server default `limit`. `/prs` reports total:64 but returns limit:50 — 14 PRs silently dropped, growing with the store. No pagination loop; the in-memory double returns everything, hiding it.

Also: the live `tokens*` kinds now call admin stats endpoints the body itself says are unimplemented server-side, so they will 404/throw in taskstore mode.

None of this breaks offline/fixtures mode, but all of it is wrong the moment taskstore mode is selected — which this PR makes possible. Verdict: COMMENT (recorded as COMMENT per the self-review API constraint). Suggest client-side windowing + pagination in TaskStoreProvider/HttpTaskStoreClient, and a PR-body update reflecting that wiring shipped here.

metrics/src/select-provider.ts
```diff
@@ -26,25 +28,25 @@ export interface ProviderEnv {
...
-  const hasPostHog =
-    typeof env.POSTHOG_PERSONAL_API_KEY === "string" &&
-    env.POSTHOG_PERSONAL_API_KEY.length > 0 &&
-    typeof env.POSTHOG_PROJECT_ID === "string" &&
-    env.POSTHOG_PROJECT_ID.length > 0;
-  if (hasPostHog) return "posthog";
+  const hasTaskStore =
+    env.METRICS_TASK_STORE_URL?.startsWith("http") === true &&
+    env.METRICS_ADMIN_URL?.startsWith("http") === true;
+  if (hasTaskStore) return "taskstore";
```
dodizzle, line 49: This wires `taskstore` mode in (priority #2, above postgres/sqlite) — contradicting the PR body's claim that select-provider wiring is deferred to MME-4.2. The prior review cycles treated "not wired in" as the reason the contract gaps weren't breaking; that rationale no longer holds. Note that in this mode the `tokens*` kinds call admin stats endpoints the body says are unimplemented server-side, so they'll 404 and throw `AdminMetricsClientError`. Either gate taskstore mode on those endpoints landing, or document the `tokens*` kinds as erroring until MME-4.2 — and update the PR body.

dodizzle (commented):
APPROVE (re-review, cycle 4)

All three findings from the prior review (cycle 3) are resolved:

1. **Date-range filtering** — `HttpTaskStoreClient.listTasks`/`listPrs` now apply client-side windowing via exported `taskAnchor`/`prAnchor`/`inWindow` helpers (the live `/tasks` and `/prs` routes ignore `from`/`to`). The recorded double now imports these same helpers instead of reimplementing them, so the cassette and live HTTP paths filter byte-identically — closing the drift that previously masked the no-op.

2. **Pagination** — `fetchAllPages` loops on `offset` (PAGE_SIZE 200) until the reported `total` is reached, with defensive termination on short/empty pages and a hard iteration guard. A new unit test proves it against a fetch double that caps at 50 server-side (130 rows -> 130 returned, multiple requests).

3. **PR body** — now accurately states the `select-provider.ts`/`server.ts` wiring shipped in this PR, with a "Known limitation (MME-4.2)" section documenting that `tokens*` kinds 404 until the admin stats endpoints land; a matching inline NOTE was added at the wiring site in `server.ts`.

`fetch` is injected for testability (no global override, honoring the repo's isolation rule). New unit tests (5) and the task-store integration suite (7) pass locally; CI green on d9ad7b3. No new issues — the fix is tightly scoped and the helper de-duplication is a genuine quality improvement over a copy-paste fix.

(Recorded as COMMENT per the GitHub self-review API constraint; the verdict is APPROVE.)

PR #824, opened 2026-06-28
feat(render-plan): upload rendered HTML and return a shareable URL with local-open fallback

dodizzle:
## Summary
- Extend `render-plan.ts` so that, when no `--out` is given, it uploads the rendered HTML to the task-store `POST /docs` endpoint (bearer auth, raw HTML body), parses the `{ id, url }` response, and prints the absolute shareable URL to stdout.
- Best-effort local open: spawns `open` (macOS) / `xdg-open` (Linux) **only** under TTY + display detection; otherwise just prints the URL (cloud/Slack contexts).
- Fully defensive fallback: when `SHIPWRIGHT_TASK_STORE_URL`/`TOKEN` are unset or the upload fails (non-2xx, network error, unparseable body, missing `url`), it writes the HTML to a local temp file and prints that path — never throws, never blocks, and the existing `--out` behavior and exit codes are unchanged.
- New injectable dependencies (`HttpClient`, `OpenEnv`, `SpawnFn`) mirror the existing `Clock` DI pattern, so all paths are testable without touching the network or spawning real processes.

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | On success, uploads HTML and prints the absolute hosted URL | MET |
| 2 | Env unset / upload fails → writes local temp file, never throws/blocks | MET |
| 3 | Local-open only under TTY + display; otherwise prints URL | MET |
| 4 | Injected fake HTTP client (no `global.fetch` override); unit tests for context-detection + open-command; net-new tests | MET |

## Test Plan
- `render-plan-upload.integration.test.ts` — injected `recordingFetch` asserts the POST endpoint (`<base>/docs`), `Authorization: Bearer <token>` header, and raw-HTML body; asserts the parsed absolute `url` is returned; plus fallback cases (non-2xx, client throws, missing `url`, env unset) each write a temp file without throwing.
- `render-plan.unit.test.ts` — added `shouldOpenLocally` (no TTY / darwin / linux+DISPLAY / linux-no-DISPLAY / non-tty) and `openCommand` (darwin→open, linux→xdg-open, other→null) cases.
- `bun test plugins/shipwright/scripts/` → 316 pass / 0 fail. Biome lint clean. `tsc --noEmit` clean.


plugins/shipwright/scripts/render-plan.ts
```diff
@@ -361,18 +514,58 @@ function main(): void {
...
+
+  const isHostedUrl = /^https?:\/\//.test(result);
+  if (isHostedUrl) {
+    maybeOpenLocally(result, {
+      platform: process.platform,
+      isTTY: Boolean(process.stdout.isTTY),
+      display: process.env.DISPLAY ?? "",
+    });
+  } else {
+    process.stderr.write(`render-plan: wrote local file ${result}\n`);
```
dodizzle, line 559: Minor: when the env vars are unset, `main()` already wrote a "…unset — writing local file" notice to stderr at the top of this branch, and this line then writes "wrote local file {result}" for the same outcome — two stderr lines for one event. Harmless and arguably informative, but could be collapsed if you want a single notice. Suggestion only.

dodizzle (commented):
APPROVE (recorded) — posted as COMMENT since GitHub rejects self-APPROVE via the API. Clean, well-tested, and contract-verified: the `POST /docs` client matches the task-store endpoint (raw-text body via `c.req.text()`, `201 { id, url }`), DI mirrors the existing `Clock` pattern, tests inject a fake `fetch` rather than overriding globals (honoring the CLAUDE.md isolation rule), the fallback never throws, and `--out`/exit-code behavior is preserved. CI green. No blocking concerns — one minor stderr-redundancy suggestion inline.

PR #832, opened 2026-06-28
fix(task-store): drop assignee=null guard from inRepoScope check

dmcaulay:
## Summary
- Remove the `task.assignee === null` guard from `inRepoScope` in `requireOwnership()` so repo-scoped tokens can PATCH tasks regardless of whether the task has an assignee
- Update JSDoc on `requireOwnership()` to reflect the corrected condition 4

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | PATCH /tasks/:id returns 200 when repo in scope, even if assignee is different | MET |
| 2 | PATCH still 403s when no assignee/claimedBy/repo match | MET |
| 3 | Existing pool-task test (assignee=null, repo in scope) continues to pass | MET |
| 4 | New smoke test: assigned task + repo-scoped token → 200 | MET |
| 5 | JSDoc updated to reflect corrected condition 4 | MET |

## Test Plan
- [x] New smoke test: PATCH on task with `assignee="other-agent"` using repo-scoped token whose repos include the task's repo → 200
- [x] Existing pool-task smoke test (assignee=null) continues to pass
- [x] Full task-store test suite: 191 pass, 0 fail
- [x] Biome lint: clean

### DIFF
--- a/task-store/src/api.smoke.test.ts
+++ b/task-store/src/api.smoke.test.ts
@@ -889,6 +889,34 @@ describe("task-store API (smoke)", () => {
     expect(res.status).toBe(200);
   });
 
+  it("PATCH /tasks/:id on assigned task with repo-scoped token returns 200 when task repo is in scope", async () => {
+    // Task is assigned to agent-2, not agent-1
+    const assignedTask = makeTask({
+      id: "assigned-1",
+      assignee: "agent-2",
+      repo: "acme-inc/backend-api",
+    });
+
+    const app = makeApp({
+      tokenService: fakeRepoAgentTokenService(["acme-inc/backend-api"]),
+      taskService: fakeTaskService({ getResult: assignedTask }),
+      scopeResolver: makeScopeResolver(["acme-inc/backend-api"]),
+    });
+
+    const res = await app.request("/tasks/assigned-1", {
+      method: "PATCH",
+      headers: {
+        Authorization: `Bearer ${AGENT_TOKEN}`,
+        "content-type": "application/json",
+      },
+      body: JSON.stringify({ status: "in_progress" }),
+    });
+
+    // Should return 200 because the task's repo is in the token's scope,
+    // even though the task is assigned to a different agent.
+    expect(res.status).toBe(200);
+  });
+
   it("GET /tasks with repo-scoped agent token passes agentScope (not assignee) to list()", async () => {
     const capturedFilters: TaskListFilters[] = [];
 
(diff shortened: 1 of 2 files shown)


task-store/src/api.smoke.test.ts
```diff
@@ -889,6 +889,34 @@ describe("task-store API (smoke)", () => {
...
+      headers: {
+        Authorization: `Bearer ${AGENT_TOKEN}`,
+        "content-type": "application/json",
+      },
+      body: JSON.stringify({ status: "in_progress" }),
+    });
+
+    // Should return 200 because the task's repo is in the token's scope,
+    // even though the task is assigned to a different agent.
+    expect(res.status).toBe(200);
```
dmcaulay, line 926: The new test only asserts `res.status === 200` but doesn't check what happened to `task.assignee`. The PATCH handler at `routes/tasks.ts` lines 236-238 contains:
```ts
if (agentId !== null && task.assignee !== null) {
  body.assignee = agentId;
}
```
This fires here — agent-2's task silently becomes agent-1's task. If the intent is that repo-scoped access implies task takeover, add:
```ts
expect((await res.json() as Task).assignee).toBe('agent-1');
```
If the intent is that the acting agent should NOT steal the task, the force-reassignment guard at `routes/tasks.ts:236-238` needs a companion condition — e.g., skip force-assign when `task.repo` is in scope but `task.assignee !== agentId`.

dmcaulay (commented):
COMMENT — Two behaviors introduced by the authorization relaxation are unverified.

**Finding 2: No PATCH-specific 403 test for repo-scoped token with wrong repo**

AC#2 ("PATCH still 403s when no assignee/claimedBy/repo match") is marked MET, but there is no PATCH test that directly exercises the wrong-repo 403 boundary. The existing test at line 707 covers GET with a repo-scoped token that has an empty scope — it returns 403. However, `requireOwnership()` was directly modified in this PR and gates all write paths. The change narrows the conditions under which 403 fires; the negative path for PATCH should be independently verified.

**Suggestion**: Add a smoke test: repo-scoped token for `['other-org/other-repo']` PATCHes a task with `repo: 'acme-inc/backend-api'` → expect 403.

dmcaulay (commented):
APPROVE — Both review findings from R1 are resolved. The `spyTaskService` correctly uses `assignedTask` as the merge base so the assignee assertion is a true regression guard. The 403 boundary test for wrong-repo PATCH is present and correct. Guard logic in `routes/tasks.ts` is precisely scoped (`ownedByAssignee || ownedByClaim`). No new issues. CI green.

PR #837, opened 2026-06-28
chore(metrics): remove event-store providers, PostHog SDK, and HogQL layer

dmcaulay:
## Summary
- Deletes all event-store backed providers (PostHog, SQLite, Postgres, SQL) and their test files from `metrics/src/`
- Removes PostHog client, HogQL query builders, local-store, cache, and validate-hogql modules
- Cleans up `api.ts` (removes batch ingest endpoint, `PostHogClientLike` DI seam, old provider fallback logic) and `server.ts` (narrows bootstrap to `fixtures` | `taskstore` modes only)
- Removes `posthog` and `pg` npm dependencies from `metrics/package.json`

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | All listed files deleted; no posthog/HogQL/event-store imports remain | ✅ MET |
| 2 | Full test suite green, no broken imports or missing symbols | ✅ MET |
| 3 | posthog and pg removed from metrics/package.json | ✅ MET |
| 4 | Batch ingest endpoint (POST /batch/) gone, smoke test deleted | ✅ MET |
| 5 | All deleted test files retired, no new tests needed | ✅ MET |

## Test Plan
- [x] 202 metrics tests pass, 0 fail
- [x] All workspace typechecks clean (`bun run --filter='*' typecheck`)
- [x] Biome lint clean (320 files, no fixes needed)


metrics/src/select-provider.ts
```diff
@@ -48,20 +40,7 @@ export function selectProviderMode(env: ProviderEnv): ProviderMode {
...
- * Returns the first non-empty value among METRICS_DATABASE_URL /
- * DATABASE_URL_METRICS, or undefined when neither is set.
- */
-export function resolvePostgresUrl(env: ProviderEnv): string | undefined {
-  const url =
-    env.METRICS_DATABASE_URL?.trim() || env.DATABASE_URL_METRICS?.trim();
-  return url || undefined;
+  // Default: return taskstore so server.ts can provide a clear error when URLs
+  // are not configured, rather than silently falling back to a removed backend.
+  return "taskstore";
```
dodizzle, line 46: `selectProviderMode` can now only return `"fixtures"` or `"taskstore"` — this default returns `"taskstore"`. The comment says "server.ts will error out if the URLs are missing," but server.ts's `console.error("FATAL: no provider configured…") + process.exit(1)` lives in the `else` after the `fixtures`/`taskstore` branches, so it's now dead code and can never run.

Result: when taskstore mode is selected with `METRICS_TASK_STORE_URL`/`METRICS_ADMIN_URL` unset or non-http, the code enters the taskstore branch and constructs `HttpTaskStoreClient("")` / `HttpAdminMetricsClient("")` with empty base URLs. The intended fast FATAL never fires; the service boots and fails later with an opaque error on the first query.

Fix options: (a) return a distinct unconfigured mode here (e.g. `"none"`) so the FATAL `else` in server.ts becomes reachable — keeps the selector as the single source of mode truth; or (b) drop the dead `else` and add an explicit URL-validity guard at the top of the taskstore branch in server.ts that exits with the helpful message when the URLs are missing/non-http.

dodizzle (commented):
Clean, well-tested removal of the event-store/PostHog/HogQL layer — no orphaned imports remain and tests were updated in lockstep. One reachable design gap (inline): the intended fail-fast-on-missing-config path in `server.ts` is dead code, so a misconfigured taskstore deployment boots and fails opaquely instead of exiting with the helpful FATAL message.

Verdict: COMMENT.

Minor (out of diff, no action required): `metrics/src/providers/task-store-provider.ts` still has header/inline comments referencing the now-deleted `sql-provider.ts` ("byte-identical to sql-provider.ts", "mirror sql-provider") — worth a refresh so future readers aren't sent looking for a deleted file.

dmcaulay (commented):
Fixed in 3465a2e. Moved the URL-validity guard to the top of the taskstore branch in server.ts —  now fires when METRICS_TASK_STORE_URL or METRICS_ADMIN_URL are missing/non-http. The dead else branch is gone. Also refreshed the stale sql-provider.ts references in task-store-provider.ts while here.

metrics/src/select-provider.ts
dmcaulay: Fixed in 3465a2e — moved the URL-validity guard to the top of the taskstore branch in server.ts (option b). Dropped the unreachable `else`. Also refreshed the stale `sql-provider.ts` comments in task-store-provider.ts.

dodizzle (approved):
Both prior findings addressed. The FATAL exit path in `server.ts` is now reachable — URL-validity guard fires before constructing clients, so a misconfigured deployment exits cleanly instead of booting with empty base URLs. Stale `sql-provider.ts` comment references in `task-store-provider.ts` cleaned up. CI green. APPROVE.

dmcaulay (commented):
APPROVE — R2, no blocking findings

All R1 findings are resolved. The cleanup is thorough — no posthog/pg/sqlite/HogQL imports remain in any production code path. `select-provider.ts`, `server.ts`, `api.ts`, and `task-store-provider.ts` are all internally consistent after the deletions. The e2e test-server wiring from the recent fix commit correctly injects `createFixtureTaskStoreProvider`. 10,000+ lines deleted, clean result.

PR #842, opened 2026-06-29
feat(admin): add public read-only task board route (PPL-2.1)

dmcaulay:
## Summary
- Add public unauthenticated read-only task board at `GET /public/tasks` in the admin service
- Scope task list to `SHIPWRIGHT_ADMIN_PUBLIC_REPO` via repo filter param on task-store queries
- Strip all create/edit/status-change controls from the rendered HTML when `readOnly=true`

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Unauthenticated `GET /public/tasks` returns 200 and renders the task list — no login redirect, no 401 | MET |
| 2 | Only tasks with `repo === SHIPWRIGHT_ADMIN_PUBLIC_REPO` appear in the response | MET |
| 3 | No create/edit/status-change controls are present in the rendered HTML | MET |
| 4 | POST/PUT/DELETE to `/public/tasks` returns 404 or 405 | MET |
| 5 | Smoke tests with injected doubles, no `mock.module()` or `global.*` | MET |

## Test Plan
- 7 new smoke tests in `admin/src/admin-ui.smoke.test.ts` (unauth 200, repo filter param, no controls in HTML, degraded mode, POST/PUT/DELETE → 404/405)
- Injected `fetchTaskStoreTasks` double verifies repo filter param is passed
- 810 tests pass, 0 fail
- `api-auth.ts` 100% line/branch coverage; `admin-ui.ts` 91.74%; `admin-ui-pages.ts` 94.12%
- Biome lint clean; all 5 packages typecheck green


admin/src/admin-ui.ts
```diff
@@ -2103,5 +2112,49 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
...
+    }
+
+    return html(
+      renderTasksPage(
+        tasks,
+        { repo: publicRepo },
+        degraded,
+        "",
+        {},
+        { total, limit: 50, page: 1 },
```
dodizzle, line 2151: `total` here comes from the task-store's actual count. When the public repo has more than 50 tasks, `renderTasksPage` renders pagination links via `makePageUrl` (line 1428 of `admin-ui-pages.ts`), which always returns `/admin/tasks?...` -- an auth-walled route. Unauthenticated visitors clicking "Next" get a login redirect.

Simplest fix: suppress pagination in read-only mode by passing `Math.min(total, 50)` as the total (so `totalPages` stays 1), or add a `hidePagination` guard around the `paginationHtml` block when `readOnly=true`.

dodizzle (commented):
COMMENT -- one finding before merge.

Pagination links in read-only mode hardcode `/admin/tasks` via the pre-existing `makePageUrl` helper, so any public board with >50 tasks will route unauthenticated visitors to the auth wall. Fix: suppress pagination when `readOnly=true` (simplest), or thread a `baseUrl` param through `makePageUrl`.

Everything else is solid: `publicNoAuthMiddleware` is a clean no-op that does not touch existing auth paths, `readOnly` is applied consistently across all control surfaces, all 5 acceptance criteria are met, and the 7 smoke tests use injected doubles with no `mock.module()`/`global.*`.

dmcaulay:
Fixed in e78d50c.

Added `|| readOnly` to the `paginationHtml` condition in `renderTasksPage` — when `readOnly=true`, pagination is suppressed entirely, so no `/admin/tasks` links can appear for unauthenticated visitors. Also added a smoke test that asserts no Next/Prev links or `/admin/tasks` hrefs appear when the task store returns `total: 500` in public mode.

dmcaulay (commented):
APPROVE — R1, one minor finding

Implementation is clean: auth properly bypassed via no-op middleware, all mutation controls suppressed in `readOnly` mode, admin toolbar/filters/pagination omitted, smoke tests cover the key scenarios with injected doubles. No security or correctness issues.

**Finding — public board silently truncates at 50 tasks with no indicator (`admin/src/admin-ui.ts` ~line 2151)**

The route fetches `limit=50` and pagination is intentionally suppressed in readOnly mode (correct — pagination links target `/admin/tasks` which is auth-walled). But when `total > 50`, the page shows at most 50 rows with no visible count or truncation notice. A visitor has no way to know the list is incomplete.

Not blocking — the public board is functional and secure — but worth a follow-up. Simple fix: render a plain-text `Showing 50 of {total} tasks` line below the table when `readOnly && total > limit`, without linking to any `/admin/` route.

PR #845, opened 2026-06-29
fix(chart): wire metrics task-store env so dashboard avoids SQLite crash-loop

dodizzle:
## Problem

`shipwright-metrics` crash-loops on boot with `SQLiteError: unable to open database file` (`SQLITE_CANTOPEN`) at `local-store.ts:94`.

Root cause is chart/code drift. Metrics **v0.91.0** (#822) **removed PostHog backend mode** and replaced it with `taskstore` mode, but the Helm chart was never updated:
- The metrics Deployment still injects the now-dead `POSTHOG_*` env and has **no** `METRICS_TASK_STORE_URL` plumbing.
- With no PostHog mode, no Postgres URL, and no task-store URL, `selectProviderMode` falls through to **`sqlite`** → opens `state/metrics.db`, whose parent dir doesn't exist in the published image (not a writable volume) → crash.

## Fix

Add the task-store wiring the new mode needs:
- `metrics.provider.taskStoreUrl` → `METRICS_TASK_STORE_URL` (plain value, like `adminUrl`)
- `metrics.provider.taskStoreToken.existingSecret/key` → `METRICS_TASK_STORE_TOKEN` (secretKeyRef, like `internalKey`)

Set `taskStoreUrl` alongside `adminUrl` to select `taskstore` mode. Includes `values.schema.json` entries, 4 helm-unittest cases (set/omit for both env vars), CHANGELOG + artifacthub annotation, and chart bump to **1.6.117**.

## Verification

- `helm unittest` — 221 passed (4 new)
- `helm lint` — clean
- `ct lint --check-version-increment` — ok (1.6.116 → 1.6.117)
- `helm template` renders both env vars when values are set, omits them by default

## Follow-up (separate, vitals-os repo)

This adds the chart *plumbing* only. The prod release also needs `infra/helm/shipwright-deploy/values-prod.yaml` to set `metrics.provider.taskStoreUrl` (+ token secret), drop the dead PostHog block, and bump the `shipwright` chart dependency to 1.6.117.

### DIFF
--- a/charts/shipwright/CHANGELOG.md
+++ b/charts/shipwright/CHANGELOG.md
@@ -10,6 +10,18 @@ independent of `appVersion`. CI enforces this with
 `ct lint --check-version-increment`. Each release here must mirror the
 `artifacthub.io/changes` annotation in `Chart.yaml`.
 
+## [1.6.117] - 2026-06-28
+
+### Added
+
+- metrics: wire `METRICS_TASK_STORE_URL` + `METRICS_TASK_STORE_TOKEN` into the
+  metrics Deployment via `metrics.provider.taskStoreUrl` and
+  `metrics.provider.taskStoreToken`. Set `taskStoreUrl` alongside `adminUrl` to
+  run the dashboard in TASKSTORE mode. Without this the provider selector falls
+  through to the bundled SQLite store, which has no writable path in the
+  published image and crash-loops on boot (`SQLITE_CANTOPEN`) — the regression
+  introduced when metrics v0.91.0 replaced PostHog mode with task-store mode.
+
 ## [1.6.116] - 2026-06-29
 
 ### Changed
--- a/charts/shipwright/Chart.yaml
+++ b/charts/shipwright/Chart.yaml
@@ -8,7 +8,7 @@ type: application
 # Chart version (SemVer). Bump on every chart change, independent of appVersion.
 # See CHANGELOG.md and the README "Versioning" section — CI enforces this bump
 # via `ct lint --check-version-increment`.
-version: 1.6.116
+version: 1.6.117
 # Version of the Shipwright application this chart deploys by default.
 appVersion: "0.1.0"
 home: https://shipwright-harness.com
@@ -28,8 +28,8 @@ annotations:
   # Artifact Hub change log for this release. Must mirror the matching version
   # entry in CHANGELOG.md (keep the two in sync on every bump).
   artifacthub.io/changes: |
-    - kind: changed
-      description: "auto-bump to chart v1.6.116 triggered by release tag admin-v0.112.0"
+    - kind: added
+      description: "metrics: wire METRICS_TASK_STORE_URL + METRICS_TASK_STORE_TOKEN so the dashboard can select TASKSTORE mode instead of crash-looping on the SQLite fallback"
 dependencies:
   # Bitnami PostgreSQL subchart. Pinned to a chart version whose default image tag
   # is concrete (NOT `latest`), because Bitnami's 2025 catalog change moved newer
--- a/charts/shipwright/templates/metrics-deployment.yaml
+++ b/charts/shipwright/templates/metrics-deployment.yaml
@@ -103,6 +103,23 @@ spec:
             - name: METRICS_ADMIN_URL
               value: {{ .Values.metrics.provider.adminUrl | quote }}
             {{- end }}
+            # Task-store mode: when taskStoreUrl is set alongside adminUrl, the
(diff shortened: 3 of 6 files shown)


dodizzle (commented):
## Review: fix(chart): wire metrics task-store env so dashboard avoids SQLite crash-loop

**Verdict: COMMENT** (self-review — dodizzle reviewing dodizzle)

### Helm fix (the stated goal) — looks correct

The core chart change is solid. `METRICS_TASK_STORE_URL` as a plain literal value and `METRICS_TASK_STORE_TOKEN` via `secretKeyRef` correctly mirrors the existing `internalKey` pattern in the same template. The `values.schema.json` `if/then` validation (require `key` when `existingSecret` is non-empty) is a nice guard. Four Helm unit tests cover the set/omit cases for both env vars. `helm lint`, `helm unittest` (221 passed, 4 new), and `ct lint --check-version-increment` all pass. The crash-loop fix is sound — `selectProviderMode` needs `METRICS_TASK_STORE_URL` to avoid falling through to SQLite, and this PR supplies it.

### Finding 1 — Bundled unrelated feature (scope mismatch)

**File:** `admin/src/admin-ui.ts`, `admin/src/admin-ui-pages.ts`, `admin/src/api-auth.ts`, `admin/src/admin-ui.smoke.test.ts`

The PR title and body describe only a Helm chart fix, but 107 of the 110 additions are a new public read-only task board (`GET /public/tasks` — unauthenticated, scoped to `SHIPWRIGHT_ADMIN_PUBLIC_REPO`). This includes a new `AdminUIDeps.publicRepo` field, a `publicNoAuthMiddleware`, a full Hono route handler, 8 smoke test cases, and doc updates to `docs/agent.md`, `docs/architecture.md`, and `docs/configuration.md`. This is a `feat(admin):` PR bundled silently inside a `fix(chart):`. The chart fix is merge-critical (stops the crash-loop in prod) and the public board is new feature scope — these should have been separate PRs. As shipped, the semantic-release commit parser will tag this as a patch fix and miss the feature in the release notes. Not a blocker if you need the chart fix out fast, but worth noting.

### Finding 2 — `taskStoreToken.key` default deviates from chart convention

**File:** `charts/shipwright/values.yaml`, line ~601

Every other `key` field in this chart defaults to `""` (empty string) — e.g. `encryptionSecret.keyEncryptionKey`, `sessionSecret.key`, `posthog.key`, `internalKey.key`. The new `taskStoreToken.key` defaults to `"METRICS_TASK_STORE_TOKEN"` (a non-empty value). This is benign at runtime because the `{{- if .Values.metrics.provider.taskStoreToken.existingSecret }}` guard means the key is only used when an existing secret is named. But it creates an inconsistency: if `existingSecret` is set without overriding `key`, the chart silently uses the env var name as the secret key rather than failing fast. Defaulting to `""` (matching all sibling fields) and letting the schema `then`-required `key` enforce it would be more consistent.

### Finding 3 — `publicNoAuthMiddleware` is a no-op that adds an invisible abstraction layer

**File:** `admin/src/api-auth.ts`

The middleware exported as `publicNoAuthMiddleware` does nothing but call `next()`. The `AdminAuthzPolicy` type is declared but never consumed anywhere in this PR. The route registration `app.get("/public/tasks", publicNoAuthMiddleware, ...)` reads as if the middleware is doing something meaningful (e.g. marking the route in an auth policy registry), but it's purely documentary. This is fine as documentation intent, but be aware that any future caller reading just the middleware name (not its body) may assume it enforces something. A comment at the call site or in the export would help: `// explicitly unauthenticated — no auth check by design, not an oversight`.

### Minor notes (not blocking)

- The `CHANGELOG.md` lists `[1.6.117]` above `[1.6.116]`, which is correct newest-first ordering. But `1.6.116` is an auto-bump entry for `admin-v0.112.0` that appears to have been inserted for a concurrent release. Confirm `1.6.116` never actually shipped as a chart release (the current chart in main is `1.6.115`); if it did, this PR would create a gap.
- The import reorder in `admin-ui.smoke.test.ts` (moving `PrListItem, PullRequestItem` import before `createAdminUIApp`) is a biome organize-imports fix — clean.
- The page title change from `Tasks — Shipwright Admin` to `Tasks — Shipwright` in the public board path is intentional; confirm the admin variant of the page still uses the original title (it does — the change is gated on `readOnly`).

### Summary

The Helm fix is correct and the crash-loop will stop once `metrics.provider.taskStoreUrl` is set in the deployment values (follow-up noted in PR body). The bundled public board feature is functionally sound and well-tested, but lives under a misleading PR title. The `taskStoreToken.key` default inconsistency is minor. Nothing here blocks merge — ship it.

PR #857, opened 2026-06-29
feat(chart): add metrics.public.{enabled,repo} to wire public proof surface (PPL-1.3)

dodizzle:
## Summary
Closes the chart-side wiring gap for the public metrics proof surface (PPL-1.2). The metrics service *code* mounts the unauthenticated, repo-scoped `/public/*` routes when `SHIPWRIGHT_METRICS_PUBLIC_MODE`/`_PUBLIC_REPO` are set — but the chart had **no way to inject them**: `metrics-deployment.yaml` rendered no `extraEnv` and there were no public values. So public mode could not be enabled via Helm.

## Changes
- `values.yaml`: new `metrics.public.{enabled,repo}` block (defaults off; additive, preserves existing behavior).
- `metrics-deployment.yaml`: render `SHIPWRIGHT_METRICS_PUBLIC_MODE=true` + `SHIPWRIGHT_METRICS_PUBLIC_REPO` when `public.enabled`; `repo` is `required` (template fails fast).
- `values.schema.json`: `public` block with `if/then` requiring `repo` (minLength 1) when `enabled=true` — schema-level fail-fast (defense-in-depth with the template `required`).
- `tests/metrics_workload_test.yaml`: enabled→both env present; default→absent; enabled+empty-repo→render fails.
- Chart version `1.6.121 → 1.6.122` + CHANGELOG + `artifacthub.io/changes` (kept in sync per chart policy).

## Verification
- `helm lint`: clean. `helm unittest`: 224/224 pass. Render confirmed for enabled (both env vars) and enabled-without-repo (rejected).

## Why now
Prerequisite for lighting up `proof.shipwrightharness.com`. The admin bootstrap counterpart is #856; the deploy-chart consumption + image bumps land in the `vitals-os` repo.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #859, opened 2026-06-29
fix(metrics): serve public dashboard assets + data under /public mount (PPL-1.4)

dodizzle:
## Summary
The merged public metrics dashboard (PPL-1.2) renders its shell but **loads no live data**. `/public/dashboard` was rendered with the global `basePath` (empty in prod), so the shared client (`dashboard/app.js`) set `window.__METRICS_BASE__=""` and fetched the **authenticated** `/metrics/*` endpoints (401) instead of the public `/public/metrics/*` ones. Its CSS/JS also resolved to the authenticated `/dashboard/*` paths. The existing smoke test only asserts the HTML shell, so it never caught this.

## Changes (`metrics/src/api.ts`)
- Render `/public/dashboard` with base `${basePath}/public` → `__METRICS_BASE__` resolves so `app.js` fetches `/public/metrics/*` (repo-scoped, no auth).
- Serve the dashboard assets under the public mount: `GET /public/dashboard/{styles.css,app.js}` (literal routes, mirroring the authenticated static-file serving; `basePath` stays the external proxy-stripped prefix, present only in rendered URLs).
- `createPublicMetricsApp` takes an optional `dashboardDir` (defaults to the bundled dir) for asset serving + testability.

## Tests (`api.public.smoke.test.ts`)
- Asserts `__METRICS_BASE__` + asset hrefs resolve to `/public`; non-empty `basePath` prefixes correctly; both static assets serve 200 with correct content types.

## Verification
Full metrics suite: 236 pass / 0 fail. Typecheck + Biome clean.

## Context
Third and final prerequisite for the public proof surface, alongside #856 (admin bootstrap, merged) and #857 (chart wiring). Without this the metrics proof page would be blank.

### DIFF
--- a/metrics/src/api.ts
+++ b/metrics/src/api.ts
@@ -1153,14 +1153,28 @@ const PUBLIC_POLICY = { kind: "public" as const };
 /**
  * Build the public, unauthenticated, repo-scoped metrics sub-app.
  *
+ * The public surface lives under a `/public` mount: JSON at /public/metrics/*,
+ * the read-only dashboard at /public/dashboard, and its client assets at
+ * /public/dashboard/{styles.css,app.js}. The dashboard is rendered with its base
+ * set to that `/public` mount so the shared client (`app.js`) fetches
+ * /public/metrics/* (repo-scoped, no auth) instead of the authenticated
+ * /metrics/* endpoints — otherwise the page renders but loads no data.
+ *
  * @param provider - a repo-scoped MetricsProvider (built in server.ts with the
  *   configured public repo). All reads are already narrowed to that repo.
- * @param basePath - optional path prefix for dashboard asset/API URLs.
+ * @param basePath - optional path prefix the whole public mount sits under.
+ * @param dashboardDir - directory holding the dashboard static assets
+ *   (styles.css, app.js); defaults to the bundled ./dashboard dir.
  */
 export function createPublicMetricsApp(
   provider: MetricsProvider,
   basePath = "",
+  dashboardDir: string = join(import.meta.dir, "dashboard"),
 ): OpenAPIHono<AuthEnv> {
+  // Base for the public dashboard's assets + client API calls. The public routes
+  // are registered literally under "/public/*", so the rendered base must resolve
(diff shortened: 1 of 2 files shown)


metrics/src/api.ts
```diff
@@ -1222,17 +1236,55 @@ export function createPublicMetricsApp(
...
+    { contentType: string; file: string; cache?: string }
+  > = {
+    "/public/dashboard/styles.css": {
+      contentType: "text/css; charset=utf-8",
+      file: "styles.css",
+      cache: "public, max-age=3600",
+    },
+    "/public/dashboard/app.js": {
+      contentType: "application/javascript; charset=utf-8",
+      file: "app.js",
```
dodizzle, line 1267: `/public/dashboard/app.js` has no `Cache-Control` header while `styles.css` above gets `public, max-age=3600`. The authenticated `/dashboard/app.js` also omits it, so this is consistent with existing behaviour — but if a CDN is in front it will serve both assets differently. Consider adding `cache: "public, max-age=3600"` here too (or removing it from `styles.css`) to keep the two consistent.

dodizzle (commented):
APPROVE — Clean, minimal fix for a silent production regression where the public dashboard was rendering a blank shell (401s on authenticated endpoints). Passing `publicBase` as `basePath` and adding literal `/public/dashboard/{styles.css,app.js}` routes is exactly right. Additive-only, no breaking changes, well-tested at the smoke layer. One cosmetic suggestion: `app.js` in `PUBLIC_STATIC_FILES` has no `Cache-Control` header while `styles.css` gets `public, max-age=3600` — worth aligning, but not a blocker.

*(Self-review: GitHub API rejects self-APPROVE, submitting as COMMENT. Verdict is APPROVE.)*

dodizzle (commented):
APPROVE — Prior suggestion addressed: `app.js` now has `Cache-Control: public, max-age=3600` matching `styles.css`. Single-line, correct fix. No new issues, no blocking concerns. CI green.

*(Self-review: GitHub API rejects self-APPROVE, submitting as COMMENT. Verdict is APPROVE.)*

PR #869, opened 2026-06-30
fix: pass dashboardDevAuth through in fixture mode (server.ts offline branch)

dodizzle:
## Summary
- Fixes a wire-up bug in `metrics/src/server.ts` where `dashboardDevAuth` was read from env but never forwarded into `deps` in the `fixtures` (offline) branch — causing `/metrics/*` to return 401 and show a red error banner on the dashboard even when `METRICS_DASHBOARD_DEV_AUTH=true`
- Adds one line: `dashboardDevAuth,` to the fixture-branch deps object, matching the existing behavior in the taskstore branch
- Adds `server-offline-mode.smoke.test.ts` to catch this class of wire-up bug at the server.ts DI layer

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `/dashboard` loads with no red error banner when `METRICS_OFFLINE=true` + `METRICS_DASHBOARD_DEV_AUTH=true` | ✅ MET |
| 2 | `/metrics/summary?preset=7d` returns 200 with no Authorization header when both env vars set | ✅ MET |
| 3 | New `server-offline-mode.smoke.test.ts` covers server.ts fixture-branch wiring | ✅ MET |
| 4 | Existing `dashboard-dev-auth.smoke.test.ts` tests pass unchanged | ✅ MET |
| 5 | No `mock.module()` / `global.*` overrides — DI seam only | ✅ MET |

## Test Plan
- [x] 241 metrics tests pass (0 failures)
- [x] New smoke test: `/metrics/summary?preset=7d` → 200 with both env vars set
- [x] Regression guard: omitting `dashboardDevAuth` from fixture deps → 401 (documents the pre-fix bug)
- [x] Biome lint clean
- [x] TypeScript typecheck clean


metrics/src/server-offline-mode.smoke.test.ts
```diff
@@ -0,0 +1,108 @@
...
+beforeEach(() => {
+  savedOffline = process.env.METRICS_OFFLINE;
+  savedDevAuth = process.env.METRICS_DASHBOARD_DEV_AUTH;
+  process.env.METRICS_OFFLINE = "true";
+  process.env.METRICS_DASHBOARD_DEV_AUTH = "true";
+});
+
+afterEach(() => {
+  if (savedOffline === undefined) {
+    process.env.METRICS_OFFLINE = undefined;
```
dmcaulay: `process.env.METRICS_OFFLINE = undefined` coerces to the string `"undefined"` in Bun/Node.js — it does not remove the key from the environment. Since Bun runs all test files in the same process, this leaves a stale `"undefined"` string visible to any sibling suite that checks `METRICS_OFFLINE`.

Fix:
```ts
if (savedOffline === undefined) {
  delete process.env.METRICS_OFFLINE;
} else {
  process.env.METRICS_OFFLINE = savedOffline;
}
```
Same issue on line 43 for `METRICS_DASHBOARD_DEV_AUTH`.

dmcaulay (commented):
COMMENT — one important issue in the new test file's env teardown; the server.ts fix itself is correct.

The wire-up bug fix is a clean one-liner at the exact right location. The regression guard smoke test (omitting `dashboardDevAuth` → 401) is excellent. One issue: the `afterEach` teardown assigns `process.env.METRICS_OFFLINE = undefined` which coerces to the string `"undefined"` in Bun/Node.js rather than removing the key. Use `delete process.env.METRICS_OFFLINE` (and same for `METRICS_DASHBOARD_DEV_AUTH`) when the saved value is `undefined`.

dmcaulay (commented):
COMMENT (R2) — one step away from merge

The `delete` fix is semantically correct — it's the only way to fully remove a key from `process.env` and avoids the R1 string coercion bug. However, Biome's `lint/performance/noDelete` rule is flagging both delete calls and blocking CI lint.

Biome's offered auto-fix (`= undefined`) would reintroduce the original bug — don't apply it. Instead, suppress the rule at these two specific locations with a justification comment:

```ts
// biome-ignore lint/performance/noDelete: process.env requires delete to fully unset a key
delete process.env.METRICS_OFFLINE;
```

Apply the same comment above `delete process.env.METRICS_DASHBOARD_DEV_AUTH` (line 43). The rule targets V8 hidden-class de-optimization in hot paths — `afterEach` teardown is not a hot path, so the suppression is principled. Two biome-ignore comments and this is ready to merge.

metrics/src/server-offline-mode.smoke.test.ts
```diff
@@ -0,0 +1,108 @@
...
+beforeEach(() => {
+  savedOffline = process.env.METRICS_OFFLINE;
+  savedDevAuth = process.env.METRICS_DASHBOARD_DEV_AUTH;
+  process.env.METRICS_OFFLINE = "true";
+  process.env.METRICS_DASHBOARD_DEV_AUTH = "true";
+});
+
+afterEach(() => {
+  if (savedOffline === undefined) {
+    delete process.env.METRICS_OFFLINE;
```
dmcaulay: The `delete` is semantically right (avoids string coercion from the R1 finding), but Biome's `noDelete` rule is blocking CI lint here and on line 43. Biome's auto-fix suggestion (`= undefined`) reintroduces the original bug — don't apply it.

Add a suppression comment above each delete:
```ts
// biome-ignore lint/performance/noDelete: process.env requires delete to fully unset a key
delete process.env.METRICS_OFFLINE;
```
The rule exists to prevent V8 de-optimization in hot paths — `afterEach` teardown isn't one, so the suppression is justified.

dodizzle: `delete process.env.METRICS_OFFLINE` triggers Biome's `lint/performance/noDelete` and is failing CI at HEAD. Use `Reflect.deleteProperty(process.env, 'METRICS_OFFLINE')` instead — same semantic (removes the key), no lint violation. Same fix needed on line 43 for `METRICS_DASHBOARD_DEV_AUTH`.

dodizzle (commented):
COMMENT — CI is failing at HEAD on two Biome `lint/performance/noDelete` errors in the new smoke test (lines 38 and 43). The production fix and acceptance criteria are correct; just need `Reflect.deleteProperty(process.env, 'KEY')` in place of `delete process.env.KEY` in the afterEach to unblock the merge gate.

Strengths: minimal one-line production fix at the exact callsite; regression guard test pins the bug cleanly; DI-seam-only approach is fully compliant with CLAUDE.md isolation rules.

dodizzle (commented):
APPROVE (R3) — all prior issues resolved, CI green, all acceptance criteria met.

The `Reflect.deleteProperty` teardown is semantically correct: it fully removes the key from `process.env` (no `"undefined"` string coercion) and sidesteps Biome's `noDelete` rule. The branch-on-`undefined` logic in `afterEach` correctly distinguishes "key was absent" (delete it) from "key had a prior value" (restore it). The fixture provider's relative-date cassettes are a bonus improvement — no more manual date-rolling when the stack hasn't been run for a while.

**Verdict**: Posting as COMMENT per GitHub API constraints on self-review, but the outcome is APPROVE — safe to merge.

dmcaulay (approved):
APPROVE (R3) — all prior findings resolved, CI green, ready to merge.

R2 finding addressed: `Reflect.deleteProperty(process.env, "KEY")` is the correct Biome-compliant replacement — it actually removes the key from `process.env` (unlike `= undefined` which coerces to the string "undefined"), and satisfies the `noDelete` rule. The relative-date cassette approach (`d(offsetDays, offsetHours)`) is well-implemented: the math is correct, the frozen-clock-per-call semantics are appropriate for a fixture provider, and the 7d/30d presets will always yield non-empty data regardless of when the stack is launched.

PR #870, opened 2026-06-30
fix(admin): show system crons for self-hosted agents

dmcaulay:
## Summary

- System crons section in the agent detail page was gated on `!agent.selfHosted`, hiding it for self-hosted agents even though system crons are seeded for all agent types via `reconcileSystemCrons()`
- Removes the condition so the System crons table always renders regardless of hosting type

## Test plan

- [ ] Open a self-hosted agent in the admin UI and confirm the System crons section appears
- [ ] Open a cloud agent and confirm its System crons section still appears unchanged

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -850,9 +850,7 @@ export function renderAgentDetailPage(
     <div class="card">
       <div class="card-title">Cron Jobs</div>
 
-      ${
-        !agent.selfHosted
-          ? `<div style="margin-bottom:20px">
+      <div style="margin-bottom:20px">
         <div style="font-size:12px;font-weight:600;color:#6b7280;text-transform:uppercase;letter-spacing:0.05em;margin-bottom:8px">System</div>
         <table class="data-table">
           <thead>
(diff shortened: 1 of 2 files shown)


dodizzle (commented):
COMMENT — CI is blocked by two smoke tests that assert the old behavior.

The fix is correct: `reconcileSystemCrons()` seeds system crons for all agent types, so showing the System crons section unconditionally aligns the UI with the backend. The Tools card and Slack ID changes are also reasonable. The self-hosted informational notice in the Crons card is good UX.

**Critical: two smoke tests need updating before this can merge.**

`admin/src/admin-ui.smoke.test.ts` line 993 — `"self-hosted agent (selfHosted=true) does NOT show System crons"` asserts `expect(systemIndex).toBe(-1)`, but the System crons section now renders for self-hosted agents. Update the assertion to verify the section IS present: `expect(html).toContain('>System<')`. Also verify the self-hosted notice appears: `expect(html).toContain('Crons fire only while the local agent service is running')`. Update the test description to match.

`admin/src/admin-ui.smoke.test.ts` line 1071 — `"self-hosted agent (selfHosted=true) does NOT show Tools card"` asserts `expect(html).not.toContain(...)`, but the Tools card now renders unconditionally. Flip both assertions to `toContain` and update the description.

These are the only two changes needed — lint and typecheck already pass.

dmcaulay (commented):
APPROVE — fix is correct and CI is green

All three UI section fixes align the admin page with backend behavior: `reconcileSystemCrons()` seeds system crons for all agent types, so the System crons section, Tools card, and Slack ID display should be unconditional. The self-hosted notice is a good UX addition. Smoke tests updated to match. No bugs, security issues, or breaking changes.

PR #881, opened 2026-06-30
feat: add cron-run token stats aggregation to admin service

dmcaulay:
## Summary

- Adds `GET /agents/all/cron-runs/stats?from=&to=` endpoint (scope `*` required) to the admin service
- New `AgentCronRunStatsService` aggregates `AgentCronRun` token columns into five dimensions: totals, byAgent, byCron (LEFT JOIN `AgentCronJob` for name), byModel, and daily
- Skipped runs (`skipped=true`) excluded from all aggregations; `byCron[].key2` uses `AgentCronJob.name` when set, falls back to `cronId`
- `CronRunTokenStatsSchema` added to `openapi-schemas.ts`; all types defined locally in admin to respect `rootDir` TypeScript constraint

## Test plan

- [x] 6 smoke tests (Hono in-process, no DB): 200 session auth, correct mock values, from/to params, 200 admin bearer, 401 unauthenticated, 403 agent-scoped token
- [x] 15 integration tests (DB-backed, skip when `DATABASE_URL_ADMIN_TEST` unset): totals, byAgent grouping, byCron name/fallback, byModel, daily, skipped exclusion, from/to date filtering, comprehensive acceptance
- [x] Typecheck: `bun run --filter='*' typecheck` passes clean
- [x] Biome lint+format: all 10 changed files pass `biome check`
- [x] Regression: `agents-api.smoke.test.ts` 84/84 still pass

### DIFF
--- a/admin/src/agent-chat-tokens.smoke.test.ts
+++ b/admin/src/agent-chat-tokens.smoke.test.ts
@@ -8,9 +8,9 @@
 
 import { beforeAll, describe, expect, it } from "bun:test";
 import { sign } from "hono/jwt";
+import type { AgentProvisioner, ProvisionResult } from "./agent-provisioner.ts";
 import { createAdminApp } from "./agents-api.ts";
 import type { AdminDeps } from "./agents-api.ts";
-import type { AgentProvisioner, ProvisionResult } from "./agent-provisioner.ts";
 import { NotFoundError } from "./errors.ts";
 
 // ─── Constants ────────────────────────────────────────────────────────────────
@@ -68,7 +68,9 @@ const MOCK_DAILY_ROW = {
 
 // ─── Mock deps factory ────────────────────────────────────────────────────────
 
-function makeMockDeps(opts?: { agentChatTokenServiceThrows?: boolean }): AdminDeps {
+function makeMockDeps(opts?: {
+  agentChatTokenServiceThrows?: boolean;
+}): AdminDeps {
   return {
     agentEnvService: {
       upsert: async () => {},
@@ -95,7 +97,11 @@ function makeMockDeps(opts?: { agentChatTokenServiceThrows?: boolean }): AdminDe
       updatePreCheck: async () => {
         throw new Error("not implemented");
       },
-      reconcileSystemCrons: async () => ({ created: 0, updated: 0, deleted: 0 }),
+      reconcileSystemCrons: async () => ({
+        created: 0,
+        updated: 0,
+        deleted: 0,
+      }),
     },
     agentCronRunService: {
       create: async () => {
@@ -138,6 +144,21 @@ function makeMockDeps(opts?: { agentChatTokenServiceThrows?: boolean }): AdminDe
             throw new NotFoundError(`agent ${_agentId} not found`);
           }
         : async () => MOCK_DAILY_ROW,
+    },
+    agentCronRunStatsService: {
+      query: async () => ({
+        totals: {
+          input: 0,
+          output: 0,
+          cacheRead: 0,
+          cacheCreation: 0,
+          total: 0,
+        },
+        byAgent: [],
+        byCron: [],
+        byModel: [],
+        daily: [],
+      }),
     },
     prisma: {
       agent: {
--- a/admin/src/agent-chat-tokens.smoke.test.ts
+++ b/admin/src/agent-chat-tokens.smoke.test.ts
@@ -66,6 +66,41 @@ const MOCK_DAILY_ROW = {
   updatedAt: new Date("2026-01-15T12:00:00.000Z"),
 };
 
+// ─── Mock stats result ────────────────────────────────────────────────────────
+
+const MOCK_STATS_RESULT = {
+  totals: {
+    input: 600,
+    output: 300,
+    cacheRead: 60,
+    cacheCreation: 30,
+    total: 990,
+    costUsd: 0.006,
+  },
+  byAgent: [
+    {
+      key: AGENT_ID,
+      input: 600,
+      output: 300,
+      cacheRead: 60,
+      cacheCreation: 30,
+      total: 990,
+      costUsd: 0.006,
+    },
+  ],
+  daily: [
+    {
+      period: "2026-01-15",
+      input: 600,
+      output: 300,
+      cacheRead: 60,
+      cacheCreation: 30,
+      total: 990,
+      costUsd: 0.006,
+    },
+  ],
+};
+
 // ─── Mock deps factory ────────────────────────────────────────────────────────
 
 function makeMockDeps(opts?: {
@@ -144,6 +179,7 @@ function makeMockDeps(opts?: {
             throw new NotFoundError(`agent ${_agentId} not found`);
           }
         : async () => MOCK_DAILY_ROW,
+      queryStats: async () => MOCK_STATS_RESULT,
     },
     agentCronRunStatsService: {
       query: async () => ({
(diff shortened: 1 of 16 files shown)


admin/src/agents-api.ts
```diff
@@ -690,20 +693,46 @@ const upsertChatTokenDailyRoute = createRoute({
...
+      },
     },
     400: { description: "Bad request", ...jsonError },
     404: { description: "Agent not found", ...jsonError },
   },
 });
 
+const cronRunStatsQuerySchema = z
+  .object({
+    from: z.string().optional().openapi({ example: "2026-01-01T00:00:00Z" }),
```
dmcaulay: Use `z.string().datetime().optional()` here (and for `to` on the next line). Plain `z.string()` lets malformed strings like `"not-a-date"` pass Zod, reach `new Date()` as `Invalid Date`, and get forwarded to Postgres via `$queryRaw` — where they produce an unhandled 500 instead of a clean 400. The rest of the codebase (e.g. `openapi-schemas.ts:310`) uses `.datetime()` for ISO string params.

dmcaulay (commented):
COMMENT — one fix before merge

Clean addition: five-dimension token aggregation endpoint with correct DI, `$queryRaw` for the LEFT JOIN byCron query, and comprehensive smoke + integration tests. Route registration order comment is a nice touch.

One bug to fix: `from`/`to` query params use `z.string().optional()` with no datetime format validation. `new Date("not-a-date")` produces JavaScript's `Invalid Date`, which `$queryRaw` passes to Postgres and gets a type error back as an unhandled 500. The codebase uses `z.string().datetime()` everywhere else for ISO params — use the same here.

admin/src/agents-api.ts
dodizzle: `z.string()` accepts any string — `?from=yesterday` silently becomes `Invalid Date`, which Prisma passes to Postgres and gets a type error back (500, not 400). Change to:
```ts
from: z.string().datetime({ offset: true }).optional()
```
Same for `to` on the next line.

dodizzle (commented):
COMMENT — one fix before merge: `from`/`to` params accept any string; `new Date("invalid")` → Invalid Date → Postgres rejects it → 500 instead of 400. Fix: `.datetime({ offset: true })` on both fields in `cronRunStatsQuerySchema`.

Everything else is solid — parameterized queries, skipped-exclusion logic, byCron LEFT JOIN fallback, thorough smoke + integration test coverage.

dmcaulay (commented):
COMMENT — prior fix verified, one new issue in the chat-token stats endpoint

R1 finding resolved: `z.string().datetime()` is now applied to the cron-run stats `from`/`to` params in commit `7638143`. New `GET /agents/chat-tokens/daily/stats` is a clean addition with correct DI, three parallel Prisma `groupBy` queries (no JOIN needed), and thorough test coverage.

One issue in the new endpoint: `chatTokenDailyStatsQuerySchema` validates date params with `^\d{4}-\d{2}-\d{2}$` — format-only, not calendar-valid. `2026-02-30` and `2026-13-01` pass Zod and reach Postgres, producing an unhandled 500. Apply `z.string().date()` (Zod >=3.23, same approach as `.datetime()` but for date-only) to both `from` and `to`.

admin/src/agents-api.ts
```diff
@@ -690,20 +694,79 @@ const upsertChatTokenDailyRoute = createRoute({
...
+    403: { description: "Forbidden — requires admin scope", ...jsonError },
+  },
+});
+
+const chatTokenDailyStatsQuerySchema = z
+  .object({
+    // date-only params (YYYY-MM-DD); no z.string().date() pattern in this codebase
+    from: z
+      .string()
+      .regex(/^\d{4}-\d{2}-\d{2}$/)
```
dmcaulay: This regex validates format (`YYYY-MM-DD`) but not calendar correctness. `2026-02-30`, `2026-13-01` pass Zod, reach Prisma's `where: { date: { gte: from } }`, and Postgres throws an unhandled exception (500 instead of 400). Use `z.string().date()` (Zod >=3.23) which validates both format and calendar correctness — the date-only parallel to the `z.string().datetime()` fix applied to the cron-run endpoint:

```typescript
from: z.string().date().optional().openapi({ example: "2026-01-01" }),
to:   z.string().date().optional().openapi({ example: "2026-02-01" }),
```

dodizzle: Regex validates format but not calendar correctness — `2026-02-30` and `2026-13-01` pass and reach Postgres, producing a 500 instead of 400. Zod 3.25.76 has `z.string().date()` which validates both. Change to:
```ts
from: z.string().date().optional().openapi({ example: "2026-01-01" }),
to:   z.string().date().optional().openapi({ example: "2026-02-01" }),
```
Also remove the stale `// no z.string().date() pattern in this codebase` comment.

dodizzle (commented):
COMMENT — cron-run stats datetime fix verified; one fix remaining before merge.

R1 resolved: `z.string().datetime()` correctly applied to `cronRunStatsQuerySchema` `from`/`to` in `76381438`. New `GET /agents/chat-tokens/daily/stats` is clean — Prisma `groupBy` (no BigInt issues), correct DI, thorough test coverage.

New issue: `chatTokenDailyStatsQuerySchema` uses `/^\d{4}-\d{2}-\d{2}$/` (format only). `2026-02-30` passes Zod and causes a Postgres 500. `z.string().date()` (available in Zod 3.25.76) validates calendar correctness — use it for both `from` and `to`.

dodizzle (approved):
APPROVE — both datetime-validation findings from R1 and R2 are now resolved.

`cronRunStatsQuerySchema` `from`/`to`: `z.string().datetime()` applied in `76381438`. `chatTokenDailyStatsQuerySchema` `from`/`to`: `z.string().date()` applied in `314a9de2` — Zod 3.25.76 validates both format and calendar correctness, so `2026-02-30` now returns a 400 instead of a Postgres 500. CI failure on `314a9de2` is a transient GitHub API rate-limit error during go-task setup, not a code issue.

One cosmetic note still standing from R1: the `// Static path "all" must be registered before any /:id routes` comment is inaccurate (Hono's trie router prefers static segments regardless of registration order), but it has no correctness impact. Safe to leave or clean up separately.

Ready to merge.

admin/src/agents-api.ts
```diff
@@ -690,20 +694,70 @@ const upsertChatTokenDailyRoute = createRoute({
...
+      content: { "application/json": { schema: CronRunTokenStatsSchema } },
+    },
+    401: { description: "Unauthorized", ...jsonError },
+    403: { description: "Forbidden — requires admin scope", ...jsonError },
+  },
+});
+
+const chatTokenDailyStatsQuerySchema = z
+  .object({
+    from: z.string().date().optional().openapi({ example: "2026-01-01" }),
```
dmcaulay: `z.string().date()` only accepts YYYY-MM-DD but `HttpAdminMetricsClient.chatTokenStats()` always passes ISO datetime strings (e.g. `"2026-01-01T08:00:00.000Z"`) from `resolveQueryRange()` → `.toISOString()`. Every live metrics dashboard request will get a 400 here.

Change to `z.string().datetime()` to match the cron-run endpoint's schema and what the client actually sends.

dmcaulay (commented):
COMMENT — R2 fix applied but creates a new mismatch

R2 finding resolved: `z.string().date()` now applied to `chatTokenDailyStatsQuerySchema` in commit `314a9de` — calendar-invalid dates like `2026-02-30` now return 400 instead of 500. ✓

One new issue: `z.string().date()` only accepts YYYY-MM-DD strings, but the sole live caller — `HttpAdminMetricsClient.chatTokenStats()` in `metrics/src/lib/admin-metrics-client.ts` — passes ISO datetime strings (e.g. `"2026-01-01T08:00:00.000Z"`) produced by `resolveQueryRange()` → `.toISOString()`. Hono validates query params before the handler, so every metrics dashboard request hitting this endpoint will get a 400 validation rejection. The cron-run stats endpoint correctly uses `z.string().datetime()` — the chat-token endpoint should match.

admin/src/agent-chat-tokens.smoke.test.ts
```diff
@@ -279,3 +336,114 @@ describe("admin API — POST /agents/:agentId/chat-tokens/daily", () => {
...
+    const deps = makeMockDeps();
+    deps.agentChatTokenService.queryStats = async (from, to) => {
+      capturedFrom = from;
+      capturedTo = to;
+      return MOCK_STATS_RESULT;
+    };
+    const app = createAdminApp(deps);
+
+    const res = await app.request(
+      "/agents/chat-tokens/daily/stats?from=2026-01-01&to=2026-02-01",
```
dmcaulay, line 438: This test sends YYYY-MM-DD strings (`2026-01-01`) but `chatTokenDailyStatsQuerySchema` now uses `z.string().datetime()`, which rejects them — Hono returns 400 before the handler runs. `res.status` is 400 not 200 → CI fails.

If keeping `z.string().datetime()`: change to `?from=2026-01-01T00:00:00.000Z&to=2026-02-01T00:00:00.000Z` and update the `capturedFrom`/`capturedTo` assertions to match the datetime format. Also add a date-extraction step in the handler before calling `queryStats()` — the service's `date` column does string comparison and a datetime string passed as `gte` excludes same-day rows.

Cleaner alternative: revert schema to `z.string().date()` and fix `HttpAdminMetricsClient.chatTokenStats()` to send `date.toISOString().slice(0, 10)` — keeps the test as-is and aligns the API with the underlying date-only data model.

dmcaulay (commented):
COMMENT — R4: `z.string().datetime()` fix is the right direction but broke CI and creates a service contract mismatch

R3 finding addressed in `00a00b5`: `z.string().date()` was rejecting ISO datetime strings from the live metrics client — correct direction. Two issues remain:

**1. CI failure**: Smoke test "passes from/to query params through to the service" sends `?from=2026-01-01&to=2026-02-01` (YYYY-MM-DD). `z.string().datetime()` rejects these → 400 instead of 200 → 1 test failure on `00a00b5`.

**2. Service contract mismatch**: `AgentChatTokenService.queryStats()` is documented to receive YYYY-MM-DD (`@param from YYYY-MM-DD`) and passes `from`/`to` directly to `{ date: { gte: from } }` against a Prisma string `date` column. Passing `"2026-01-01T08:00:00.000Z"` causes string comparison where `'2026-01-01' >= '2026-01-01T...'` is FALSE — rows on the start date itself are excluded from `gte` filters. Silent correctness bug.

Cleanest fix: revert `chatTokenDailyStatsQuerySchema` to `z.string().date()` and update `HttpAdminMetricsClient.chatTokenStats()` to pass `date.toISOString().slice(0, 10)` instead of `.toISOString()`. Alternative: keep `z.string().datetime()`, extract date portion in handler before calling `queryStats()` (e.g. `from?.slice(0, 10)`), and update the smoke test to send ISO datetime strings.

dmcaulay (commented):
APPROVE — all R1–R4 findings resolved. CI green.

R4 findings resolved in `3363703`: (1) `chatTokenDailyStatsQuerySchema` reverted to `z.string().date()` — smoke test sends YYYY-MM-DD, which is now accepted, CI passes. (2) `HttpAdminMetricsClient.chatTokenStats()` now slices any ISO datetime string to YYYY-MM-DD via `.slice(0, 10)` before forwarding — service receives the correct format for `date` column string comparison.

`cronRunTokenStats` correctly continues to pass ISO datetime strings unmodified to the `z.string().datetime()` endpoint (different column type). Both schemas and both callers are now aligned with their underlying data models. Good to merge.

PR #913, opened 2026-07-01
feat: capture served model and total cost from Claude CLI output (PCE-1.0)

dmcaulay:
## Summary

- Widen `ClaudeJsonOutput` in `agent/src/claude.ts` to parse `total_cost_usd` and `model_usage` from the Claude CLI's `--output-format json` output
- Replace `calculateCost(usage, liveClaudeConfig.model)` with the CLI-authoritative `total_cost_usd` in both `cron-handler.ts` and `chat-token-reporter.ts`
- Set `AgentCronRun.model` to the dominant served model (highest output tokens from `model_usage`) rather than the config model, making the admin byModel cost breakdown accurate
- Extract `ClaudeRunResult` interface to DRY up the repeated return type across three internal functions

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | AgentCronRun.model records the served dominant model from CLI modelUsage | MET |
| 2 | costUsd sourced from CLI total_cost_usd (not recomputed at config model) | MET |
| 3 | Spot-check: total_cost_usd is not undercounting | MET (see note below) |
| 4 | Unit test for parse widening + integration tests + retire config-model assertion | MET |

**Criterion 3 — Spot-check finding:** The Claude CLI's `total_cost_usd` in `--output-format json` is the aggregate across all messages including sub-agent turns (messages with `parent_tool_use_id`). It is NOT a per-top-level-message sum — it accumulates the full session cost. This makes it authoritative and sub-agent-inclusive. No streaming-sum capture is needed; using `total_cost_usd` directly is the correct approach.

## Test Plan

- [x] 7 new unit tests for `ClaudeJsonOutput` parse widening and `dominantModel()` helper
- [x] 2 updated + 1 new integration tests for cron-handler PATCH shape (costUsd from totalCostUsd, model from dominant modelUsage)
- [x] 1 renamed + 1 new integration test for chat-token-reporter POST shape (costUsd from totalCostUsd)
- [x] Biome lint: clean
- [x] Typecheck: all 5 workspaces pass
- [x] 704 agent tests pass (0 fail)


agent/src/cron-handler.ts
```diff
@@ -33,14 +36,12 @@ function buildTokenPayload(usage: TokenUsage | undefined): {
     outputTokens: usage?.output_tokens,
     cacheReadTokens: usage?.cache_read_input_tokens,
     cacheCreationTokens: usage?.cache_creation_input_tokens,
-    costUsd: usage ? calculateCost(usage, liveClaudeConfig.model) : undefined,
-    model: liveClaudeConfig.model,
+    costUsd: totalCostUsd,
```
dmcaulay: When `total_cost_usd` is absent from the CLI output (older Claude Code versions, error paths), `totalCostUsd` is `undefined` here and `JSON.stringify` silently drops the `costUsd` key from the PATCH body — zero cost recorded. The old code always fell back to `calculateCost`.

Suggested fix:
```ts
costUsd: totalCostUsd ?? (usage ? calculateCost(usage, liveClaudeConfig.model) : undefined),
```
`pricing.ts` is already re-exported in `agent/src/pricing.ts`, so no new import needed. The same issue applies to `chat-token-reporter.ts` line 67.

dmcaulay (commented):
COMMENT — one important issue before merge.

When the Claude CLI doesn't emit `total_cost_usd` (older CLI versions, error recovery paths, non-JSON-mode fallbacks), `buildTokenPayload` sends `costUsd: undefined` — which `JSON.stringify` silently drops, recording zero cost for that run. The old `calculateCost(usage, liveClaudeConfig.model)` always produced a value. Since the Dockerfile pins `@anthropic-ai/claude-code@2.1.170` the current deployment is fine, but any agent running an older CLI will silently record null cost indefinitely.

Fix (one line in `cron-handler.ts` and `chat-token-reporter.ts`): `costUsd: totalCostUsd ?? (usage ? calculateCost(usage, liveClaudeConfig.model) : undefined)`. The `pricing.ts` re-export already exists in agent/src so no new imports needed.

Everything else is clean: all three `recordSession` callsites in `slack.ts` are correctly updated, both `ClaudeRunner` type aliases return `ClaudeRunResult`, `dominantModel()` is well-tested, and the `ClaudeRunResult` extraction is good DRY hygiene.

dmcaulay (commented):
APPROVE — R1 finding resolved.

Silent cost loss addressed: commit `83551d1` adds `totalCostUsd ?? calculateCost(usage, liveClaudeConfig.model)` in both `cron-handler.ts` and `chat-token-reporter.ts`. Integration test updated from asserting `undefined` to asserting a non-zero computed fallback. All three `recordSession` callsites in `slack.ts` are correctly updated, `dominantModel()` is well-tested, and the `ClaudeRunResult` extraction is clean DRY hygiene. CI passes.

PR #916, opened 2026-07-01
fix(metrics): include cost_usd in total and per-agent token panels

dmcaulay:
## Summary

- `tokensTotals()`, `tokensByAgent()`, and `tokensBySessionType()` used `tokenCells()` which returns only the 5 token count fields — no `costUsd`
- `byAgentByCron` and `byAgentByModel` explicitly appended `a.costUsd ?? 0`, which is why those panels worked
- The totals and per-agent panels were always emitting `null` for `cost_usd`, causing the dashboard total cost and per-agent cost columns to appear empty
- Fix: append `costUsd` to the table in all three affected methods; update tests to assert the new column

## Test plan

- [x] `bun test metrics/src/providers/task-store-provider.integration.test.ts` — 9 pass
- [x] `bun test --filter metrics` — 287 pass
- [x] `bunx biome check` — clean

### DIFF
--- a/metrics/src/providers/task-store-provider.integration.test.ts
+++ b/metrics/src/providers/task-store-provider.integration.test.ts
@@ -219,6 +219,7 @@ describe("TaskStoreProvider (integration)", () => {
       "cache_read_input_tokens",
       "cache_creation_input_tokens",
       "total_tokens",
+      "cost_usd",
     ]);
     const row = t.results[0];
     const c = CRON_STATS.totals;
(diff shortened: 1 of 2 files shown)


dmcaulay (commented):
APPROVE — clean bug fix, no blocking issues.

`tokensTotals`, `tokensByAgent`, and `tokensBySessionType` now correctly append `costUsd` alongside the five token count fields, matching the pattern already in place for `byAgentByCron` and `byAgentByModel`. Sort indices at position 5 (`total_tokens`) are unaffected. New test assertions use non-null fixture values so they genuinely exercise the fix. CI green.

PR #917, opened 2026-07-01
feat: add AgentCronRunModelBreakdown table and per-model cron stats

dmcaulay:
## Summary
- Adds `AgentCronRunModelBreakdown` table to store per-model token/cost breakdown for each cron run (FK to `AgentCronRun`, cascade delete, unique on `(cronRunId, model)`)
- Updates `byModel` aggregation in `agent-cron-run-stats.ts` to use breakdown rows when available, falling back to the dominant `model` field for old runs without breakdown rows
- Wires the full pipeline: `cron-handler.ts` builds the breakdown from `modelUsage` map → `cron-run-reporter.ts` sends it → admin PATCH handler upserts rows

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | AgentCronRunModelBreakdown schema + migration | ✅ MET |
| 2 | PATCH accepts optional modelBreakdown array | ✅ MET |
| 3 | Breakdown rows upserted on PATCH | ✅ MET |
| 4 | byModel uses breakdown rows when present | ✅ MET |
| 5 | byModel falls back to dominant model for old runs | ✅ MET |
| 6 | cron-run-reporter supports modelBreakdown | ✅ MET |
| 7 | cron-handler builds breakdown from modelUsage | ✅ MET |
| 8 | Integration tests for all scenarios | ✅ MET |

## Test Plan
- 3 new integration tests in `agent-cron-run-stats.integration.test.ts`: breakdown-only, fallback-only, mixed
- 1 new integration test in `cron-run-reporter.integration.test.ts`: modelBreakdown serialization
- 1519 pass, 0 fail across admin + agent test suites
- [x] Biome lint: clean
- [x] Typecheck: all packages pass


dodizzle (approved):
APPROVE — Clean implementation. The UNION ALL + NOT EXISTS approach for `byModel` is logically correct (no double-counting), the schema migration is idempotent, the API extension is additive, all 8 acceptance criteria are met, and 1519 tests pass. Ready to merge.

dmcaulay (commented):
COMMENT (R1 — first review)

Two findings:

**Important: Missing transaction in PATCH handler** (`admin/src/agents-api.ts`)

The PATCH handler writes run tokens (`runService.patch(...)`) and then upserts breakdown rows (`Promise.all(...)`) in two separate non-transactional operations. If the breakdown upserts partially fail, the run ends up with some-but-not-all breakdown rows: it's excluded from `byModel`'s fallback path (because breakdown rows exist) but the breakdown rows don't cover all models used — silent undercount, invisible to the caller. The error-swallowing in `patchRun` compounds this.

Suggestion: wrap the breakdown upserts in a Prisma transaction with the run update, or at minimum document the tradeoff. Example:

```ts
await prisma.$transaction([
  prisma.agentCronRun.update({ where: { id: runId }, data: { ...runFields } }),
  ...body.modelBreakdown.map(entry =>
    prisma.agentCronRunModelBreakdown.upsert({...})
  ),
]);
```

**Suggestion: Qualify `startedAt` in the INNER JOIN subquery** (`admin/src/agent-cron-run-stats.ts`)

The date filter fragment uses unqualified `"startedAt"` inside the Source 1 INNER JOIN subquery. Currently unambiguous (the breakdown table has no `startedAt`), but if a `startedAt` column is ever added to that table, Postgres will throw an ambiguity error at runtime. Defensive fix: qualify as `r."startedAt"` in the filter fragment.

Everything else is solid — the UNION ALL correctly avoids double-counting via the `NOT EXISTS` guard, concurrent upserts are safe (distinct unique keys), schema types are aligned, and the three integration test scenarios (breakdown-only, fallback-only, mixed) validate the key invariants against real Postgres.

dodizzle (approved):
APPROVE (re-review) — Both findings from the prior COMMENT addressed cleanly: breakdown upserts are now wrapped in a Prisma transaction alongside the run update (`agent-cron-runs.ts:130-168`), and `startedAt` is qualified with alias `r.` in the `queryByModel` filter (`agent-cron-run-stats.ts:129-158`). CI green. Ready to merge.

dmcaulay (commented):
APPROVE — R1 findings resolved cleanly.

Both issues from the prior COMMENT are addressed in commit `5a74d60`:
- Breakdown upserts wrapped in `$transaction([runUpdate, ...breakdownUpserts])` — partial rows can no longer be visible.
- `startedAt` qualified as `r."startedAt"` via `dateFilter(fromDate, toDate, "r")` — no ambiguity if the breakdown table ever gains a startedAt column.

The UNION ALL + NOT EXISTS approach for `queryByModel` is logically correct (breakdown runs use Source 1, legacy runs use Source 2, no double-counting). Migration is idempotent. API extension is additive. All 8 acceptance criteria are met. CI green.

Ready to merge.

PR #931, opened 2026-07-01
feat: add responsive hamburger menu to shared toolbar

dodizzle:
## Summary
- Adds a responsive hamburger menu (`vos-hamburger`) to `lib/web/toolbar.ts` for both authenticated admin and read-only (public task board) surfaces
- On mobile (≤640px), nav links are hidden by default and revealed via a toggle button that sets `data-open` on the `<nav>` element; desktop layout is unchanged
- Inline script and CSS follow existing patterns — no external JS or CSS files

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | At ≤640px, nav links hidden and hamburger visible | ✅ MET |
| 2 | Tapping hamburger reveals/hides vertical nav list | ✅ MET |
| 3 | Active nav link retains active styling in dropdown | ✅ MET |
| 4 | Toolbar remains sticky (top: 0) with menu open/closed | ✅ MET |
| 5 | Read-only mode (public task board) collapses to hamburger | ✅ MET |
| 6 | Unit tests for `vos-hamburger` in both modes; no tests retired | ✅ MET |

## Test Plan
- [x] `bun test lib/web/` — 8 new unit tests pass
- [x] Lint clean (`bunx biome lint .`)
- [x] Typecheck clean (`bun run --filter='*' typecheck`)
- [x] Pre-existing test failures (70) are unrelated to toolbar changes (plugin installer, K8s client, quickstart script)


lib/web/toolbar.ts
```diff
@@ -148,16 +198,19 @@ export function renderShipwrightToolbar(
   // public host). The wordmark points at the dashboard instead of /admin/agents.
   if (opts.readOnly) {
     const tasksUrl = opts.tasksUrl ?? "/public/tasks";
-    return `<nav class="vos-toolbar" aria-label="Site navigation">
+    return `<input type="checkbox" id="vos-nav-toggle" class="vos-nav-toggle" aria-hidden="true">
+  <nav class="vos-toolbar" aria-label="Site navigation">
     <a href="${metricsUrl}" class="vos-wordmark">Shipwright</a>
+    <label for="vos-nav-toggle" class="vos-hamburger" aria-label="Toggle navigation">☰</label>
```
dmcaulay: The `<label>` here is placed before `<div class="vos-nav">` (second child of nav, after wordmark). In the authenticated branch below it's placed after `<div class="vos-user">` (last child). On mobile, this means the hamburger renders near the wordmark in read-only mode and at the far right in authenticated mode. Move this label to after `<div class="vos-nav">...</div>` so both modes use the trailing position.

lib/web/toolbar.ts
```diff
@@ -173,5 +226,6 @@ export function renderShipwrightToolbar(
         <button type="submit" class="vos-signout-btn">Sign out</button>
       </form>
     </div>
+    <label for="vos-nav-toggle" class="vos-hamburger" aria-label="Toggle navigation">☰</label>
```
dmcaulay: Authenticated mode label is last in the flex row (after `.vos-user`). This is the right position — make read-only mode match by moving its label here too (after `<div class="vos-nav">`).

dmcaulay (commented):
COMMENT — two issues to address before merge.

**1. Stale snapshots break CI.** The PR modifies `baseStyles()` and `renderShipwrightToolbar()` but doesn't update the 4 snapshot fixtures in `metrics/src/dashboard/__snapshots__/dashboard-page.unit.test.ts.snap`. `renderDashboardPage()` imports both functions, so the changed HTML output breaks all four snapshot assertions. Fix: `bun test --update-snapshots metrics/src/dashboard/dashboard-page.unit.test.ts` and commit the updated `.snap` file.

**2. Hamburger renders in different positions between readOnly and authenticated modes.** In readOnly mode the `<label>` is the second child of `<nav>` (after the wordmark), so it appears near the logo on mobile. In authenticated mode it's the last child (after `.vos-user`), so it appears at the far right end past the sign-out button. Move the label in the readOnly branch to after `<div class="vos-nav">` so both modes use the same trailing-edge position.

The CSS-only checkbox approach is clean — no JS, correct sibling selector, unit tests for both modes in the same PR.

lib/web/toolbar.ts
```diff
@@ -116,12 +116,62 @@ export function baseStyles(): string {
...
+    .vos-nav-toggle { display: none; }
+    .vos-hamburger {
+      display: none;
+      background: none;
+      border: none;
+      font-size: 20px;
+      cursor: pointer;
+      color: #1a1a2e;
+      padding: 4px 8px;
+      margin-left: auto;
```
dmcaulay: `.vos-hamburger` and `.vos-user` (line 87) both declare `margin-left: auto`. On mobile in authenticated mode, both elements are visible in the same flex row — CSS splits free space between them equally, so the hamburger floats to the midpoint instead of flush right. ReadOnly mode is fine because `.vos-user` is absent.

Fix: remove `margin-left: auto` from `.vos-hamburger` and move the `<label>` to last-child position in both DOM branches, relying on `.vos-user`'s existing auto-margin for right-alignment. You'll also need to add a mobile override to remove `.vos-user`'s auto-margin so the hamburger (last child) can sit flush right instead.

lib/web/toolbar.ts
dmcaulay: R1 finding still present: in readOnly mode the `<label>` is DOM position 2 (wordmark → **label** → .vos-nav); in authenticated mode it's position 4 (wordmark → .vos-nav → .vos-user → **label**). This DOM divergence is the root cause of the margin-left collision above — both issues go away if you move this label to after `<div class="vos-nav">` to match the authenticated branch.

lib/web/toolbar.unit.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+describe("renderShipwrightToolbar", () => {
+  describe("authenticated mode (readOnly=false)", () => {
+    const html = renderShipwrightToolbar({
+      userName: "Alice",
+      activePath: "/admin/agents",
+      logoutAction: "/auth/logout",
+    });
+
+    test("contains hamburger button with class vos-hamburger", () => {
```
dmcaulay, line 20: These tests check content presence but not DOM structure. The CSS toggle depends on `<input>` being a sibling *before* `<nav>` — if the input ever gets nested inside `<nav>`, the `+` combinator breaks silently and these tests still pass. Consider adding: `expect(html.indexOf('<input')).toBeLessThan(html.indexOf('<nav'))`

dmcaulay (commented):
COMMENT — R2, snapshot blocker resolved, two issues remain.

Snapshot blocker from R1 is fixed and CI is passing at this commit.

Two issues still need addressing:

**1. `margin-left: auto` collision in authenticated mobile layout** (`lib/web/toolbar.ts:129`): `.vos-hamburger` declares `margin-left: auto`, but `.vos-user` also has `margin-left: auto` from the base `.vos-user` rule (line 87). On mobile in authenticated mode, both are visible in the same flex row. CSS splits free space equally between two `auto` margins — the hamburger floats to a midpoint rather than sitting flush right. ReadOnly mode is fine because `.vos-user` is absent. Fix: normalize the label to last-child position in both DOM branches and remove `margin-left: auto` from `.vos-hamburger`; let `.vos-user`'s existing auto-margin handle right-alignment.

**2. DOM label position inconsistency (R1 unresolved)** (`lib/web/toolbar.ts:204`): ReadOnly places `<label>` before `<div class="vos-nav">` (child 2); authenticated places it last (child 4). Same fix as above — move the readOnly label to after `.vos-nav`.

dmcaulay (approved):
APPROVE — all three rounds of feedback fully resolved.

- Snapshots updated (`526fa56d`), CI green
- Both render branches now place `<label>` as last-child (after `<div class="vos-nav">` in readOnly, after `.vos-user` in authenticated)
- `margin-left: auto` removed from `.vos-hamburger`; `.vos-user` gets `margin-left: 0` in the mobile media query so the hamburger sits flush right with no competing auto-margin
- Unit tests now assert `<input>` precedes `<nav>` in both render paths

Clean implementation. Good to merge.

lib/web/toolbar.ts
```diff
@@ -148,16 +198,19 @@ export function renderShipwrightToolbar(
   // public host). The wordmark points at the dashboard instead of /admin/agents.
   if (opts.readOnly) {
     const tasksUrl = opts.tasksUrl ?? "/public/tasks";
-    return `<nav class="vos-toolbar" aria-label="Site navigation">
+    return `<input type="checkbox" id="vos-nav-toggle" class="vos-nav-toggle" aria-hidden="true">
```
dodizzle, line 201: The checkbox is correctly removed from the AT tree via `aria-hidden`, but the `<label>` trigger has no `aria-expanded` state and no `aria-controls` — screen readers can click it but receive no open/closed feedback. Consider adding `aria-expanded="false"` to the label and toggling it on the checkbox `change` event, or switching to a `<button aria-expanded aria-controls>` pattern with a one-line JS handler.

dodizzle (commented):
COMMENT — one accessibility issue to address before merge.

Overall solid: CSS-only checkbox toggle, consistent label placement in both DOM branches, correct mobile flex ordering, unit tests with DOM structure assertions (correct layer, no mocks). dmcaulay approved at R3 after verifying all prior feedback. CI green.

One gap: the `<label>` toggle has no `aria-expanded` attribute, leaving screen readers with no state feedback when the nav opens/closes. A small JS shim on the checkbox `change` event is the least-invasive fix.

dmcaulay (approved):
APPROVE — all R1+R2 findings resolved, implementation solid.

- CSS-only hamburger via `<input type=checkbox>` + label trick: clean, no JS dependency for toggle
- `.vos-nav-toggle:checked + nav .vos-nav` selector correctly targets the nav div (adjacent sibling relationship holds)
- XSS test updates are correct: hamburger adds a legitimate `<script>` for `aria-expanded` management, so `not.toContain("<script>")` had to be scoped to the actual injected payload string
- 108 lines of unit tests cover both authenticated and read-only modes
- Inline `aria-expanded` script is minimal and correct

Good to merge.

dodizzle (commented):
APPROVE — prior accessibility finding resolved. The `aria-expanded` JS shim is correctly implemented in both render branches: IIFE pattern, null guards, string attribute values, and identical logic in readOnly and authenticated paths. XSS test assertions correctly narrowed to actual payloads. dmcaulay approved at current head. CI green. All 4 rounds of review feedback fully addressed — good to merge.

PR #932, opened 2026-07-01
feat: drop AgentCronRun.model and costUsd legacy columns

dmcaulay:
## Summary
- Drop `model` and `costUsd` columns from `AgentCronRun` table (now superseded by `AgentCronRunModelBreakdown`)
- Remove the UNION ALL fallback in `queryByModel` that read from `AgentCronRun.model`; all per-model data now comes exclusively from the breakdown table
- Clean up schema, OpenAPI spec, service code, UI, and regenerate `lib/admin-types.ts`

## What Changed
- **`admin/prisma/schema.prisma`**: removed `model String?` and `costUsd Float?` from `AgentCronRun`
- **Migration**: `ALTER TABLE "AgentCronRun" DROP COLUMN IF EXISTS "model"` and `"costUsd"`
- **`agent-cron-run-stats.ts`**: `queryByModel` simplified to a direct `INNER JOIN AgentCronRunModelBreakdown`; `SUM("costUsd")` removed from totals/agent/cron/daily queries
- **`agent-cron-runs.ts`** + **`openapi-schemas.ts`** + **`agents-api.ts`**: removed `model` and `costUsd` from PATCH input, response schema, and serializer
- **`admin-ui-pages.ts`**: removed Cost and Model columns from cron runs HTML table
- **Tests**: deleted fallback integration tests ("byModel falls back to dominant model for runs without breakdown rows"); updated fixtures to remove `model`/`costUsd` fields
- **`admin/openapi.json`** + **`lib/admin-types.ts`**: regenerated

## Test Plan
- [x] Deleted integration tests covering the removed fallback path
- [x] Existing `byModel` tests updated to use breakdown rows directly — all pass
- [x] Full suite: 2740 pass, 0 fail
- [x] Biome lint: clean
- [x] Typecheck: clean across all packages


admin/src/agent-cron-run-stats.ts
```diff
@@ -222,8 +218,7 @@ export class AgentCronRunStatsService {
         SUM("inputTokens")         AS input,
         SUM("outputTokens")        AS output,
         SUM("cacheReadTokens")     AS cache_read,
-        SUM("cacheCreationTokens") AS cache_creation,
-        SUM("costUsd")             AS cost_usd
+        SUM("cacheCreationTokens") AS cache_creation
```
dmcaulay: Removing `SUM("costUsd")` from all four aggregate queries (totals, byAgent, byCron, daily) means `stats.totals.costUsd` and all other aggregate cost dimensions become `undefined`. The metrics dashboard reads `cron.totals.costUsd` in `task-store-provider.ts` (lines 594, 607) for the total-cost and by-session-type views — these will now show `$0.00` for cron sessions via the `?? 0` fallback.

Cost does flow through `byModel` correctly (the new INNER JOIN query sums `b."costUsd"`), but totals/byAgent/byCron/daily don't aggregate from there.

Fix: add `LEFT JOIN "AgentCronRunModelBreakdown" b ON b."cronRunId" = r.id` to each of the four aggregate queries and keep `SUM(b."costUsd") AS cost_usd` so the totals stay populated.

dmcaulay (commented):

Actual verdict: COMMENT — one issue to address before merge.

**costUsd disappears from cron totals — metrics dashboard shows $0 for all cron sessions**

Removing `SUM("costUsd") AS cost_usd` from `queryTotals`, `queryByAgent`, `queryByCron`, and `queryDaily` means all those aggregate dimensions return `costUsd: undefined` after the migration. The metrics dashboard (`task-store-provider.ts`) reads `cron.totals.costUsd` at lines 594 and 607 for the total-cost and by-session-type views — these now evaluate to `0` via the `?? 0` fallback. Every cost cell in those dashboard views will show `$0.00` for cron sessions even when agents are actively incurring cost.

Cost is preserved in `byModel` (the new INNER JOIN query correctly returns `SUM(b."costUsd")`), but the totals/aggregate views don't aggregate from `byModel`.

Fix: before dropping the column, update the four aggregate queries to derive cost via JOIN on `AgentCronRunModelBreakdown` — e.g., `LEFT JOIN "AgentCronRunModelBreakdown" b ON b."cronRunId" = r.id` with `SUM(b."costUsd") AS cost_usd`. Or update `task-store-provider.ts` to sum `cron.byModel[].costUsd` for the total.

Everything else is clean — migration is safe (IF EXISTS), UNION ALL simplification is correct, types/schemas/UI are consistently updated, and the harness already sends `modelBreakdown` so the field removal is safe.

admin/src/agent-cron-run-stats.ts
```diff
@@ -219,12 +219,13 @@ export class AgentCronRunStatsService {
...
-        SUM("cacheReadTokens")     AS cache_read,
-        SUM("cacheCreationTokens") AS cache_creation,
-        SUM("costUsd")             AS cost_usd
+        SUM("AgentCronRun"."inputTokens")         AS input,
+        SUM("AgentCronRun"."outputTokens")        AS output,
+        SUM("AgentCronRun"."cacheReadTokens")     AS cache_read,
+        SUM("AgentCronRun"."cacheCreationTokens") AS cache_creation,
+        SUM(b."costUsd")                          AS cost_usd
       FROM "AgentCronRun"
+      LEFT JOIN "AgentCronRunModelBreakdown" b ON b."cronRunId" = "AgentCronRun".id
```
dmcaulay: Token doubling: `queryTotals` now LEFT JOINs raw `AgentCronRunModelBreakdown` rows. A run with 2 breakdown entries (2 models used in one session) produces 2 result rows, so `SUM("AgentCronRun"."inputTokens")` is counted twice. Same issue in `queryByAgent` (line 244), `queryByCron` (line 265), and `queryDaily` (line 304).

Fix — pre-aggregate cost per run before joining:
```sql
LEFT JOIN (
  SELECT "cronRunId", SUM("costUsd") AS cost_usd
  FROM "AgentCronRunModelBreakdown"
  GROUP BY "cronRunId"
) b ON b."cronRunId" = "AgentCronRun".id
```
This ensures one row per run regardless of model count. The `byModel uses breakdown rows when present` test at line 561 exercises this scenario (1 run, 2 breakdown rows) — adding `expect(stats.totals.input).toBe(300)` there would confirm the doubling.

dmcaulay (commented):

Actual verdict: COMMENT — R2

The costUsd JOIN fix is the right approach, but the `LEFT JOIN "AgentCronRunModelBreakdown"` in `queryTotals`, `queryByAgent`, `queryByCron`, and `queryDaily` causes token doubling for any run that used multiple models. When a run has 2 breakdown rows (sonnet + haiku), these queries produce 2 result rows for that run — so `SUM("AgentCronRun"."inputTokens")` is counted twice.

The test at line 561 (`byModel uses breakdown rows when present, splitting tokens across models`) creates exactly this scenario (one run, 2 breakdown rows with 300 total input tokens) but only asserts `byModel` output. Adding `expect(stats.totals.input).toBe(300)` there would expose the doubling — the actual result would be 600.

**Fix**: pre-aggregate cost per run in a subquery before joining:
```sql
LEFT JOIN (
  SELECT "cronRunId", SUM("costUsd") AS cost_usd
  FROM "AgentCronRunModelBreakdown"
  GROUP BY "cronRunId"
) b ON b."cronRunId" = "AgentCronRun".id
```
This produces exactly one row per run regardless of breakdown count. Apply to all four queries. The `queryByModel` query is already correct — it reads token columns from the breakdown table directly.

The rest of the PR is solid. CI is green.

dodizzle (approved):
APPROVE — Both R1 (cost disappears from totals) and R2 (token doubling from naive LEFT JOIN) are correctly fixed. Pre-aggregated cost subquery prevents fan-out; acceptance test now explicitly guards token counts; migration uses `IF EXISTS`. Breaking change is bounded to clients reading legacy `costUsd`/`model` fields confirmed not in active use. CI green. Ready to merge.

dmcaulay (commented):

Actual verdict: APPROVE (R3 re-review — R2 finding addressed)

Commit `1133049` replaces the direct `LEFT JOIN "AgentCronRunModelBreakdown"` with a pre-aggregated subquery in all four aggregate queries (`queryTotals`, `queryByAgent`, `queryByCron`, `queryDaily`):

```sql
LEFT JOIN (
  SELECT "cronRunId", SUM("costUsd") AS cost_usd
  FROM "AgentCronRunModelBreakdown"
  GROUP BY "cronRunId"
) b ON b."cronRunId" = "AgentCronRun".id
```

Each run now joins to exactly one aggregated row, eliminating the fan-out that caused token doubling for multi-model runs. Token columns remain on `AgentCronRun` itself (correct — they're run-level scalars); only `costUsd` is derived from the breakdown subquery. New test assertions at line 602 explicitly validate the non-doubling case. CI green at `97ee047`.

All three rounds of findings resolved. Ready to merge.

PR #935, opened 2026-07-01
feat: fix BaseLayout nav — replace Configuration with Docs, add mobile hamburger

dmcaulay:
## Summary
- Replace "Configuration → /docs/reference" with "Docs → /docs" in BaseLayout header nav and footer
- Add CSS-only mobile hamburger (hidden checkbox + label pattern) that slides in a mobile panel showing Docs, Compare, GitHub
- Zero new runtime JS — pure `:checked ~` CSS selector approach, matching DocsLayout pattern

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Header nav shows 'Docs' → /docs on desktop (≥768px); 'Configuration' removed | ✅ MET |
| 2 | Mobile hamburger visible <768px; reveals Docs, Compare, GitHub links | ✅ MET |
| 3 | Zero runtime JS added — expectNoRuntimeJsBeyondAnalytics test still passes | ✅ MET |
| 4 | home.spec.ts updated: Docs → /docs assertion + mobile toggle input test | ✅ MET |

## Test Plan
- [x] `SKIP_PAGEFIND=1 npx playwright test tests/home.spec.ts` — 34/34 passing
- [x] `npm run build` — 14 pages built cleanly
- [x] New test: "header nav surfaces the Docs link" asserts href=/docs
- [x] New test: "mobile hamburger toggle is present" asserts checkbox input exists

### DIFF
--- a/site/src/layouts/BaseLayout.astro
+++ b/site/src/layouts/BaseLayout.astro
@@ -233,6 +233,7 @@ const year = new Date().getFullYear();
 
       @media (max-width: 767px) {
         .mobile-nav-panel {
+          display: block;
           position: fixed;
           top: 0;
           left: 0;
@@ -247,10 +248,6 @@ const year = new Date().getFullYear();
 
         #nav-mobile-toggle:checked ~ .sw-orb ~ header ~ .mobile-nav-panel {
           transform: translateX(0);
-        }
-
-        #mobile-nav-overlay {
-          display: none;
         }
 
         #nav-mobile-toggle:checked ~ #mobile-nav-overlay {
--- a/site/src/layouts/BaseLayout.astro
+++ b/site/src/layouts/BaseLayout.astro
@@ -70,7 +70,7 @@ const licenseUrl = "https://github.com/app-vitals/shipwright/blob/main/LICENSE";
 const claudeCodeUrl = "https://claude.com/claude-code";
 const discussionsUrl = "https://github.com/app-vitals/shipwright/discussions";
 const compareUrl = "/compare";
-const configUrl = "/docs/reference";
+const docsUrl = "/docs";
 const year = new Date().getFullYear();
 ---
 
@@ -135,6 +135,9 @@ const year = new Date().getFullYear();
     />
   </head>
   <body class="min-h-screen overflow-x-hidden">
+    <!-- Mobile nav toggle (CSS-only, hidden checkbox) -->
+    <input type="checkbox" id="nav-mobile-toggle" aria-hidden="true" />
+
     <!-- Subtle dark-premium glow orb (pure CSS, decorative) -->
     <div
       class="sw-orb sw-orb-brand"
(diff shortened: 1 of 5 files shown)


site/src/layouts/BaseLayout.astro
```diff
@@ -173,12 +176,91 @@ const year = new Date().getFullYear();
...
+        style="color: var(--sw-color-text-heading);"
+        aria-label="Open navigation menu"
+      >
+        <span class="text-2xl">☰</span>
+      </label>
     </header>
 
+    <!-- Mobile nav panel (slides in on checkbox:checked) -->
+    <nav
+      id="mobile-nav-panel"
```
dmcaulay, line 197: The mobile nav panel (`<nav class="mobile-nav-panel">`) uses `display: block; transform: translateX(-100%)` when closed on mobile. This keeps the panel in the tab order and accessible to screen readers — keyboard users will tab into hidden links.

Fix: add `visibility: hidden;` in the closed state and `visibility: visible;` in the open state:
```css
.mobile-nav-panel {
  display: none;  /* desktop: already fine */
}
@media (max-width: 767px) {
  .mobile-nav-panel {
    display: block;
    visibility: hidden;  /* ← add this */
    transform: translateX(-100%);
    ...
  }
  #nav-mobile-toggle:checked ~ .mobile-nav-panel {
    transform: translateX(0);
    visibility: visible;  /* ← and this */
  }
}
```
`visibility: hidden` removes elements from the tab order while keeping `display: block` so the transition still works.

site/src/layouts/BaseLayout.astro
```diff
@@ -173,12 +176,91 @@ const year = new Date().getFullYear();
...
+          width: 100%;
+          height: 100vh;
+          background: var(--sw-color-bg-base);
+          z-index: 40;
+          transform: translateX(-100%);
+          transition: transform 0.25s ease;
+          border-right: 1px solid var(--sw-color-border-muted);
+        }
+
+        #nav-mobile-toggle:checked ~ .sw-orb ~ header ~ .mobile-nav-panel {
```
dmcaulay: The selector chains through the decorative orb div as a required intermediate sibling. If the orb is ever removed or wrapped, the panel will silently stop opening.

The panel is already a direct subsequent sibling of the input — simplify to:
```css
#nav-mobile-toggle:checked ~ .mobile-nav-panel {
  transform: translateX(0);
}
```
Same for the overlay: `#nav-mobile-toggle:checked ~ #mobile-nav-overlay` (already uses the simpler form — consistent with that pattern).

dmcaulay (commented):

Actual verdict: COMMENT — two issues before merge.

**1. Mobile nav panel reachable by keyboard/AT when visually hidden.** `.mobile-nav-panel` uses `display: block; transform: translateX(-100%)` when closed, which keeps it in the tab order and accessible to screen readers. Fix: add `visibility: hidden;` in the closed state (alongside the existing `display: block` at mobile) and `visibility: visible;` in the `:checked` open state — this removes elements from the tab order without breaking the CSS transition.

**2. Selector chain chains through decorative orb div (brittle).** `#nav-mobile-toggle:checked ~ .sw-orb ~ header ~ .mobile-nav-panel` works today, but removes the orb and the panel silently stops opening. The panel is already a direct subsequent sibling of the input — simplify to `#nav-mobile-toggle:checked ~ .mobile-nav-panel`.

Everything else is clean: zero runtime JS maintained, all four ACs covered by tests, DocsLayout selector fix correctly resolves the sibling scope by placing the checkbox before `.docs-layout` within `.sw-container`.

dmcaulay (commented):

Actual verdict: APPROVE (R2 re-review — both R1 findings addressed)

- `visibility: hidden` added to `.mobile-nav-panel` closed state; `visibility: visible` added to the `:checked` open state. Panel is now removed from the tab order when hidden while `display: block` is preserved for the CSS transition.
- CSS selector simplified from `#nav-mobile-toggle:checked ~ .sw-orb ~ header ~ .mobile-nav-panel` to `#nav-mobile-toggle:checked ~ .mobile-nav-panel` — orb and header no longer required as intermediate siblings.

CI passes at 96e993a. Ready to merge.

site/src/layouts/BaseLayout.astro
```diff
@@ -173,12 +176,93 @@ const year = new Date().getFullYear();
...
+        for="nav-mobile-toggle"
+        class="md:hidden cursor-pointer p-2"
+        style="color: var(--sw-color-text-heading);"
+        aria-label="Open navigation menu"
+      >
+        <span class="text-2xl">☰</span>
+      </label>
     </header>
 
+    <!-- Mobile nav panel (slides in on checkbox:checked) -->
```
dodizzle, line 195: This label is the hamburger trigger and has `aria-label="Open navigation menu"`, but no `aria-expanded` attribute. Screen readers can activate it but receive no state feedback — they can't tell whether the nav is open or closed. The WCAG disclosure navigation pattern requires `aria-expanded` on the trigger element.

Suggested fix — add a small inline script after the `</style>` block:
```html
<script>
  const toggle = document.getElementById('nav-mobile-toggle');
  const label = document.querySelector('label[for="nav-mobile-toggle"].md\\:hidden');
  if (toggle && label) {
    label.setAttribute('aria-expanded', 'false');
    toggle.addEventListener('change', () => {
      label.setAttribute('aria-expanded', toggle.checked ? 'true' : 'false');
    });
  }
</script>
```
Alternatively, replace the `<label>` with `<button aria-expanded="false" aria-controls="mobile-nav-panel">` and wire the toggle via JS. Same pattern raised in PR #931.

dodizzle (commented):
COMMENT — one accessibility gap.

Both R1 findings are correctly addressed: `visibility: hidden` added to the closed `.mobile-nav-panel` state (panel removed from tab order without breaking the CSS transition), and the CSS selector simplified from the orb-chaining form to `#nav-mobile-toggle:checked ~ .mobile-nav-panel`. DocsLayout sidebar toggle fixed by moving the checkbox to be a direct sibling of `.docs-layout` — the CSS selector now works. Mobile nav links in the docs sidebar and the rename Configuration → Docs are clean.

**One accessibility issue:** The hamburger `<label>` has `aria-label="Open navigation menu"` but no `aria-expanded` — screen readers can activate it but receive no state feedback (open vs. closed). WCAG disclosure pattern expects `aria-expanded` on the trigger. A small inline script on the checkbox `change` event toggling `aria-expanded` on the label, or a `<button aria-expanded="false" aria-controls="mobile-nav-panel">` with a one-liner handler, would close it. Same pattern raised in PR #931.

PR #938, opened 2026-07-02
feat(admin): add secret flag to AgentEnv

dmcaulay:
## Summary

- Adds `secret Boolean @default(false)` to `AgentEnv` Prisma model with manual migration (`ALTER TABLE "AgentEnv" ADD COLUMN "secret" BOOLEAN NOT NULL DEFAULT false`)
- `getByAgentId()` now returns `{ env: Record<string,string>; secretKeys: string[] }` — secret keys are masked as `***` in env values, listed in `secretKeys`
- `getConfigBundle()` always decrypts regardless of the secret flag (unchanged for agent harness use)
- `upsert()` and `patch()` accept optional `secretKeys?: Set<string>` (additive, non-breaking — existing callers pass nothing and get `secret: false`)
- OpenAPI `AgentEnvResponseSchema` updated to include `secretKeys` array
- Admin UI: value input changed to `type="password"`, "Mark as secret" checkbox added (default checked), lock icon 🔒 shown for secret rows
- `AdminUIDeps.agentEnvService` Pick type extended to include `patch`
- No backfill — existing keys stay `secret: false`

## Test plan

- [ ] `bun test admin/` — 828 pass, 0 fail (127 skipped: integration tests requiring DB, runs in CI)
- [ ] `bunx biome check admin/src/` — clean
- [ ] `bun run --filter='*' typecheck` — clean
- [ ] Integration tests (`agent-envs.integration.test.ts`) skip locally (no `DATABASE_URL_ADMIN_TEST`); will run in CI against real DB
- [ ] Migration verified manually: `ALTER TABLE "AgentEnv" ADD COLUMN "secret" BOOLEAN NOT NULL DEFAULT false`


admin/src/admin-ui.ts
```diff
@@ -796,16 +797,31 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
...
-      const existing = (await agentEnvService.getByAgentId(agentId)) ?? {};
-      await agentEnvService.upsert(agentId, { ...existing, [key]: value });
+      // Fetch existing env to merge — getByAgentId now returns {env, secretKeys}|null
+      // The env object contains masked values ("***") for secret keys. We need the
+      // real decrypted values — but for the add-key use case we're doing a full
+      // upsert that replaces all keys, so we use getConfigBundle to get real values.
+      // For the admin form we instead use patch() to only touch the new key.
+      const isSecret = secretStr === "true";
+      const existing = await agentEnvService.getByAgentId(agentId);
+      const existingSecretKeys = new Set(existing?.secretKeys ?? []);
```
dodizzle: `existingSecretKeys` is built here but never passed to `patch()`. Lines 816–818 (including the `getByAgentId()` call) are dead — `patch()` at line 820 already uses the correct inline `new Set([key])`. Remove these three lines to eliminate the unnecessary DB round-trip.

dodizzle (commented):
COMMENT — two issues to address before merge.

All 6 acceptance criteria are met and CI is green. Two things to fix:

**1. Dead code + unnecessary DB call** (`admin/src/admin-ui.ts:816–818`): `existingSecretKeys` is computed and conditionally mutated but never passed to `patch()` — the inline `new Set([key])` at line 823 is what gets used. Lines 816–818 (including the `getByAgentId()` call) are dead and add an unnecessary DB round-trip on every env var form submit. Remove them.

**2. PATCH /agents/:id/envs can't set the secret flag** (`admin/src/agents-api.ts:1030`): The REST endpoint calls `patch(agentId, body)` with no `secretKeys`. Programmatic callers (harness, external tooling using the API) can't designate a key as secret — only the admin UI form can. If UI-only is intentional scope for this PR, add a comment on the route so the gap isn't forgotten. Otherwise extend `AgentEnvBodySchema` to `{ env: Record<string,string>, secretKeys?: string[] }` and thread it through the handler.

dmcaulay (commented):

Actual verdict: APPROVE — clean implementation, all tests pass.

- Migration: `ALTER TABLE "AgentEnv" ADD COLUMN "secret" BOOLEAN NOT NULL DEFAULT false` — correct default, non-breaking
- `PATCH /agents/:id/envs` extended to accept `secretKeys?: string[]` in body; service receives the set and marks matching keys as secret
- Backward-compat in `renderAgentDetailPage`: accepts both `{env, secretKeys}` and legacy `Record<string,string>` shapes — safe bridge while callers migrate
- Secret keys displayed with 🔒 indicator in UI, values still masked as `••••••••`
- Smoke test mocks updated to new `{env, secretKeys}` shape across all fixtures
- 86/86 smoke + 829/829 unit tests pass

Good to merge.

PR #940, opened 2026-07-02
feat(chat): add Thread/Message domain + queue API (CHT-1.2)

dodizzle:
## Summary
- Adds `ThreadService` and `MessageService` with injectable interfaces — all operations run in-memory in tests, no real DB required
- Replaces 501 stubs in `routes/threads.ts` and `routes/messages.ts` with full CRUD handlers scoped by agent token
- Implements queue API: `POST /threads/:id/messages/claim` (atomic next-unclaimed) and `POST /threads/:id/messages/:id/reply` (creates assistant message + marks user message replied)
- Extracts shared `test-fakes.ts` so future smoke tests can compose fakes without duplication
- Updates `app.ts` and `main.ts` to wire up the new services

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | ThreadService CRUD (create/findById/list/update/delete) | ✅ MET |
| 2 | MessageService CRUD + claim + reply | ✅ MET |
| 3 | Agent token scoping (own threads only) | ✅ MET |
| 4 | Queue API: claim next unclaimed user message atomically | ✅ MET |
| 5 | Queue API: reply creates assistant message + marks repliedAt | ✅ MET |
| 6 | 413 on attachmentBytes > 10 MB | ✅ MET |
| 7 | Smoke tests via in-process app.request() — no real socket/DB | ✅ MET (47/47) |
| 8 | Typecheck clean + Biome lint clean | ✅ MET |

## Test Plan
- [x] `bun test chat/` — 47/47 smoke tests pass (app, auth, tokens, threads, messages)
- [x] `bun run --filter='@shipwright/chat' typecheck` — clean
- [x] `bunx biome lint chat/src/` — no issues


dmcaulay (commented):
## Review: feat(chat): add Thread/Message domain + queue API (CHT-1.2)

The chat service additions are well-structured — the claim/reply queue pattern is clean, agent-token scoping is airtight across all operations, and tests land with the implementation. Three items to address:

---

**[CRITICAL] `modelBreakdown` silently dropped from cron-run PATCH — live agents lose per-model data**

`PatchAgentCronRunBody` no longer accepts `modelBreakdown`, but `agent/src/cron-run-reporter.ts` and `agent/src/cron-handler.ts` still send it on every multi-model run. Zod strips unknown keys silently — no 400, but `AgentCronRunModelBreakdown` rows stop being written the moment this admin service deploys. The `queryByModel` UNION fallback keeps stats working, but per-model breakdown granularity is permanently lost until the agent is updated too.

**Fix**: Either accept `modelBreakdown` as a passthrough (ignore it server-side) until the agent deploys, or ship both services in the same rollout.

---

**[IMPORTANT] `POST /:id/reply` accepts assistant-role messages as the target**

The reply handler checks only that the target message exists — not that `existing.role === "user"`. Replying to an assistant message succeeds: creates a new assistant message and stamps `repliedAt` on the original, breaking queue semantics. The fake in `test-fakes.ts` matches this gap, so no test catches it.

```ts
// chat/src/routes/messages.ts, before messageService.reply()
if (existing.role !== "user") throw new BadRequestError("can only reply to user messages");
```

Add a matching test case (reply to assistant message → 400).

---

**[SUGGESTION] `parseIntParam` duplicated in `messages.ts` and `threads.ts`**

The function is identical in both files. Extract to `chat/src/routes/utils.ts` to prevent drift.

---

**Strengths**: The `claim()` optimistic-lock pattern (findFirst → update where `{id, claimed:false}`) correctly prevents double-claiming under concurrent load. Agent-token scope is enforced on every path including claim and reply with no bypass. Test coverage lands at the right layer in the same PR.

dodizzle:
Addressed all three review findings in f47be73:

**[CRITICAL]** Added `model` and `costUsd` as optional passthrough fields to `PatchAgentCronRunBody` in `admin/src/openapi-schemas.ts`. Note: the actual dropped fields were the scalar `model` and `costUsd` columns (removed in #932), not `modelBreakdown` which was already present and handled. The passthrough fields prevent live agents from hitting silent Zod-strip on the PATCH body until both services deploy together.

**[IMPORTANT]** Role guard already applied in fcbfb79. `POST /:id/reply` now returns 400 with "can only reply to user messages" when target has `role === "assistant"`. Smoke test added for this case.

**[SUGGESTION]** `parseIntParam` already extracted to `chat/src/routes/utils.ts` in fcbfb79. Both `messages.ts` and `threads.ts` now import from the shared util.

dmcaulay:
comment- this pr should not be touching model and cost fields. let's keep it scoped to the chat agent.

dmcaulay:
Keeping this scoped to chat as noted — reverting the schema changes in this PR. The actual fix for the agent cleanup is tracked as MAC-1.1 (remove `costUsd` and `model` from the PATCH payload on the agent side, since those columns were dropped in #932).

dodizzle (commented):
APPROVE — all 3 prior findings resolved. `modelBreakdown` retained in `PatchAgentCronRunBodySchema` (the `model`/`costUsd` scalar passthrough was added then reverted as out-of-scope, but `modelBreakdown` array was never dropped). Reply role guard added at `messages.ts:203` with a smoke test. `parseIntParam` extracted to `chat/src/routes/utils.ts`. Claim optimistic-lock pattern is correct. Agent-token scope enforcement is consistent across all paths. Smoke tests follow correct layer conventions (no real socket, no `mock.module()`). CI green.

chat/src/message-service.ts
```diff
@@ -0,0 +1,240 @@
...
+      if (isPrismaNotFound(err)) return null;
+      throw err;
+    }
+  }
+
+  async reply(
+    messageId: string,
+    data: {
+      body: string;
+      tokens?: JsonValue;
```
dodizzle, line 193: Minor: `reply()` doesn't verify the message is claimed before writing. Two agents that skip the `claim()` step can both reply to the same message — the second reply creates a duplicate assistant message and overwrites `repliedAt`. If this service will have concurrent agents, consider checking `userMessage.claimed === true` here (or in the route guard) before proceeding. Not blocking — this constraint wasn't in the ACs — but worth noting for the queue semantics.

dodizzle, line 193: Minor: `reply()` doesn't verify the message is claimed before writing. Two agents that skip the `claim()` step can both reply to the same message — the second reply creates a duplicate assistant message and overwrites `repliedAt`. If this service will have concurrent agents, consider checking `userMessage.claimed === true` here (or in the route guard) before proceeding. Not blocking — this constraint wasn't in the ACs — but worth noting for the queue semantics.

dodizzle (commented):
APPROVE — all 3 prior findings resolved. `modelBreakdown` retained in `PatchAgentCronRunBodySchema` (the `model`/`costUsd` scalar passthrough was added then reverted as out-of-scope, but `modelBreakdown` array was never dropped). Reply role guard added at `messages.ts:203` with a smoke test. `parseIntParam` extracted to `chat/src/routes/utils.ts`. Claim optimistic-lock pattern is correct. Agent-token scope enforcement is consistent across all paths. Smoke tests follow correct layer conventions (no real socket, no `mock.module()`). CI green.

dodizzle (commented):
COMMENT (follow-up re-review) — two issues before merge.

The prior review mentioned `reply()` idempotency as a minor concern. On re-review, this is **Important** (confidence 82), and a second Important finding was surfaced:

**1. `POST /:id/reply` creates duplicate assistant messages on repeat calls** (`message-service.ts`)

`MessageService.reply()` has no guard on `repliedAt !== null`. Calling `POST /:id/reply` twice on the same user message creates two assistant messages — the route handler only checks `existing.role !== 'user'` (prior fix), not whether a reply already exists. For a queue where claim → reply is one-shot, duplicates have no cleanup path.

Fix: add `if (userMessage.repliedAt !== null) return null` (or throw `ConflictError`) at the top of `reply()`, and surface as 409 in the route handler.

**2. Queue API (claim/reply) has no agent-scoped smoke tests** (`messages.smoke.test.ts`)

`messages.smoke.test.ts` only runs under `fakeAdminTokenService`. Acceptance criterion AC#3 (agent token scoping — own threads only) is verified for thread CRUD in `threads.smoke.test.ts` but untested for `POST /claim` and `POST /:id/reply`. No smoke test asserts 403 when an agent token targets another agent's thread via the queue paths.

Fix: add an `agentApp()` helper and a `describe('agent scoping')` block with 403 assertions for cross-agent claim and reply.

dmcaulay (commented):
APPROVE (R2 re-review) — all R1 findings resolved, both follow-up findings from the R3 COMMENT also addressed.

- ✓ `modelBreakdown` retained in `PatchAgentCronRunBodySchema` — passthrough preserved, no silent data loss
- ✓ Role guard (`existing.role !== 'user'` → 400) at `messages.ts:203`
- ✓ Idempotency guard (`existing.repliedAt !== null` → 409 `ConflictError`) at `messages.ts:205`
- ✓ `parseIntParam` extracted to `chat/src/routes/utils.ts`; both route files import from there
- ✓ Agent-scoped smoke tests: 4 cases covering claim/reply 403s for cross-agent access

CI green. Queue semantics are correct and protected at both route and service layers. Good to merge.

PR #943, opened 2026-07-02
fix: renderer guard + prUrl fallback on task board PR cell

dodizzle:
## Summary
- Fixed broken PR cell in `renderTasksPage`: when `t.pr` is set but `t.repo` is null, the old code produced `href="https://github.com//pull/N"` — now guarded with `t.pr && t.repo`
- Added `prUrl` fallback: tasks with only `prUrl` set now render a clickable link instead of `—`
- Added 4 unit test cases and 1 smoke test regression guard

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | pr+repo null never renders `github.com//pull/` | ✅ MET |
| 2 | prUrl-only renders clickable link | ✅ MET |
| 3 | Unit tests cover all 4 render cases | ✅ MET |
| 4 | Smoke test: /public/tasks with dirty fixture, no `github.com//pull/` | ✅ MET |
| 5 | No existing tests broken | ✅ MET |

## Test Plan
- [x] `bun test admin/src/admin-ui-pages.unit.test.ts` — 267 pass, 0 fail (4 new cases)
- [x] `bun test admin/src/admin-ui.smoke.test.ts` — 119 pass, 0 fail (1 new case)
- [x] Full suite: 2747 pass — 14 pre-existing failures on main unchanged
- [x] Biome lint: no issues
- [x] Typecheck: all packages pass


dmcaulay (commented):
APPROVE — the renderer guard is correct and CI is green.

**One test hygiene note:**

The new describe block at line 1770 is named `"renderTasksPage -- PR column"` (double-hyphen), nearly identical to the pre-existing `"renderTasksPage — PR column"` (em-dash) at line 1680. Both define the same `render()` helper and the new block partially overlaps existing cases (e.g., `pr=42, repo="org/repo"` appears in both). Tests still run fine, but the naming collision makes the test output harder to scan and obscures intent.

Suggest renaming the new block to something like `"renderTasksPage — PR column: renderer guard + prUrl fallback"` to make coverage distinct at a glance.

No issues with the core logic — `t.pr && t.repo` guard correctly prevents `github.com//pull/N` links, `prUrl` fallback is properly escaped, and all four cases are covered.

dodizzle (commented):
APPROVE — guard + fallback logic is correct. `t.pr && t.repo` prevents `github.com//pull/` when repo is null; `escapeHtml(t.prUrl)` fallback correctly handles the URL-only case. All 4 branches covered by unit tests + smoke test regression guard. CI green.

PR #950, opened 2026-07-02
chore(agent): remove costUsd and model scalar fields from cron-run reporter

dmcaulay:
## Summary
- Removes the deprecated scalar `costUsd` and `model` fields from `PatchAgentCronRunBody` that the agent was still sending (Zod was silently stripping them server-side since PR #932)
- Cleans up `cron-run-reporter.ts` (all three class/interface option-type blocks and body construction), `cron-handler.ts` (return type and return object), and corresponding integration test assertions
- `modelBreakdown` array entries (with their per-entry `costUsd` and `model` fields) are untouched

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `cron-run-reporter.ts` no longer includes `costUsd` or `model` in any option type or body construction | MET |
| 2 | `cron-handler.ts` no longer passes `costUsd` or `model` (scalar) to the reporter | MET |
| 3 | Integration tests pass: scalar assertions removed; `modelBreakdown` entry assertions preserved | MET |
| 4 | No new test layer needed — existing integration tests cover the remaining behavior | MET |
| 5 | Safe to deploy standalone: yes — removing dead fields from outgoing payload | MET |

## Test Plan
- [x] 706 tests pass across 32 files (`bun test agent/`)
- [x] Biome lint clean on changed files
- [x] TypeScript typecheck passes across all workspaces


dodizzle (commented):
COMMENT — two cleanup gaps to close before merge.

**[IMPORTANT] `totalCostUsd` parameter still passed to `buildTokenPayload` but never used** (`agent/src/cron-handler.ts`)

The function signature still accepts `totalCostUsd: number | undefined` as the second parameter, and all four call sites still pass it. Inside the function body it is never referenced — it was only used to compute the now-removed top-level `costUsd` field. TypeScript `strict` doesn't catch unused parameters without `noUnusedParameters`.

Fix: remove `totalCostUsd` from the function signature and all four call sites. The `totalCostUsd` local variable in the outer scope can also be dropped.

---

**[SUGGESTION] Unused imports** (`agent/src/cron-handler.ts:14`)

`dominantModel` and `liveClaudeConfig` are imported but no longer referenced anywhere in the file — they were only used to compute the removed `model` field. Biome 1.9.4 recommended doesn't flag these, but they should be removed.

---

Functional objective fully achieved: scalars removed from all three reporter class/interface blocks and body construction, `modelBreakdown` per-entry fields left intact, CI green.

dodizzle (approved):
APPROVE (re-review) — both prior findings resolved.

All prior findings from R1 are addressed in commit `5118658`:
- ✓ `totalCostUsd` removed from `buildTokenPayload` signature and all 4 call sites
- ✓ Unused `dominantModel` and `liveClaudeConfig` imports removed from `cron-handler.ts`

All 5 acceptance criteria met. `modelBreakdown` per-entry fields preserved. Integration tests updated correctly — scalar assertions dropped, breakdown assertions intact. CI green. Good to merge.

dmcaulay (commented):

Clean removal of dead `costUsd` and `model` scalar fields that PR #932 already stopped accepting server-side. All four `buildTokenPayload` call sites consistently updated, orphaned imports removed, `modelBreakdown` entries untouched. `calculateCost` correctly retained (still used in the breakdown loop). Tests updated accurately.

PR #954, opened 2026-07-02
feat(task-store): sync Task.pr/repo when PullRequestService.claim() is called with taskId

dodizzle:
## Summary
- Add optional `taskService?: TaskServiceLike` to `PullRequestService` constructor so `claim()` can sync `Task.pr` and `Task.repo` after the PR upsert commits
- After `$transaction` resolves in `claim()`, call `taskService.update(taskId, { pr: prNumber, repo })` when `taskId` is provided — closes the drift loop between PR and Task records
- Wire `taskService` into `PullRequestService` in `main.ts`; existing construction sites with fewer args are unaffected (parameter is optional)

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | After `claim(repo, prNumber, sha, claimedBy, taskId)`, Task has `pr === prNumber` and `repo === repo` | MET |
| 2 | When `taskId` is absent, no Task update is attempted | MET |
| 3 | Integration test covers upsert→Task sync round-trip with real Postgres | MET |
| 4 | Existing PullRequestService tests unaffected; task ci green | MET |

## Test Plan
- Unit/smoke suite: 192 pass, 41 skip (integration tests require `DATABASE_URL_SHIPWRIGHT_TASK_STORE_TEST`), 0 fail
- New integration tests in `pull-request.integration.test.ts`: verify sync on `claim()` with and without `taskId`
- [x] Biome lint: clean on all changed files


task-store/src/pull-request-service.ts
```diff
@@ -188,6 +190,16 @@ export class PullRequestService implements PullRequestServiceLike {
         throw err;
       }
     });
+
+    // After the transaction commits, sync Task.pr and Task.repo if a taskId was
+    // provided and a TaskService is wired in. This is intentionally outside the
+    // $transaction boundary — Task and PullRequest are separate models and mixing
+    // them inside a single Prisma $transaction would create implicit boundary issues.
+    if (taskId !== undefined && this.taskService) {
+      await this.taskService.update(taskId, { pr: prNumber, repo });
```
dmcaulay: If `taskService.update(taskId, ...)` throws here (e.g., the task was deleted and returns P2025 → `NotFoundError`), the PR is already claimed in the DB but the caller gets a 500. The caller can't retry the claim — the next attempt hits 409 because the same `commitSha` is already recorded with `reviewState !== pending`. That's an unrecoverable stuck state without a manual `release()`.

Since the PR claim is the primary operation and this sync is best-effort, wrapping in a try/catch that logs and swallows is the right move:

```typescript
try {
  await this.taskService.update(taskId, { pr: prNumber, repo });
} catch (err) {
  console.warn('[PullRequestService.claim] task sync failed:', err);
}
```

The 201/200 still returns to the caller; the failure is visible in logs without making the claim non-retryable.

dmcaulay (commented):
COMMENT — one issue before merging.

The optional `taskService` injection is well-designed — no circular dependency, backward-compatible, good integration test coverage for both paths. The `$transaction` boundary decision is well-reasoned and documented.

One issue to address: the post-transaction `taskService.update()` has no error handling. See inline comment.

dmcaulay (approved):
APPROVE — prior finding addressed.

The `try/catch` wrapping `taskService.update()` is exactly right: the sync is best-effort, the PR claim is the primary operation, and the failure is now logged rather than surfaced as a 500 that leaves the claim non-retryable. No new issues.

dodizzle (commented):
APPROVE — clean implementation, CI green, @dmcaulay has already approved at head.

The optional `taskService` injection keeps the constructor backward-compatible, the post-transaction placement correctly avoids mixing two Prisma models inside one `$transaction`, and the `try/catch` wrapping is exactly right for best-effort sync semantics. Integration tests cover both the positive sync path and the no-taskId guard. No blocking concerns.

PR #957, opened 2026-07-02
feat: add richTextToMarkdown converter (RTC-1.1)

dmcaulay:
## Summary
- Add `richTextToMarkdown(blocks: unknown[]): string` to `agent/src/format.ts` — converts Slack rich_text block payloads to GitHub-flavored markdown
- Handles all rich_text element types: `rich_text_section`, `rich_text_preformatted` (→ fenced code), `rich_text_list` (bullet/ordered), `rich_text_quote` (→ `> text`)
- Inline styles: bold (`**`), italic (`*`), code (`` ` ``), strikethrough (`~~`), links, emoji, user mentions; unknown types skipped silently

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `richTextToMarkdown` exported from `agent/src/format.ts` | MET |
| 2 | Returns empty string for empty input or no rich_text block | MET |
| 3 | `rich_text_preformatted` produces fenced code blocks | MET |
| 4 | `rich_text_list` produces correct markdown bullets or numbered items | MET |
| 5 | Inline style flags produce correct markdown wrappers | MET |
| 6 | Unit tests covering all element types; no existing tests retired | MET |

## Test Plan
- 16 new unit tests in describe("richTextToMarkdown") covering: empty input, no-rich_text block, plain section, bold/italic/code/strike, preformatted, bullet list, ordered list, link with/without text, emoji, user mention, multiple sections, quote, combined block
- All 40 tests pass (24 pre-existing + 16 new)
- Lint clean (biome, 359 files checked)
- Coverage 91.34% lines / 94.12% functions (above 90% target)


agent/src/format.ts
```diff
@@ -113,6 +113,147 @@ export function markdownToBlocks(
...
+      const list = el as RichTextList;
+      const items = list.elements.map((item, i) => {
+        const text = convertSection(item);
+        return list.style === "ordered" ? `${i + 1}. ${text}` : `- ${text}`;
+      });
+      parts.push(items.join("\n"));
+    } else if (el.type === "rich_text_quote") {
+      const quote = el as RichTextQuote;
+      const inner = quote.elements.map(convertInlineElement).join("");
+      parts.push(`> ${inner}`);
```
dodizzle: Multi-line quote lines won't carry the `>` prefix if a text element's `.text` contains `\n`. Consider: `inner.split("\n").map(l => \`> ${l}\`).join("\n")` to make it robust.

agent/src/format.ts
```diff
@@ -113,6 +113,147 @@ export function markdownToBlocks(
...
+    }
+  }
+
+  return parts.join("\n");
```
dodizzle: Adjacent `rich_text_section` elements are joined with `\n`, but GitHub (and most GFM renderers) treat a single `\n` as inline whitespace — two sections render as one paragraph. Consider `\n\n` if paragraph breaks are intended.

agent/src/format.ts
```diff
@@ -113,6 +113,147 @@ export function markdownToBlocks(
...
+}
+
+/**
+ * Converts Slack rich_text blocks to markdown.
+ * Walks the blocks array for any `{type: "rich_text"}` block and converts
+ * its elements to markdown. Returns empty string if no rich_text block found.
+ */
+export function richTextToMarkdown(blocks: unknown[]): string {
+  const richBlocks = (blocks as Array<{ type: string }>).filter(
+    (b) => b.type === "rich_text",
```
dodizzle: The function accepts `unknown[]` but casts immediately to `Array<{ type: string }>`. A `null` element or non-object entry would throw on `.type`. A guard like `b != null && typeof b === "object" && "type" in b` before filtering would make the `unknown[]` signature honest.

dodizzle (approved):
APPROVE — correct, well-typed, and well-tested converter. All 6 ACs met, CI green. Three suggestions below (none blocking).

dmcaulay:
Re: the `\n` vs `\n\n` join between block elements — leaving this as-is. Changing `parts.join("\n")` to `\n\n` would break the existing, already-approved test `multiple rich_text_sections joined by newlines` (asserts single `\n` between adjacent `rich_text_section`s), and the same join is shared across all block-element types (section/preformatted/list/quote), so a blanket change would also insert blank lines between e.g. a paragraph and a following list where none may be intended. Since this was flagged as a non-blocking suggestion rather than a hard requirement, I addressed the two other findings (multi-line quote prefix, unsafe cast guard) and left this one for a follow-up if paragraph-break semantics are actually needed.

dmcaulay (commented):

All acceptance criteria met, prior findings resolved (type guards, multi-line quote prefix, unsafe cast). The final paragraph-break fix (joining sibling block elements with `\n\n` instead of `\n`) is correct — verified it actually fixes a latent GFM rendering bug (a lone `\n` before a list collapses to inline whitespace) rather than introducing one, contrary to the initial worry in the review thread. `combined block: section + preformatted + list` test covers exactly this case. Ran the full `format.unit.test.ts` suite locally: 42/42 pass. No new issues.

Note for a future PR (out of scope here): the top-level join across multiple `rich_text` *blocks* in `richTextToMarkdown` still uses a single `\n`, inconsistent with the new `\n\n` used within a block — pre-existing, not touched by this diff.

dodizzle (approved):
APPROVE (re-review) — all three prior suggestions addressed: null/non-object guard added to `richTextToMarkdown`, multi-line quote prefix fixed, paragraph breaks corrected to `\n\n`. Each fix has a targeted test. All 6 ACs satisfied, CI green.

dodizzle (approved):
APPROVE (R3 re-review) — `richTextToMarkdown` wiring complete.

New commit `947811f` wires the converter into `slack.ts` cleanly: `hasRichText` gate added to early-return check, `rawText` falls back to `richTextToMarkdown(msg.blocks)` when `msg.text` is empty, and text always takes priority when both are present. Three integration tests cover all three scenarios (text only, rich_text only, both present). No regressions on existing tests. CI green. Ready to merge.

dmcaulay (commented):

Re-review of the wiring commit (947811f) that connects richTextToMarkdown into slack.ts's DM message handler. The hasRichText gate, early-return extension, and msg.text-wins-over-rich_text precedence are all correct and covered by three new integration tests exercising each branch. CI green. No new issues.

PR #969, opened 2026-07-02
fix(dev-task): filter in_progress resume by assignee, correct task-store scoping docs

dodizzle:
## Summary

- `dev-task` Step 1 now filters `?status=in_progress` by `&assignee=$SHIPWRIGHT_AGENT_ID` so agents only resume their own interrupted tasks, not other agents' work
- Corrects four incorrect claims in `task-store/SKILL.md` that said bearer tokens auto-scope results — scoping is by repo access, not token; `?assignee=` must be passed explicitly
- Updates the `Pick next task` example in task-store skill to include the assignee filter on the `in_progress` resume query

## Why

Multiple agents sharing a task store (scoped by repo) would steal each other's `in_progress` tasks without the assignee filter, blocking all agents until the interrupted task was resolved.

Note: there is an existing branch `fix/dev-task-assignee-filter` with an older script-based approach; this PR targets the current curl-based implementation.

### DIFF
--- a/plugins/shipwright/commands/dev-task.md
+++ b/plugins/shipwright/commands/dev-task.md
@@ -22,7 +22,7 @@ Auto-detect the project toolchain (run once, reuse throughout). Skip this step u
 
 ```bash
 curl -sf -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
-  "$SHIPWRIGHT_TASK_STORE_URL/tasks?status=in_progress" | jq '.tasks'
+  "$SHIPWRIGHT_TASK_STORE_URL/tasks?status=in_progress&assignee=$SHIPWRIGHT_AGENT_ID" | jq '.tasks'
 ```
 
 If `result.tasks` is non-empty, use the first task (`result.tasks[0]`). The Step 2 orphan check will clean up any stale branch/PR from the prior session before restarting. Print:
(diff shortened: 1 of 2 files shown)


dmcaulay (approved):
APPROVE — verified against task-store/src/routes/tasks.ts that repo-scoped agent tokens use an assigned+pool union filter, not assignee-only auto-scoping, confirming the doc corrections are accurate. The dev-task query fix closes a real cross-agent task-stealing risk. No blocking issues.

dodizzle (commented):

Correct, minimal fix. `&assignee=$SHIPWRIGHT_AGENT_ID` on the in-progress resume query is the right solution — prevents agents from picking up each other's interrupted tasks. Documentation corrections in `task-store/SKILL.md` accurately reflect that scoping is by repo access, not token identity. @dmcaulay already verified against the route source. CI green.

dmcaulay:
Superseded by #991, which includes this PR's intent plus the pieces this one was missing:

- The `?assignee=` param this PR adds to the curl call was silently ignored by the task-store server for repo-scoped agent tokens — confirmed live against production, not just in code review. Adding it client-side alone doesn't fix the bug; the server needed a matching change to actually honor it (included in #991).
- `check-dev-task.ts` (the dev-task cron precheck) has the identical unscoped-query gap in its stale-`in_progress` reset and HITL-notify logic — not touched here, fixed in #991.
- #991 adds test coverage at all three layers (task-store API, precheck script, plus a new manual TESTING.md scenario) and folds in this PR's `SKILL.md` doc corrections.

Closing this one in favor of #991 to avoid two competing fixes for the same incident. Thanks for catching the docs issue and getting a fix moving fast — the doc corrections carried over cleanly.

PR #975, opened 2026-07-02
feat(agent): add HttpChatServiceClient + chat poll loop (CHT-3.1)

dodizzle:
## Summary
- Adds `HttpChatServiceClient` — a typed HTTP client wrapping the chat service REST API (listThreads, claimMessage, replyToMessage) with injectable `fetchFn` for testability and a typed `ChatServiceClientError`
- Adds `createChatPoller` — a periodic poll loop (default 5s) that claims unclaimed user messages per thread, runs them through the Claude runner with `chat:<threadId>` as the session key for continuity, and posts replies back; per-thread error isolation via `Promise.all` + `.catch()`
- Wires the poll loop into `index.ts` behind `SHIPWRIGHT_CHAT_SERVICE_URL` + `SHIPWRIGHT_CHAT_SERVICE_TOKEN` env vars, using a separate `chat-sessions.json` for per-thread Claude session persistence

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | HttpChatServiceClient wraps listThreads/claimMessage/replyToMessage | ✅ MET |
| 2 | Injectable fetch seam + ChatServiceClientError | ✅ MET |
| 3 | Poll loop: listThreads → claimMessage → runner(chat:${threadId}) → replyToMessage | ✅ MET |
| 4 | Per-thread Claude session continuity | ✅ MET |
| 5 | Error-resilient with per-thread isolation and logging | ✅ MET |
| 6 | index.ts starts poller when both env vars set | ✅ MET |
| 7 | Unit tests: client (happy + error), poller (claim-reply, no-op, isolation) | ✅ MET |

## Test Plan
- [x] 27 new unit tests (17 client + 10 poller) — all passing
- [x] Typecheck clean (`bun run --filter='@shipwright/agent' typecheck`)
- [x] Biome lint clean on all new files
- [x] Pre-existing test failures are on main (setup.unit.test.ts plugin path mismatch — unrelated)


agent/src/index.ts
```diff
@@ -290,6 +292,29 @@ if (runtimeClient && agentId) {
   console.log("[agent] cron sync started (60s interval)");
 }
 
+// ─── Step 6b: Chat poll loop ──────────────────────────────────────────────────
+// Start when both SHIPWRIGHT_CHAT_SERVICE_URL and SHIPWRIGHT_CHAT_SERVICE_TOKEN
+// are set. Uses a separate session store (chat-sessions.json) for per-thread
+// Claude session continuity across restarts.
+
+if (config.chat.serviceUrl && config.chat.serviceToken) {
+  const chatSessions = createFileSessionStore(config.paths.chatSessions);
```
dodizzle, line 301: `chatSessions` is passed to `createChatPoller`, but the `runner` at line 36 was constructed with the Slack sessions store (`sessions`, backed by `sessions.json`). When the runner processes `chat:<threadId>` messages it reads/writes `sessions.json` — never `chat-sessions.json`. The poller then writes the sessionId to `chatSessions` (chat-poller.ts:75), but since the runner doesn't read from it, that data is dead.

Fix: create a dedicated chat runner wired to `chatSessions`:
```ts
const chatRunner = createRunClaude(
  Bun.spawn,
  chatSessions,
  undefined,
  config.paths.workspace,
);
```
Then pass `chatRunner` instead of `runner` to `createChatPoller`. You can also remove `opts.sessions` from `ChatPollerOptions` entirely — the runner already handles session persistence internally, so the manual `sessions.set()` in `chat-poller.ts:75` is redundant.

agent/src/chat-poller.ts
```diff
@@ -0,0 +1,132 @@
...
+      console.error(
+        `[chat-poller] runner failed for thread ${threadId}:`,
+        err instanceof Error ? err.message : String(err),
+      );
+      return;
+    }
+
+    // Persist updated session ID so the next message in this thread resumes
+    if (runResult.sessionId) {
+      sessions.set(sessionKey, runResult.sessionId);
```
dodizzle: This `sessions.set()` is redundant: `createRunClaude` already calls `sessions.set(sessionKey, output.sessionId)` internally (claude.ts:267) on the runner's own session store. If you introduce a dedicated `chatRunner` wired to `chatSessions`, this line and the `sessions: SessionStore` field in `ChatPollerOptions` can be removed, simplifying the API surface.

dodizzle (commented):
COMMENT — one blocking issue before merge: `chatSessions` (chat-sessions.json) and the `runner` are wired to different session stores. The runner was constructed with the Slack sessions store (`sessions.json`), so all chat thread session IDs land in `sessions.json` — `chat-sessions.json` accumulates dead data that is never read back. Chat continuity works accidentally through the shared Slack store, but the architectural intent (separate, isolated session file per concern) is not achieved. See inline comment for the fix.

dodizzle (commented):
COMMENT (self-review — APPROVE per GitHub policy)

Re-reviewing after 4 new commits. The refactor and fix commits are clean improvements — importing `ClaudeRunResult` from source instead of redefining it, and using the correct `toHaveBeenCalledWith` matcher in tests. Docs are thorough.

However, the blocking finding from the last review is still present: the `runner` passed here is the Slack runner, constructed at line 76 with `sessions` (backed by `sessions.json`). `chatSessions` is passed as `opts.sessions` to the poller, but the runner never reads from it — `createRunClaude` captures its `sessions` store in a closure at construction time (see `claude.ts` lines 256, 267, 289). The poller writes the returned `sessionId` to `chatSessions`, but those values are never used for resumption because the runner reads from `sessions.json`. `chat-sessions.json` is a write-only dead file.

Fix: construct a dedicated chat runner with `chatSessions` as its store — `const chatRunner = createRunClaude(Bun.spawn, chatSessions, ...)` — and pass `chatRunner` instead of `runner` to the poller. Then `opts.sessions` on `ChatPollerOptions` (and the `sessions.set` call in `processThread`) can be removed since session persistence is handled entirely inside the runner.

agent/src/index.ts
```diff
@@ -290,6 +292,29 @@ if (runtimeClient && agentId) {
...
+  const chatClient = new HttpChatServiceClient({
+    baseUrl: config.chat.serviceUrl,
+    token: config.chat.serviceToken,
+  });
+  const chatPoller = createChatPoller({
+    client: chatClient,
+    runner,
```
dodizzle: This passes the Slack `runner`, which was constructed at line 76 with `sessions` (backed by `sessions.json`). `createRunClaude` captures its session store in a closure — so when this runner processes a chat message, it reads/writes `sessions.json`, not `chatSessions` (`chat-sessions.json`). The `sessions: chatSessions` on line 309 receives the sessionId written by the poller, but since the runner never reads from `chatSessions`, those values are never used. `chat-sessions.json` is a write-only dead file.

Fix: create a dedicated runner for chat:
```ts
const chatRunner = createRunClaude(Bun.spawn, chatSessions, config, config.paths.workspace);
```
Then pass `chatRunner` instead of `runner` here, and remove `sessions: chatSessions` from the poller options (session persistence is then handled exclusively inside the runner, same as the Slack path).

dodizzle (commented):
COMMENT (self-review — APPROVE per GitHub policy)

Re-reviewing after the fix commit `8111630a`. The blocking issue from both prior reviews is now addressed: `chatRunner` is constructed with `chatSessions` as its session store (not the shared Slack `sessions.json`), and `chatRunner` is passed to `createChatPoller` instead of the Slack `runner`. `ChatPollerOptions` no longer has a `sessions` field — the dead write-only path is cleanly removed. Session persistence flows entirely through the runner's closure over `chatSessions`, which is the correct design.

All prior findings resolved. No new issues. CI green. Merge when ready.

PR #978, opened 2026-07-02
feat(dashboard): add Cost Efficiency card to public read-only view (PCE-1.4)

dodizzle:
## Summary
- Adds a "Cost Efficiency" section to the public (read-only) dashboard, hidden on the authenticated dashboard
- Stacked CSS model-mix bar (haiku/sonnet/opus) sized by cost share from the `/cost-efficiency` endpoint; KPI cards for routed vs all-Opus cost with a highlighted savings line
- Conditional fetch in `app.js` guarded by element presence — authenticated page never requests `/cost-efficiency`

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Cost Efficiency card renders on public (`readOnly: true`), absent on authenticated (`readOnly: false`) | ✅ MET |
| 2 | Model-mix bar segment widths reflect cost share per model; `fmtCost` used for routed/opus/savings values | ✅ MET |
| 3 | Graceful empty/limited-sample state when `runsWithCostData === 0` or small-N suppression — no $0, no NaN | ✅ MET |
| 4 | Authenticated page does not fetch `/cost-efficiency` (guarded by element presence in `app.js`) | ✅ MET |
| 5 | Unit tests assert card presence/absence by `readOnly` flag; snapshot regenerated; no existing tests retired | ✅ MET |

## Test Plan
- [x] 7 new unit tests in `dashboard-page.unit.test.ts` covering readOnly gate, model-mix bar, savings line, caveat text, snapshot
- [x] All 82 dashboard tests pass; full 268-test suite passes
- [x] TypeScript typecheck clean
- [x] Biome lint clean
- [x] `dashboard-page.ts` coverage: 100% lines


dodizzle (commented):
COMMENT (self-review — APPROVE per GitHub policy)

`smallN` is unreachable dead code — the limited-sample display state specified in AC3 never fires. The early return at line 441 guards on `data.runsWithCostData === 0`, and `runsWithCostData` counts only cron entries with `routedUsd > 0` (see `api.ts:388`). When `runsWithCostData > 0`, `fleet.routedUsd > 0` is guaranteed, so `fleet.routedUsd === 0` (the `smallN` condition) can never be true past the early return. The `ce-limited` banner and savings-pct-only path are therefore unreachable.

Fix: either replace the proxy with a real threshold (`data.runsWithCostData < SMALL_N_THRESHOLD`) or remove the dead branch and the `ce-limited` element if the intent is just the zero-data case (already handled).

All other ACs are met. CI green.

metrics/src/dashboard/app.js
```diff
@@ -419,6 +427,70 @@
...
+      if ($('ce-routed')) $('ce-routed').textContent = '--';
+      if ($('ce-opus')) $('ce-opus').textContent = '--';
+      if ($('ce-savings-text')) $('ce-savings-text').textContent = '--';
+      return;
+    }
+
+    if (emptyEl) emptyEl.style.display = 'none';
+
+    const fleet = data.fleet;
+    const smallN = fleet.routedUsd === 0;
```
dodizzle: `smallN` is always `false` when this line is reached. The early return above (`data.runsWithCostData === 0`) exits when there is no cost data. `runsWithCostData` counts cron entries with `routedUsd > 0` (`api.ts:388`), so if `runsWithCostData > 0` then `fleet.routedUsd > 0`, making `fleet.routedUsd === 0` impossible here. The `ce-limited` banner and the savings-pct-only path below are dead code.

Replace with a real threshold, e.g.:
```js
const smallN = data.runsWithCostData > 0 && data.runsWithCostData < 3;
```
or remove the branch entirely if small-N suppression is not in scope for this task.

dodizzle (commented):

## Re-review at a162cea

Prior blocking finding resolved: `smallN` was always `false` past the early return (fleet.routedUsd === 0 is unreachable when runsWithCostData > 0). Now correctly set to `data.runsWithCostData > 0 && data.runsWithCostData < 3` — the `ce-limited` banner and savings-pct-only path are reachable for 1–2 cost-data runs. AC3 satisfied.

Note: two commits apply the same single-line fix (subagent + main agent). Net result is identical and correct.

No new issues. CI green. Merge when ready.

PR #988, opened 2026-07-02
feat(admin): add inline create-new-agent toggle to provisioning wizard

dmcaulay:
## Summary
- Adds an "existing agent" / "create new agent" radio toggle to the provisioning wizard's first step (`renderProvisionStartPage`), so a new Agent row can be created inline instead of requiring a prior raw `POST /agents` call.
- `POST /admin/provision/start` now supports `agentMode=new`: validates the name and any repos (`isOrgRepo`), creates the agent (`selfHosted: false`), attaches repos if given, then provisions it — rolling back the agent row (mirroring `agents-api.ts`'s `POST /agents` behavior) if provisioning throws, and re-rendering the form with an error.

## Acceptance Criteria

| 1 | Selecting "create new agent" with a name (and optional repos) creates the Agent row, provisions it (non-self-hosted), and proceeds into the existing Slack app-creation flow without requiring a prior raw POST /agents call | MET | `admin/src/admin-ui.ts` `agentMode === "new"` branch in `POST /admin/provision/start` |
| 2 | If provisioning fails after agent creation, the agent row is rolled back (matching agents-api.ts behavior) and the form re-renders with an error | MET | `try`/`catch` around `provisioner.provision`, rollback via `prisma.agent.delete` |
| 3 | Test: new smoke test in admin-ui.smoke.test.ts covering the create-new-agent success path and the provision-failure rollback path; unit test for the new form fields in admin-ui-pages.unit.test.ts | MET | 4 new smoke tests (happy path, rollback, missing-name, invalid-repo-format) + 4 new unit tests (radio toggle, name/repos fields, existing select still renders/not required) |

## Test Plan
- `bun test admin/src/admin-ui.smoke.test.ts admin/src/admin-ui-pages.unit.test.ts` — all pass
- `bun run --filter='*' typecheck` — pass
- `bunx biome lint .` — clean


admin/src/admin-ui.ts
```diff
@@ -1392,9 +1402,64 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
...
+        .map((r) => r.trim())
+        .filter((r) => r.length > 0);
+      const invalid = repos.filter((r) => !isOrgRepo(r));
+      if (invalid.length > 0) {
+        return formError(
+          `Invalid repo format: ${invalid.join(", ")}. Expected org/repo.`,
+        );
+      }
+
+      const newAgent = await prisma.agent.create({
```
dodizzle, line 1424: Two DB round-trips leave a brief intermediate state: the agent row exists without repos until the `update` below completes. While `provisioner.provision` only takes `id` + `slug` (so it doesn't read repos), a single atomic create is cleaner and future-proof:

```ts
const newAgent = await prisma.agent.create({
  data: { name, selfHosted: false, ...(repos.length > 0 && { repos }) },
});
```

Then remove the conditional `prisma.agent.update` block (lines 1430–1437).

dodizzle (commented):
COMMENT — One improvement before merge: the new agent is created without repos and then repos are attached in a separate `prisma.agent.update` call, leaving a brief window where the agent row exists with an empty repos list. Collapsing to a single atomic `prisma.agent.create` with repos inline is cleaner. Server-side logic is solid (validation, rollback, Slack flow), UI toggle is correct, and 4 smoke tests cover all branches including rollback. CI green.

dmcaulay (commented):
Verdict: APPROVE. Directly and correctly addresses @dodizzle's prior finding: the two-step create-then-update that left a brief window with an empty repos list is now a single atomic prisma.agent.create() with repos passed inline. Rollback-on-provision-failure logic is unaffected and still correct. Verified locally: `bun test admin/src/admin-ui.smoke.test.ts admin/src/admin-ui-pages.unit.test.ts` — 409 pass, 0 fail. No issues.

dodizzle (approved):
APPROVE (re-review). The prior COMMENT finding — two-step agent create leaving a brief intermediate state with no repos — is directly addressed: `prisma.agent.create({ data: { name, selfHosted: false, ...(repos.length > 0 && { repos }) } })` now passes repos inline in a single atomic call, removing the conditional `update`. Smoke test updated to assert `createArgs.repos` and remove the `update` mock. Rollback-on-provision-failure logic is unaffected and still correct. CI green. No new issues.

dodizzle (approved):
APPROVE (re-review, independent verification). All three acceptance criteria confirmed: (1) atomic `prisma.agent.create` with repos inline — no intermediate state; (2) provision-failure rollback via `try/catch` → `prisma.agent.delete` matching agents-api.ts behavior; (3) full smoke + unit test coverage across all server-side branches. Minor observation: rollback smoke test retains an orphaned `update` mock (never called in production path) — harmless. CI green. No new issues. Ready to merge.

admin/src/admin-ui.ts
```diff
@@ -1392,9 +1402,57 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
...
+        data: { name, selfHosted: false, ...(repos.length > 0 && { repos }) },
+      });
+      agentId = newAgent.id;
+
+      try {
+        await provisioner.provision(newAgent.id, { slug: newAgent.name });
```
dmcaulay, line 1430: Orphaned agent + K8s workload risk: `prisma.agent.create()` (line 1424) and `provisioner.provision()` (this line) are both committed side effects at this point, but the only rollback path below covers a `provisioner.provision` failure. If the later Slack `createAppManifest` call fails (the block below this one, outside this diff), the newly created agent row and its provisioned K8s workload are left behind with no cleanup — the handler just returns `formError(msg)`. A retry with the same `newAgentName` would create a second orphan, since nothing here is idempotent on the name. Consider extending the rollback (or moving the Slack call inside this try/catch scope) so a downstream Slack failure also cleans up the agent it just created.

dmcaulay (commented):
COMMENT (re-review). Prior two passes at this commit found no issues and both were APPROVE-recommended (independently reconfirmed by @dodizzle twice on GitHub). On this third pass I found one new edge case worth flagging before this ships broadly: in `agentMode=new`, the agent row is created and its K8s workload provisioned *before* the Slack `createAppManifest` call — but only a `provisioner.provision` failure is rolled back. If Slack fails afterward, the newly created agent + K8s workload is left orphaned with no rollback, and a retry with the same name would create a second orphan since the flow isn't idempotent on `newAgentName`. Not covered by the acceptance criteria or the new tests (none exercise create-succeeds-then-Slack-fails). Not blocking — the existing-agent path is unaffected, CI is green, and the stated acceptance criteria are met — but worth a follow-up before wider use.

dodizzle (approved):
APPROVE (re-review). The Slack-manifest-failure rollback finding from the prior cycle is directly addressed: `agentMode=new` now catches `createAppManifest` errors, deprovisions the K8s workload, deletes the agent row, and re-renders with an error — in that order, each wrapped in `.catch()` to log but not swallow cleanup failures. New smoke test verifies both `deprovision` and `delete` are called with the correct agent ID when Slack throws. All three acceptance criteria confirmed. CI green. Two human APPROVE reviews from @dodizzle already on file.

dmcaulay:
Verdict: APPROVE (re-review). Confirmed the edge-case finding from the prior self-review — a Slack `createAppManifest` failure after the agent row + K8s workload were already created (in `agentMode=new`) left both orphaned, with a same-name retry creating a second orphan — is resolved at HEAD (`c1305c1`). The Slack-failure catch path now mirrors the existing provisioner-failure rollback: deprovisions the K8s workload and deletes the agent row (both error-swallowed with logging) before returning the form error. New smoke tests cover the rollback branch. Verified locally: `bun test admin/src/admin-ui.smoke.test.ts admin/src/admin-ui-pages.unit.test.ts` — 410 pass, 0 fail. `tsc --noEmit` and `biome lint` clean. No new issues.

PR #991, opened 2026-07-02
fix: scope dev-task task-store queries to the calling agent

dmcaulay:
## Summary

Supersedes #969 with the complete fix — that PR only patched the client-side query string, which does not work against the current server, and missed two other call sites with the same bug.

- **task-store server**: `GET /tasks` now honors an explicit `?assignee=` as an additional AND filter on top of `agentScope`'s repo-pool OR-visibility. Previously the server silently discarded any caller-supplied `?assignee=` whenever the token had repo scope — confirmed live against production, not just in theory.
- **`dev-task.md` Step 1**: passes `?assignee=$SHIPWRIGHT_AGENT_ID` server-side (now functional) and re-filters client-side as a backstop before treating anything as "an interrupted task to resume."
- **`check-dev-task.ts`**: the precheck's stale-`in_progress` reset/stamp loop and HITL-pending notification logic had the identical unscoped-query gap — could reset or notify on another agent's task. Now filtered to `deps.agentId`.
- **`task-store/skills/task-store/SKILL.md`**: corrected four scoping claims — three said bearer tokens auto-scope results to the calling agent (they don't; visibility is by repo access, filtering requires explicit `?assignee=`), and one (`?ready=true`) was rewritten by #969 to claim the same broad pool visibility as the other endpoints, which is actually wrong for that endpoint — `listReady()` stays strict (own tasks + unassigned pool only), confirmed via `task-service.ts` and its existing test.

**Not touching plugin/package versions in this PR** — an earlier revision corrected `plugin.json`/`version.txt` back to `4.29.x` (reverting PLG-1.2/#962's reset to `0.1.0`). That's been dropped: Dave is addressing the version-numbering question separately, so this PR now leaves every version-bearing file at main's current value (`0.1.0`) and only carries the assignee-scoping fix.

## Why

Real incident: an agent resumed and began implementing another agent's actively-`in_progress` task (different `assignee`, no relation) because `GET /tasks?status=in_progress` returns the whole repo's task pool for repo-scoped tokens, and dev-task Step 1 took `result.tasks[0]` unfiltered.

## Test Plan
- [x] `bun test task-store/src/` — 204 pass, 59 skip (DB-gated integration tests, will run in CI)
- [x] `bun test plugins/shipwright/scripts/check-dev-task.test.ts` — 20 pass, including 5 new cross-agent scoping regression tests
- [x] `tsc --noEmit` clean in both `task-store/` and `plugins/shipwright/`
- [x] `bunx biome lint` clean on all touched files
- [x] `bun run scripts/sync-version.ts --check` passes — version files match main, unmodified
- [x] New manual scenario added: TESTING.md Scenario 43
- [x] Confirmed live against production that the previous `?assignee=` query param was silently ignored (the exact fix #969 relied on)


dmcaulay (commented):
Verdict: APPROVE. Thorough, correctly-reasoned fix for a real cross-agent data-safety bug: a repo-scoped agent token's caller-supplied ?assignee= was silently discarded server-side, letting one agent resume/reset/notify on another agent's in-progress task. The narrowing-only security reasoning (an explicit ?assignee= can only narrow agentScope's already-visible OR set, never widen it) is sound and consistently applied across all three call sites (task-store route, dev-task.md Step 1, check-dev-task.ts precheck). Verified locally: `bun test task-store/src/api.smoke.test.ts` (39 pass) and `bun test plugins/shipwright/scripts/check-dev-task.test.ts` (20 pass) both green, matching the PR's stated results. Version bump to 0.1.1 verified consistent across all 6 files (package.json x4, plugin.json, marketplace.json, version.txt). No issues.

PR #996, opened 2026-07-02
fix: scope check-deploy's active-deploy guard per repo

dmcaulay:
## Summary
- `check-deploy.ts`'s "deploying guard" looped over every configured repo and returned `exit 1` (nothing to do) the moment **any** repo had an active Deploy workflow run — blocking ready-to-deploy PRs in every other, independent repo too.
- Now it collects a `busyRepos` set per repo instead of short-circuiting globally, and the main PR-scanning loop skips only the busy repo(s). A ready PR in a repo with no active deploy still surfaces as a candidate even while a different repo is mid-deploy.
- Updated the plugin constitution's precheck table to describe the per-repo scoping.

## Context
Per Dan: the deploy skill/precheck was too conservative — "we should be able to merge unless there is an active deploy on the repo. It doesn't need to be across all repos." Confirmed `deploy.md` itself has no equivalent pre-flight gate (it only monitors its own PR's post-merge Deploy run), so this precheck was the only place enforcing (over-broad) concurrency.

## Test plan
- [x] `bun test plugins/shipwright/scripts/check-deploy.ts plugins/shipwright/scripts/check-deploy.unit.test.ts` — 40 pass, 0 fail
- [x] Added two tests: a busy repo no longer blocks a ready PR in a different repo, and a single busy repo still correctly exits 1 (no regression to single-repo behavior)
- [x] `bun run --filter='*' typecheck` — all workspaces pass
- [x] `bunx biome check` on changed files — clean

### DIFF
--- a/plugins/shipwright/CLAUDE.md
+++ b/plugins/shipwright/CLAUDE.md
@@ -122,7 +122,7 @@ favors permissive: false positives are visible and self-correcting; false negati
 | Script | Guards | What it checks |
 |---|---|---|
 | `check-review.ts` | `review` cron | Open PRs with unreviewed commits (by headRefOid dedup against task store `/prs` records); respects `allow_self_review` policy |
-| `check-deploy.ts` | `deploy` cron | Open PRs with `APPROVED` review decision and green CI; respects `allow_self_review` for self-authored PRs |
+| `check-deploy.ts` | `deploy` cron | Open PRs with `APPROVED` review decision and green CI; respects `allow_self_review` for self-authored PRs; skips a repo with an active Deploy workflow run, scoped per repo — a busy repo does not block ready PRs in other configured repos |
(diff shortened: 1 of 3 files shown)


dmcaulay (commented):
Verdict: APPROVE. Correctly-scoped fix: the deploying guard's early `return` on the first busy repo blocked ready PRs in every other, independent repo — replaced with a per-repo `busyRepos` set so the main scanning loop only skips the actually-busy repo(s). Verified the base-branch logic did in fact short-circuit globally (confirmed via `git show origin/main`), and the new per-repo skip (`if (busyRepos.has(repo)) continue;`) matches the PR's stated behavior. Two new unit tests cover both the multi-repo fix (busy repo doesn't block a ready PR elsewhere) and the single-repo regression case (still exits 1 when the only repo is busy). CI green on f3750a2. No API/DB/architecture changes, no breaking changes. No issues.

PR #1027, opened 2026-07-03
feat: per-chat token/cost tracking and display

dodizzle:
## Summary
- Add `GET /threads/:id/stats` endpoint to the chat service returning per-thread message count, input/output token totals, and cumulative cost
- Update `ChatMessage.tokens` type from `number` to a structured `MessageTokens` object with `input_tokens`, `output_tokens`, and cache fields
- Display per-message token/cost badges and a thread-level stats summary in the admin thread detail page

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Structured token tracking (input/output/cache) per message | ✅ MET |
| 2 | Thread stats API (messageCount, totalInputTokens, totalOutputTokens, totalCostUsd) | ✅ MET |
| 3 | Per-message token/cost badge in admin UI | ✅ MET |
| 4 | Thread-level aggregate stats shown in thread header | ✅ MET |
| 5 | Auth enforcement on stats endpoint | ✅ MET |
| 6 | Tests at correct layer | ✅ MET |

## Test Plan
- [x] `GET /threads/:id/stats` returns 200 with correct shape for existing thread
- [x] Returns 404 for unknown thread
- [x] Returns 403 when agent token scoped to a different agent's thread
- [x] Returns 401 without bearer token
- [x] Returns zeros when messages have no token data
- [x] Admin UI shows per-message token badge ("N in / N out · $X.XXXX") on assistant messages
- [x] Admin UI thread header shows aggregated totals formatted with k-suffix
- [x] No badge rendered when `tokens` is null
- [x] Pre-ship typecheck and lint pass clean across all 6 packages


dodizzle (commented):
APPROVE (self-review, posted as COMMENT — GitHub blocks self-APPROVE via API for the PR author. Verdict is APPROVE.)

Clean, correct implementation of per-chat token/cost tracking. All 6 acceptance criteria met. Two non-blocking suggestions below.

**Strengths:**
- The prior `tokens: number | null` type was a fiction — Prisma schema uses `Json?` and the agent always wrote structured objects. The new `MessageTokens` interface correctly aligns the TypeScript type with runtime reality.
- The `typeof m.tokens === "object"` guard before casting to `MessageTokens` safely handles any legacy messages that may have stored numeric values.
- Test coverage is thorough: 401, 403, 404, 200-with-data, and 200-with-zeros all exercised at the smoke layer via `app.request()`, per the project's test-system doc.
- `NoopChatClient.getThreadStats()` stub means degraded-mode admin routes don't regress.

chat/src/routes/threads.ts
```diff
@@ -76,6 +76,16 @@ export function createThreadsRoutes(
     return c.json(thread, 201);
   });
 
+  // ─── Stats ─────────────────────────────────────────────────────────────────
+  app.get("/:id/stats", async (c) => {
+    const thread = await threadService.findById(c.req.param("id"));
+    if (!thread) throw new NotFoundError("thread not found");
+    enforceAgentScope(c.get("agentId"), thread.agentId);
+    const stats = await threadService.getStats(c.req.param("id"));
```
dodizzle: The route calls `threadService.findById()` here (to check existence + enforce agent scope), then `threadService.getStats()` internally does another `prisma.thread.findUnique()` (thread-service.ts:125). That's a redundant DB round-trip per stats request. Could pass the already-fetched thread into `getStats()` or restructure to avoid the double lookup. Low impact for admin-only tooling, but worth noting.

chat/src/thread-service.ts
```diff
@@ -111,6 +120,46 @@ export class ThreadService implements ThreadServiceLike {
...
+    if (!thread) return null;
+
+    const agg = await this.prisma.message.aggregate({
+      where: { threadId },
+      _count: { id: true },
+      _sum: { costUsd: true },
+    });
+
+    // tokens is a JSON blob — can't aggregate in SQL, so sum in JS.
+    const tokenRows = await this.prisma.message.findMany({
```
dodizzle, line 138: The `findMany` loads every message's `tokens` JSON blob into memory to sum in JS. Correct since JSON columns can't be aggregated in SQL — but for threads with many messages (thousands), this will be slow. Fine for admin-only display usage now, but worth documenting the constraint or adding a row cap if chat threads are expected to grow large.

PR #1035, opened 2026-07-03
feat: stamp plugin version into the agent image at build time (PVS-1.2)

dmcaulay:
## Summary
- Stamp the correct plugin version into the agent image at Docker build time (`ARG AGENT_VERSION` + `sync-version.ts` RUN step in `agent/Dockerfile`'s build stage), fed from a bare-version parse step in `build-agent.yml`.
- Closes the "deployed image is one cycle behind" gap left by PVS-1.1 (`sync-plugin-version.yml`), which still lands the git-committed version asynchronously for external `claude plugin install` — that path is unchanged.
- `ARG AGENT_VERSION` defaults to `0.0.0-dev` so `ci.yml`'s PR-time `docker build` (no build-arg) doesn't break.

## Acceptance Criteria
- [x] `build-agent.yml` passes `--build-arg AGENT_VERSION=<bare semver>` to `docker build`, computed via shell string stripping
- [x] `agent/Dockerfile`'s build stage declares `ARG AGENT_VERSION`, copies `scripts/`, and runs `sync-version.ts` after `plugins/`/`.claude-plugin/` are copied in, before the runtime stage copies them out
- [x] `sync-plugin-version.yml` (PVS-1.1) is unmodified — still the source of truth for the git-committed version; header comment updated to describe the new split of responsibilities
- [x] Existing `test-sync-plugin-version.sh` still passes (the reused prefix-stripping logic is already covered there); no new bash test file needed since no new pure-logic branch was introduced beyond that
- [ ] Manual verification after merge: confirm the next `build-agent.yml` run's image has `version.txt` matching that run's own `agent-v*` tag

## Test Plan
- `bash .github/workflows/test-sync-plugin-version.sh` — 10/10 pass (unaffected, confirms shared logic still correct)
- Simulated `sync-version.ts` invocation locally against copied fixture files with both a real version and the `0.0.0-dev` default — both produce valid output
- YAML validated via `Bun.YAML.parse`
- `docker build` itself cannot be exercised in this sandbox (no docker) — relies on `ci.yml`'s `agent-docker-build` job for real validation

### DIFF
--- a/.github/workflows/build-agent.yml
+++ b/.github/workflows/build-agent.yml
@@ -33,6 +33,19 @@ jobs:
           default_bump: patch
           fetch_all_tags: true
 
+      # Bare semver off the tag (agent-v0.117.0 → 0.117.0), stamped into the
+      # image at build time below so the deployed artifact's plugin.json
+      # always matches the tag it was built for — see agent/Dockerfile's
+      # "sync plugin version" step. This does not touch git; the git-committed
+      # version (needed for the public plugin marketplace listing) is synced
+      # asynchronously by .github/workflows/sync-plugin-version.yml, which
+      # this same tag also triggers.
+      - name: Parse bare version from tag
+        id: version
+        run: |
+          TAG="${{ steps.tag.outputs.new_tag }}"
+          echo "version=${TAG#agent-v}" >> "$GITHUB_OUTPUT"
+
       - name: Authenticate to GCP
         uses: google-github-actions/auth@v2
         with:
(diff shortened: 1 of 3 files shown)


dodizzle (approved):
APPROVE — Well-scoped fix to the version-lag problem. The `0.0.0-dev` default ensures `ci.yml` PR builds succeed without a real tag; the `scripts/` COPY is correctly placed after `plugins/` so `sync-version.ts` targets the right files; CI is green. No blocking concerns.

PR #1080, opened 2026-07-04
chore(chart): bump chart version to 1.6.220

dmcaulay:
## Chart version bump: 1.6.219 → 1.6.220

**Triggering tag:** `task-store-v0.53.0`
**New chart version:** `1.6.220`

### What happens next

1. This PR is auto-merged (squash) once CI passes.
2. The merge triggers `chart-release.yml`, which packages and publishes
   chart v1.6.220 to the Helm repository.

---
_Auto-generated by `.github/workflows/auto-bump-chart.yml`._

PR #1087, opened 2026-07-04
fix: add --admin to self-merge of auto-generated version-bump PRs

dmcaulay:
## Problem

Follow-up to #1083. That PR fixed the `gh pr checks --watch` race (waiting for checks to register before watching them), and it worked — PR #1084 shows the workflow correctly reaching "Checks passed — merging". But the plain `gh pr merge --squash` call then failed:

```
X Pull request app-vitals/shipwright#1084 is not mergeable: the base branch policy prohibits the merge.
To use administrator privileges to immediately merge the pull request, add the `--admin` flag.
```

These version-bump PRs are self-authored (opened by the same bot account the workflow runs as), so no external reviewer can approve them, and branch protection has no bypass for that without `--admin`. The previous code comment assumed a ruleset bypass already applied to self-authored PRs on plain merge — #1084 shows that assumption was wrong; it was simply never reached before because the check-registration race failed first.

## Fix

Add `--admin` to both `gh pr merge` calls (`auto-bump-chart.yml`, `sync-plugin-version.yml`), same as the manual admin-merges used to clear the backlog (#1080, #1082, #1084) while this was being diagnosed.

## Test plan

- [ ] Next tag push creates a version-bump PR that merges automatically end-to-end with zero manual intervention

### DIFF
--- a/.github/workflows/auto-bump-chart.yml
+++ b/.github/workflows/auto-bump-chart.yml
@@ -253,11 +253,11 @@ jobs:
       # Merge directly once required checks pass, rather than GitHub's native
       # queued auto-merge (`gh pr merge --auto`). Native auto-merge only
       # re-evaluates mergeability on its own schedule, which can lag well past
-      # this workflow's 10-minute wait when the merge relies on a repo
-      # ruleset bypass (self-authored PR, no external reviewer) instead of a
-      # genuine approval — `gh pr checks --watch` blocks on required checks,
-      # then an immediate `gh pr merge` (no `--auto`) applies the bypass right
-      # away, the same way a manual merge click does.
+      # this workflow's 10-minute wait. `--admin` is required because this PR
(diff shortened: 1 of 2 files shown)


dmcaulay (commented):
Verdict: APPROVE (posted as COMMENT — self-authored PR). Minimal, well-justified fix: adds `--admin` to both `gh pr merge --squash` calls (auto-bump-chart.yml, sync-plugin-version.yml) so the automated version-bump workflows can actually complete once checks pass. Root cause is verified against real evidence — PR #1084 reached "Checks passed — merging" (confirming #1083's check-registration fix works) but then failed on the plain merge with "the base branch policy prohibits the merge," and was ultimately merged manually via `--admin` by @dmcaulay, exactly matching the failure mode described here. Same one-line change applied identically to both affected workflows, comments updated to match corrected reasoning. CI green, no blocking issues.

PR #1135, opened 2026-07-04
fix: retry stale-base merge race in sync-plugin-version and auto-bump-chart

dmcaulay:
## Summary
- Both `sync-plugin-version.yml` and `auto-bump-chart.yml` trigger off the same tag pushes and race to merge onto `main`, so either can fail with a transient `GraphQL: Base branch was modified. Review and try the merge again.` error when the other lands first — leaving a green-checks PR stuck open until a human notices.
- Wrap the `gh pr merge --squash --admin` call in both workflows in a 3-attempt retry loop with a 5s sleep, gated by a new `is_stale_base_error()` helper that matches only that specific transient GraphQL error text. Any other failure (real conflict, auth error, branch protection rejection) still fails immediately, unchanged from before.
- Mirror the same `is_stale_base_error()` logic and test cases into `test-sync-plugin-version.sh` and `test-auto-bump-chart.sh`, matching this repo's existing convention of duplicating inline workflow bash logic into a paired test script.

## Acceptance Criteria
| 1 | sync-plugin-version.yml merge wrapped in retry loop (2-3 attempts, short sleep) | MET | `MAX_ATTEMPTS=3` for-loop with `sleep 5` |
| 2 | auto-bump-chart.yml merge wrapped in retry loop (2-3 attempts, short sleep) | MET | Same pattern |
| 3 | Retry only on stale-base GraphQL error; other failures fail immediately | MET | `is_stale_base_error()` grep-matches specific error text |
| 4 | Test coverage for the new logic | MET | Both test scripts updated with 3 new assertions each; full suites pass (13/13, 16/16) |

## Test Plan
- `bash .github/workflows/test-sync-plugin-version.sh` — 13/13 passed
- `bash .github/workflows/test-auto-bump-chart.sh` — 16/16 passed
- Manually simulated the retry loop against a mocked `gh` failing twice with the stale-base error then succeeding on the 3rd attempt — confirmed it retries and exits 0
- `bunx biome lint .` — clean

### DIFF
--- a/.github/workflows/auto-bump-chart.yml
+++ b/.github/workflows/auto-bump-chart.yml
@@ -258,6 +258,19 @@ jobs:
       # protection otherwise rejects the merge outright — confirmed via
       # PR #1084, where `gh pr merge --squash` (no `--admin`) failed with
       # "the base branch policy prohibits the merge" even after checks passed.
+      #
+      # Retry on stale-base races: this workflow and sync-plugin-version.yml
+      # both trigger off the same tag pushes and race to merge onto main.
+      # When the other workflow's PR lands first, main moves out from under
+      # this PR (no textual conflict — just a stale base ref) and
+      # `gh pr merge` fails with a transient GraphQL error: "Base branch was
+      # modified. Review and try the merge again." Confirmed via run
+      # 28703925995 (agent-v0.129.0, 2026-07-04 10:54). Retry a few times
+      # with a short sleep to absorb that window. Any other failure (real
+      # conflict, auth error, branch protection rejection, etc.) is NOT
+      # retried — fail immediately so a human is alerted, same as before this
+      # change. If the stale-base error persists past all retries, the step
+      # still ultimately fails for the same reason.
       - name: Wait for checks and merge
         if: steps.idempotency.outputs.skip != 'true' && steps.commit-push.outputs.push_skipped != 'true'
         timeout-minutes: 10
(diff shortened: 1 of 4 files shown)


dmcaulay (commented):
APPROVE — Small, well-scoped, well-tested fix for a real transient stale-base merge race between sync-plugin-version.yml and auto-bump-chart.yml. Verified the retry loop's control flow under GitHub Actions' default `set -e` semantics directly (the `MERGE_OUTPUT=$(...) && { ...; exit 0; }` pattern correctly avoids triggering `set -e` on failure, `continue` retries as expected, non-stale-base errors exit immediately). Both `test-sync-plugin-version.sh` (13/13) and `test-auto-bump-chart.sh` (16/16) pass locally. No blocking issues found.

.github/workflows/auto-bump-chart.yml
```diff
@@ -281,4 +294,35 @@ jobs:
...
           echo "Checks passed — merging ${PR_URL}."
-          gh pr merge "$PR_URL" --squash --admin
+
+          # Detects the transient stale-base GraphQL error (see comment
+          # above). Mirrors .github/workflows/test-auto-bump-chart.sh — keep
+          # both copies in sync.
+          is_stale_base_error() {
+            local output="$1"
+            if echo "$output" | grep -qi "Base branch was modified" || \
+               echo "$output" | grep -qi "try the merge again"; then
```
dodizzle, line 304: The second OR clause (`grep -qi "try the merge again"`) is broader than needed. Any GitHub error that happens to contain that phrase — even a non-transient one — would trigger up to 3 silent retries before alerting a human. `"Base branch was modified"` alone is specific and sufficient to identify the stale-base race.

Same concern applies in `sync-plugin-version.yml` at the equivalent line.

```bash
# Suggested: use only the unambiguous clause
if echo "$output" | grep -qi "Base branch was modified"; then
  return 0
fi
```

.github/workflows/sync-plugin-version.yml
```diff
@@ -243,5 +256,36 @@ jobs:
...
-          gh pr merge "$PR_URL" --squash --admin \
-            --subject "chore(plugin): sync version to ${VERSION} [skip ci]"
+
+          # Detects the transient stale-base GraphQL error (see comment
+          # above). Mirrors .github/workflows/test-sync-plugin-version.sh —
+          # keep both copies in sync.
+          is_stale_base_error() {
+            local output="$1"
+            if echo "$output" | grep -qi "Base branch was modified" || \
+               echo "$output" | grep -qi "try the merge again"; then
```
dodizzle, line 266: Same second-OR-clause concern as `auto-bump-chart.yml` — `grep -qi "try the merge again"` is broader than necessary. Suggest removing it and relying solely on `"Base branch was modified"` for the same reasons.

dodizzle (commented):
COMMENT — One concern before merging: the second OR clause (`grep -qi "try the merge again"`) in `is_stale_base_error()` is broader than necessary and could silently retry non-transient merge failures. `"Base branch was modified"` alone is unambiguous and sufficient. The retry loop control flow itself is correct under `set -eo pipefail`; no issues with the loop logic or parity between the two workflows.

PR #1139, opened 2026-07-05
fix: tighten stale-base error match to a single unambiguous phrase

dmcaulay:
## Summary
- Follow-up to #1135 per [dodizzle's review comment](https://github.com/app-vitals/shipwright/pull/1135#issuecomment-review): the second OR clause in `is_stale_base_error()` (`grep -qi "try the merge again"`) is broader than necessary and could silently retry a non-transient merge failure that happens to contain that generic phrase.
- The actual GraphQL error text always includes `"Base branch was modified"`, so matching on that phrase alone is sufficient and unambiguous.
- Applied identically in both `auto-bump-chart.yml` and `sync-plugin-version.yml`, plus their respective test scripts (kept in sync per the existing comment convention).
- Added a regression test case in both test scripts confirming a message containing "try the merge again" but not "Base branch was modified" now correctly returns false.

## Test plan
- [x] `bash .github/workflows/test-auto-bump-chart.sh` — 17/17 pass
- [x] `bash .github/workflows/test-sync-plugin-version.sh` — 14/14 pass

### DIFF
--- a/.github/workflows/auto-bump-chart.yml
+++ b/.github/workflows/auto-bump-chart.yml
@@ -266,11 +266,16 @@ jobs:
       # `gh pr merge` fails with a transient GraphQL error: "Base branch was
       # modified. Review and try the merge again." Confirmed via run
       # 28703925995 (agent-v0.129.0, 2026-07-04 10:54). Retry a few times
-      # with a short sleep to absorb that window. Any other failure (real
-      # conflict, auth error, branch protection rejection, etc.) is NOT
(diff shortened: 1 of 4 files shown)


dodizzle (approved):
APPROVE — Correct narrowing of an over-broad error match. The `is_stale_base_error()` two-clause OR introduced in #1135 is simplified to match only "Base branch was modified", which is unambiguous and always present in the real GraphQL error. Regression test cases added to both `.sh` mirror scripts. No blocking concerns.

PR #1151, opened 2026-07-05
feat: add shared Sentry module to @shipwright/lib

dmcaulay:
## Summary
- Add `lib/sentry.ts` to `@shipwright/lib` exporting `initSentry(opts, sentryClient?)`, `scrubEvent`, `scrubLog`, and `SECRET_ENV_VARS`
- `initSentry` is a no-op (zero telemetry) when `SENTRY_DSN` is unset; when set, it wires `enableLogs`, a console logging integration, environment fallback (`SENTRY_ENVIRONMENT ?? NODE_ENV ?? "production"`), a `service` tag, and `beforeSend`/`beforeSendLog` scrub hooks
- `scrubEvent`/`scrubLog` unconditionally strip `Authorization`/`Cookie` request headers and recursively redact any string matching a currently-set secret-shaped env var (cycle-safe, depth-capped)
- Added `@sentry/bun` as a dependency of `lib/package.json` and exported `./sentry` in its exports map

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | `initSentry` never calls `sentryClient.init` when `SENTRY_DSN` is unset | MET |
| 2 | `initSentry` calls `sentryClient.init` with `enableLogs`, service tag, both scrub hooks wired, when `SENTRY_DSN` is set | MET |
| 3 | `scrubEvent`/`scrubLog` strip `Authorization`/`Cookie` headers unconditionally and redact live secret env values with no false positives | MET |
| 4 | New `lib/sentry.unit.test.ts` — pure/DI-injected unit tests only, no `mock.module()`/`global.*` overrides | MET |

## Test Plan
- `bun test lib/sentry.unit.test.ts` — 11 pass, 100% line/function coverage
- `bunx biome lint .` — clean
- `bun run --filter='*' typecheck` — clean across all workspaces
- `bun scripts/check-config-docs.ts`, `bun run scripts/sync-version.ts --check`, `bun plugins/shipwright/scripts/check-banned-strings.ts` — all pass
- Full `bun test` run has 15 pre-existing failures in `agent/`/`admin/`/`metrics/` (Kubernetes CA fallback, plugin installer exec mocking, chat-token-reporter) unrelated to this `lib/`-only change

### DIFF
--- a/lib/sentry.ts
+++ b/lib/sentry.ts
@@ -13,24 +13,17 @@ type SentryLog = NonNullable<BunOptions["beforeSendLog"]> extends (
   ? L
   : never;
 
-/** Options accepted by {@link initSentry}. */
 export interface InitSentryOptions {
-  /** Value tagged onto every event/log as `service` (e.g. "metrics", "agent"). */
   service: string;
 }
 
-/** The subset of the @sentry/bun client surface initSentry depends on — injectable for tests. */
+/** Narrowed to the one method initSentry calls, so tests can inject a fake without mock.module(). */
 export interface SentryClient {
   init: (options: Record<string, unknown>) => void;
 }
 
-/**
- * Fixed list of secret-shaped env vars redacted from Sentry events/logs. Any string value
- * (anywhere, nested) that exactly matches the *current* value of one of these is replaced
- * with "[Filtered]". Unset/empty vars are never matched, so there is no false-positive
- * redaction when a secret simply isn't configured in the current environment.
- */
-const SECRET_ENV_VARS = [
+/** Secret-shaped env vars redacted from Sentry events/logs when currently set to a non-empty value. */
+export const SECRET_ENV_VARS = [
   "ANTHROPIC_API_KEY",
   "CLAUDE_CODE_OAUTH_TOKEN",
   "GH_APP_PRIVATE_KEY",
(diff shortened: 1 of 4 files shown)


lib/sentry.ts
```diff
@@ -0,0 +1,147 @@
...
+function redactSecrets<T>(
+  value: T,
+  secrets: string[],
+  seen: WeakSet<object>,
+  depth: number,
+): T {
+  if (secrets.length === 0) return value;
+
+  if (typeof value === "string") {
+    return (secrets.includes(value) ? "[Filtered]" : value) as unknown as T;
```
dodizzle: **Secret scrubbing uses exact equality, not substring matching.**

`secrets.includes(value)` is `Array.prototype.includes` — it returns `true` only when the entire string *exactly equals* a known secret. A string like `"Failed to authenticate: Bearer ghp_secretvalue"` (by far the most common leakage scenario) is not redacted because it is not equal to `"ghp_secretvalue"`.

Suggested fix:
```ts
if (typeof value === "string") {
  if (secrets.length === 0) return value as unknown as T;
  let result = value;
  for (const secret of secrets) {
    if (result.includes(secret)) {
      result = result.split(secret).join("[Filtered]");
    }
  }
  return result as unknown as T;
}
```
Also add a test case where the secret appears embedded in a longer string to lock in the behavior.

dodizzle (commented):
COMMENT — One important security gap: `redactSecrets` uses `secrets.includes(value)` (exact equality) rather than substring matching, so secrets embedded in error messages pass through unredacted. See inline comment.

Everything else is clean — good DI design, solid test coverage, correct cycle-safe scrubbing logic.

dodizzle (approved):
APPROVE — prior finding addressed: `redactSecrets` now uses substring-based matching (`result.includes(secret)` + `result.split(secret).join("[Filtered]")`) rather than exact equality, and the new test case confirms embedded secrets are redacted correctly. All four acceptance criteria are met and CI is green. Ready to merge.

PR #1212, opened 2026-07-06
feat: add check-learn-dream precheck

dmcaulay:
## Summary
- Add `check-learn-dream.ts` precheck for the learn-dream cron, using the same anchor-file pattern as `check-docs-freshness.ts`
- Reads `state/learn-dream-last-run.json` for a last-run timestamp and compares it against session transcript mtimes under `~/.claude/projects/<sanitized-workspace-path>/*.jsonl`, exiting 1 when nothing is newer (skip the cron) and 0 otherwise (permissive on first run or any read failure)
- Refresh `docs/architecture.md` to reflect the new precheck script

## Acceptance Criteria
| 1 | Reads last-run anchor from state/learn-dream-last-run.json (absent -> exit 0) | MET | check-learn-dream.ts reads anchor; exits 0 permissively when null |
| 2 | Exits 1 when no transcript newer than anchor | MET | newerCount===0 -> exit 1, empty output |
| 3 | Exits 0 with summary when at least one transcript newer | MET | exit 0 with count summary |
| 4 | check-learn-dream.test.ts, deps-injection, sibling naming, covers scenarios | MET | 6 tests covering no-anchor, anchor-newer-than-all, empty-list, anchor-older-than-one, count-summary, and permissive-null-read |

## Test Plan
- `bun test plugins/shipwright/scripts/check-learn-dream.test.ts` — 6/6 pass
- `bunx tsc --noEmit` (plugins/shipwright) — clean
- `bunx biome lint .` — clean
- Full `bun test` — 3265 pass, 15 pre-existing failures (confirmed identical on `main`, unrelated to this change: kubernetes-client, plugin installation, chat-token-reporter)

### DIFF
--- a/plugins/shipwright/scripts/check-learn-dream.ts
+++ b/plugins/shipwright/scripts/check-learn-dream.ts
@@ -81,13 +81,15 @@ export async function run(deps: Deps): Promise<RunResult> {
 
 /**
  * Derive the Claude Code project directory name for a workspace path, the
- * same way Claude Code itself does: the absolute path with `/` replaced by
- * `-` (e.g. `/data/agent-home/workspace` → `-data-agent-home-workspace`).
+ * same way Claude Code itself does: the absolute path with `/` and `.`
+ * replaced by `-` (e.g. `/data/agent-home/workspace` →
+ * `-data-agent-home-workspace`; `/home/user/my.workspace` →
+ * `-home-user-my-workspace`).
  */
 export function sanitizeWorkspacePathForClaudeProjects(
   workspacePath: string,
 ): string {
-  return workspacePath.replace(/\//g, "-");
+  return workspacePath.replace(/[/.]/g, "-");
 }
 
 function buildProductionDeps(): Deps {
--- a/plugins/shipwright/scripts/check-learn-dream.ts
+++ b/plugins/shipwright/scripts/check-learn-dream.ts
@@ -56,12 +56,18 @@ export async function run(deps: Deps): Promise<RunResult> {
 
   // Read failure — unknown state, exit permissively
   if (mtimes === null) {
-    return { exit: 0, output: "" };
+    return {
+      exit: 0,
+      output: "Transcript read failed — running permissively.",
+    };
   }
 
   // Anchor unparsable — unknown state, exit permissively
   if (Number.isNaN(anchorMs)) {
-    return { exit: 0, output: "" };
+    return {
+      exit: 0,
+      output: "Anchor unparsable — running permissively.",
+    };
   }
 
   const newerCount = mtimes.filter((mtime) => mtime > anchorMs).length;
--- a/plugins/shipwright/scripts/check-learn-dream.ts
+++ b/plugins/shipwright/scripts/check-learn-dream.ts
@@ -1,0 +1,150 @@
+#!/usr/bin/env bun
+/**
+ * plugins/shipwright/scripts/check-learn-dream.ts
+ *
+ * Pre-check for the learn-dream cron.
+ *
+ * Reads state/learn-dream-last-run.json to get the last-run timestamp.
+ * - If absent → exit 0 (first run, always worth running)
+ * - If no session transcript is newer than the anchor → exit 1 (nothing new to learn from)
+ * - If at least one transcript is newer → exit 0 with a short summary as stdout
+ *
+ * This is a coarser gate than the other prechecks in this plan — it can only
+ * rule out the zero-activity case, not judge whether the activity contains a
+ * worthwhile learning. That judgement is left to the learn-dream skill itself.
+ *
+ * Exit 0 + summary → transcript activity since last run, run the dream job
+ * Exit 1 + no output → nothing to do
+ *
+ * Usage:
+ *   bun plugins/shipwright/scripts/check-learn-dream.ts
+ */
+
+import { existsSync, readdirSync, readFileSync, statSync } from "node:fs";
+import { homedir } from "node:os";
+import { join } from "node:path";
+import { resolveWorkspacePath } from "./check-helpers.ts";
+
+// ─── Types ────────────────────────────────────────────────────────────────────
+
+interface Deps {
+  readLastRunAnchor: () => string | null;
+  listTranscriptMtimes: () => number[] | null;
+}
+
+interface RunResult {
+  exit: 0 | 1;
+  output: string;
+}
+
+// ─── Core logic ───────────────────────────────────────────────────────────────
+
+export async function run(deps: Deps): Promise<RunResult> {
+  const lastRun = deps.readLastRunAnchor();
+
+  // First run — no anchor, always worth checking
+  if (lastRun === null) {
+    return {
+      exit: 0,
+      output: "No last-run anchor found — running full learn-dream check.",
+    };
+  }
+
+  const anchorMs = Date.parse(lastRun);
+
+  const mtimes = deps.listTranscriptMtimes();
+
+  // Read failure — unknown state, exit permissively
+  if (mtimes === null) {
+    return { exit: 0, output: "" };
+  }
+
+  // Anchor unparsable — unknown state, exit permissively
+  if (Number.isNaN(anchorMs)) {
+    return { exit: 0, output: "" };
+  }
+
+  const newerCount = mtimes.filter((mtime) => mtime > anchorMs).length;
+
+  // No transcript activity since last run — nothing to do
+  if (newerCount === 0) {
+    return { exit: 1, output: "" };
+  }
+
+  return {
+    exit: 0,
+    output: `${newerCount} transcript file(s) newer than last run — running learn-dream check.`,
+  };
+}
+
+// ─── Production deps ──────────────────────────────────────────────────────────
+
+/**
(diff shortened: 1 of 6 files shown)


plugins/shipwright/scripts/check-learn-dream.ts
```diff
@@ -0,0 +1,150 @@
...
+): string {
+  return workspacePath.replace(/\//g, "-");
+}
+
+function buildProductionDeps(): Deps {
+  const cwd = process.cwd();
+
+  return {
+    readLastRunAnchor: (): string | null => {
+      const anchorPath = join(cwd, "state", "learn-dream-last-run.json");
```
dodizzle, line 106: The anchor path is read here, but nothing in the codebase writes `state/learn-dream-last-run.json`. The `learn-dream` command and `learning-dreamer` agent have no step that creates or updates this file. Because the file never exists, `readLastRunAnchor()` always returns `null`, and the precheck always exits 0 (first-run permissive path) — every cron firing still reaches the full LLM session. Add a write step to whichever command/agent runs the dream job (e.g. after the learning-dreamer agent completes, write `{ "lastRun": new Date().toISOString() }` to this path).

dodizzle (commented):
COMMENT — The core logic is correct and the tests are solid, but `state/learn-dream-last-run.json` is never written by any skill (`learn-dream` command and `learning-dreamer` agent have no step that writes this file). The precheck will always hit the `null` → exit 0 path, so every cron firing still reaches the full LLM session — the guard provides no protection until something writes the anchor. Either add a write step to the `learn-dream` command / `learning-dreamer` agent, or note the gap so it can be tracked.

plugins/shipwright/scripts/check-learn-dream.ts
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * Derive the Claude Code project directory name for a workspace path, the
+ * same way Claude Code itself does: the absolute path with `/` replaced by
+ * `-` (e.g. `/data/agent-home/workspace` → `-data-agent-home-workspace`).
+ */
+export function sanitizeWorkspacePathForClaudeProjects(
+  workspacePath: string,
+): string {
+  return workspacePath.replace(/\//g, "-");
```
dodizzle: The Claude Code project directory encoding rule replaces both `/` and `.` with `-` (per `plugins/shipwright/skills/investigate-cron/SKILL.md` line 55), but this only replaces `/`. A workspace path with dots (e.g. `/home/user/my.workspace`) produces the wrong directory name, causing `listTranscriptMtimes` to find nothing and return `null` — silently falling back to permissive exit 0 on every firing.

```ts
return workspacePath.replace(/[/\.]/g, "-");
```

dodizzle (commented):
COMMENT — one bug in `sanitizeWorkspacePathForClaudeProjects`: it only replaces `/` with `-` but the Claude Code encoding rule (per `investigate-cron/SKILL.md` line 55) requires replacing both `/` and `.` with `-`. On any workspace path containing dots this will silently produce the wrong transcript directory and fall back to permissive exit 0 on every cron firing. Fix: use a regex that matches both `/` and `.`. Everything else looks good — clean deps-injection design, solid test coverage, well-documented anchor write contract.

dmcaulay (commented):
APPROVE — Both prior findings are correctly resolved: `sanitizeWorkspacePathForClaudeProjects` now replaces both `/` and `.` (matching `investigate-cron/SKILL.md`'s documented encoding rule), with new tests covering dotted paths; and `learn-dream.md` now has a step instructing the command to write `state/learn-dream-last-run.json` on successful completion, ordered to avoid falsely advancing the anchor on partial failures. 9/9 tests pass locally, lint clean, CI green. No new issues found.

dodizzle (commented):
**Re-review after a19d8fa** — prior finding (dot-to-dash encoding in `sanitizeWorkspacePathForClaudeProjects`) has been addressed.

**New finding: precheck script not wired into the cron definition**

The learn-dream entry in `admin/src/system-crons.ts` (line 76) has no `preCheck` field, so `check-learn-dream.ts` is never invoked by the cron system. Every other cron with a precheck script declares it via `preCheck` (e.g. `check-docs-freshness.ts` at line 72 just above). Without this wiring, the cron fires unconditionally — the precheck script exists and is tested, but has no effect.

Fix: add `preCheck: "shipwright:check-learn-dream.ts"` to the learn-dream cron entry:
```ts
{
  name: "learn-dream",
  schedule: "0 3 * * *",
  preCheck: "shipwright:check-learn-dream.ts",
  prompt: "/shipwright:learn-dream --since 1d --review",
  silent: true,
  enabled: false,
}
```

**COMMENT** — one wiring step missing before this PR achieves its stated purpose.

dodizzle (commented):
**Re-review after a19d8fa** — prior finding (dot-to-dash encoding in `sanitizeWorkspacePathForClaudeProjects`) has been addressed.

**New finding: precheck script not wired into the cron definition**

The learn-dream entry in `admin/src/system-crons.ts` (line 76) has no `preCheck` field, so `check-learn-dream.ts` is never invoked by the cron system. Every other cron with a precheck script declares it via `preCheck` (e.g. `check-docs-freshness.ts` at line 72 just above). Without this wiring, the cron fires unconditionally — the precheck script exists and is tested, but has no effect.

Fix: add `preCheck: "shipwright:check-learn-dream.ts"` to the learn-dream cron entry:
```ts
{
  name: "learn-dream",
  schedule: "0 3 * * *",
  preCheck: "shipwright:check-learn-dream.ts",
  prompt: "/shipwright:learn-dream --since 1d --review",
  silent: true,
  enabled: false,
}
```

**COMMENT** — one wiring step missing before this PR achieves its stated purpose.

dodizzle (approved):
All prior findings resolved. APPROVE.

**Prior findings:**
- Dot-encoding bug in `sanitizeWorkspacePathForClaudeProjects` — fixed (regex now `/[/.]/g`)
- `preCheck: "shipwright:check-learn-dream.ts"` missing from learn-dream cron — fixed (added at `admin/src/system-crons.ts:78`)

Clean implementation following established precheck patterns. Tests comprehensive (6 unit tests, all scenarios covered), CI green.

dodizzle (commented):
**Re-review after 1f2b676** — prior findings (dot-encoding, preCheck wiring) are confirmed resolved.

**New finding: permissive fallback paths suppress the cron instead of running it**

Two fallback paths in `check-learn-dream.ts` return `{ exit: 0, output: "" }`: when `listTranscriptMtimes()` is `null` (read failure, line 54) and when `anchorMs` is `NaN` (unparsable anchor, line 60). In `cron-handler.ts:253`, the handler evaluates `if (checkExitCode !== 0 || !output)` after `output = checkOutput.trim()` — empty string trims to `""`, so `!output` is `true`, and the session is suppressed with `"preCheck:no-work"`. The comments say "unknown state, exit permissively" but the actual effect is the opposite: the job doesn't run.

**Fix**: Return a non-empty output string in both fallback branches:
```ts
// line 54:
return { exit: 0, output: "Transcript read failed — running permissively." };
// line 60:
return { exit: 0, output: "Anchor unparsable — running permissively." };
```

**COMMENT** — one fix needed before merge.

plugins/shipwright/scripts/check-learn-dream.ts
```diff
@@ -0,0 +1,152 @@
...
+  if (lastRun === null) {
+    return {
+      exit: 0,
+      output: "No last-run anchor found — running full learn-dream check.",
+    };
+  }
+
+  const anchorMs = Date.parse(lastRun);
+
+  const mtimes = deps.listTranscriptMtimes();
```
dodizzle, line 55: This returns empty `output` on exit 0. `cron-handler.ts:253` evaluates `if (checkExitCode !== 0 || !output)` — empty string trims to `""`, so `!output` is `true` and the cron session is suppressed (not run permissively as intended).

Fix: `return { exit: 0, output: "Transcript read failed — running permissively." };`

plugins/shipwright/scripts/check-learn-dream.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  // Read failure — unknown state, exit permissively
+  if (mtimes === null) {
+    return { exit: 0, output: "" };
+  }
+
```
dodizzle, line 64: Same issue — empty output on exit 0 causes the cron handler to suppress the session rather than run permissively.

Fix: `return { exit: 0, output: "Anchor unparsable — running permissively." };`

dodizzle (approved):
APPROVE — all prior findings resolved.

**Prior findings:**
- Dot-encoding bug in `sanitizeWorkspacePathForClaudeProjects` — fixed (regex `/[/.]/g`)
- `preCheck: "shipwright:check-learn-dream.ts"` missing from learn-dream cron — fixed
- Permissive fallback paths returning `{ exit: 0, output: "" }` suppressing the cron — fixed; both branches now return descriptive non-empty output strings

Test coverage is comprehensive: 8 tests covering no-anchor, anchor-newer-than-all, empty-list, anchor-older-than-one, count-summary, permissive-null-read, unparsable-anchor, and path sanitization with slash/dot/mixed patterns. CI green.

PR #1219, opened 2026-07-06
feat: wire Sentry into metrics and consolidate onError

dmcaulay:
## Summary
- Wire Sentry into the metrics service: `initSentry({service:"metrics"})` in `server.ts`, and conditional `@sentry/hono/bun` middleware mount (only when `SENTRY_DSN` is set)
- Extend the shared `makeOnError` factory in `metrics/src/lib/errors.ts` to call `Sentry.captureException` for 5xx-mapped and unhandled errors (not 4xx, not the "Malformed JSON" 400 special case)
- Consolidate `createMetricsApp` and `createPublicMetricsApp` in `metrics/src/api.ts` onto the single `makeOnError` factory, removing the duplicate inline `onError` blocks
- Add `@sentry/bun`/`@sentry/hono` deps and an optional `SENTRY_DSN`/`SENTRY_ENVIRONMENT` env passthrough to the metrics Helm chart (mirroring the existing task-store pattern)

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Both metrics app factories use the single makeOnError factory; no inline onError duplication remains | MET |
| 2 | Sentry.captureException fires from makeOnError only when SENTRY_DSN is set, verified via unit test with injected fake client | MET |
| 3 | Existing error-path behavior (status codes, ApiError mapping, "Malformed JSON" -> 400 special case) is unchanged — regression-covered | MET |
| 4 | Test decision: extended metrics/src/lib/errors unit tests to cover the Sentry-capture branch | MET |

## Test Plan
- `bun test metrics/` — 285/285 pass
- Full repo `bun test` — 3276 pass / 15 fail (same 15 pre-existing failures confirmed present on `main`, unrelated to this change — agent plugin install, HttpKubernetesClient, HttpChatTokenReporter)
- `bun run --filter='*' typecheck` — clean
- `bunx biome lint .` — clean
- `bun scripts/check-config-docs.ts`, `sync-version.ts --check` — pass


dmcaulay:
Shipwright cleanup — resuming task from prior session

PR #1221, opened 2026-07-06
feat: wire Sentry into agent (SEN-1.5)

dmcaulay:
## Summary
- Calls `initSentry({ service: "agent" })` in `agent/src/index.ts`, placed right after the console-timestamp monkeypatch so Sentry's `consoleLoggingIntegration` wraps on top of it
- Adds an optional `sentryClient` param to `startHealthServer()` in `agent/src/health.ts`; the `POST /cron` catch block now calls `sentryClient?.captureException(err)` before the existing `console.error` and 500 response (skipped for the typed `ValidationError` / 422 branch)
- Adds `@sentry/bun` to `agent/package.json`; documents `SENTRY_DSN`/`SENTRY_ENVIRONMENT` for the agent in `docs/configuration.md` (agent is provisioned per-tenant by `KubernetesAgentProvisioner`, not chart-provisioned, so operators set `SENTRY_DSN` directly in the agent's own env — same as `ANTHROPIC_API_KEY`)

## Acceptance Criteria

| 1 | initSentry call placed after console monkeypatch, verified by code-order or test | MET | `agent/src/index.ts` — `initSentry(...)` sits directly after the monkeypatch loop |
| 2 | With SENTRY_DSN unset, cron behavior/console output unchanged | MET | `sentryClient?.captureException` no-ops when undefined; test covers this explicitly |
| 3 | With SENTRY_DSN set, cron error triggers captureException before console.error/response | MET | `health.ts` catch block order verified + test |
| 4 | Unit test on cron error path asserting captureException via injected fake client; no tests retired | MET | 3 new tests added to `cron-handler.integration.test.ts`, no removals |

## Test Plan
- `bun test agent/` — 88/88 pass in the affected suites, including 3 new Sentry-wiring tests
- `cd agent && bun run typecheck` — clean
- `bunx biome lint .` — clean
- `bun scripts/check-config-docs.ts` — all env vars documented
- `bun run scripts/sync-version.ts --check` — passes
- `bun test` (repo-wide) — 3272 pass / 15 fail, matching the pre-existing 15 failures on `main` (confirmed via baseline run) — no regressions
- Coverage: `agent/src/health.ts` at 100% lines/funcs


agent/src/index.ts
```diff
@@ -120,7 +129,14 @@ const cronDeps: CronHandlerDeps = {
...
   process.env.SHIPWRIGHT_HEALTH_PORT ?? DEFAULT_HEALTH_PORT,
 );
-startHealthServer(healthPort, analytics.summarize, cronDeps, slackClock);
+startHealthServer(
+  healthPort,
+  analytics.summarize,
+  cronDeps,
+  slackClock,
+  undefined,
+  process.env.SENTRY_DSN ? Sentry : undefined,
```
dodizzle, line 138: This ternary is evaluated once at module load, before the config-sync loop runs. If `SENTRY_DSN` arrives via the 60-second config-sync bundle (`Object.assign(process.env, bundle.env)`), Sentry stays inactive for the pod's lifetime with no warning. Docs say to set it like `ANTHROPIC_API_KEY` (direct pod env), but a silent failure here is easy to miss. Consider: `if (!sentryInitialized && process.env.SENTRY_DSN) { console.warn('[agent] SENTRY_DSN set after startup — Sentry is inactive; restart required to enable') }` in the config-sync handler, or add a line to docs/configuration.md noting that config-sync delivery won't activate Sentry.

dodizzle (commented):
COMMENT — All 4 acceptance criteria met and CI is green. One issue before merge: both `initSentry()` and the `process.env.SENTRY_DSN ? Sentry : undefined` ternary are evaluated at startup before the 60-second config-sync loop runs. If an operator sets `SENTRY_DSN` via the agent config bundle rather than as a pod env var, Sentry silently stays inactive for the pod's lifetime with no warning. A log warning after config sync ("SENTRY_DSN set via config-sync but Sentry was not initialized at startup — restart required") or an explicit note in `docs/configuration.md` that this env var must be present at pod startup would close the gap. The docs already say to set it like `ANTHROPIC_API_KEY`, so a one-liner clarifying config-sync doesn't work for this var would be enough.

dodizzle (approved):
APPROVE — All acceptance criteria met. `initSentry` is correctly placed after the console monkeypatch; unhandled cron errors are captured via the injected `sentryClient`; `ValidationError` is explicitly excluded; docs now clearly state `SENTRY_DSN` must be present at pod startup (config-sync will not work). CI green.

PR #1236, opened 2026-07-06
feat(ci): debounce chart-bump bursts into a single PR

dmcaulay:
## Why

Shipwright services release independently, so it's common for two or three
release tags to land within seconds/minutes of each other (e.g.
`admin-v0.156.0` immediately followed by `agent-v0.144.0`). Today each tag
triggers its own `auto-bump-chart.yml` run — its own PR, its own merge, its
own `chart-release.yml` publish, its own downstream vitals-os bump — for what
is really one release event. That's a lot of merge/Deploy noise for no added
safety.

## What changed

`auto-bump-chart.yml` now runs in two jobs:

1. **`debounce`** — sleeps 90s in a `chart-bump-debounce` concurrency group
   with `cancel-in-progress: true`. A newer tag arriving during that sleep
   cancels the older run outright, so only the *last* tag in a burst survives.
2. **`bump-chart`** — unchanged concurrency semantics (queued, never
   cancelled, so a run in progress always finishes). It now collects every
   release tag that landed since the previous chart-bump commit (not just its
   own triggering tag), and credits all of them in the PR body, commit
   message, CHANGELOG entry, and Artifact Hub `changes` annotation.

Net effect: a burst of releases collapses into one PR, one merge, one
`chart-release.yml` run, and one downstream vitals-os bump — same traceability
(every triggering tag is still listed), fewer deploys.

Known tradeoff (documented in the workflow's header comment): if tags land
continuously faster than the 90s window, no bump ever lands. Not a concern at
current release cadence (a handful of tags/day); flagged for anyone revisiting
this later.

## Test plan

- [x] `bash .github/workflows/test-auto-bump-chart.sh` — 23/23 passing,
      including new tests for `collect_batch_tags` (multi-tag batching +
      no-prior-bump fallback) and multi-tag annotation/changelog formatting.
- [x] Validated the new YAML structure (job graph, `needs`, concurrency
      groups) parses correctly.
- [x] Manually replayed the tag-collection and formatting logic against
      scratch git repos matching the real `admin-v0.156.0` /
      `agent-v0.144.0` burst that prompted this change.
- [ ] Can't fully exercise the live GitHub Actions burst/cancellation
      behavior without pushing real tags — see the manual test plan in the
      workflow's header comment for how to do that safely with dummy tags.

🤖 Generated with Claude Code (Doc, App Vitals agent)

### DIFF
--- a/.github/workflows/auto-bump-chart.yml
+++ b/.github/workflows/auto-bump-chart.yml
@@ -7,25 +7,51 @@ name: Auto-Bump Chart Version
 # successful image push. This workflow detects those tags and
 # opens a chart-version bump PR so no manual bump-chart.sh invocation is needed.
 #
+# Debounce: services release independently, so it's common for two or three
+# tags to land within seconds/minutes of each other (e.g. admin-v0.156.0
+# immediately followed by agent-v0.144.0). Bumping the chart once per tag in
+# that situation produces a chart-bump PR/merge/publish cycle per tag for what
+# is really one release event — multiplying merge noise and Deploy runs for no
+# added safety. The `debounce` job absorbs bursts: it sleeps for
+# DEBOUNCE_SECONDS, and its `chart-bump-debounce` concurrency group
+# (cancel-in-progress: true) means a newer tag arriving during that sleep
+# cancels the older run outright — only the *last* tag in a burst survives to
+# reach the `bump-chart` job. `bump-chart` then credits every release tag that
+# landed since the previous chart-bump commit (not just its own triggering
+# tag) so the batch is still fully traceable, just collapsed into a single
+# version bump.
+#
+# Known limitation: if tags land continuously faster than DEBOUNCE_SECONDS
+# apart, no run ever clears the debounce window and no bump lands. This is an
+# accepted tradeoff — release cadence here is a handful of tags a day, not a
+# continuous stream. If that changes, add a max-wait fallback.
+#
 # Flow:
-#   tag push (agent-v* / admin-v* / metrics-v*)
-#     → this workflow
-#       → bump patch version in Chart.yaml
-#       → open PR (chore/chart-v{new_version})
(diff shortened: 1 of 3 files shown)


.github/workflows/auto-bump-chart.yml
```diff
@@ -72,6 +117,51 @@ jobs:
...
+                | awk -F'|' -v since="$SINCE" '$1 > since { print $2 }')
+            done
+          fi
+
+          # Always include the triggering tag even if it fell outside the
+          # window (clock skew, first-ever bump) — never silently drop the
+          # tag that fired this run.
+          TRIGGER="${GITHUB_REF_NAME}"
+          FOUND=false
+          for T in "${TAGS[@]:-}"; do
```
dodizzle, line 153: `${TAGS[@]:-}` expands to a single empty string `""` when `TAGS` is empty (the no-`SINCE` case), so the loop iterates once with `T=""`. It works correctly — `"" != TRIGGER`, so TRIGGER still gets appended — but the intent is "skip the loop if the array is empty". `${TAGS[@]:+${TAGS[@]}}` makes that explicit and avoids the ghost iteration.

dodizzle (approved):
APPROVE — Clean, well-tested debounce implementation. The two-job design (cancellable `debounce` + queued `bump-chart`) correctly absorbs tag bursts while preserving full traceability. The `needs: debounce` dependency ensures `bump-chart` only runs after a complete quiet period, and cancellation of `debounce` cleanly skips `bump-chart` with no half-written state. All 23 shell tests pass. One low-confidence suggestion: `${TAGS[@]:-}` in the fallback loop emits one empty string when `TAGS` is empty — functionally harmless (empty string != trigger tag, `sort -u` would deduplicate) but `${TAGS[@]:+${TAGS[@]}}` is more explicit. No blocking concerns.

PR #1252, opened 2026-07-06
feat: consolidate cron-run token accounting onto AgentCronRunModelBreakdown

dmcaulay:
## Summary
- `AgentCronRunModelBreakdown` becomes the sole source of truth for cron-run token/cost accounting — dropped the redundant `inputTokens`/`outputTokens`/`cacheReadTokens`/`cacheCreationTokens` columns from `AgentCronRun`, backfilling a synthetic `legacy-unattributed` breakdown row (cost $0) for any pre-existing run that had token data but no breakdown row.
- Rewrote `queryTotals`/`queryByAgent`/`queryByCron`/`queryDaily` in `agent-cron-run-stats.ts` to a direct `LEFT JOIN AgentCronRunModelBreakdown` + `SUM` (no subquery, no `COALESCE`); `queryByModel`/`queryByCronModel` keep their existing shape.
- Updated the two consumers that read/wrote the dropped columns directly: `agent-cron-runs.ts`'s `patch()` (now ignores the legacy top-level token fields for backward compat with older agent builds) and `admin-ui-pages.ts`'s per-run token cell (now sums `modelBreakdown`). Also fixed `agents-api.ts`'s `serializeCronRun()`, a third consumer discovered while validating the column drop, which read the same columns directly in the API response serializer.

## Note
Pre-2026-07-01 runs will show `model = legacy-unattributed` and `cost = $0`. This is not new data loss — the real per-run model name and cost for those runs were already unrecoverable, dropped by an earlier migration (`20260701030000_drop_cron_run_legacy_columns`) before `AgentCronRunModelBreakdown` existed. Token counts were never dropped and are preserved verbatim by this migration's backfill.

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Migration backfills breakdown rows + drops 4 columns | MET |
| 2 | schema.prisma no longer declares the 4 columns | MET |
| 3 | queryTotals/ByAgent/ByCron/Daily → direct LEFT JOIN+SUM; queryByModel/ByCronModel unchanged shape | MET |
| 4 | agent-cron-runs.ts patch() no longer maps legacy fields onto runData | MET |
| 5 | admin-ui-pages.ts per-run cell sums modelBreakdown | MET |
| 6 | Test updates: fixtures via breakdown rows, legacy-backfill integration test, unit test for cell, stale assertions removed | MET |
| 7 | PR description notes legacy-unattributed/$0 caveat | MET (this note, above) |

## Test Plan
- `bun test` — full repo suite green; the only failures present (kubernetes-client CA fallback, plugin-install shell-out tests, chat-token cost fallback) are pre-existing on `main` and unrelated to this change.
- `bun run --filter='*' typecheck` — all packages pass.
- New integration test asserting a pre-backfill legacy row (no breakdown rows pre-migration) surfaces correctly post-backfill as `legacy-unattributed`/$0.
- New unit test for the per-run token cell summing `modelBreakdown`.


admin/src/agent-cron-runs.ts
```diff
@@ -120,6 +120,11 @@ export class AgentCronRunService {
       throw new NotFoundError(`cron run ${runId} not found`);
     }
 
+    // The top-level token fields (inputTokens/outputTokens/…) are still accepted
+    // on the input for backward compatibility with older agent builds, but they
```
dodizzle, line 124: Worth adding a TODO here pointing to `agent/src/cron-handler.ts`'s `buildTokenPayload()` follow-up, so this deferred cleanup doesn't get lost. Something like: `// TODO(CTT-cleanup): remove top-level token field acceptance once all agent builds send modelBreakdown only.` The current comment explains the why well; a TODO ensures the follow-up is findable.

dodizzle (commented):
COMMENT — one behavior gap before merge.

The `patch()` method now silently drops top-level `inputTokens`/`outputTokens`/`cacheReadTokens`/`cacheCreationTokens` PATCH body fields (columns dropped). Any agent build using the old `buildTokenPayload()` path (sends top-level token fields only, no `modelBreakdown`) will have token data silently discarded post-merge — those runs will show em-dash in the UI and null in the API. The PR body acknowledges this as intentional/deferred. If the agent-side sender rollout is confirmed, this can merge as-is; otherwise a TODO comment pointing to the `agent/src/cron-handler.ts` follow-up would prevent the deferred cleanup from going stale.

Everything else is clean: migration is idempotent, all six query methods are unified to the same LEFT JOIN pattern, API response shape is backward-compatible (same nullable fields, computed from breakdown), and the test coverage is solid including the legacy-backfill integration test.

Prior findings: none — this is a first review.

dmcaulay (commented):
APPROVE — clean consolidation of cron-run token accounting onto `AgentCronRunModelBreakdown`. Migration is idempotent, all six stats query methods are consistently rewritten to the same LEFT/INNER JOIN pattern with no double-counting risk (1:1 join key on `[cronRunId, model]`), and both remaining consumers (`admin-ui-pages.ts` cell, `agents-api.ts` `serializeCronRun()`) are updated. The one gap from the prior review round — `patch()` silently dropping legacy top-level token fields — is now resolved with an explicit comment and a `TODO(CTT-cleanup)` pointing at the agent-side follow-up. Test coverage is solid: new legacy-backfill integration test, per-run-cell unit test, and fixtures updated across all four affected test files. CI green.

(Posted as COMMENT per self-review policy — GitHub blocks self-APPROVE via API. Verdict is APPROVE, no blocking concerns.)

dodizzle (approved):
APPROVE — Prior finding resolved: the `TODO(CTT-cleanup)` comment in `agent-cron-runs.ts` now documents the silent token-field drop and points to the agent-side follow-up. Migration is idempotent (INSERT uses NOT EXISTS, DROP uses IF EXISTS), all six stat queries use a consistent `LEFT JOIN AgentCronRunModelBreakdown` with no double-counting risk, `serializeCronRun()` correctly returns null (not 0) for runs with no breakdown rows, and `list()` always includes `modelBreakdown: true`. Legacy-backfill integration test is a solid addition. CI green. No blocking concerns.

dodizzle (approved):
APPROVE — Re-review at same commit (31b8f71) confirms prior verdict. All acceptance criteria met: migration is idempotent (NOT EXISTS guard), four token columns dropped, all six stat queries use a consistent `LEFT JOIN AgentCronRunModelBreakdown`, `patch()` ignores legacy top-level token fields with a `TODO(CTT-cleanup)` comment pointing to the agent-side follow-up, UI token cell sums `modelBreakdown` correctly. Test coverage solid across integration, unit, and smoke layers. CI green. No blocking concerns.

dodizzle (approved):
APPROVE — Prior finding resolved: `TODO(CTT-cleanup)` comment added to `agent-cron-runs.ts` `patch()` documenting the silent token-field drop and pointing to `agent/src/cron-handler.ts`'s `buildTokenPayload()` follow-up. All 6 acceptance criteria met. Migration is idempotent (NOT EXISTS guard), all six stat queries use consistent LEFT JOIN pattern, API contract preserved via `serializeCronRun()` summing from breakdown, test coverage comprehensive. CI green.

dodizzle (approved):
APPROVE — clean consolidation. Token accounting now lives exclusively in `AgentCronRunModelBreakdown`; migration, query rewrites, all three consumers, and tests are all consistent. No blocking issues.

PR #1271, opened 2026-07-08
fix: filter GET /mcp/tools through the tool allowlist

dmcaulay:
## Summary
- `GET /mcp/tools` in `mcp-server/src/index.ts` now filters through `allowedTools()` (the same filter `mcp-server.ts` already uses for the stdio/Streamable HTTP MCP surface) instead of mapping over the raw 25-tool `generatedTools` array.
- The HTTP discovery endpoint now matches the actual MCP protocol surface: 9 tools (`tasks_list`, `tasks_create`, `tasks_bulk`, `tasks_distinct`, `tasks_get`, `tasks_update`, `prs_list`, `prs_get`, `prs_update`) instead of leaking pipeline-internal lifecycle ops, `tasks_delete`, and all `tokens_*` routes.

## Acceptance Criteria

| 1 | GET /mcp/tools filters through allowedTools() instead of the raw generatedTools array | MET | `mcp-server/src/index.ts`: `const tools = allowedTools(generatedTools);` |
| 2 | Response only lists the 9 allowed tools | MET | New smoke test asserts `toHaveLength(9)` and all 9 expected names present |
| 3 | Test asserting the endpoint returns exactly the allowlisted set and excludes at least one lifecycle/token/destructive op | MET | `mcp-server/src/index.smoke.test.ts` asserts every `EXCLUDED_TOOLS` entry is absent from the response |

## Test Plan
- `bun test mcp-server/src/index.smoke.test.ts` — 3/3 pass (new test written RED-first, confirmed failing against the unfixed handler before the fix)
- `task lint`, `task typecheck`, `task check-config-docs`, `task check-version-sync`, `task check-strings` — all pass
- `task test` (full suite) — 15 pre-existing failures unrelated to this change (metrics integration tests hitting live endpoints, reproduced identically on `main`)

### DIFF
--- a/mcp-server/src/index.smoke.test.ts
+++ b/mcp-server/src/index.smoke.test.ts
@@ -1,5 +1,18 @@
 import { describe, it, expect } from "bun:test";
 import { app } from "./index.js";
+import { EXCLUDED_TOOLS } from "./tool-allowlist.js";
+
+const ALLOWED_TOOL_NAMES = [
+  "tasks_list",
+  "tasks_create",
+  "tasks_bulk",
+  "tasks_distinct",
+  "tasks_get",
+  "tasks_update",
+  "prs_list",
+  "prs_get",
+  "prs_update",
+] as const;
 
 describe("mcp-server", () => {
   it("exports an app", () => {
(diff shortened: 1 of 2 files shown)


dmcaulay (commented):
APPROVE — small, correctly-scoped fix. `/mcp/tools` now reuses the existing `allowedTools()` filter (same one already applied to the stdio/Streamable HTTP MCP transport) instead of leaking pipeline-internal, destructive, and token-management tools. Verified the resulting allowed set is exactly the 9 tools claimed (25 generated - 16 excluded), and the new smoke test asserts both the allowed set and absence of every excluded tool. Ran the test locally — 3/3 pass. No blocking concerns.

PR #1286, opened 2026-07-08
feat: add shipwright-loop system cron and scheduling dispatch

dmcaulay:
## Summary
- Adds a `shipwright-loop` entry to `SYSTEM_CRONS` (1-minute schedule, `enabled: false`, `silent: true`, no `preCheck`) per the system-cron-change rule that new system crons always ship disabled.
- In `syncCrons()`, routes the `shipwright-loop` job to a new `handleLoopCronRequest()` handler instead of the generic `handleCronRequest()`, since the loop's multi-invocation, drain-until-dry shape doesn't fit the single-preCheck/single-prompt model.
- Extracts the scheduling decision into a pure, unit-tested `classifyCronJobsForScheduling()` (agent/src/loop-cron-classifier.ts): when this agent's own fetched jobs include an enabled `shipwright-loop`, the five pipeline phase jobs (dev-task, review, patch, review-patch, deploy) are excluded from independent node-cron scheduling — they remain loop-config-only. Unmigrated agents (no loop job, or present-but-disabled) see zero behavior change.
- Adds `resolveLoopPhaseToggles()`, which resolves dev-task/review/patch/deploy as four independent, non-mutually-exclusive phase booleans and deliberately never reads `shipwright-review-patch`'s enabled flag (its review-vs-patch selection is redundant with the loop's own per-phase toggling).
- `handleLoopCronRequest()` is a placeholder that logs resolved phase toggles — the full multi-invocation orchestration is WL-3.3.

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | SYSTEM_CRONS includes shipwright-loop, enabled:false, 1-min schedule | MET |
| 2 | syncCrons() dispatches shipwright-loop to a distinct handler | MET |
| 3 | Independent scheduling of the 5 pipeline jobs skipped only when shipwright-loop enabled for this agent | MET |
| 4 | Loop toggle logic treats 4 phases independently, never reads review-patch's enabled flag | MET |
| 5 | Classification logic extracted to pure unit-tested function; migrated/unmigrated/review-patch-independence cases covered; no existing syncCrons tests retired | MET |

## Test Plan
- `agent/src/loop-cron-classifier.unit.test.ts` — 10 new unit tests covering unmigrated (loop absent/disabled), migrated (loop enabled) scheduling, disabled-job exclusion, null-name fallthrough, and explicit `shipwright-review-patch`-has-no-effect assertions for the toggle resolver.
- `admin/src/system-crons.unit.test.ts` — extended existing suite for the new `shipwright-loop` entry (count, names, schedule, enabled, silent, no preCheck).
- `bun test` (agent + admin packages) and `typecheck` both pass; two pre-existing, unrelated full-suite flakes were confirmed to reproduce on clean `main` (tracked separately) and are not caused by this change.


agent/src/loop-cron-classifier.ts
```diff
@@ -0,0 +1,137 @@
...
+// ─── Loop cron handler (placeholder) ───────────────────────────────────────────
+
+/**
+ * PLACEHOLDER for the shipwright-loop dispatch handler. Resolves the current
+ * phase toggles and logs them — the actual multi-invocation drain-until-dry
+ * orchestration (claiming work, firing /shipwright:dev-task, /shipwright:review,
+ * /shipwright:patch, /shipwright:deploy per enabled phase, looping until dry)
+ * is WL-3.3's job and explicitly out of scope here.
+ */
+export async function handleLoopCronRequest<T extends CronJobLike>(
```
dodizzle, line 132: WL-3.3 note: `handleLoopCronRequest(jobs)` in the caller closes over the `jobs` snapshot from the `syncCrons()` invocation that first registered this task. Because the cron is only registered once (`!cronTasks.has(id)` guard), subsequent sync cycles update a different `jobs` binding — the registered callback never sees it. For the placeholder (logs only) this is fine, but the WL-3.3 orchestrator will need fresh state each invocation. Recommend accepting `(runtimeClient, agentId)` or a `getJobs()` callback instead of a snapshot, or re-fetching inside the handler.

dodizzle (approved):
Clean implementation of the loop cron scaffolding. All five acceptance criteria met, zero behavioral change for unmigrated agents, CI green.

One heads-up for WL-3.3: `handleLoopCronRequest(jobs)` captures the `jobs` snapshot from the syncCrons() call that first registered the loop task. Phase-toggle changes won't be visible until the agent restarts (or the loop cron is re-registered). The placeholder only logs so this is inert now, but WL-3.3's actual orchestrator will need to re-fetch jobs inside the handler (or accept a runtimeClient/agentId ref) to see live enable/disable changes without a restart.

APPROVE — no blocking concerns.

PR #1294, opened 2026-07-08
feat: port check-helpers.ts shared utilities into agent/src

dmcaulay:
## Summary
- Add `agent/src/check-helpers.ts`, a native 1:1 port of `plugins/shipwright/scripts/check-helpers.ts` — the shared task-store client, gh CLI/GraphQL helpers, `getCurrentUser`, `readAllowSelfReview`, `resolveWorkspacePath`/`resolveAllRepos`, and `isMergeOnlyUpdate` — for in-process use by the upcoming precheck ports (WL-2.2), without a subprocess hop
- The only intentional behavior adjustment: `createTaskStoreClient` gains an optional injectable `fetchFn` param (defaults to global `fetch`) so tests can inject a fake fetch instead of overriding `globalThis.fetch`, per this repo's test-isolation rule
- `plugins/shipwright/scripts/check-helpers.ts` and its test file are left completely untouched — still required by the plugin's own check-*.ts scripts, review-patch's internal precheck orchestration, and the manual-only ship-loop skill
- Refresh `docs/migration.md` to note the new agent/src copy

## Acceptance Criteria
| 1 | agent/src gains a native module exposing the same functions as check-helpers.ts, adjusted only for import paths/module boundaries — no behavior change | MET | `agent/src/check-helpers.ts` ports all 11 exports 1:1; only deviation is the optional `fetchFn` DI param, defaulting to global fetch |
| 2 | plugins/shipwright/scripts/check-helpers.ts is untouched | MET | `git diff main...HEAD -- plugins/shipwright/scripts/check-helpers.ts` is empty |
| 3 | Existing unit tests ported alongside relocated functions, adapted to new module path; no coverage lost, no old tests retired | MET | `agent/src/check-helpers.unit.test.ts` (44 tests) covers every case in the plugin's suite plus more; plugin test file untouched |

## Test Plan
- [x] `bun test agent/src/check-helpers.unit.test.ts` — 44 pass / 0 fail
- [x] `task lint` — clean
- [x] `task typecheck` — all 7 workspaces pass
- [x] Coverage: 87.04% lines (vs. 56.10% baseline on the original plugin file — net improvement, no coverage lost)
- [x] Confirmed plugin source + test file byte-identical to `main`

### DIFF
--- a/agent/src/check-helpers.ts
+++ b/agent/src/check-helpers.ts
@@ -1,0 +1,384 @@
+/**
+ * agent/src/check-helpers.ts
+ *
+ * Shared helpers ported from plugins/shipwright/scripts/check-helpers.ts —
+ * a native, directly-importable copy for the agent runtime. The plugin
+ * scripts remain the source of truth for agents that still run the
+ * check-*.ts scripts as separate processes; this module exists so
+ * in-process precheck ports (WL-2.2) can import the same logic without a
+ * subprocess hop.
+ *
+ * Behavior is unchanged from the plugin source except createTaskStoreClient,
+ * which gains an optional injectable fetch function so callers can supply a
+ * fake fetch in tests instead of overriding globalThis.fetch (agent/src test
+ * isolation rule: no global.fetch/global.* overrides). When no fetchFn is
+ * passed, it falls back to global fetch — identical behavior to the plugin
+ * version.
+ *
+ * Covers:
+ * - Workspace path resolution (WORKSPACE_PATH env var or cwd heuristic)
+ * - Org/repo resolution from repos/ dir
+ * - gh CLI execution helper
+ */
+
+import { spawnSync } from "node:child_process";
+import { existsSync, readFileSync, readdirSync } from "node:fs";
+import { join } from "node:path";
+
+// ─── Task types ───────────────────────────────────────────────────────────────
+
+export type TaskStatus =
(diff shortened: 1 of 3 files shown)


dmcaulay (commented):
Verdict: APPROVE — Verified byte-for-byte behavioral fidelity to plugins/shipwright/scripts/check-helpers.ts (plugin file diff is empty); the only intentional deviation is the DI'd `fetchFn` param on `createTaskStoreClient`, matching the test-isolation rule. Ran the new suite (44/44 pass), lint and typecheck clean. Module has no consumers yet — deliberately, ahead of WL-2.2. No blocking issues.

dmcaulay:
Follow-up flagged on WL-2.2 (the next step in this port): shipwright #1333 (merged after this PR was opened) added `isCleanApproveBody` to `plugins/shipwright/scripts/check-helpers.ts`, fixing a bug where `check-deploy.ts`/`check-patch.ts` didn't recognize narrative "Verdict: APPROVE" self-reviews. This port (WL-2.1) predates that fix, so `agent/src/check-helpers.ts` doesn't have it yet — no action needed here, but WL-2.2 (porting `check-deploy.ts`/`check-patch.ts` into `agent/src`) needs to bring `isCleanApproveBody` along too, not re-derive the old logic. Tracked in the WL-2.2 task-store note.

PR #1299, opened 2026-07-09
feat: add repos to agent config response schema + handler

dodizzle:
## Summary
- Add `repos: string[]` to `AgentConfigResponse` (schema + interface + handler) so `GET /agents/:id/config` — the self-serve endpoint agent-scoped tokens can already call — returns the agent's scoped repo list, sourced from the existing `prisma.agent.findUnique` call (no new query).
- Regenerate `admin/openapi.json` and `lib/admin-types.ts` so downstream consumers (the `agent` package's generated OpenAPI client) pick up the new required field, and update `agent` package test fixtures that construct `AgentConfigResponse` literals accordingly.
- Refresh `docs/agent-api.md` to document the new `repos` field on the config response.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `AgentConfigResponseSchema` includes `repos: z.array(z.string())` | ✅ MET |
| 2 | `AgentConfigResponse` interface includes `repos: string[]` | ✅ MET |
| 3 | `GET /agents/:id/config` response body includes `repos: agent.repos`, no new Prisma query | ✅ MET |
| 4 | `GET /agents/:id` (admin-only route) and its `isAdmin` gate unchanged; no write path for repos added | ✅ MET |
| 5 | No new test file needed — smoke test coverage deferred to dependent task ACR-1.2 | ✅ MET |

## Test Plan
- [x] Existing `admin/src/api.smoke.test.ts` suite passes unchanged (14 pass / 0 fail)
- [x] `bun run --filter='*' typecheck` passes across all workspaces
- [x] `bunx biome lint .` passes with no issues
- [x] Full `bun test` run confirmed at parity with `main` (3538 pass / 14 pre-existing unrelated failures, verified identical on both branches)

### DIFF
--- a/admin/src/api.ts
+++ b/admin/src/api.ts
@@ -42,6 +42,7 @@ export interface AgentConfigResponse {
   env: Record<string, string>;
   allowedTools: string[];
   plugins: AgentPlugin[];
+  repos: string[];
 }
 
 interface AgentEnvServiceLike {
@@ -67,7 +68,7 @@ interface AgentCronJobServiceLike {
 
 interface PrismaLike {
   agent: {
-    findUnique(args: { where: { id: string } }): Promise<{ id: string } | null>;
+    findUnique(args: { where: { id: string } }): Promise<{ id: string; repos: string[] } | null>;
   };
   agentPlugin: {
     findMany(args: {
@@ -196,6 +197,7 @@ export function createAgentRuntimeApp(deps: AgentRuntimeDeps): OpenAPIHono {
           ? { marketplace: "shipwright", plugin: p.name }
           : { plugin: p.name.slice(0, at), marketplace: p.name.slice(at + 1) };
       }),
+      repos: agent.repos,
     };
 
     return c.json(response, 200);
--- a/admin/openapi.json
+++ b/admin/openapi.json
@@ -1541,12 +1541,23 @@
             "items": {
               "$ref": "#/components/schemas/AgentConfigPlugin"
             }
+          },
+          "repos": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "example": [
+              "org/repo1",
+              "org/repo2"
+            ]
           }
         },
         "required": [
           "env",
           "allowedTools",
-          "plugins"
+          "plugins",
+          "repos"
         ]
       },
       "RuntimeError": {
--- a/admin/src/api.smoke.test.ts
+++ b/admin/src/api.smoke.test.ts
@@ -34,6 +34,7 @@ interface MockPlugin {
 interface MockAgent {
   id: string;
   name: string;
+  repos: string[];
 }
 
 // ─── Mock factories ───────────────────────────────────────────────────────────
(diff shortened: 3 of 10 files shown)


admin/src/api.ts
```diff
@@ -196,6 +197,7 @@ export function createAgentRuntimeApp(deps: AgentRuntimeDeps): OpenAPIHono {
           ? { marketplace: "shipwright", plugin: p.name }
           : { plugin: p.name.slice(0, at), marketplace: p.name.slice(at + 1) };
       }),
+      repos: agent.repos,
```
dodizzle, line 200: This makes the response satisfy `AgentConfigResponseSchema` (which now requires `repos`) only when `agent.repos` is actually populated. Nothing here or in `@hono/zod-openapi`'s `app.openapi()` validates the outgoing response against the schema, and the existing smoke-test mock in `admin/src/api.smoke.test.ts` (`MockAgent`, unmodified by this PR) doesn't supply `repos` — so those tests currently exercise a response that's missing this now-required field, and nothing asserts on `body.repos`. Safe in production (Prisma always returns the column, default `[]`), but worth a smoke-test assertion so a future refactor can't silently drop this line without a test catching it.

dodizzle (commented):
Verdict: COMMENT — Small, well-scoped additive change with green CI (repos threaded cleanly through interface, Zod schema, generated OpenAPI/types, handler, docs, and every downstream fixture). One important issue: `AgentConfigResponseSchema.required` now includes `repos`, but responses aren't runtime-validated against that schema, and the existing `admin/src/api.smoke.test.ts` mock (`MockAgent`/`makeMockPrisma`, not touched by this diff) still omits `repos` — I verified directly that the handler's JSON response in that case silently drops the `repos` key entirely, and no test in the file asserts on `body.repos`. Production is unaffected today (Prisma's real client always returns `repos` via the schema default `[]`), but the test suite now gives false confidence for this field. Worth a quick assertion added to the existing `GET /:id/config` smoke tests, or an explicit note that this is a known, accepted gap ahead of ACR-1.2.

dodizzle (commented):
APPROVE — Small, additive `repos: string[]` field on `GET /agents/:id/config`, reusing the existing Prisma query with no new DB round-trip. Non-breaking (no current consumer reads `.repos` yet, so no rolling-deploy hazard), covered by two new smoke tests plus an extended combined-app regression test, and OpenAPI spec/types/docs updated in lockstep. CI green. No blocking issues.

dodizzle (commented):
Verdict: APPROVE — Re-review: the prior finding (schema declares `repos` required but the smoke-test mock silently produced a response missing it) is directly closed — `MockAgent`/`buildApp()` now thread a `repos` option through the mock agent, the base config test asserts `body.repos`, a new test covers the empty-array case, and the combined-app regression test asserts `body.repos` too. No production code changed since the last review (verified via diff), smoke suite passes 15/15 at this commit, CI and pr-title-lint green. No blocking issues.

PR #1300, opened 2026-07-09
fix: mask agent secret env vars and stop minting duplicate provisioning tokens

dmcaulay:
## Summary

Two related bugs found while diagnosing a prod agent crash-loop (chart 1.6.309 deploy):

- **Secret masking gap**: `GET /agents/:id/envs` returned real secrets (Slack bot/app/signing/client-secret tokens, GH tokens, Claude OAuth tokens, task-store/agent API keys, etc.) in cleartext for every agent except `SENTRY_DSN`, because none of the three provisioning-flow `patch()` calls in `admin-ui.ts` passed a `secretKeys` set. Extracted the canonical list out of `lib/sentry.ts` into a new shared `lib/secret-env-vars.ts` and wired it in everywhere secrets get written during provisioning.
- **Duplicate token minting**: the `/admin/provision/xapp-token` step independently minted a *second*, different `SHIPWRIGHT_AGENT_API_KEY` and `SHIPWRIGHT_TASK_STORE_TOKEN` into AgentEnv on every new K8s-managed agent — duplicating what `provisioner.provision()` had already minted straight into the K8s Secret moments earlier in the same flow. Its idempotency check looked at AgentEnv (always empty at that point for a fresh agent) instead of the Secret, so the check never actually prevented the double-mint. Removed the minting entirely: K8s-managed agents already get both credentials from the Secret via `secretKeyRef`, and self-hosted agents never read either (no containerized entrypoint, GitHub-backed task-store CLI instead of a bearer token). Also dropped the now-redundant `SHIPWRIGHT_TASK_STORE_URL` write, since the Deployment manifest already sets that unconditionally for every K8s-managed agent.

## Test plan

- [x] `bun run typecheck` clean across all 7 workspace packages
- [x] `bun test admin/src/ lib/` — 1186 pass, 0 fail (rewrote/removed 5 tests that asserted the old dual-mint behavior)
- [x] biome-formatted
- [x] Backfilled `secretKeys` for all 7 already-provisioned K8s-managed agents in prod (verified zero value changes, only the `secret` flag flipped) and removed the now-orphaned duplicate `SHIPWRIGHT_AGENT_API_KEY`/`SHIPWRIGHT_TASK_STORE_TOKEN`/`SHIPWRIGHT_TASK_STORE_URL` entries from AgentEnv where present

https://claude.ai/code/session_01KcART2JJwCDBHL8FL3DDFB

### DIFF
--- a/admin/src/provision-callback.smoke.test.ts
+++ b/admin/src/provision-callback.smoke.test.ts
@@ -69,7 +69,11 @@ async function makeProvisionStateCookie(
 
 interface MockState {
   upsertCalls: Array<{ agentId: string; env: Record<string, string> }>;
-  patchCalls: Array<{ agentId: string; env: Record<string, string> }>;
+  patchCalls: Array<{
+    agentId: string;
+    env: Record<string, string>;
+    secretKeys: string[];
+  }>;
   reconcileCalls: string[];
   createTokenCalls: Array<{ agentId: string; label?: string }>;
 }
@@ -165,8 +169,16 @@ function makeMockDeps(
       upsert: async (agentId: string, env: Record<string, string>) => {
         state.upsertCalls.push({ agentId, env });
       },
-      patch: async (agentId: string, env: Record<string, string>) => {
-        state.patchCalls.push({ agentId, env });
+      patch: async (
+        agentId: string,
+        env: Record<string, string>,
+        secretKeys?: Set<string>,
+      ) => {
+        state.patchCalls.push({
+          agentId,
+          env,
+          secretKeys: secretKeys ? [...secretKeys] : [],
+        });
       },
       deleteKey: async () => {},
       getConfigBundle: async () => null,
@@ -297,8 +309,13 @@ describe("GET /admin/provision/complete — OAuth callback", () => {
 
     // SLACK_BOT_TOKEN should have been stored via patch()
     expect(state.patchCalls.length).toBeGreaterThan(0);
-    const storedEnv = state.patchCalls[state.patchCalls.length - 1].env;
-    expect(storedEnv).toHaveProperty("SLACK_BOT_TOKEN", "xoxb-mock-bot-token");
+    const lastPatch = state.patchCalls[state.patchCalls.length - 1];
+    expect(lastPatch.env).toHaveProperty(
+      "SLACK_BOT_TOKEN",
+      "xoxb-mock-bot-token",
+    );
+    // SLACK_BOT_TOKEN must be marked as a secret so it's masked on read
+    expect(lastPatch.secretKeys).toContain("SLACK_BOT_TOKEN");
   });
 
   it("absent cookie → renders error page (not blank form)", async () => {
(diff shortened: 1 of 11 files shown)


admin/src/provision-callback.smoke.test.ts
```diff
@@ -517,24 +515,19 @@ describe("POST /admin/provision/xapp-token", () => {
...
-    );
-    expect(apiKeyPatch).toBeDefined();
-    expect(apiKeyPatch?.env.SHIPWRIGHT_AGENT_API_KEY).toBe(
-      "raw_test_token_abc123def456",
-    );
+    // Neither key gets minted here — K8s-managed agents already have both from
+    // provisioner.provision() (Secret is the source of truth); self-hosted
+    // agents mint their own token separately from the agent detail page.
+    expect(
+      state.patchCalls.some((c) => "SHIPWRIGHT_AGENT_API_KEY" in c.env),
```
dodizzle, line 554: Good assertion for the dedup fix, but nothing here (or elsewhere in this file) checks that `agentEnvService.patch` was called with `secretKeys` containing `SLACK_APP_TOKEN` — the mock `patch: async (agentId, env) => {...}` at line 168 drops the third argument entirely, so the masking fix this PR is primarily about has no test coverage. Worth extending `MockState.patchCalls` to capture `secretKeys` and asserting it here and at the other two `patch()` call sites (config-save, slack-bot-token).

dodizzle (commented):
Verdict: COMMENT — Correct, well-scoped security fix (real prod bug: secrets returned in cleartext from GET /agents/:id/envs, plus a duplicate/orphaned token mint on every provision) with green CI, ran the changed test files locally (471 pass), typecheck and lint clean, and no dangling references to the removed `taskStoreProvisioningClient` wiring (still legitimately used by the K8s provisioning path). One important gap: the test suite thoroughly covers the token-dedup half of the fix (explicit assertions that no second key is minted) but not the masking half — every mock `agentEnvService.patch` touched in this PR (`provision-callback.smoke.test.ts`, `admin-ui.smoke.test.ts`) only captures `(agentId, env)`, silently dropping the third `secretKeys` argument (compiles because it's optional on the real `AgentEnvService.patch` signature). So nothing in this diff actually verifies that `new Set(SECRET_ENV_VARS)` is passed through from `admin-ui.ts`'s three provisioning `patch()` call sites — the exact behavior this PR exists to fix.

dodizzle (approved):
Verdict: APPROVE — Follow-up to the prior COMMENT review: the one important gap flagged (mock `agentEnvService.patch` implementations silently dropping the third `secretKeys` argument, so nothing verified the masking half of the fix) is now closed. `MockState.patchCalls` in both `provision-callback.smoke.test.ts` and `admin-ui.smoke.test.ts` captures `secretKeys`, and each of the three provisioning `patch()` call sites now has an assertion that the right keys are marked secret (e.g. `expect(lastPatch.secretKeys).toContain("SLACK_BOT_TOKEN")`, `expect(slackAppTokenPatch?.secretKeys).toContain("SLACK_APP_TOKEN")`). Verified against the real `agentEnvService.patch()` implementation (`admin/src/agent-envs.ts`), which sets `secret: secretKeys?.has(key) ?? false` on both create and update. Ran the full changed-file test suite locally — 471 pass, 0 fail — and `bun run --filter='@shipwright/admin' typecheck` clean. No new issues.

PR #1323, opened 2026-07-09
feat: delete agent PVC on deprovision

dodizzle:
## Summary
- `KubernetesAgentProvisioner.deprovision()` now deletes the agent's PVC (in addition to the Deployment and Secret it already deleted), swallowing a 404 the same way the existing best-effort deletes do — deleted agents no longer leak 40Gi+ persistent volumes indefinitely.
- `AgentProvisioner.deprovision()` gains an optional `opts?: { slug?: string }` param mirroring `provision()`'s shape, since PVC naming can depend on slug when a `pvcName` template is configured. Both callers (`agents-api.ts` DELETE `/agents/:id`, `admin-ui.ts` POST `/admin/agents/:id/delete` and the Slack-manifest-failure rollback path) now pass the agent's name as `slug`.
- `provision()`'s rollback path is untouched — it intentionally never deletes the PVC (data-safety policy for a failed *retry*, as opposed to a deliberate full agent delete).

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | deprovision() calls k8s.deletePvc() with the correctly-resolved PVC name (default + templated naming), swallows 404 | MET |
| 2 | deprovision() signature gains optional `{ slug?: string }`; both callers pass the agent's name as slug | MET |
| 3 | Unit test: deprovision() calls deletePvc with expected name under default and templated naming | MET |
| 4 | Integration test: already-absent-PVC (404) path is idempotent | MET |
| 5 | Pre-existing test that asserted the PVC survives deprovision is updated (not left contradicting new behavior) | MET |

## Test Plan
- `admin/src/agent-provisioner.unit.test.ts` — new tests assert `deletePvc` is called with the correct name under default naming and under a `pvcName` template + slug.
- `admin/src/agent-provisioner.integration.test.ts` — updated the existing "leaves PVC" test to assert the PVC is now deleted, extended another deprovision test to assert PVC deletion, and added a dedicated idempotent-already-absent-PVC test plus a templated-name deprovision test.
- [x] `bun run --filter='*' typecheck` passes across all workspaces
- [x] `bun test admin/src/agent-provisioner.unit.test.ts admin/src/agent-provisioner.integration.test.ts` passing (integration suite requires `DATABASE_URL_ADMIN_TEST`, present in CI)
- [x] Full `bun test` run: only pre-existing, unrelated failures (confirmed identical on `main`) — a K8s in-cluster CA fallback test and several `installPlugins` tests, both environment-dependent in this sandbox
- [x] Docs refreshed (`docs/agent-api.md`, `docs/agent.md`, `docs/deploy-kubernetes.md`) to reflect PVC deletion on deprovision


dmcaulay (approved):
APPROVE — deprovision() now deletes the agent's PVC (idempotent, respects templated naming), all acceptance criteria met with matching test coverage. One narrow, untested edge case noted below the confidence bar: admin-ui.ts's danger-zone delete tolerates a missing Agent row and falls back to slug: undefined, which could leave a templated-name PVC undeleted on a specific retry sequence (agents-api.ts's DELETE endpoint doesn't have this gap since it 404s first). Worth a follow-up if templated PVC naming is in use, not blocking.

PR #1348, opened 2026-07-09
feat: add deleteAgentFully idempotent orchestration for agent deletion

dodizzle:
## Summary
- Adds `deleteAgentFully()` in `admin/src/agent-deletion.ts` — the single shared orchestration for fully deleting an agent, consolidating what today are two independently-drifting entry points (`DELETE /agents/:id` and the admin-ui "danger zone" handler).
- Orchestrates four idempotent cleanup steps (K8s deprovision, task-store token revocation, chat-service token revocation + thread deletion, optional Slack app deletion) and deletes the Agent DB row **last**, only once every attempted step succeeds — making the whole operation safely retryable after a partial failure.
- Wiring the two existing entry points into this function is out of scope here (tracked separately as ADC-4.1/ADC-4.2); this PR only adds the shared function.

## Acceptance Criteria

| 1 | Happy-path: agentDeleted true, all steps completed, manualStepsRequired correct, Agent row deleted | MET | agent-deletion.unit.test.ts:129-201, 203-242 |
| 2 | Partial-failure: agentDeleted false, row survives, failed names step+error, other steps still run | MET | agent-deletion.unit.test.ts:248-290, 292-343 |
| 3 | Retry: second call completes remaining steps and deletes row, no re-run errors | MET | agent-deletion.unit.test.ts:349-401 |
| 4 | Slack-without-token: manual checklist entry added, agentDeleted not forced false | MET | agent-deletion.unit.test.ts:407-454 |
| 5 | New coverage only, no retired tests | MET | agent-deletion.ts and agent-deletion.unit.test.ts are new files |

## Test Plan
- `bun test admin/src/agent-deletion.unit.test.ts` — 9/9 passing, 100% line coverage on the new module (all collaborators injected as fakes per the repo's no-global-mocking test-isolation rule)
- `bunx biome lint .` — clean
- `bun run --filter='*' typecheck` — clean across all 7 workspaces
- Full `bun test` — only pre-existing, unrelated failure (`kubernetes-client.integration.test.ts`, an environment-dependent in-cluster CA fixture issue that also fails on `main`)


dodizzle:
Shipwright cleanup — resuming task from prior session

PR #1378, opened 2026-07-10
feat: build loop orchestrator with busy-guard and drain-until-dry (WL-3.3)

dmcaulay:
## Summary
- Implements the `shipwright-loop` drain-until-dry orchestrator (`agent/src/loop-orchestrator.ts`): a busy-guarded closure that, per tick, reads the four independent phase toggles (dev-task/review/patch/deploy — never `shipwright-review-patch`), collects each enabled phase's WL-2.2 candidates, calls `selectNextWorkItem` once, and dispatches the correct one-shot command, repeating until nothing remains.
- Extends `CronRunReporter` (`createRun`/`completeRun`/`skipRun`) with an optional `phase` param so every real dispatch reports its own tagged `AgentCronRun` row — the admin API already persists this field (added by WL-3.4). A phase with no candidates never creates a row; `skipRun` is used only when a dispatched command reports `[silent]` (no work found once it actually ran).
- Removes the WL-3.1 placeholder `handleLoopCronRequest` from `loop-cron-classifier.ts` and wires the real orchestrator into `agent/src/index.ts`'s cron-sync loop (constructed once, lazily, and reused across ticks so the busy flag persists).

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Tick no-ops on concurrent busy flag | MET |
| 2 | Four independent phases, never reads/invokes review-patch | MET |
| 3 | Per-phase qualification → merged PR list → selectNextWorkItem called exactly once | MET |
| 4 | Winning item's type/phase tag determines exact command | MET |
| 5 | Busy flag released even on thrown error | MET |
| 6 | Each dispatch reports its own tagged createRun/completeRun pair | MET |
| 7 | Empty-candidate phases create zero rows; skipRun only for dispatched-but-no-work | MET |
| 8 | Fully unit-tested via injected deps, no real Claude/HTTP | MET |

## Test Plan
- `agent/src/loop-orchestrator.unit.test.ts` — 15 unit tests covering: concurrent-invocation no-op, phase-disabled skip, per-phase command routing (all 4 phases), globally-oldest cross-phase FIFO drain, review-patch irrelevance (enabled and disabled), tagged run reporting, idle-tick zero-rows noise guard, `[silent]`-triggered skipRun, busy-flag release on qualification-throw, and busy-flag release + failed-run reporting on a runner-throw during dispatch.
- Full `agent/` suite: 919 pass / 14 fail (pre-existing, environment-only failures — `installPlugins` needs the `claude` CLI binary, `HttpChatTokenReporter` needs a pricing service; reproduced identically on a clean `main` checkout).
- `tsc --noEmit` clean, `biome lint` clean.


dodizzle (approved):
Verdict: APPROVE — Clean, well-tested implementation of the drain-until-dry loop orchestrator; verified the busy-guard, four-phase toggle reading, single selectNextWorkItem call, per-dispatch tagged run reporting, and noise-guard against the source (work-selector.ts, loop-cron-classifier.ts, the four check-*.ts providers, and CronRunReporter). Ran the new 15-test unit suite (all pass), full agent suite (1211 pass / 13 pre-existing environment-only installPlugins failures, reproduced identically on main), tsc --noEmit, and biome lint — all clean.

PR #1432, opened 2026-07-11
feat: make check-docs-freshness.ts and research-docs --auto repo-aware

dmcaulay:
## Summary
- `check-docs-freshness.ts` now iterates every repo under `repos/` (via a new `resolveRepoDirs(workspacePath)` helper in `check-helpers.ts`, sharing scan logic with the existing `resolveRepos`/`resolveAllRepos`) instead of assuming `process.cwd()` is a single target repo — fixes the `shipwright-docs-freshness` cron being a silent no-op when the agent's cwd is the workspace root (not a git repo).
- Each repo's own `state/docs-last-synced.json` anchor is read/written independently; repos with no `docs/` directory are skipped cleanly with no spurious anchor I/O.
- `research-docs.md`'s Auto Mode documents how per-repo iteration is driven: parses the repo list from the precheck's stdout (which becomes the cron prompt) when available, falling back to iterating `repos/*` directly for manual runs; Steps A1-A8 now run per-repo, Step A9 prints one aggregated summary.
- `docs/agent.md`'s cron table entry refreshed to describe the new multi-repo behavior.

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | check-docs-freshness.ts iterates repos/* instead of a single process.cwd() | MET (via `resolveRepoDirs`, same resolution pattern as sibling prechecks' `resolveAllRepos`) |
| 2 | Each repo's own state/docs-last-synced.json read/written independently | MET |
| 3 | Repos with no docs/ directory skipped cleanly | MET |
| 4 | research-docs.md documents per-repo iteration driver | MET |
| 5 | Existing single-repo behavior unchanged | MET |
| 6 | Unit test coverage mirrors check-deploy.ts's/check-review.ts's resolveAllRepos pattern | MET |

## Test Plan
- `check-docs-freshness.test.ts`: preserved all single-repo assertions, added multi-repo iteration, independent per-repo anchors, clean no-docs skip, all-no-docs, one-repo-git-failure-doesn't-block-another, first-run, and repo-identity-in-output tests.
- `check-helpers.unit.test.ts`: added `resolveRepoDirs` coverage mirroring `resolveRepos`/`resolveAllRepos` tests.
- `research-docs.unit.test.ts`: added assertions for the new per-repo iteration language.
- [x] `task typecheck`, `task lint`, `task check-strings`, `task check-version-sync`, `task check-config-docs` pass
- [x] `bun test plugins/shipwright/` — 781/781 pass
[...]

PR #1433, opened 2026-07-11
fix: exempt third-party review findings addressed via author reply in check-patch

dmcaulay:
## Summary
- `check-patch.ts`'s `hasUnaddressedFindings`/`hasMergeOnlyStaleFindings` treated a third-party reviewer's non-empty review body as an unaddressed finding forever, even after all inline threads were resolved — because `updatePullRequestReview` (the existing self-review "Verdict: APPROVE" workaround) can only edit a review's own author's body, not a third party's (concrete case: shipwright PR #1432, `dodizzle`'s review).
- Adds `isAddressedByAuthorReply`: a qualifying review's body-only finding is now treated as addressed when all its inline threads are resolved AND the PR author has posted a subsequent PR-level reply after the review's `submittedAt` — the only available signal for a third-party review that can't be body-edited.
- Deliberately does **not** use `dismissPullRequestReview` (a more consequential, untested-permissions action flagged by the task as needing separate research before adoption).
- Syncs `patch.md` Step 3a with the same exclusion, per the plugin's Precheck Contract (skill and precheck must not diverge).

## Acceptance Criteria
| 1 | Third-party review body finding resolves once threads resolved + PR-author reply posted (PR #1432 scenario) | MET | `isAddressedByAuthorReply()` wired into `hasUnaddressedFindings`; test "mirrors PR #1432" exits 1 |
| 2 | Does not use `dismissPullRequestReview` | MET | No references anywhere in the diff |
| 3 | Genuinely unaddressed findings still trigger exit 0 (no reply / stale reply / unresolved threads / non-author reply) | MET | 4 regression-guard tests all exit 0 as expected |
| 4 | `hasMergeOnlyStaleFindings` gets the same treatment | MET | Same exemption applied; 2 dedicated tests |
| 5 | Tests land with the code (unit layer) | MET | 8 new tests in `check-patch.test.ts`, 42/42 passing |
| 6 | `patch.md` skill doc kept in sync with precheck logic | MET | Step 3a updated with matching exclusion clause |

## Test Plan
- `bun test plugins/shipwright/scripts/check-patch.test.ts` — 42/42 passing (8 new)
- `bun test plugins/shipwright/` — full plugin suite, 767/767 passing
- `bunx biome lint .` — clean
- `bunx tsc --noEmit` (plugins/shipwright) — clean

### DIFF
--- a/plugins/shipwright/commands/patch.md
+++ b/plugins/shipwright/commands/patch.md
@@ -144,12 +144,13 @@ A PR has **unaddressed findings** when ANY of the following are true:
 - At least one inline thread has `isResolved == false`
 - At least one review with `state == "COMMENTED"` or `state == "CHANGES_REQUESTED"` has a
   non-empty `body` (a review body without matching inline threads is itself a finding),
-  excluding clean-APPROVE reviews (see below)
+  excluding clean-APPROVE reviews (see below) and reviews addressed via a subsequent author
+  reply (see below)
 
 A PR has **no findings** (skip it) when ALL of the following are true:
 - All inline threads are resolved (`isResolved == true` for every thread)
 - No COMMENTED or CHANGES_REQUESTED review has a non-empty body, other than clean-APPROVE
-  reviews (see below)
+  reviews (see below) and reviews addressed via a subsequent author reply (see below)
 
 **Clean-APPROVE exclusion**: A review is excluded from the body check above when its body is
 a clean APPROVE verdict, matched either by:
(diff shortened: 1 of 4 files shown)


dmcaulay (commented):
Verdict: APPROVE — Well-scoped fix for check-patch.ts's perpetual-trigger bug on third-party review findings (traces directly to the PR #1432 false-positive case). isAddressedByAuthorReply is correctly scoped to the PR author's own replies, still requires all inline threads resolved, and deliberately avoids the more consequential dismissPullRequestReview per the task's own guidance. patch.md kept in sync with the precheck logic. 47/47 tests pass, CI green. No blocking concerns.

PR #1467, opened 2026-07-13
fix: pass PINNED_LINES via env to avoid backtick command-substitution crash

dmcaulay:
## Summary
- Fixed `auto-bump-chart.yml`'s "Open bump PR" step, where `PINNED_LINES="${{ steps.pin.outputs.pinned_lines }}"` spliced backtick-containing markdown directly into a double-quoted bash string. GitHub Actions injects the raw expression text into the script source before bash parses it, so bash treated the backticks as command substitution and crashed with "command not found" (exit 127) before `gh pr create` ever ran — this is what left the chart-v1.6.341 bump branch pushed with no PR opened.
- Fix: route `PINNED_LINES` through the step's `env:` block (same pattern already used for `GH_TOKEN`) instead of interpolating it into the script body. The heredoc then references `${PINNED_LINES}` as a single runtime parameter expansion, which is never re-parsed for command substitution.
- Mirrored the same PR-body-building logic into `test-auto-bump-chart.sh` as a new `build_pr_body()` function, and added a regression test that feeds it backtick-laden `pinned_lines` content, asserting it doesn't crash, renders the backticks literally, and produces no "command not found" artifacts.

## Acceptance Criteria

| 1 | PINNED_LINES constructed without exposing backtick content to bash expansion in a double-quoted string | MET | Moved to `env: PINNED_LINES: ${{ steps.pin.outputs.pinned_lines }}`, referenced via `${PINNED_LINES}` param expansion |
| 2 | test-auto-bump-chart.sh's mirrored logic updated identically | MET | New `build_pr_body()` mirrors the exact heredoc, accepting pinned_lines as a real function param |
| 3 | Regression test with backticks in pinned-lines content, asserting no error/command substitution | MET | New `=== build_pr_body tests ===` section: exit-0, literal backtick rendering, and no "command not found" assertions |
| 4 | Test decision: unit-layer only, extends existing suite | MET | Extends `test-auto-bump-chart.sh` in place; no existing tests retired |

## Test Plan
- Reproduced the original bug locally (`PINNED_LINES="..."` with literal backticks → `command not found`, exit 127) to confirm the fix mechanism.
- `bash .github/workflows/test-auto-bump-chart.sh` — 61/61 tests passing, including the 4 new regression tests.
- Validated workflow YAML parses correctly after the edit.


dodizzle (approved):
Verdict: APPROVE — Minimal, correctly root-caused fix (routes PINNED_LINES through env: like the existing GH_TOKEN pattern instead of splicing a backtick-laden GHA expression into a double-quoted bash string) with a verified regression test (61/61 passing) and green CI. No blocking concerns.

PR #1470, opened 2026-07-13
fix: force NODE_ENV=test in workspace test invocations

dmcaulay:
## Summary
- Prefix `NODE_ENV=test` on the two full-workspace `bun test` invocation sites (Taskfile.yml's `test`/`test:coverage` tasks and package.json's root `test` script) so `buildSentryInitOptions()`'s test guard in `lib/sentry.ts` holds even when a parent process already exports `NODE_ENV` (e.g. `NODE_ENV=production`), instead of relying on Bun's env-inheritance-dependent auto-set behavior.
- Add `scripts/test-env-guard.unit.test.ts` as net-new config-drift regression coverage, asserting every full-workspace `bun test` invocation in both files carries the prefix.
- Refresh `docs/observability.md`'s two references to the guard being "auto-set by `bun test`" to reflect the explicit prefix.

## Acceptance Criteria
| 1 | Taskfile.yml's `test` task runs `NODE_ENV=test bun test` | MET | Taskfile.yml:27 |
| 2 | Taskfile.yml's `test:coverage` task runs `NODE_ENV=test bun test --coverage --coverage-reporter=lcov` | MET | Taskfile.yml:32 |
| 3 | package.json's root `test` script is `NODE_ENV=test bun test` | MET | package.json:16 |
| 4 | New unit test asserts every `bun test` invocation in Taskfile.yml/package.json is NODE_ENV=test-prefixed; no existing tests retired | MET | scripts/test-env-guard.unit.test.ts (new, 90 lines) |

## Test Plan
- TDD red-green-refactor: wrote `scripts/test-env-guard.unit.test.ts` first, confirmed it failed against the unfixed Taskfile.yml/package.json, then applied the two-file fix and confirmed green.
- `bunx biome lint .` — clean
- `bun run --filter='*' --sequential typecheck` — clean across all 7 workspaces
- `bun test` — 3860 pass / 15 fail; the 15 failures are pre-existing (metrics/admin full-suite flakes), reproduce identically on unmodified `main`, unrelated to this diff
- Coverage gate (`task test:coverage` + `scripts/check-coverage.ts`): aggregate 78.99% lines, below the 80% threshold — pre-existing sandbox limitation (task-store integration tests require a live Postgres DB unavailable in this environment), not attributable to this change

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -24,12 +24,12 @@ tasks:
   test:
     desc: Run all tests
     cmds:
-      - bun test
+      - NODE_ENV=test bun test
 
   test:coverage:
     desc: Run all tests with coverage, then gate on aggregate 80/80 lines/functions (scripts/check-coverage.ts)
     cmds:
-      - bun test --coverage --coverage-reporter=lcov
+      - NODE_ENV=test bun test --coverage --coverage-reporter=lcov
       - bun run scripts/check-coverage.ts
 
   check-strings:
--- a/docs/observability.md
+++ b/docs/observability.md
@@ -6,7 +6,7 @@ Everything below this line through [Configuration reference](#configuration-refe
 
 ## Overview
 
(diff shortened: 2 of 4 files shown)


dodizzle (approved):
Verdict: APPROVE — Correctly scoped fix (prefixes NODE_ENV=test on both real full-workspace bun test invocation sites so the Sentry test guard holds regardless of inherited shell env), with matching regression coverage and no unintended change to other test behavior. Verified per-workspace package.json test scripts left unprefixed aren't a real gap since no test file imports the service entrypoints that call initSentry(). CI green at the reviewed commit.

PR #1512, opened 2026-07-14
test: fix setup.unit.test.ts marketplace-dir isolation leak (T-002)

dmcaulay:
## Summary
- Fixes a test isolation leak in `agent/src/setup.unit.test.ts`: 4 of 6 `installPlugins` calls in the "local marketplace" describe block omitted the `extraMarketplaceDirs` argument, silently falling through to `discoverBakedMarketplaces(BAKED_MARKETPLACES_ROOT)` — a real `existsSync`/`readdirSync` against `/opt/shipwright/marketplaces` on disk.
- Adds an explicit `NO_EXTRA_MARKETPLACES` constant and passes it at every call site in that describe block, plus a new regression test asserting deterministic behavior regardless of what's actually on disk at `BAKED_MARKETPLACES_ROOT`.
- No production behavior change — `agent/src/setup.ts`'s `installPlugins` already had the injection seam (`extraMarketplaceDirs`); only the test file needed to consistently use it.

## Acceptance Criteria
| 1 | installPlugins accepts an injected marketplace-directory path instead of reading /opt/shipwright/marketplaces/ directly | MET | Pre-existing `extraMarketplaceDirs` param (setup.ts:305, unchanged); confirmed no production call site relies on removing it |
| 2 | setup.unit.test.ts no longer depends on ambient sandbox marketplace state | MET | All 6 `installPlugins` calls in the local-marketplace describe block now explicitly pass `NO_EXTRA_MARKETPLACES`; new regression test added |
| 3 | Verification command `bun test agent/src/setup.unit.test.ts` passes with all cases green | MET | 10 pass, 0 fail, 18 expect() calls |

## Test Plan
- `bun test agent/src/setup.unit.test.ts` — 10 pass, 0 fail (was 5 pass / 4 fail before the fix — this sandbox has a real baked marketplace at `/opt/shipwright/marketplaces/vitals-os` that was leaking into the previously-unfixed tests).
- `task lint` — clean, no fixes needed.
- `task typecheck` — all 7 workspaces pass.
- Full `agent/` suite: 903 pass / 10 fail — all 10 failures confirmed pre-existing on clean `main` (unrelated env/network dependent integration tests), not introduced by this change.

### DIFF
--- a/agent/src/setup.unit.test.ts
+++ b/agent/src/setup.unit.test.ts
@@ -7,6 +7,13 @@
 import { describe, expect, it } from "bun:test";
 import { discoverBakedMarketplaces, installPlugins } from "./setup.ts";
 
+// Explicit empty extra-marketplace list — every call in this describe block
+// must pass this (or an intentional array) instead of omitting the arg.
+// Omitting it falls through to discoverBakedMarketplaces(BAKED_MARKETPLACES_ROOT),
+// which reads the REAL /opt/shipwright/marketplaces directory on disk and makes
+// these tests depend on ambient sandbox/image state.
+const NO_EXTRA_MARKETPLACES: string[] = [];
+
 describe("installPlugins — local marketplace", () => {
   it("registers the marketplace before installing plugins", async () => {
     const calls: Array<{ cmd: string; args: string[] }> = [];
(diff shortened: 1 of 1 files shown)


dodizzle (approved):
Verdict: APPROVE — Test-only fix verifiably closes the marketplace-dir isolation leak at every installPlugins call site in the describe block; new regression test correctly encodes the contract; no production code touched.

PR #1515, opened 2026-07-14
test: fix K8s CA path and ANTHROPIC_MODEL isolation leaks (T-004)

dmcaulay:
## Summary
- Make `HttpKubernetesClient`'s default CA-loading path injectable via a new `saDir` option, so the "missing default CA path" integration test no longer depends on whatever happens to be mounted at the real in-cluster SA path in the sandbox
- Inject `liveClaudeConfig` via `setLiveClaudeConfig()` in `chat-token-reporter.integration.test.ts`'s `beforeEach`/`afterEach`, so the pricing-fallback test no longer depends on the ambient `ANTHROPIC_MODEL` env var (and doesn't leak state into sibling suites)
- Refresh `docs/agent.md` to document the `liveClaudeConfig`/`setLiveClaudeConfig()` fallback behavior

## Acceptance Criteria
| 1 | kubernetes-client.ts's CA-loading path is injectable so the 'missing default CA path' test doesn't depend on the sandbox's real filesystem | MET | New `saDir` option added to `HttpKubernetesClientOpts`, threaded through `readCa(caPath, saDir)`; test now injects a fresh `mkdtempSync` empty dir instead of relying on ambient `/var/run/secrets/...` |
| 2 | chat-token-reporter.integration.test.ts injects liveClaudeConfig via setLiveClaudeConfig() in beforeEach | MET | `beforeEach` now calls `setLiveClaudeConfig({ model: "claude-sonnet-4-6" })`; `afterEach` restores original value to prevent cross-suite leakage |
| 3 | Verification command passes regardless of ambient sandbox env | MET | `NODE_ENV=test bun test admin/src/kubernetes-client.integration.test.ts agent/src/chat-token-reporter.integration.test.ts` → 57 pass, 0 fail |

## Test Plan
- [x] Reproduced both failures pre-fix (this sandbox is a real k8s pod with a real CA mounted at the default SA path, and `ANTHROPIC_MODEL` is set to a model not in the pricing table — both concretely leaked into the tests before this fix)
- [x] `bun test admin/src/kubernetes-client.integration.test.ts agent/src/chat-token-reporter.integration.test.ts` — 57 pass, 0 fail
- [x] `task lint`, typecheck for `@shipwright/admin` and `@shipwright/agent` — clean
- [x] Full `bun test` run — only pre-existing unrelated `installPlugins` failures (verified present on `main` too)


dodizzle (approved):
Verdict: APPROVE — Clean, narrowly-scoped test-isolation fix; saDir is purely additive and the liveClaudeConfig injection uses the repo's established mutable-config-ref pattern with proper afterEach restoration.

PR #1522, opened 2026-07-14
fix: remove incorrect local dependency re-check from selectNextWorkItem

dmcaulay:
## Summary
- Removes `isTaskUnblocked()`/`tasksById` local dependency re-validation from `selectNextWorkItem()` — the `tasks` list it received is already the ready-only candidate set (task-store's `?ready=true` query, itself server-side dependency-satisfaction filtered), so the local re-check could never find a satisfied dependency in that same list and silently dropped every ready task that had one.
- Task-side selection now trusts `status === "pending"` + `createdAt` age comparison only, matching the trust level PR candidates already get in the loop below.
- Replaces the 6 unit tests that locked in the old (incorrect) re-validation contract with a single regression-guard test asserting a ready task with a dependency absent from the candidate list is still selected.
- Refreshes `docs/agent.md`'s `loop-orchestrator.ts` entry to reflect the corrected dependency-filtering model.

## Acceptance Criteria
| 1 | Task-side loop no longer calls `isTaskUnblocked`/builds `tasksById`; selection purely on age comparison | MET | `work-selector.ts` task loop now mirrors the PR loop's simple age comparison |
| 2 | `isTaskUnblocked()` deleted entirely, zero remaining references | MET | Function + `TERMINAL_TASK_STATUSES` removed; repo-wide grep confirms zero references |
| 3 | 6 obsolete tests deleted; new regression test added for satisfied-dep-absent-from-list scenario | MET | 6 tests removed, 1 regression test added, adjacent test fixed to drop its now-invalid dependency-blocking assertion |

## Test Plan
- `bun test agent/src/work-selector.unit.test.ts` — 12/12 pass, 100% line/function coverage on `work-selector.ts`
- Full `agent/` suite: 899/908 pass; the 9 failures are pre-existing `installPlugins` sandbox flakes in `setup.integration.test.ts`, reproduced identically on a clean `main` checkout — unrelated to this change
- `bun run --filter='*' --sequential typecheck` — clean across all 7 packages
- `bunx biome lint .` — clean


dodizzle (approved):
Verdict: APPROVE — Verified, well-tested bug fix removing a redundant client-side dependency re-check that was strictly stricter than (and thus in conflict with) the server-side ready=true filter. No dangling references, no API surface change, full alignment with acceptance criteria.

PR #1513, opened 2026-07-14
feat: thread selected item id into loop-orchestrator dispatch

dmcaulay:
## Summary
- `dispatch(phase)` in `agent/src/loop-orchestrator.ts` now accepts `(phase, itemId)` and appends the winning `WorkItem`'s id to the runner message (`${PHASE_COMMANDS[phase]} ${itemId}`), so the loop and the invoked one-shot command agree on which item to work.
- `runLoopTick` derives `itemId` alongside its existing `phase` derivation: `item.task.id` for a task, `item.pr.id` (already `org/repo#number`-formatted) for a PR.
- No changes to `selectNextWorkItem`, `work-selector.ts`, or any `check-*.ts` candidate-collection function — purely additive threading through the existing dispatch call site.
- `docs/agent.md` refreshed to reflect that dispatch commands now include the winning item's id.

## Acceptance Criteria

| # | Criterion | Status | Evidence |
|---|-----------|--------|----------|
| 1 | dispatch call always includes the winning item's id appended to the phase's command string | MET | `runLoopTick` computes `itemId` and passes it to `dispatch(phase, itemId)`; `dispatch` builds `` `${PHASE_COMMANDS[phase]} ${itemId}` `` |
| 2 | Task items pass task.id; PR items pass pr.id (no new fields needed) | MET | `item.type === "task" ? item.task.id : item.pr.id` — reuses existing `id` fields on `WorkTaskCandidate`/`WorkPrCandidate` |
| 3 | dispatch's shape changes only by threading the item through — no change to selectNextWorkItem/work-selector.ts/check-*.ts | MET | Only `agent/src/loop-orchestrator.ts` and its test file changed (plus docs) |
| 4 | Extend loop-orchestrator.unit.test.ts's dispatch-path fixtures to assert item id in constructed message, no tests retired | MET | 7 existing tests updated to assert `"<command> <itemId>"`; test count unchanged (18 tests, all passing) |

## Test Plan
- `cd agent && bun test loop-orchestrator.unit.test.ts` — 18 pass, 0 fail
- `bunx biome lint` on changed files — clean
- `cd agent && bunx tsc --noEmit` — clean
- [x] Pre-commit checks passing

Generated with [Claude Code](https://claude.com/claude-code)

PR #1548, opened 2026-07-14
feat: convert dev-task.md to explicit-target-only (WLS-3.1)

dmcaulay:
## Summary
- Convert `dev-task.md` to explicit-target-only: a task-id argument is now required, and the command no longer self-selects work via the interrupted-task resume check or the `GET /tasks?ready=true` ready-queue scan.
- No-argument invocation responds `[silent]` and stops immediately, with zero task-store queries.
- Given an explicit task id, Step 1 fetches `GET /tasks/{task-id}` directly and, for a `pending` task, validates every dependency against the same satisfaction rules the task store applies for `?ready=true` before proceeding to claim.
- A claim `409` now responds `[silent]` and stops — the "loop back to Step 1 and pick the next ready task" retry is removed, since this command targets exactly one explicit task id.
- Updated `dev-task.test.ts` to drop the old self-search fixtures and add coverage for the required-argument contract, the no-argument `[silent]` path, the explicit fetch/validate/claim flow, and the claim-409 exit-clean path.

This lands after WLS-2.1 (merged, PR #1513), which already threads the winning task id into the loop-orchestrator's `/shipwright:dev-task` dispatch — confirmed via `agent/src/loop-orchestrator.ts` — so this change does not go silently dead.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | No-argument invocation → `[silent]`, zero task-store queries | MET |
| 2 | Fetches `GET /tasks/{id}` directly; validates pending status + satisfied dependencies before claiming | MET |
| 3 | Claim 409 → `[silent]`, stop, no retry against a different task | MET |
| 4 | `argument-hint` and prose both state the argument is required | MET |
| 5 | Test suite updated: retired self-search fixtures, added no-arg and 409-exit-clean cases | MET |

## Test Plan
- `NODE_ENV=test bun test plugins/shipwright/commands/dev-task.test.ts plugins/shipwright/commands/dev-task.unit.test.ts` — 39/39 pass
- `NODE_ENV=test bun test plugins/shipwright` — 731/731 pass (full plugin suite, no regressions)
- `bun run lint` — clean
- `bun run typecheck` — clean across all workspaces


dodizzle (commented):
Verdict: COMMENT — dev-task.md and its tests correctly implement the WLS-3.1 explicit-target-only contract (required task-id argument, fetch/validate/dependency-check/claim flow, clean 409 stop), but this PR doesn't update the live `shipwright-dev-task` system cron's precheck (`plugins/shipwright/scripts/check-dev-task.ts:81`). That precheck's output still reads "Pick the next ready task from the task store and execute via /shipwright:dev-task" with no task id — since `agent/src/cron-handler.ts` replaces the cron prompt entirely with the precheck's stdout, and a no-argument `/shipwright:dev-task` now silently no-ops by design (AC #1), this risks silently stopping the queue drain for any agent still on the legacy `shipwright-dev-task` cron (as opposed to the newer loop-orchestrator path, which already threads an explicit id and is unaffected). The plugin's own CLAUDE.md requires flagging exactly this kind of system-cron migration risk before shipping. Recommend updating check-dev-task.ts's output to include a fetched task id (or migrating/retiring that cron) before or alongside this merge.

dmcaulay:
Addressed @dodizzle's finding on the `shipwright-dev-task` system cron precheck.

Verified the premise: `check-dev-task.ts` (plugin script) was still handing off `"Pick the next ready task from the task store and execute via /shipwright:dev-task"` with no task id, and `agent/src/cron-handler.ts` does replace the cron's prompt entirely with the precheck's stdout (`prompt = output` at line ~267). Combined with this PR's AC #1, that means a bare `/shipwright:dev-task` invocation from the legacy `shipwright-dev-task` cron would silently no-op — confirmed `shipwright-dev-task` is still `enabled: true` in `admin/src/system-crons.ts`, so this was a live risk, not a hypothetical one.

Fix (`fdcff1f4`): updated `plugins/shipwright/scripts/check-dev-task.ts` to select the oldest ready task by `addedAt` and emit `Execute /shipwright:dev-task {id} — the next ready task from the task store.` instead of the id-less prompt. Selection mirrors the FIFO ordering `agent/src/work-selector.ts`'s `selectNextWorkItem()` already uses for the loop-orchestrator path (oldest `addedAt`/`createdAt` first), and the resulting handoff matches how `agent/src/loop-orchestrator.ts` dispatches `/shipwright:dev-task {itemId}` — so both the legacy cron and the loop-orchestrator path now agree on selection order and always pass an explicit id. Added two unit tests covering the id-in-prompt behavior and multi-task FIFO ordering; all 22 tests in `check-dev-task.unit.test.ts` pass, along with the full `bun run lint` and `bun run test` suites (3899 pass / 0 fail).

dodizzle (approved):
Verdict: APPROVE — Explicit-target-only contract correctly implemented (AC #1-5 met), the prior review's live-cron precheck risk is verified fixed in this commit, and test coverage for the new contract is solid. One non-blocking suggestion: `TESTING.md`'s cross-agent leak-protection QA scenario (lines ~1130-1155) instructs invoking `/shipwright:dev-task` with no arguments to test resume/assignee-filtering — that path is now an immediate `[silent]` no-op with zero queries, so the scenario can no longer be reproduced as written. Worth a follow-up doc update, not blocking this PR.

PR #1558, opened 2026-07-15
feat: add explicit-target mode to patch.md, remove self-scan

dmcaulay:
## Summary
- `patch.md` now requires an explicit `org/repo#number` argument-hint (`<org/repo#number>`) instead of the optional `[org/repo#number]`; a no-argument invocation responds `[silent]` and stops (new Step 0), with no more scan of all own open PRs via `gh pr list --author`.
- Step 2 ("Resolve Target PR") unconditionally fetches the single passed PR via `gh pr view`; the three-condition qualification logic (unaddressed review findings, merge conflicts, failing CI) and the pre-work claim-409 handling in Steps 4/5/6 are unchanged, now evaluated against that one PR.
- `docs/agent.md`'s cron table marks the standalone `shipwright-patch` cron as deprecated in favor of `shipwright-loop`, which already threads explicit PR ids per WLS-2.1.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | patch.md requires org/repo#number; no-argument invocation responds [silent] with no GitHub scan | MET |
| 2 | Three-condition qualification logic preserved, evaluated against the one passed PR | MET |
| 3 | Claim 409 continues to respond cleanly, no regression | MET |
| 4 | argument-hint and prose both state the argument is required | MET |
| 5 | Test decision: retire scan fixtures, add explicit-target fixtures, keep per-condition coverage | MET |

## Test Plan
- `patch.content.test.ts` gained a new 5-test describe block asserting the required argument-hint, required prose, `[silent]` no-arg stop, removal of the multi-repo scan, and the single-PR `gh pr view` fetch in Step 2.
- All 11 pre-existing claim-409/claim-lock (CLM-2.1) tests in the same file are untouched and still pass, since that logic is unchanged.
- [x] `bun test plugins/shipwright/` — 728 pass, 0 fail
- [x] `tsc --noEmit` (plugin package) — clean


dodizzle (commented):
Verdict: COMMENT — This change silently breaks the default-enabled `shipwright-review-patch` cron: it dispatches `/shipwright:patch` with no arguments (see review-patch.md Step 3b), and patch.md's new Step 0 now goes `[silent]` on empty `$ARGUMENTS`. After this merges, the patch phase of `review-patch` will always no-op, even when check-patch.ts correctly detects real work — with no error surfaced. `review-patch.md` needs to pass an explicit target (or this needs to be gated until review-patch.md is retired in WLS-4.1). The patch.md rewrite itself is clean and well-tested.

dmcaulay (commented):
Verdict: APPROVE — The follow-up fix correctly and completely addresses the prior review finding (review-patch.md's Step 3b no longer dispatches a bare `/shipwright:patch`; it now captures and forwards check-patch.ts's literal `/shipwright:patch {org}/{repo}#{number}` output, which is now emitted on all 4 exit-0 paths). Traced the dispatch mechanics end-to-end (cron-handler.ts's `prompt = output` precheck override, loop-orchestrator's already-compatible `candidateId()`-based ids) — no remaining breaking-change gaps. No new issues found.

PR #1565, opened 2026-07-15
feat(agent): make claude session timeout configurable via SHIPWRIGHT_CLAUDE_TIMEOUT_MS

EugeneTrapeznikov:
## Problem

The `claude -p` session hard timeout is hardcoded to 30 minutes (`agent/src/claude.ts`, `timeoutMs: number = 30 * 60 * 1000`) with no override — no caller passes the parameter and no env var reads it.

I hit it when a dev-task worker finished its implementation, opened its PR, then polled CI — and was SIGKILLed at exactly 1800s mid-poll, before it could mark its task complete. The task stayed `in_progress`, the stale-claim reaper reset it to `pending` after the claim TTL, and the loop dispatched a full duplicate run of an already-complete task (repeating every ~45 min until an operator intervened with `POST /tasks/:id/complete`). A tunable timeout lets operators size sessions to their tasks.

## Changes

Makes the `claude -p` session hard timeout configurable via a new `SHIPWRIGHT_CLAUDE_TIMEOUT_MS` env var. The value is parsed and validated in the central `agent/src/config.ts` via a new `positiveIntMs` helper (positive-integer only; unset, non-numeric, non-integer, zero, or negative all fall back to the existing `1_800_000` default, so **default behavior is unchanged**) and threaded into both `createRunClaude` call sites in `index.ts` — the task worker and the chat runner — so it uniformly governs every Claude session. Documented in `docs/configuration.md`, the rendered site config reference (`site/src/data/config-vars.ts`), and `state/dev-agent.env.example`, including the operational caveat to raise `SHIPWRIGHT_TASK_STORE_CLAIM_TTL_MS` in step (preserving the ~5-min buffer above the timeout, matching the invariant documented in `stale-claim-reaper.ts`) to keep the stale-claim reaper from re-dispatching duplicates. Adds 5 config unit tests covering default, valid override, and each rejection path.

## Verification

Beyond unit tests (full agent suite: 919 pass with the var unset — behavior identical to main), we verified empirically against the real code with a stub `claude` binary on PATH:

- `SHIPWRIGHT_CLAUDE_TIMEOUT_MS=3000` + a 60s-sleeping stub → `ClaudeTimeoutError` at 3004ms.
- `=20000` + a stub exiting at 5s → returns normally at 5017ms (no early fire).
- Config resolution matrix (fresh process per case): unset/"abc"/"-5"/"0"/"1.5" → 1800000; "600000" → 600000.
- Same kill test driven with the chat-runner call shape → 2006ms kill (both call sites thread the value).
- Inside a container built from the agent Dockerfile: env var read from container env, kill at 4021ms with `=4000`.

### DIFF
--- a/agent/src/config.ts
+++ b/agent/src/config.ts
@@ -1,7 +1,24 @@
 import { join } from "node:path";
 
+// Default hard timeout for a single `claude -p` session — mirrors the default
+// in createRunClaude (agent/src/claude.ts). Kept in sync so behavior is
+// unchanged when SHIPWRIGHT_CLAUDE_TIMEOUT_MS is unset.
+const DEFAULT_CLAUDE_TIMEOUT_MS = 30 * 60 * 1000; // 30 minutes
+
 function optional(key: string): string | undefined {
   return process.env[key];
+}
+
+/**
+ * Reads a positive-integer millisecond env var, falling back to `fallback`
+ * when the value is unset, non-numeric, non-integer, zero, or negative — so a
+ * malformed override never silently disables or shortens the timeout.
+ */
+function positiveIntMs(key: string, fallback: number): number {
+  const raw = optional(key);
+  if (raw === undefined) return fallback;
+  const parsed = Number(raw);
+  return Number.isInteger(parsed) && parsed > 0 ? parsed : fallback;
 }
 
 function buildConfig(agentHome: string) {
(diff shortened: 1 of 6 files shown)


dmcaulay (approved):
Verdict: APPROVE — Small, additive, backward-compatible config change; verified both createRunClaude call sites thread the new timeoutMs arg into the correct positional slot, positiveIntMs validation is defensive, and docs/tests are updated in lockstep. No blocking issues.

dmcaulay:
i've also queued up some work to make sure the reaper will not reclaim tasks with a higher timeout.

PR #1562, opened 2026-07-15
refactor: strip unused status/branch/dependencies fields from WorkTaskCandidate

dmcaulay:
## Summary
- Removed `status`, `branch`, and `dependencies` from `WorkTaskCandidate` (`agent/src/work-selector.ts`) — only `id` and `createdAt` remain, the fields actually used for age-based selection.
- Also removed the now-dead `task.status !== "pending"` filter inside `selectNextWorkItem()`: task-store's `?ready=true` endpoint already guarantees `status === "pending"` and dependency satisfaction server-side, mirroring LPF-1.4's removal of the redundant local dependency re-check. This was necessary for the interface change to typecheck, and matches the task's stated premise.
- Updated `toWorkTaskCandidate()` in `check-dev-task.ts` to only map `id`/`createdAt`.
- Updated `work-selector.unit.test.ts`, `check-dev-task.unit.test.ts`, and `loop-orchestrator.unit.test.ts` fixtures/tests to drop references to the removed fields.
- Refreshed `docs/agent.md`: fixed a stale "checks dependencies" description on `check-dev-task.ts`, added a dedicated `work-selector.ts` Key Files entry, and updated `loop-orchestrator.ts`'s description to reflect the refactored responsibility boundary.

## Acceptance Criteria

| # | Criterion | Status | Evidence |
|---|-----------|--------|----------|
| 1 | WorkTaskCandidate no longer declares status, branch, or dependencies — only id/createdAt remain | MET | `agent/src/work-selector.ts` interface reduced to `id`/`createdAt` |
| 2 | toWorkTaskCandidate() only maps remaining fields; no unused local variables | MET | `check-dev-task.ts` mapper returns `{ id, createdAt }` only |
| 3 | Delete "candidate carries branch and dependencies through" test; remaining tests don't reference removed fields | MET | Test deleted; `work-selector.unit.test.ts`/`loop-orchestrator.unit.test.ts` fixtures updated |

## Test Plan
- `bun test src/work-selector.unit.test.ts src/check-dev-task.unit.test.ts src/loop-orchestrator.unit.test.ts` — 40/40 pass
- Full `agent/` suite — 912/912 pass
- `tsc --noEmit` across all workspace packages — clean
- `bunx biome lint .` — clean
- [x] Pre-commit checks passing

Generated with [Claude Code](https://claude.com/claude-code)

PR #1575, opened 2026-07-15
fix: consume /prs ready-filter in loop candidate collectors, drop selector's claim check

dmcaulay:
## Summary
- `check-patch.ts` and `check-deploy.ts` now query the task-store's `/prs` endpoint with `ready=true` (LPF-2.1's filter), so a resolved `null` record doubles as a claim gate — by patch/deploy phase a record should always exist, so `null` unambiguously means "currently claimed"
- `check-review.ts` keeps a non-`ready`-filtered query (it needs to distinguish "no record yet" from "claimed record") but adds an explicit inline `record.claimedBy != null` exclusion, achieving the same outcome
- `selectNextWorkItem` (work-selector.ts) no longer checks `pr.claimedBy` — `WorkPrCandidate` no longer carries the field — since all three collectors now only ever return unclaimed candidates. The selector is now a pure oldest-of-two-already-filtered-lists comparison, matching its docstring's stated intent
- `docs/agent.md` refreshed to document the new `ready=true` filter and check-review's exception

## Acceptance Criteria
| 1 | check-review.ts/check-patch.ts/check-deploy.ts's PR-fetching calls now request only unclaimed PR records via LPF-2.1's filter | MET | check-patch.ts and check-deploy.ts's `buildProductionDeps` pass `ready: true` to `createPrRecordQuery`; check-review.ts performs an equivalent inline `record.claimedBy != null` check since it needs to distinguish "no record" from "claimed" |
| 2 | selectNextWorkItem no longer checks pr.claimedBy — WorkPrCandidate's claimedBy field removed if nothing else reads it | MET | `claimedBy` removed from `WorkPrCandidate` type and the `if (pr.claimedBy != null) continue;` line in work-selector.ts; grep confirms it's retained only in check-review.ts's PrRecord (which still reads it) |
| 3 | Test decision: update work-selector.unit.test.ts and check-review/patch/deploy suites to reflect the new contract | MET | work-selector.unit.test.ts's claim-exclusion tests replaced with age-only-selection tests; new regression tests added in check-deploy/patch (null-as-claimed via ready=true) and check-review (inline claimedBy exclusion) |

## Test Plan
- `NODE_ENV=test bun test agent` — 1227 pass, 0 fail
- `bunx biome lint .` — clean
- Coverage gate (aggregate 80/80) fails in-sandbox due to missing test DB (skips integration tests) — reproduces identically on clean main (47.81% vs this branch's 47.99%), not a regression


dmcaulay (commented):
Verdict: COMMENT — the ready=true collapse of "no record" into "claimed" silently starves patch/deploy candidacy for self-authored PRs when allow_self_review is false, and fails closed on transient task-store errors (see inline comments). Also: check-patch.ts's module docstring (line 15, "the task-store /prs record is only consulted for the age field, not for qualification") is now stale — this PR makes the record directly gate qualification.

agent/src/check-deploy.ts
```diff
@@ -231,14 +235,21 @@ export async function getDeployCandidates(
...
+            // Query failed → treated the same as a ready=true-filtered
+            // "claimed" result below (createPrRecordQuery's production
+            // implementation never actually throws, but a caught error here
+            // must not silently add a possibly-claimed PR as a candidate).
           }
+          // queryPrRecord IS configured but returned null — by deploy phase a
+          // task-store record should always exist (review always claims
+          // first), so a ready=true-filtered null means "currently claimed".
+          // Skip.
+          if (!record) continue;
```
dmcaulay: This treats a `ready=true`-filtered `null` as "claimed" and skips the PR. That collapses two different cases: (1) genuinely claimed, and (2) no record was ever created because `check-review.ts` skips self-authored PRs entirely when `allow_self_review: false` (review never calls `claim()` for them, and `claim()` is the only path that creates a `PullRequest` record). Since `check-deploy.ts` hard-filters to `pr.author.login === currentUser`, every deploy candidate here is self-authored — under that policy setting, this phase goes permanently dark. It also fails closed on any `queryPrRecord` error/misconfig, since `createPrRecordQuery` swallows those into `null` too — previously a null record did not exclude the PR. Suggest mirroring check-review.ts's explicit `claimedBy != null` check instead of relying on `ready=true`.

agent/src/check-patch.ts
```diff
@@ -302,14 +306,20 @@ export async function getPatchCandidates(
...
-        // Query failed → fall back to PR createdAt below.
+        // Query failed → treated the same as a ready=true-filtered "claimed"
+        // result below (createPrRecordQuery's production implementation
+        // never actually throws, but a caught error here must not silently
+        // add a possibly-claimed PR as a candidate).
       }
+      // queryPrRecord IS configured but returned null — by patch phase a
+      // task-store record should always exist (review always claims first),
+      // so a ready=true-filtered null means "currently claimed". Skip.
+      if (!record) continue;
```
dmcaulay: Same issue as check-deploy.ts: this collapses "no record" and "claimed" into one skip condition. `check-patch.ts` only ever considers own PRs (`listOwnOpenPrs`), so if `allow_self_review: false` means review never claims these PRs, they never get a task-store record and are silently excluded from patch candidacy forever — plus any transient queryPrRecord failure now fails closed instead of falling back to createdAt as before.

agent/src/work-selector.ts
```diff
@@ -14,10 +14,18 @@
  * fixtures.
  *
  * `tasks` is already the ready-only candidate set (task-store's ?ready=true
- * query), so status === "pending" and dependency satisfaction are both
- * guaranteed server-side before this function ever sees the list — a task
- * is selectable purely on createdAt age, the same trust level PR candidates
- * already get below.
+ * query), so dependency satisfaction is computed server-side before this
+ * function ever sees the list — a task is selectable purely on
+ * status === "pending" and createdAt age.
```
dmcaulay, line 19: This docstring update (and the mirrored `work-selector.ts` row in docs/agent.md) is stale in the same way check-patch.ts's docstring was before this commit fixed it: (1) it says PR candidates are pre-filtered by check-deploy/check-patch "via its `?ready=true` filter" — but this same commit removed `ready: true` from both, so no production caller passes it anymore; all three collectors now use the explicit inline `claimedBy` check. (2) it says a task is "selectable purely on `status === \"pending\"` and createdAt age" — but `WorkTaskCandidate` has no `status` field (dropped in #1562) and the loop below never checks it; selection is purely `createdAt`. Worth a follow-up pass to match the accurate descriptions already given to check-deploy.ts/check-patch.ts/check-review.ts.

dmcaulay (commented):
Verdict: APPROVE — both prior findings (claimed/no-record collapse starving self-authored PRs, fail-closed on transient query errors) are correctly fixed in e563b7b1 with matching regression tests, and CI is green. One new low-impact issue: work-selector.ts's docstring (and its docs/agent.md row) is now stale in the same way check-patch.ts's was — it still describes check-deploy/check-patch as filtering via `?ready=true`, but this same commit dropped `ready=true` from both in favor of an explicit `claimedBy` check; it also claims task selection checks `status === "pending"`, but that field was removed from WorkTaskCandidate and the code never checks it. Comment-only, no behavioral effect — see inline note.

PR #1592, opened 2026-07-15
feat: add listForAgent() cross-cron query to AgentCronRunService

dodizzle:
## Summary
- Add `AgentCronRunService.listForAgent(agentId, opts?)` to `admin/src/agent-cron-runs.ts` — queries an agent's `AgentCronRun` rows across every cron it owns (unlike `.list()`, which is pinned to a single `cronId`)
- Includes each run's `cron: { id, name, schedule }` relation and `modelBreakdown` in the Prisma query so a future cron-logs page can render without N+1 lookups
- Supports optional `cronId` narrowing (foreign cronId yields an empty result, not an error) and `outcome` filtering, with `"skipped"` matching the `skipped` boolean flag regardless of the `outcome` column value — mirrors the existing `cronRunOutcomeLabel()` display convention

## Acceptance Criteria

| 1 | listForAgent(agentId) returns all runs across every cron for that agent, sorted startedAt desc, paginated like .list() | MET | `$transaction([findMany, count])`, default limit 20 / offset 0, `{ items, total, limit, offset }` |
| 2 | opts.cronId narrows to a single cron, still scoped to agentId (foreign cronId → empty result, not error) | MET | WHERE agentId AND cronId; no ownership lookup/throw |
| 3 | outcome === "skipped" → skipped=true regardless of outcome column; other value → skipped=false AND outcome=value; omitted → no filter | MET | Explicit branching + dedicated tests for both cases |
| 4 | Each item includes cron's id, name, schedule | MET | `AgentCronRunWithCron` type + `include: { cron: { select: {...} } }` |
| 5 | Integration tests covering the 5 named scenarios; no existing tests retired | MET | 9 new tests added; all 20 pre-existing tests untouched |

## Test Plan
- 9 new integration tests in `admin/src/agent-cron-runs.integration.test.ts` (multi-cron aggregation, cronId narrowing incl. cross-agent empty-result, normal outcome filter, `"skipped"` outcome regardless of outcome column, cross-agent isolation, pagination, modelBreakdown + cron together, empty-agent case)
- `bun run typecheck` clean across all workspaces
- `bunx biome lint .` clean
- Full monorepo test suite: 3867 pass, 0 fail (integration tests in this file require a real Postgres connection not available in the sandbox — they report skipped locally via `describeOrSkip`; CI runs them against real Postgres)
- Docs refreshed: `docs/agent.md` updated with the new service method/types (auto-refresh commit)


dodizzle (commented):
Verdict: APPROVE — listForAgent() cross-cron query and the unified cron-logs page both cleanly satisfy their acceptance criteria (CRL-1.1 + CRL-1.2), with solid integration/smoke/unit coverage and no dead code left from the old per-cron route. No blocking issues found.

PR #1593, opened 2026-07-15
feat: repoint admin task-detail Added field from addedAt to createdAt

dmcaulay:
## Summary
- Repoints the admin task-detail "Added" date field from `task.addedAt` to `task.createdAt` — same displayed concept, now sourced from the DB-guaranteed field
- Removes `addedAt` from the `TaskItem` interface in `admin-ui-pages.ts` since it's no longer referenced in this file (`createdAt` was already present, used by an existing separate "Created" field)
- Part of Group 4 (LPF-4.x) removing the `addedAt` field entirely in favor of `createdAt`, per `planning/loop-fixes/PLAN.md`

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Interface types createdAt instead of/in addition to addedAt | MET |
| 2 | 'Added' date field renders task.createdAt | MET |
| 3 | addedAt removed from file if unreferenced | MET |
| 4 | Test fixtures/assertions updated addedAt→createdAt (unit-layer) | MET |

## Test Plan
- Updated `admin-ui-pages.unit.test.ts` timezone-formatting fixtures (TZ-1, TZ-2) to use `createdAt` instead of `addedAt`
- `bun test admin/src/admin-ui-pages.unit.test.ts` — 349 pass, 0 fail
- `tsc --noEmit` in `admin/` — clean

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -222,7 +222,6 @@ export interface TaskItem {
   complexity?: number | null;
   hitl?: boolean | null;
   hours?: number | null;
-  addedAt?: string | null;
   startedAt?: string | null;
   completedAt?: string | null;
   blockedAt?: string | null;
@@ -1881,7 +1880,7 @@ export function renderTaskDetailPage(
     .join("\n");
 
   const datesRows = [
-    dateField("Added", task.addedAt),
+    dateField("Added", task.createdAt),
     dateField("Started", task.startedAt),
     dateField("Claimed", task.claimedAt),
     dateField("Last Heartbeat", task.heartbeatAt),
--- a/admin/src/admin-ui-pages.unit.test.ts
+++ b/admin/src/admin-ui-pages.unit.test.ts
@@ -2444,7 +2444,7 @@ describe("renderTaskDetailPage — timezone formatting", () => {
         id: "TZ-1",
         title: "Timezone test",
         status: "pending",
-        addedAt: "2025-01-16T05:00:00Z", // Jan 16 UTC, Jan 15 Pacific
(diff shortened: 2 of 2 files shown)


admin/src/admin-ui-pages.ts
```diff
@@ -1881,7 +1880,7 @@ export function renderTaskDetailPage(
     .join("\n");
 
   const datesRows = [
-    dateField("Added", task.addedAt),
+    dateField("Added", task.createdAt),
```
dodizzle, line 1883: The pre-existing "Created" row a few lines below (line 1889, unchanged by this PR) already renders `dateField("Created", task.createdAt)`. Now that "Added" also sources `task.createdAt`, both rows will always display the exact same timestamp under two different labels — before this PR they showed distinct values (`addedAt` vs `createdAt`). Was this duplication intended, or should one of the two rows be removed?

dodizzle (commented):
Verdict: COMMENT — The 'Added' row now renders the same value as the pre-existing 'Created' row (both show task.createdAt), a duplicate-timestamp UX regression worth confirming before merge.

dmcaulay:
Confirmed: after this change, both "Added" (line 1883) and "Created" (line 1889) render `task.createdAt`, so they'll always show the same timestamp under two labels.

This is intentional, scoped behavior for LPF-4.4, per `planning/loop-fixes/PLAN.md`:

> **LPF-4.4 — admin.** The task-detail "Added" field repoints from `task.addedAt` to `task.createdAt` — same displayed concept, reliable source.

The task's approved scope is the data-source repoint only; it doesn't call for removing or merging either row, and none of the PR's Acceptance Criteria touch the "Created" row. Collapsing the two into one would be a UI polish change beyond what LPF-4.4 was asked to do — worth a follow-up if the team wants to clean it up, but not part of this PR.

PR #1647, opened 2026-07-16
feat: group session dependencies by ready/in_progress/blocked/closed state

dmcaulay:
## Summary
- The session detail page's Dependencies card flattened all tasks' dependencies into a single deduped bullet list — no way to see which task needed which dependency, or what's actually actionable right now.
- Added a **Dependency graph** card grouped by **Ready / In Progress / Blocked / Closed** — the same taxonomy already used by the Tasks page's state tabs (task-store's `listReady`/`listBlocked` semantics). Each card shows what it needs, and same-branch tasks (bundled into one PR) get a matching color-coded chip/border even across groups. Only tasks that participate in a dependency edge are shown here; unrelated tasks stay in the plain task table.
- The **task table above it** (previously split into just "Open"/"Closed") now uses the same Ready/In Progress/Blocked/Closed grouping and summary line, so both views speak the same language.
- Dependency ids outside the session render as plain unlinked text inline in a participating task's "needs" list; in-session dependency ids link to that task's detail page.

## Test plan
- [x] `bun test admin/src/admin-ui-pages.unit.test.ts` — 368 pass, covering state-group classification/ordering for both the task table and the dependency graph, in-session vs external dependency linking, unrelated-task exclusion, branch-bundle color matching across groups, empty-group omission, cycle safety, and XSS escaping
- [x] `task typecheck` — all workspaces pass
- [x] `bunx biome check` on both changed files — clean (pre-existing formatting drift elsewhere in the test file left untouched)

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -1986,10 +1986,171 @@ const SESSION_CLOSED_STATUSES = new Set([
 ]);
 
 /**
+ * A node in the session's dependency graph. `title`/`status`/`branch` are
+ * null for a dependency id that isn't one of this session's own tasks (an
+ * "external" prerequisite from another session) — we only know its id.
+ */
+interface DependencyNode {
+  id: string;
+  title: string | null;
+  status: string | null;
+  branch: string | null;
+  dependsOn: string[];
+}
+
+/**
+ * Groups a session's tasks into topological "waves": wave N contains every
+ * node whose dependencies are all satisfied by nodes in waves < N. Only
+ * tasks that participate in a dependency edge are included (either they
+ * declare a dependency, or another task in the session declares them as
+ * one) — tasks with no relationships are noise here and stay in the plain
+ * task table above.
+ *
+ * Layering uses round-based Kahn's algorithm: each round admits every node
+ * whose deps already have an assigned wave, so a node's wave number equals
+ * 1 + max(dep wave) without an explicit longest-path pass. A dependency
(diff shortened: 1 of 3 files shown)


dmcaulay (commented):
Verdict: APPROVE — Topological wave graph replaces the flat dependency list correctly; round-based Kahn's layering, cycle-guard, and escaping all check out, with solid unit test coverage. No blocking issues.

dmcaulay (commented):
Verdict: APPROVE — Redesign correctly mirrors task-store's ready/in_progress/blocked/closed semantics (verified against ready.ts, blocked-by.ts, task-service.ts), escaping and internal/external dependency-link distinction are both correct, and tests were rewritten to match. No blocking issues.

dmcaulay (commented):
Verdict: APPROVE — Moves the ready/in_progress/blocked/closed grouping to the task table (as intended) while leaving the dependency graph card's existing grouping unchanged. classifyTaskState is consistent across both call sites, the grouping Map is exhaustive with no double-counting, and no stale references to the old open/closed split remain. No blocking issues.

dmcaulay (commented):
Verdict: APPROVE — Test-only fix for the CI failure on the prior commit (stale 'open/closed' smoke-test wording after the production grouping change). Assertions correctly match classifyTaskState's rules, no other stale references remain, CI is green.

dodizzle (approved):
Verdict: APPROVE — Well-tested, non-breaking refactor; classifyTaskState verified against task-store's actual listBlocked semantics, escaping is thorough, no blocking issues.

PR #1648, opened 2026-07-16
feat(agent): plugin-declared per-command pre-run hooks (command hooks)

EugeneTrapeznikov:
## What this adds

**Command hooks** — plugin-declared, per-command **pre-run** hooks, executed as guards around the shared Claude runner.

A hook is attached to a fully-qualified slash command by directory convention, resolved across **all installed plugins** (same `installed_plugins.json` / plugin-cache resolution `preCheck` already uses), plus a workspace escape hatch:

```
<plugin root>/hooks/<plugin>:<command>.pre/*     # executable(s), filename order
<workspace>/state/hooks/<plugin>:<command>.pre/* # operator escape hatch, runs last
```

Execution order: plugin hooks first (alphabetical by plugin, then filename), workspace hooks last. Each hook runs sequentially via a single `withCommandHooks(runner)` decorator applied where the shared runner is constructed in `index.ts` — because the attachment unit is the slash command parsed from the outgoing message itself, the same hooks fire uniformly for **loop dispatches, generic cron jobs, Slack invocations, and the chat poll loop**, with zero per-dispatch-site code. The parser understands each path's message framing (the `[Cron job: …]` tag from `formatCronMessage`, Slack's `[<name>]: ` sender prefix / thread-message line / mention token / thread-context block).

Hooks receive: the raw outgoing message on stdin, `cwd` = workspace, and `env` = live `process.env` (mirroring `preCheck`'s passthrough, so runtime config-sync rotations are visible) plus `SHIPWRIGHT_HOOK_EVENT=pre` and `SHIPWRIGHT_HOOK_COMMAND=<plugin>:<command>`.

## Off by default

No hook directories installed anywhere ⇒ the decorator is a **byte-identical passthrough**. Existing installs see zero behavior change.

## Fail-closed guard semantics

Pre hooks answer "is it safe/ready to run this command." A nonzero exit or timeout throws a typed `HookError` (hook name, exit code, stderr) and **the session does not run**. `HookError` propagates as an ordinary runner error through the **existing** error paths — `loop-orchestrator.ts`, `cron-handler.ts`, and `chat-poller.ts` already catch runner errors and record a failed run, so this PR makes **zero changes** to `loop-orchestrator.ts`, `cron-handler.ts`, `cron-run-reporter.ts`, or any schema. A failed pre hook is visible today as a failed `AgentCronRun` with the hook's name and stderr in the error payload. No modes, no config: candidacy gating ("is there work?") stays `preCheck`'s job; hooks are a readiness guard.

## Timeout

Per-hook timeout via `SHIPWRIGHT_HOOK_TIMEOUT_MS` (default 120000 ms), following the existing `SHIPWRIGHT_*_MS` convention. On expiry the hook gets SIGTERM, then SIGKILL after a short grace — a SIGTERM-ignoring hook cannot hang the agent. Documented in `docs/configuration.md`, the site config reference, and `state/dev-agent.env.example`.

## Security stance

Hooks are arbitrary executables at the **plugin/workspace trust tier** — the tier that already executes `preCheck` scripts and authors session prompts; installing a plugin already means trusting its code. Hooks are deliberately **not declarable from the remote admin config bundle**: the bundle stays a data-only surface, and no new remote-code-execution vector is introduced. (The bundle can still select which plugins install, exactly as it can for `preCheck`-carrying plugins today — unchanged boundary.)

## Motivating use-case

Attaching environment-preparation guards to stock commands **without forking their markdown**. Example: a deployment that provisions an ephemeral dev/test environment can ship a `hooks/shipwright:dev-task.pre/10-reset-env.sh` in its **own** plugin to reset that environment before every dev-task session — including manually-invoked ones — while `dev-task.md` stays byte-identical to upstream. Cross-plugin attachment is intentional: a sidecar-directory design (rather than frontmatter in the command's own file) is what makes hooking another plugin's command possible without maintaining a drifting copy of a large prompt file.

## Hook-authoring note (documentation guidance)

A hook must not leave background children attached to its stdio. The runner waits on the hook's stderr pipe as well as its exit; a hook that daemonizes a child without redirecting the child's output (`>/dev/null 2>&1` or `setsid`/`nohup` with redirection) keeps the pipe open and reads as a timeout.

## Tests

29 unit tests in `agent/src/hooks.unit.test.ts` (injected spawner/fs doubles, house style):
- resolution order: own-plugin / cross-plugin / workspace, filename order, non-executable files skipped, manifest + plugin-cache resolution
- message parsing for every dispatch framing (loop, cron tag, Slack DM / thread / mention / thread-context block — commands quoted inside history do **not** fire), fully-qualified-command requirement
- guard-abort on nonzero exit (runner never called), sequential abort, dangling-symlink fail-closed
- **real-spawn** timeout test: an actual `trap '' TERM` process is SIGKILLed within the grace window
- `HookError` surfacing as a failed run through the real `cron-handler` error path, and session suppression through the real chat-poller path
- no-hooks and non-command passthrough identity

`bun test agent/` 974 pass / 0 fail; `tsc --noEmit` clean; no new biome diagnostics vs `main`.

Scope note: `pre` event only. `post` hooks and `hooks:` frontmatter sugar for the own-command case are possible follow-ups, deliberately kept out of this PR.

### DIFF
--- a/agent/src/hooks.ts
+++ b/agent/src/hooks.ts
@@ -162,6 +162,10 @@ function listPluginRoots(config: CommandHooksConfig, fs: HookFs): PluginRoot[] {
     config.pluginManifestPath ??
     join(homedir(), ".claude", "plugins", "installed_plugins.json");
 
+  // An absent manifest means no plugins are installed — the no-hooks
+  // passthrough case, not an error.
+  if (!fs.existsSync(manifestPath)) return roots;
+
   try {
     const manifest = JSON.parse(fs.readFileSync(manifestPath, "utf-8")) as {
       version: number;
@@ -175,9 +179,9 @@ function listPluginRoots(config: CommandHooksConfig, fs: HookFs): PluginRoot[] {
       }
     }
   } catch (err) {
-    console.warn(
-      `[agent:hooks] failed to read installed_plugins.json: ${String(err)}`,
-    );
+    // Fail closed: a manifest that exists but cannot be read/parsed makes
+    // hook resolution unknowable — the session must not run ungated.
+    throw new HookError(manifestPath, undefined, String(err));
   }
   return roots;
 }
--- a/agent/src/hooks.ts
+++ b/agent/src/hooks.ts
@@ -1,0 +1,330 @@
+/**
+ * agent/src/hooks.ts
+ *
+ * Command hooks — plugin-declared, per-command pre-run hooks executed as
+ * fail-closed guards around the shared Claude runner. Hook dirs named
+ * `hooks/<plugin>:<command>.pre/` are resolved across all installed plugins
+ * (same manifest/pluginCacheDir resolution preCheck uses), then the workspace
+ * escape hatch `state/hooks/<plugin>:<command>.pre/`. With no hook dirs
+ * installed the decorator is a byte-identical passthrough.
+ */
+
+import { existsSync, readFileSync, readdirSync, statSync } from "node:fs";
+import { homedir } from "node:os";
+import { join } from "node:path";
+import type { ClaudeRunResult } from "./claude.ts";
+
+// ─── Types ────────────────────────────────────────────────────────────────────
+
+/** The claude runner signature the decorator preserves. */
+export type ClaudeRunner = (
+  message: string,
+  sessionKey?: string,
+) => Promise<ClaudeRunResult>;
+
+/** Filesystem surface used for hook resolution — injectable for tests. */
+export interface HookFs {
+  existsSync: (path: string) => boolean;
+  statSync: (path: string) => { isDirectory: () => boolean; mode: number };
+  readdirSync: (path: string) => string[];
+  readFileSync: (path: string, encoding: "utf-8") => string;
+}
+
+const nodeFs: HookFs = { existsSync, statSync, readdirSync, readFileSync };
+
+export interface CommandHooksConfig {
+  /** Workspace root — hook cwd, and the base for the `state/hooks` escape hatch. */
+  workspace?: string;
+  /** Plugin cache dir — overridable for testing, like cron-handler's preCheck. */
+  pluginCacheDir?: string;
+  /** installed_plugins.json path — the production resolution source. */
+  pluginManifestPath?: string;
+  /** Per-hook timeout. Defaults to SHIPWRIGHT_HOOK_TIMEOUT_MS (120000ms). */
+  timeoutMs?: number;
+  /** Grace period between SIGTERM and SIGKILL on timeout. Default 5000ms. */
+  killGraceMs?: number;
+  /** Process spawner — injectable for tests. Defaults to Bun.spawn. */
+  spawner?: typeof Bun.spawn;
+  /** Filesystem surface — injectable for tests. Defaults to node:fs. */
+  fs?: HookFs;
+}
+
+/** The plugin + command a slash-command message invokes. */
+export interface ParsedCommand {
+  plugin: string;
+  command: string;
+}
+
+const DEFAULT_HOOK_TIMEOUT_MS = 120_000;
+const DEFAULT_KILL_GRACE_MS = 5_000;
+
+// ─── Errors ─────────────────────────────────────────────────────────────────
+
+/**
+ * Thrown when a pre hook exits nonzero, times out, or cannot be resolved.
+ * Fail-closed: the session is suppressed.
+ */
+export class HookError extends Error {
+  constructor(
+    readonly hookName: string,
+    readonly exitCode: number | undefined,
+    readonly stderr: string,
+    readonly timedOut: boolean = false,
+  ) {
+    super(
+      timedOut
+        ? `pre hook "${hookName}" timed out${stderr ? `: ${stderr}` : ""}`
+        : `pre hook "${hookName}" failed (exit ${exitCode})${stderr ? `: ${stderr}` : ""}`,
+    );
+    this.name = "HookError";
+  }
+}
+
+// ─── Command parsing ────────────────────────────────────────────────────────
+
+// A fully-qualified slash command: /<plugin>:<command>, terminated by
+// whitespace or end-of-line.
+const COMMAND_RE = /^\/([A-Za-z0-9._-]+):([A-Za-z0-9._-]+)(?:\s|$)/;
+
+// Wrapper lines the dispatch paths prepend on their own line(s) before the
+// substantive message: cron-handler's `[Cron job: <id>] Current time: …` and
+// slack.ts's `[Thread message — …]` line. Skipped so the command is still
+// resolvable from the outgoing message alone.
+const WRAPPER_LINE_RE = /^\[(?:Cron job:|Thread message)/;
+
+// slack.ts's app_mention handler may prepend a `[Thread context]` …
+// `[end thread context]` block of quoted history — never the message being
+// sent, so it is skipped wholesale (a slash command inside it must NOT fire).
+const THREAD_CONTEXT_START = "[Thread context]";
+const THREAD_CONTEXT_END = "[end thread context]";
+
+// slack.ts prefixes every human message with `[<display name>]: ` and an
+// app_mention's text retains the bot's `<@U…>` mention token — both may
+// precede the slash command on the substantive line.
+const SENDER_PREFIX_RE = /^\[[^\]]+\]:\s*/;
+const MENTION_TOKEN_RE = /^(?:<@[^>\s]+>\s*)+/;
+
+/**
+ * Parse the leading fully-qualified slash command from an outgoing runner
+ * message. Returns undefined when the message does not invoke one (⇒ the
+ * decorator passes through untouched).
+ */
+export function parseCommand(message: string): ParsedCommand | undefined {
+  let inThreadContext = false;
+  for (const rawLine of message.split("\n")) {
+    const line = rawLine.trim();
+    if (inThreadContext) {
+      if (line === THREAD_CONTEXT_END) inThreadContext = false;
+      continue;
+    }
+    if (line === "") continue;
+    if (line === THREAD_CONTEXT_START) {
+      inThreadContext = true;
+      continue;
+    }
+    if (WRAPPER_LINE_RE.test(line)) continue;
+    const stripped = line
+      .replace(SENDER_PREFIX_RE, "")
+      .replace(MENTION_TOKEN_RE, "");
+    const match = COMMAND_RE.exec(stripped);
+    return match ? { plugin: match[1], command: match[2] } : undefined;
+  }
+  return undefined;
+}
+
+// ─── Hook resolution ────────────────────────────────────────────────────────
+
+interface PluginRoot {
+  name: string;
+  root: string;
+}
+
+/**
+ * Enumerate installed plugins as `{ name, root }`, using the same
+ * pluginCacheDir / installed_plugins.json sources cron-handler's preCheck
+ * resolution uses.
+ */
+function listPluginRoots(config: CommandHooksConfig, fs: HookFs): PluginRoot[] {
+  const roots: PluginRoot[] = [];
+
+  if (config.pluginCacheDir) {
+    if (!fs.existsSync(config.pluginCacheDir)) return roots;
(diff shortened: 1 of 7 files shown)


agent/src/hooks.ts
```diff
@@ -0,0 +1,330 @@
...
+    };
+    for (const [key, entries] of Object.entries(manifest.plugins)) {
+      const installPath = entries?.[0]?.installPath;
+      if (installPath) {
+        // Manifest keys are `<name>@<source>` — the plugin name is the prefix.
+        roots.push({ name: key.split("@")[0], root: installPath });
+      }
+    }
+  } catch (err) {
+    console.warn(
```
dmcaulay: `HookError`'s own doc comment promises fail-closed behavior when a hook "cannot be resolved," but a manifest read/parse failure here is caught, warned, and swallowed — `listPluginRoots` returns `[]`, so every plugin's guard hooks silently no-op and the session runs ungated. Compare to `listHookScripts` just below, which correctly throws `HookError` on an unstat-able entry (fail-closed). It also diverges from `preCheck`'s equivalent resolution in `cron-handler.ts:179-198`, where a manifest parse failure leaves the job skipped (no session runs) rather than running unguarded. Suggest throwing a resolution error here instead of returning `[]`.

dmcaulay (commented):
Verdict: COMMENT — Well-designed, well-tested dispatch-layer gate, but `listPluginRoots` fails open on a broken plugin manifest, contradicting the feature's own fail-closed guarantee (see inline comment). Fix before merge.

agent/src/hooks.ts
EugeneTrapeznikov: Good catch — agreed, this was fail-open and inconsistent with both `listHookScripts` and `preCheck`'s resolution. Fixed in 7672d9a: a manifest that exists but can't be read or parsed now throws `HookError` (session blocked, surfaces as a failed run through the existing error paths). An absent manifest remains the no-plugins-installed passthrough case — that boundary is now explicit via an `existsSync` guard. Added two tests covering both sides: absent manifest ⇒ passthrough, corrupt manifest ⇒ `HookError` with the runner never invoked.

dmcaulay:
Stepping back from the implementation review — a couple of design questions before we land this.

**1. Do we need this at all four dispatch surfaces, or is cron (which already has a gate) the only place this has mattered in practice?**

We have a `preCheck` gate on crons — it's the mechanism we use for standalone crons like `check-test-readiness.ts`, `check-docs-freshness.ts`, `check-learn-dream.ts`, and `check-error-patrol.ts`. That said, we're in the middle of changing phase gating: the four per-phase crons (dev-task/patch/review/deploy) and their `preCheck` scripts are being retired in favor of `shipwright-loop`'s in-process qualification (`WL-2.2` already ported the four qualification functions directly into `agent/src`; `WLS-5.1` removes the legacy scripts).

We can extend `preCheck` by allowing multiple scripts and give `shipwright-loop`'s own phase dispatch (`loop-orchestrator.ts`) a dedicated extension point using the phase `preCheck` and allowing multiple scripts there.

That leaves ambiguous, human-typed dispatches (Slack DMs, mentions) as the one case where the command really can't be known ahead of time and does need to be parsed out of the message — that's where the Skill-based alternative in point 2 is worth weighing against this PR's regex approach.

**2. The message-parsing approach to detect a command feels fragile, and I'm not sure it beats a Skill.**

`parseCommand`'s regex requires the fully-qualified command as the literal first token on the line, after stripping only known dispatch-added prefixes. In practice that's easy to accidentally dodge — e.g. Slack's own client intercepts a message starting with `/` as an attempted native slash command, so people commonly type around it (a leading backtick, or "please run /shipwright:review") — and none of those match, so the hook silently never fires. There's also no registry check against real commands (any `/word:word`-shaped text matches, valid or not) and no visibility into what actually triggered — if a hook doesn't fire because the message didn't parse as expected, nothing surfaces that.

Compare that to just giving the agent a Skill it can invoke itself once a session is already running — e.g. `/shipwright:dev-task ABC-1.1, spin up an ephemeral environment for this and use it for testing` could resolve to the agent calling an "ephemeral environment" skill directly, with no external parsing step at all. That also sidesteps two gaps we found in this PR's design: the hook has no access to the real session/conversation id for reuse decisions (only available *inside* a session), and no return channel to hand connection info back (hook stdout isn't captured) — a skill's tool result naturally solves both, since it's already part of the conversation the agent can act on.

And if the goal is "don't make people type the ephemeral-env instruction every time" — that's a `dev-task` prompt-customization problem, not a dispatch-gating problem. Worth solving separately (we already want a way to extend the built-in command prompts with instructions like this) rather than building an external hook layer to enforce it.

Curious how you're thinking about these two — happy to dig into either further.

dmcaulay:
i updated the comment again ^^
the `shipwright-loop` migration is confusing for our agents because it's hard for them to determine what's new vs legacy.
things are shifting this week 😄

agent/src/hooks.unit.test.ts
```diff
@@ -0,0 +1,693 @@
...
+    expect(runner).toHaveBeenCalledTimes(1);
+  });
+
+  test("real spawn: a SIGTERM-ignoring hook is SIGKILLed after the grace period", async () => {
+    const cacheDir = tmp();
+    const dir = join(cacheDir, "custom", "hooks", "shipwright:dev-task.pre");
+    mkdirSync(dir, { recursive: true });
+    writeExec(
+      join(dir, "10-stubborn.sh"),
+      "#!/usr/bin/env bash\ntrap '' TERM\nwhile true; do sleep 0.05; done\n",
```
dodizzle: This test deliberately omits the injected spawner to exercise real SIGTERM/SIGKILL signal behavior — a genuine process spawn. `docs/test-readiness/test-system.md` scopes unit tests to "no process spawning" and says a spawning test should move to integration. No other `*.unit.test.ts` file in the repo does a real spawn (the equivalent real-process testing for `preCheck` lives in `cron-handler.smoke.test.ts`, the smoke layer). Consider splitting this test out into `hooks.integration.test.ts` (or alongside the smoke file) while keeping the rest of the fixture-based resolution tests here.

dodizzle (commented):
Verdict: COMMENT — The fail-closed fix (7672d9a) correctly resolves the prior review's finding, and the timeout/kill-grace logic and env passthrough check out. Two issues to fix before merge: the PR title is currently failing the required `pr-title-lint` gate (`agent:` isn't a recognized Conventional Commit type — e.g. `feat(agent): ...`), and one new test in `hooks.unit.test.ts` does a real (non-injected) process spawn, which crosses the unit-layer's no-process-spawning boundary per `docs/test-readiness/test-system.md` — consider moving it to an integration/smoke file alongside `cron-handler.smoke.test.ts`.

agent/src/hooks.unit.test.ts
EugeneTrapeznikov: Moved that test to a new agent/src/hooks.smoke.test.ts in 0d55c90f, following the real-process preCheck precedent in cron-handler.smoke.test.ts. Assertions unchanged; the remaining withCommandHooks tests stay in hooks.unit.test.ts behind the injected spawner double.

EugeneTrapeznikov:
Addressed both review items in 0d55c90f:

- **PR title**: retitled to `feat(agent): plugin-declared per-command pre-run hooks (command hooks)` so it passes `pr-title-lint` (Conventional Commit type).
- **Test layer**: moved the one real-spawn test (SIGTERM→SIGKILL grace) out of `hooks.unit.test.ts` into a new `agent/src/hooks.smoke.test.ts`, following the real-process precedent in `cron-handler.smoke.test.ts`. Assertions unchanged; all other fixture-based `withCommandHooks` tests remain in the unit file behind the injected spawner double.

EugeneTrapeznikov:
> Do we need this at all four dispatch surfaces, or is cron (which already has a gate) the only place this has mattered in practice?

Agreed, and your read matches our production data: 100% of our real hook firings are loop/phase dispatches; the Slack surfaces are theoretical for us. We'd happily narrow this PR to a phase-dispatch extension point in loop-orchestrator (multi-script, preCheck-style semantics), which also aligns with WL-2.2/WLS-5.1 instead of instrumenting crons that are being retired.

That also dissolves point 2's parsing concerns entirely: at the loop's dispatch site the phase is structured data, so `parseCommand` and its wrapper-stripping disappear from the PR — the regex only ever existed to reconstruct, from message text, a fact the dispatcher had moments earlier.

> The message-parsing approach to detect a command feels fragile, and I'm not sure it beats a Skill.

The fragility examples are fair (and the silent no-match path doubly so). Where we'd distinguish: a Skill is agent-invoked inside a running session, so it's a capability, not an invariant — our use case is an operator-enforced, fail-closed guard ("this session must not start unless X exit-0'd") that has to hold even when the model skips a step. For the ephemeral-env example with session ids and a return channel, agreed a Skill is the better shape there. And yes — "don't retype the instruction every time" is prompt customization, a separate problem.

If the loop extension point direction works for you, we'll rework the PR to that scope — we've started sketching it.

dmcaulay:
Glad the phase-dispatch direction works, and appreciate you already checking it against the in-process phase-qualification work and the legacy-precheck cleanup — that's exactly the alignment we were hoping for.

One ask before you sink more time into it: **please hold off starting implementation until we've landed the current `shipwright-loop` migration.** `loop-orchestrator.ts`'s phase-dispatch call site — the exact code you'd be extending — has several changes actively in flight right now (threading the selected item ID and the real cron ID through the dispatch path for attribution), plus the cleanup that removes the legacy `preCheck` scripts and nulls their refs hasn't landed yet. Building against that file today means building against a moving target and likely rebasing more than once.

We'll follow up here once the migration settles and the extension point is stable ground to build on. Wanted to flag this now rather than let you sketch further against code that's about to shift under you.

agent/src/hooks.ts
```diff
@@ -0,0 +1,334 @@
...
+// precede the slash command on the substantive line.
+const SENDER_PREFIX_RE = /^\[[^\]]+\]:\s*/;
+const MENTION_TOKEN_RE = /^(?:<@[^>\s]+>\s*)+/;
+
+/**
+ * Parse the leading fully-qualified slash command from an outgoing runner
+ * message. Returns undefined when the message does not invoke one (⇒ the
+ * decorator passes through untouched).
+ */
+export function parseCommand(message: string): ParsedCommand | undefined {
```
dodizzle, line 112: `parseCommand`'s thread-context state machine exits `inThreadContext` on any trimmed line that exactly equals `[end thread context]` (or enters it on `[Thread context]`). But `agent/src/slack.ts:560-576` builds `historyLines` from raw, unsanitized Slack message text (`m.text`) joined with `\n` — Slack messages routinely contain embedded newlines, so any thread participant can type a multi-line message whose internal line happens to read `[end thread context]`, `[Thread context]`, or a bare `/plugin:command` line. When quoted into a later prompt's `[Thread context]` block, the parser desyncs: it exits context-tracking on the fake boundary and immediately returns based on the very next line — which is still inside quoted, participant-influenceable history, not the real live command. This can make the fail-closed hook gate either silently not fire for a genuine live command, or fire based on a stale/quoted command instead of the actual one being dispatched — undermining the feature's own stated fail-closed guarantee. New logic introduced by this PR; no test coverage for multi-line historical message bodies (the existing quoted-command test at `hooks.unit.test.ts:630` only covers a single-line quoted message).

Suggestion: make `historyLines` construction in `slack.ts` robust to embedded newlines (collapse/escape internal `\n` before joining), or make `parseCommand`'s boundary matching insensitive to attacker-controlled quoted content — only trust boundary markers at the position the dispatch code itself inserted them.

dodizzle (commented):
Verdict: COMMENT — Both prior findings (PR title, test-layer boundary) are correctly resolved in 0d55c90f. One new issue: `parseCommand`'s thread-context tracking (`hooks.ts:112-133`) can desync on multi-line Slack message content quoted into `[Thread context]` blocks, since `slack.ts` joins raw unsanitized `m.text` (which can contain embedded newlines) without escaping — see inline comment.

dmcaulay:
closing in favor of extending prechecks.

EugeneTrapeznikov:
> closing in favor of extending prechecks.

sounds good

dmcaulay:
all clear. feel free to pick up the prechecks work.

PR #1676, opened 2026-07-16
feat: report failed cron ticks via cronRunReporter and Sentry

dodizzle:
## Summary
- The cron-sync scheduler's catch block in `agent/src/index.ts` previously only did a bare `console.error` when a scheduled tick's dispatch threw — no `cronRunReporter` call, so a failing tick never showed up in the admin run-history UI or Slack, only in raw pod stdout.
- Wraps that catch block (which covers both the `dispatch === "loop"` path and the regular `handleCronRequest` path) with `cronRunReporter.createRun` / `completeRun(..., "failed", { error })` calls, mirroring the pattern `loop-orchestrator.ts`'s `dispatch()` already uses, and adds `Sentry.captureException(err)` alongside the existing `console.error` (which is retained, not replaced).
- **Extraction note:** `index.ts` has no existing test seam (it's the startup entrypoint with top-level side-effecting code), so the catch-and-report logic was extracted into a new small, pure, testable module `agent/src/cron-failure-reporter.ts` exporting `reportCronFailure(cronId, err, deps)`. `index.ts`'s catch block now just calls this function, passing `cronRunReporter ?? new NoopCronRunReporter()`, `sentryClient: process.env.SENTRY_DSN ? Sentry : undefined`, and `SystemClock()`.

## Acceptance Criteria
| 1 | createRun + completeRun("failed", {error}) called for both dispatch paths | MET | Single catch block wraps both `dispatch === "loop"` and `handleCronRequest` calls with `reportCronFailure()`, which calls `cronRunReporter.createRun` then `completeRun(..., "failed", { error: message })` |
| 2 | Sentry.captureException(err) called in the same catch block | MET | `cron-failure-reporter.ts` calls `sentryClient?.captureException(err)`; wired from `index.ts` only when `SENTRY_DSN` is set |
| 3 | Existing console.error retained | MET | `console.error(\`[cron] job ${cronId} failed:\`, message)` preserved verbatim inside the extracted function |
| 4 | Test asserting completeRun("failed") + captureException on injected fakes; extraction noted | MET | `agent/src/cron-failure-reporter.unit.test.ts` — 4 tests with fake `CronRunReporter` and fake `ErrorCapturingClient` |

## Test Plan
- New `agent/src/cron-failure-reporter.unit.test.ts`: asserts `createRun`/`completeRun("failed", {error})` pair on an injected fake `CronRunReporter`, non-Error stringification, `captureException` call on an injected fake `ErrorCapturingClient`, and that omitting `sentryClient` doesn't throw.
- Full agent test suite: 980 pass / 0 fail.
- `bunx biome lint .`: clean across 475 files.
- `tsc --noEmit` (agent): clean.


agent/src/index.ts
```diff
@@ -409,10 +410,11 @@ if (runtimeClient && agentId) {
               );
             }
           } catch (err) {
```
dodizzle, line 406: Both call paths this catch wraps already report the failure via `cronRunReporter` before rethrowing:
- `handleCronRequest`'s runner-scope try/catch (`agent/src/cron-handler.ts:301-306`) calls `cronRunReporter?.completeRun(jobId, runId, ..., "failed", {...})` on the `runId` it created earlier, then `throw err`.
- `loop-orchestrator.ts`'s `dispatch()` (`agent/src/loop-orchestrator.ts:159-169`) does the same with its own `runId` and a `phase` tag, then `throw err`.

Neither error is caught in between, so both propagate here, where `reportCronFailure()` does its own fresh, unconditional `createRun`/`completeRun("failed")` — no `runId` reuse, no signal the failure was already reported. Net effect: one real failure produces two separate `AgentCronRun` "failed" rows (one phase-tagged from the inner call, one untagged from this outer call), corrupting run-history and inflating failure-rate metrics on essentially every dispatch failure.

Suggest having `handleCronRequest`/`dispatch` mark/type the rethrown error as "already reported" and skip the outer createRun/completeRun in that case, or scope this catch to only the parts of the callback outside their own try/catch.

dodizzle (commented):
Verdict: COMMENT — Every dispatch failure now creates two AgentCronRun "failed" rows: both `handleCronRequest` and `loop-orchestrator.ts`'s `dispatch()` already report the failure via `cronRunReporter.completeRun(..., "failed", ...)` before rethrowing, and this new outer catch in `index.ts` unconditionally calls `reportCronFailure()`, which does its own fresh `createRun`/`completeRun("failed")` pair on top. See inline comment.

agent/src/index.ts
```diff
@@ -403,10 +404,11 @@ if (runtimeClient && agentId) {
...
-            console.error(
-              `[cron] job ${id} failed:`,
-              err instanceof Error ? err.message : String(err),
-            );
+            await reportCronFailure(id, err, {
```
dmcaulay, line 407: Both call paths this catch wraps already report the failure via `cronRunReporter` before rethrowing:
- `handleCronRequest`'s runner-scope try/catch (`agent/src/cron-handler.ts:301-306`) calls `cronRunReporter?.completeRun(jobId, runId, ..., "failed", {...})`, then `throw err`.
- `loop-orchestrator.ts`'s `dispatch()` (`agent/src/loop-orchestrator.ts:159-169`) does the same with its own `runId` and a `phase` tag, then `throw err`.

Neither error is caught in between, so both propagate here, where `reportCronFailure()` does its own fresh, unconditional `createRun`/`completeRun("failed")` — no `runId` reuse, no "already reported" signal. Net effect: one real failure produces two separate `AgentCronRun` "failed" rows (one phase-tagged from the inner call, one untagged from this outer call), corrupting run-history and inflating failure-rate metrics on essentially every dispatch failure.

This is a regression newly introduced by this PR (the prior code was a bare `console.error`, no reporter call), and it was already raised in an earlier review round on this exact PR without being addressed since (only a no-op merge-from-main commit landed after).

Suggest having `handleCronRequest`/`dispatch` mark the rethrown error as "already reported" so this outer catch can skip its own createRun/completeRun in that case, or scope this catch to only the parts of the callback outside their own try/catch.

dmcaulay (commented):
Verdict: COMMENT — Every dispatch failure now creates two AgentCronRun "failed" rows: both `handleCronRequest` and `loop-orchestrator.ts`'s `dispatch()` already report the failure via `cronRunReporter.completeRun(..., "failed", ...)` before rethrowing, and this outer catch in `index.ts` unconditionally calls `reportCronFailure()`, which does its own fresh `createRun`/`completeRun("failed")` pair on top. This was already flagged in a prior review round on this PR and hasn't been addressed since (only a no-op merge-from-main landed). See inline comment.

agent/src/cron-failure-reporter.ts
```diff
@@ -0,0 +1,106 @@
...
+  deps: ReportCronFailureDeps,
+): Promise<void> {
+  const { cronRunReporter, sentryClient, clock } = deps;
+  const message = err instanceof Error ? err.message : String(err);
+
+  console.error(`[cron] job ${cronId} failed:`, message);
+
+  sentryClient?.captureException(err);
+
+  if (isCronRunFailureReported(err)) {
```
dodizzle, line 95: This dedup check only catches failures already tagged by cron-handler.ts's runner-scope catch or loop-orchestrator.ts's dispatch() catch. But handleCronRequest's post-run scope calls completeRun(..., "completed", ...) *before* attempting Slack delivery (channel post, DM open, dispatchMarkers) — if that delivery step throws, the error is never caught inside handleCronRequest and never tagged here. It reaches index.ts's outer catch untagged, so reportCronFailure creates a fresh "failed" row on top of the already-correct "completed" row for the same tick. Same duplicate-row bug class as the one this commit fixes, just via a sibling path (cron-handler.ts's Slack-delivery throws, e.g. around the "could not open DM" error).

agent/src/cron-failure-reporter.unit.test.ts
```diff
@@ -0,0 +1,200 @@
...
+    client: {
+      captureException: (err: unknown) => {
+        capturedErrors.push(err);
+      },
+    },
+    capturedErrors,
+  };
+}
+
+describe("reportCronFailure", () => {
```
dodizzle, line 84: These tests validate the tagging primitive in isolation (manually constructing errors and calling markCronRunFailureReported/reportCronFailure directly) but nothing here drives handleCronRequest or loop-orchestrator.ts's dispatch() through an actual failure to confirm tagging happens before the error reaches index.ts's outer catch — and nothing covers the post-completion Slack-delivery-failure path where no tagging occurs at all. Since the PR's whole purpose is a cross-module dedup contract, consider an integration-level test that exercises that contract end-to-end rather than just the primitive.

dodizzle (commented):
Verdict: COMMENT — the dedup fix correctly resolves the runner-scope duplicate-row bug two prior reviews flagged, but the post-completion Slack-delivery-failure path in cron-handler.ts still isn't tagged, so it can still produce a spurious duplicate "failed" AgentCronRun row alongside the correct "completed" one. Worth one more patch cycle before merge.

dmcaulay (approved):
Verdict: APPROVE — Both rounds of prior findings (the outer-catch duplicate-AgentCronRun-row bug, and the post-completion Slack-delivery-failure dedup gap dodizzle flagged) are independently re-verified as fully fixed via the `markCronRunFailureReported`/`Symbol.for` tagging mechanism, with genuine end-to-end test coverage in `cron-handler.smoke.test.ts`. Also checked and ruled out a potential Sentry double-capture concern — neither `cron-handler.ts` nor `loop-orchestrator.ts` call Sentry directly, so `reportCronFailure` remains the sole capture point. No new issues found.

PR #1679, opened 2026-07-16
feat: scope review/patch/deploy candidate repo lists to agent config, read live

dodizzle:
## Summary
- Threads a `getScopedRepos: () => string[]` dependency (defaulting to WL-4.2's live `agentReposRef` singleton) through `CheckReviewDeps`, `CheckPatchDeps`, and `CheckDeployDeps`
- `getReviewCandidates` / `getPatchCandidates` / `getDeployCandidates` filter the local-clone repo list against `getScopedRepos()` at the top of every call — not once at `buildProductionDeps` time — so a repo cloned locally but outside the agent's configured scope never appears as a candidate, and a later scope change takes effect on the very next call
- Exports `agentReposRef` as a process-wide singleton from `agent-repos-ref.ts` so `index.ts`'s `syncConfig()` and the three `check-*.ts` `buildProductionDeps` functions share the exact same live ref, without threading a new parameter through `loop-orchestrator.ts`

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `CheckReviewDeps`/`CheckPatchDeps`/`CheckDeployDeps` gain `getScopedRepos`; `buildProductionDeps` wires WL-4.2's live ref | ✅ MET |
| 2 | Candidate functions filter the local-clone list against `getScopedRepos()` at call time, every invocation | ✅ MET |
| 3 | Unit test per consumer proving exclusion + live re-evaluation between calls | ✅ MET |
| 4 | In-scope behavior unchanged; no change to `loop-orchestrator.ts`/`work-selector.ts` required | ✅ MET |
| 5 | Safe to deploy standalone (purely additive) | ✅ MET |

## Test Plan
- [x] Unit tests added to `check-review.unit.test.ts`, `check-patch.unit.test.ts`, `check-deploy.unit.test.ts`: exclusion of an out-of-scope repo, and live re-evaluation (scope changes between two calls change the second call's result)
- [x] Unit test added to `agent-repos-ref.unit.test.ts` covering the new `agentReposRef` singleton
- [x] `task typecheck` passes across all 7 workspace packages
- [x] `task lint` passes
- [x] `NODE_ENV=test bun test agent/src` — 982/983 pass (1 pre-existing failure in `claude.unit.test.ts`, unrelated to this change and reproducible on `main`)
- [x] Aggregate coverage improves from 77.12% → 79.09% (still below the 80% gate, but this is a pre-existing repo-wide gap unrelated to this PR, not a regression)


agent/src/check-review.ts
```diff
@@ -132,6 +146,7 @@ export async function buildProductionDeps(opts: {
   return {
     getCurrentUser,
     isSelfReviewAllowed: readAllowSelfReview(workspacePath),
+    getScopedRepos: opts.getScopedRepos ?? agentReposRef.get,
```
dodizzle, line 196: `getScopedRepos` defaults to `agentReposRef.get`, which is only populated by `index.ts`'s `syncConfig()` on a successful config-bundle fetch. Its 404 branch (`agent/src/index.ts:234-241`) returns without calling `agentReposRef.set(...)`. Before this PR, an empty/unpopulated ref had no effect on candidate collection — filtering was purely by local clone directory. Now, if config sync never succeeds (e.g. agent provisioned without a config bundle yet), `getScopedRepos()` returns `[]` forever and every repo silently drops out of review/patch/deploy candidacy, indistinguishable from "no work found." Same pattern applies to `check-patch.ts:441` and `check-deploy.ts:328`. Consider failing open to the full local-clone list when the ref has never been successfully populated, or emitting a recurring (not one-time) warning.

dodizzle (commented):
Verdict: COMMENT — Clean singleton migration and well-tested scope filtering, but defaulting getScopedRepos to agentReposRef.get means a persistent config-sync failure (e.g. a 404 on the agent's config bundle) leaves the ref permanently empty, silently excluding every repo from review/patch/deploy candidacy agent-wide — a new failure mode this PR introduces since scope was never previously a candidacy gate.

dodizzle (commented):
Verdict: APPROVE — Clean, well-tested fail-open fix for the scope-sync silent-failure gap flagged in the prior self-review. Verified hasScopeSynced() covers every non-set() path in syncConfig(), fail-open logic is consistent across all three consumers, and no breaking change to the only production caller.

PR #1689, opened 2026-07-16
feat: add live branch/PR reality check before worktree creation in dev-task.md

dmcaulay:
## Summary
- Replace the status-gated Step 1/Step 2 "in_progress → Orphan Check" logic in `dev-task.md` — now unreachable in the normal cron-driven path since WLS-3.1 landed (work-selector.ts only ever hands out `pending` tasks) — with an unconditional Branch/PR Reality Check that runs right before worktree creation in Step 4, regardless of task-store status.
- The new check inspects local branch, remote branch (`git ls-remote --heads origin`), and open PR (`gh pr list --head ... --state open`) state directly. Complete-and-correct prior work (diff satisfies acceptance criteria, CI green if a PR exists) skips destructive recovery and PATCHes the task store to match reality; incomplete/stale work closes the PR and deletes the branch (remote + local) before falling through to the existing fresh-worktree flow.
- Reuses the exact `--repo` derivation pattern from the existing Step 4 stale-bundle-branch check (`git remote get-url origin | sed ...`) rather than inventing a new idiom.

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Check local branch / remote branch / open PR before `git worktree add -b {branch}`, regardless of task-store status | MET |
| 2 | Complete-and-correct: skip destructive recovery, rebase or resume-from-PR, PATCH task-store to match reality | MET |
| 3 | Incomplete/stale: close PR + delete branch (remote+local), proceed to fresh worktree creation | MET |
| 4 | Reuses existing Step 4 `--repo` derivation pattern | MET |
| 5 | Test decision: retire stale in_progress/409 tests (already retired by WLS-3.1), add new coverage for reality-check ordering, complete/incomplete paths, `--repo` derivation | MET |

## Test Plan
- `bun test plugins/shipwright/commands/` — 223 pass, 0 fail
- `bun test` (full suite) — 4036 pass, 0 fail
- `bun run lint` — clean
- `bun run typecheck` — clean (all 7 packages)
- `task check-strings`, `check-config-docs`, `sync-version --check` — all pass

### DIFF
--- a/plugins/shipwright/commands/dev-task.md
+++ b/plugins/shipwright/commands/dev-task.md
@@ -277,12 +277,19 @@ diff unrelated to the task, or CI red with no clear fix path): treat exactly lik
 from a crashed session — clean up before starting fresh.
 1. If an open PR exists, close it: `gh pr close {number} --repo "$GH_REPO" --comment "Shipwright cleanup — prior attempt incomplete/stale, restarting"`
 2. If a remote branch exists, delete it: `git push origin --delete {branch}`
-3. If a local branch exists, delete it: `git -C ${SHIPWRIGHT_REPO_DIR:-$HOME/src}/{repo} branch -D {branch}`
-4. Print:
+3. If a worktree for `{branch}` already exists (a crashed session — exactly the scenario this
+   reality check recovers from — can leave one checked out on disk), remove it first: git
+   refuses to force-delete a branch checked out in any worktree, so this must happen before
+   step 4's `branch -D`.
+   ```bash
+   git -C ${SHIPWRIGHT_REPO_DIR:-$HOME/src}/{repo} worktree remove ${SHIPWRIGHT_WORKTREE_DIR:-$HOME/worktrees}/{repo}-{branch-slug} --force
+   ```
+4. If a local branch exists, delete it: `git -C ${SHIPWRIGHT_REPO_DIR:-$HOME/src}/{repo} branch -D {branch}`
+5. Print:
    ```
    ⚠ {branch} had incomplete/stale prior work — cleaned up (PR closed, branch deleted). Starting fresh.
    ```
-5. Fall through to the standard fresh-start flow below (branch now absent on remote).
+6. Fall through to the standard fresh-start flow below (branch now absent on remote).
 
 ---
 
--- a/plugins/shipwright/commands/dev-task.md
+++ b/plugins/shipwright/commands/dev-task.md
@@ -248,25 +248,26 @@ against the task's acceptance criteria before deciding how to proceed:
    correct" — treat it the same as the incomplete/stale case below.
 
 **Complete and correct** (diff satisfies every acceptance criterion, and CI is green if a PR
-exists): skip destructive recovery — do not close the PR or delete the branch.
-- **Local-only case** (branch exists, no PR yet): rebase is enough — no PATCH of `branch` or
-  `pr` is needed since neither changed; still refresh `heartbeatAt` (Step 5b already does
-  this before dispatch). Continue to the standard "branch DOES exist on remote" flow below
-  (or, if the branch is local-only and never pushed, treat it as already checked out and
-  skip straight to Step 5 — no worktree recreation needed against a fresh `origin/main`
-  base).
-- **PR case**: resume from the existing PR rather than rebuilding it. PATCH the task store to
-  match reality so downstream steps (Step 9, Step 9b, Step 10a) see the correct state:
+exists): skip destructive recovery — do not close the PR or delete the branch. Both shortcuts
+below bypass Step 4's `worktree add` block, so each must bind `{worktree-path}` itself
+(reuse an on-disk worktree if the prior session left one, else create it) before handing off.
+- **Local-only** (branch exists, no PR, remote branch present): rebase is enough — no PATCH
+  needed; still refresh `heartbeatAt` (Step 5b). Continue to the "branch DOES exist on
+  remote" flow below, which binds `{worktree-path}` via the tracking-branch `worktree add`.
+  - **Never pushed** (no remote branch): reuse the on-disk worktree, or
+    `worktree add {worktree-path} {branch}` (no `-b`/`origin/main` — branch exists locally).
+    Bind `{worktree-path}`, go straight to Step 5.
+- **PR case**: resume from the existing PR. PATCH the task store to match reality:
   ```bash
   curl -sf -X PATCH -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
     -H "Content-Type: application/json" \
     "$SHIPWRIGHT_TASK_STORE_URL/tasks/{id}" \
     -d "{\"status\": \"pr_open\", \"pr\": {existing-pr-number}}" | jq .
   ```
-  Then skip Step 5 (implementation) entirely and continue from Step 6 (Simplify) using the
-  existing PR's branch, or — if Step 6 onward were already fully done too (CI green, nothing
-  left to verify) — skip straight to Step 10 (Handoff) since the task is effectively already
-  shipped. Print:
+  Reuse the on-disk worktree, or fetch + tracking-add one for the PR branch (same pair as
+  "branch DOES exist on remote" below). Bind `{worktree-path}` — Step 6's diff/Edit work
+  runs from it. Skip Step 5, continue from Step 6 on the PR's branch — or go straight to
+  Step 10 if Step 6+ was already done (CI green). Print:
   ```
   ↩ {branch} already has complete, correct work ({If PR: "PR #{number}, CI green" | "local branch, no PR yet"}).
   Resuming from existing state — skipping destructive recovery.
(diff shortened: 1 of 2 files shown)


plugins/shipwright/commands/dev-task.md
```diff
@@ -208,12 +197,98 @@ architecture to implementation. Auto-fix all quality issues.
...
+  ↩ {branch} already has complete, correct work ({If PR: "PR #{number}, CI green" | "local branch, no PR yet"}).
+  Resuming from existing state — skipping destructive recovery.
+  ```
+
+**Incomplete, stale, or the diff doesn't match the brief** (missing acceptance criteria,
+diff unrelated to the task, or CI red with no clear fix path): treat exactly like an orphan
+from a crashed session — clean up before starting fresh.
+1. If an open PR exists, close it: `gh pr close {number} --repo "$GH_REPO" --comment "Shipwright cleanup — prior attempt incomplete/stale, restarting"`
+2. If a remote branch exists, delete it: `git push origin --delete {branch}`
+3. If a local branch exists, delete it: `git -C ${SHIPWRIGHT_REPO_DIR:-$HOME/src}/{repo} branch -D {branch}`
```
dodizzle: `git branch -D {branch}` will fail here if `{branch}` is currently checked out in a worktree (git refuses to force-delete a branch checked out anywhere, even with `-D`) — and a crashed session, which is exactly what this reality check is designed to recover from, plausibly left such a worktree on disk. The "if the worktree already exists, remove it first" step is only reached later, via the fresh-start fallthrough after this cleanup path — too late to prevent the failure here.

Suggest removing any matching worktree (e.g. `git worktree remove {worktree-path} --force`, or scan `git worktree list` for `{branch}`) before this `branch -D` call.

dodizzle (commented):
Verdict: COMMENT — In the new incomplete/stale cleanup path, `git branch -D {branch}` runs before the worktree-removal step, which is positioned later in the file. If a crashed session left `{branch}` checked out in a worktree — the exact scenario this PR targets — `branch -D` will error out and the recovery never completes. See inline comment.

dmcaulay (commented):
Verdict: COMMENT — The worktree-removal-before-branch-delete fix addressing the prior review is correct, but the new 'complete and correct' resume shortcuts (local-only and PR case) skip Step 4's worktree-add logic entirely without ever binding {worktree-path}, leaving Step 5/Step 6 without a concrete working directory. See inline comment.

plugins/shipwright/commands/dev-task.md
```diff
@@ -208,12 +197,105 @@ architecture to implementation. Auto-fix all quality issues.
...
+   reuse the Step 9b Actions-API pattern (`gh api "repos/$GH_REPO/actions/runs?head_sha=..."`)
+   against the PR's head SHA. A PR with failing or still-running CI is not "complete and
+   correct" — treat it the same as the incomplete/stale case below.
+
+**Complete and correct** (diff satisfies every acceptance criterion, and CI is green if a PR
+exists): skip destructive recovery — do not close the PR or delete the branch.
+- **Local-only case** (branch exists, no PR yet): rebase is enough — no PATCH of `branch` or
+  `pr` is needed since neither changed; still refresh `heartbeatAt` (Step 5b already does
+  this before dispatch). Continue to the standard "branch DOES exist on remote" flow below
+  (or, if the branch is local-only and never pushed, treat it as already checked out and
```
dmcaulay: Both "complete and correct" resume shortcuts bypass Step 4's worktree-add logic that normally establishes `${SHIPWRIGHT_WORKTREE_DIR}/{repo}-{branch-slug}` as the working directory.

The local-only-unpushed case here says to "skip straight to Step 5 — no worktree recreation needed," but Step 5 dispatches a subagent with `Working directory: {worktree-path}` (line 385), a placeholder that's never assigned anywhere in this shortcut. The PR case below (lines 258-273) says to "continue from Step 6 (Simplify)," but Step 6 runs `git diff main...HEAD` and applies Edit-tool fixes — which requires actually being in a worktree with that branch checked out — and no `worktree add --track` runs before that jump either.

A downstream dev-task-executing agent following this prose literally has no instruction to attach/create a worktree before either shortcut hands off, and could end up operating in the wrong directory (e.g. the read-only `repos/{repo}` clone). Suggest explicitly instructing a worktree create/attach step (reusing one already on disk from a crashed session where applicable) before proceeding to Step 5 or Step 6 in both branches.

plugins/shipwright/commands/dev-task.md
```diff
@@ -208,12 +197,105 @@ architecture to implementation. Auto-fix all quality issues.
...
+  skip straight to Step 5 — no worktree recreation needed against a fresh `origin/main`
```
dodizzle: This local-only "skip straight to Step 5" bypasses every `worktree add` (lines 302/328/334) and the working-directory binding at line 343. Step 5 then dispatches a subagent with `Working directory: {worktree-path}` (line 385) — a placeholder never bound on this path, so the agent has no working directory and would fall back to the workspace root / read-only `repos/{repo}` clone (forbidden by CLAUDE.md). Before skipping to Step 5, bind the worktree explicitly: confirm the prior session's worktree at `${SHIPWRIGHT_WORKTREE_DIR}/{repo}-{branch-slug}` (or `git worktree add` it against the local branch), set `{worktree-path}`, then proceed.

plugins/shipwright/commands/dev-task.md
```diff
@@ -208,12 +197,105 @@ architecture to implementation. Auto-fix all quality issues.
...
+  base).
+- **PR case**: resume from the existing PR rather than rebuilding it. PATCH the task store to
+  match reality so downstream steps (Step 9, Step 9b, Step 10a) see the correct state:
+  ```bash
+  curl -sf -X PATCH -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
+    -H "Content-Type: application/json" \
+    "$SHIPWRIGHT_TASK_STORE_URL/tasks/{id}" \
+    -d "{\"status\": \"pr_open\", \"pr\": {existing-pr-number}}" | jq .
+  ```
+  Then skip Step 5 (implementation) entirely and continue from Step 6 (Simplify) using the
```
dodizzle: The PR-case "continue from Step 6 (Simplify)" also skips all of Step 4's `worktree add` commands and the line-343 CWD binding. Step 6 runs `git diff main...HEAD` and applies Edit-tool fixes, both of which require the worktree CWD — but "using the existing PR's branch" never instructs the agent to locate or create a worktree for it. After the task-store PATCH, ensure a worktree exists (`git worktree add ... origin/{branch} --track -b {branch}` if the prior session's is gone) and state that subsequent operations run from it.

plugins/shipwright/commands/dev-task.md
```diff
@@ -208,12 +197,105 @@ architecture to implementation. Auto-fix all quality issues.
...
+3. **If a PR exists**, additionally check its CI status before treating it as complete —
+   reuse the Step 9b Actions-API pattern (`gh api "repos/$GH_REPO/actions/runs?head_sha=..."`)
+   against the PR's head SHA. A PR with failing or still-running CI is not "complete and
+   correct" — treat it the same as the incomplete/stale case below.
+
+**Complete and correct** (diff satisfies every acceptance criterion, and CI is green if a PR
+exists): skip destructive recovery — do not close the PR or delete the branch.
+- **Local-only case** (branch exists, no PR yet): rebase is enough — no PATCH of `branch` or
+  `pr` is needed since neither changed; still refresh `heartbeatAt` (Step 5b already does
+  this before dispatch). Continue to the standard "branch DOES exist on remote" flow below
```
dodizzle: Edge case: this local-only path says "Continue to the standard 'branch DOES exist on remote' flow," which reuses the reality-check `git ls-remote` result. If the local branch exists but the remote branch is absent (pushed then deleted externally between sessions), that reused result routes to the "branch does NOT exist on remote" path (line 300) → `worktree add origin/main -b {branch}`, which fails because the local branch already exists. The never-pushed sub-case is handled; the pushed-then-deleted edge isn't. Consider treating remote-absent as a fresh start (delete/rename local branch first) or re-pushing before the standard flow.

dodizzle (commented):
Verdict: COMMENT — Both "complete and correct" resume shortcuts (local-only skip-to-Step-5 and PR-case continue-from-Step-6) hand off downstream without ever creating or binding a worktree, so `{worktree-path}` (line 385) is never resolved and Step 5/Step 6 have no valid working directory. The worktree-remove-before-`branch -D` fix from the prior review is correctly applied and CI is green.

plugins/shipwright/commands/dev-task.md
```diff
@@ -208,17 +197,126 @@ architecture to implementation. Auto-fix all quality issues.
...
+The reality check above already ran `git pull` and `git ls-remote --heads origin {branch}`
+for the local/remote signals — reuse that same remote-branch result here (indicates a
+bundled task joining an existing PR) rather than re-querying:
+
+**If the branch does NOT exist on remote** (new task, standard flow) — but first check
+whether a local branch `{branch}` already exists (from the reality check's `branch --list`
+result above). This happens when the branch was pushed then deleted externally between
+sessions (a remote-absent edge case, distinct from the never-pushed local-only case already
+handled above under "Complete and correct"):
+- **Local branch exists, remote does not** (pushed-then-deleted): do not run `origin/main -b
```
dmcaulay: This new "Local branch exists, remote does not" branch can't actually execute. The reality check's gate above ("If any of the three exist, assess whether the existing work is complete and correct") fires unconditionally whenever a local branch exists alone — that's exactly the "Local-only … Never pushed" sub-case under "Complete and correct," which short-circuits straight to Step 5 without ever reaching this section. If that same local-only branch is instead judged incomplete/stale, the cleanup steps above explicitly delete the local branch *before* falling through here. Either way, by the time control reaches this "does NOT exist on remote" section, a previously-existing local branch has already been resolved — there's no path that arrives here with the local branch still present.

Suggest removing this branch (it's dead prose describing a case the control flow already prevents), or restructuring the assessment gate if the intent was actually different.

dmcaulay (commented):
Verdict: COMMENT — The prior worktree-binding gap in the resume shortcuts is now fixed. This commit's other fix, for the remote-absent (pushed-then-deleted) edge case, adds a branch that's unreachable given the reality check's existing control flow — see inline comment.

dmcaulay:
Addressed the latest review round.

**Already fixed by earlier commits on this branch** (their inline threads are resolved on GitHub):
- @dodizzle 12:45 — worktree-remove-before-\`branch -D\` ordering in the incomplete/stale cleanup path (fixed in 11fa8da2).
- @dmcaulay 16:25 / @dodizzle 19:19 — \`{worktree-path}\` binding gap in both "complete and correct" resume shortcuts (fixed in 26427ec).

**This commit** (7114ff1) addresses the remaining thread flagged by @dmcaulay 20:21 / thread `PRRT_kwDOSy4Ihs6RlCXC`: removed the unreachable "Local branch exists, remote does not" bullet under the remote-absent section. Confirmed via control-flow trace that a local branch can never be present at that point — it's either never-existing, resolved by the local-only "complete and correct" shortcut, or deleted in the incomplete/stale cleanup — all of which precede this section. Simplified the surrounding prose to match. Lint and content tests pass; no test referenced the removed bullet.

dmcaulay (commented):
Verdict: APPROVE — Traced the full Branch/PR Reality Check control flow against this commit's claim that a local branch is guaranteed absent by the time execution reaches the standard fresh-worktree flow; confirmed true (every path either binds `{worktree-path}` explicitly via the complete-and-correct shortcuts, or resolves the local branch during incomplete/stale cleanup, before the fall-through). The unreachable dead-code branch from the prior round is correctly removed. No findings above the confidence threshold remain — clean to merge.

PR #1715, opened 2026-07-16
feat: narrow TaskService.list()'s blockedBy dependency lookup to referenced IDs only

dodizzle:
## Summary
- `TaskService.list()` now scopes its `blockedBy` dependency lookup to `task.findMany({ where: { id: { in: depIds } } })` (deduped union of `pageTasks[*].dependencies`), replacing an unconditional whole-table `task.findMany()`.
- Mirrors the existing scoped-lookup pattern already used in `TaskService.get()`.
- `listReady()` and `listBlocked()` are unchanged — they intentionally remain whole-table since they need the full dependency graph.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `GET /tasks` still returns identical `blockedBy` results for every existing case (HITL gate, terminal-status dependency, same-branch bundled dependency, missing/unknown dependency) | MET |
| 2 | The dependency lookup query is scoped to `id IN (...)` — no unconditional `task.findMany()` remains in `list()` | MET |
| 3 | `listReady()`/`listBlocked()` untouched | MET |
| 4 | Test decision: unit test asserting narrowed query resolves a same-page dependency correctly; real-DB integration test covering a dependency outside the current page | MET |

## Test Plan
- `task-store/src/task-service.unit.test.ts` — new suite asserting the scoped dependency lookup resolves out-of-page dependencies, dedupes shared dependency IDs into a single `id IN (...)` call, and short-circuits (no query) when no page task has dependencies.
- `task-store/src/tasks.integration.test.ts` — new real-DB cases: a dependency that falls outside the current pagination window (still resolves via ID-based lookup) and a same-page dependency sanity check.
- [x] `task lint` — clean
- [x] `task typecheck` (task-store) — clean
- [x] `bun test task-store/` — 316 pass, 0 fail, 99 skip (skips are pre-existing DB-gated integration tests; the two new integration cases are included in that DB-gated set and will run in CI against the provisioned test DB)

### DIFF
--- a/task-store/src/task-service.ts
+++ b/task-store/src/task-service.ts
@@ -153,8 +153,16 @@ export class TaskService implements TaskServiceLike {
     const limit = filters.limit ?? 50;
     const offset = filters.offset ?? 0;
 
-    // Load all tasks for dependency resolution (computeBlockedBy needs the full graph).
-    const [pageTasks, total, allTasks] = await this.prisma.$transaction([
+    // pageTasks + total are the only two queries that share the `where`
+    // filter, so batching them in a transaction is what actually mattered
+    // here — the old third whole-table findMany() depended on nothing else
+    // in the array and gained no consistency guarantee from sharing a
+    // transaction with it (Postgres already gives each individual query its
+    // own consistent snapshot). Splitting it out lets us compute depIds from
+    // the real pageTasks result first, then scope the dependency lookup by
+    // id — mirroring get()'s `where: { id: { in: task.dependencies } }`
+    // pattern — instead of always pulling every row in the table.
+    const [pageTasks, total] = await this.prisma.$transaction([
       this.prisma.task.findMany({
         where,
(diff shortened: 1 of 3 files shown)


dodizzle (commented):
Verdict: APPROVE — Scoped dependency lookup preserves identical blockedBy semantics for every case (HITL, terminal, same-branch bundled, missing/unknown, out-of-page) with no meaningful new race risk from splitting the $transaction array, and is backed by tests that genuinely validate the scoping behavior. (Posted as COMMENT — self-authored PR, GitHub blocks self-APPROVE via the API.)

dmcaulay (approved):
Verdict: APPROVE — Scoped dependency lookup correctly mirrors the existing get() pattern, is provably sufficient since computeBlockedBy only resolves direct dependency IDs (no transitive lookup), and is backed by tests that genuinely exercise cross-page and same-page dependency resolution. No breaking changes, no regressions.

PR #1718, opened 2026-07-16
feat: metrics client sends updatedSince/repo server-side instead of full-fetch-then-filter

dodizzle:
## Summary
- `HttpTaskStoreClient.listTasks()`/`listPrs()` now pass `updatedSince` (mapped from `params.from`) and `repo` (from `params.repo`) into the server-side query string via `fetchAllPages()`'s `baseParams`, instead of relying purely on client-side filtering after fetching every row.
- The server already supports both params (TSF-1.1, merged/deployed) — this wires the metrics client to actually send them, cutting down the amount of data fetched over the wire and paginated round-trips.
- `taskAnchor()`/`prAnchor()`/`inWindow()`/`matchesRepo()` are unchanged and still run as the precise final filter — `updatedSince` is a conservative, inclusive-only pre-filter with no window-end equivalent, so client-side filtering remains necessary for correctness. Updated the stale comments that claimed these params were fully ignored server-side.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | listTasks/listPrs requests include updatedSince/repo in the query string | MET |
| 2 | Dashboard /metrics/summary, /trends, /features, /queue return identical results to before, just over fewer paginated round-trips | MET |
| 3 | No changes to taskAnchor()/prAnchor()/inWindow()/matchesRepo() or to the offline/recorded-fixture provider path | MET |
| 4 | Extend metrics/src/lib/task-store-client.unit.test.ts to assert the built request URL includes updatedSince/repo; no changes to task-store-provider.integration.test.ts; no existing tests retired | MET |

## Test Plan
- Added 4 unit tests in `task-store-client.unit.test.ts` asserting `updatedSince`/`repo` are present in every paginated request URL when passed, and absent when omitted, for both `listTasks()` and `listPrs()`.
- All pre-existing pagination/window-filtering/repo-filtering tests remain green (byte-identical filtering behavior).
- [x] `bun run --filter='*' --sequential typecheck` clean across all workspaces
- [x] `NODE_ENV=test bun test --coverage` — 4024 pass, 0 fail
- [x] `task-store-client.ts` coverage: 95.4% (repo-wide aggregate coverage gate is at 79.27%, a pre-existing gap unrelated to this 2-file change)

### DIFF
--- a/metrics/src/lib/task-store-client.ts
+++ b/metrics/src/lib/task-store-client.ts
@@ -104,11 +104,12 @@ export interface TaskStoreClient {
 
 // ─── Window filtering ─────────────────────────────────────────────────────────
 //
-// The live task store ignores `from`/`to` query params (the `/tasks` and `/prs`
-// routes read only `limit`/`offset`), so the client must window-filter the rows
-// it fetches. These helpers are the single source of truth for that filtering —
-// the recorded test double imports them so the cassette path and the live HTTP
-// path apply byte-identical windowing.
+// The live task store accepts `updatedSince` (mapped from `from`) as a coarse,
+// inclusive-only pre-filter — it has no window-end (`to`) equivalent and is not
+// a precise sync anchor. The client must still window-filter the rows it
+// fetches to apply the exact [from, to] boundary. These helpers are the single
+// source of truth for that filtering — the recorded test double imports them
+// so the cassette path and the live HTTP path apply byte-identical windowing.
 
 /** Pick the timestamp a task is "anchored" to for window filtering. */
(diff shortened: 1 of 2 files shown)


dodizzle (commented):
Verdict: APPROVE — Small, correctly scoped query-optimization change. Verified server-side updatedSince/repo support directly against task-store routes/services; updatedAt is bumped on the same write that sets anchor fields, so the inclusive gte pre-filter can't exclude anything the existing client-side filter would keep. No regressions found.

PR #1369, opened 2026-07-09
feat: unify claim TTL to one 35-minute constant for Task + PR

dmcaulay:
## Summary
- Retires `SHIPWRIGHT_TASK_STORE_TASK_CLAIM_TTL_MS`; `StaleClaimReaper` and `PullRequestService.claimNext()` now both read the single `SHIPWRIGHT_TASK_STORE_CLAIM_TTL_MS` env var (default `2_100_000` ms / 35 min — the 30-minute `claude -p` hard timeout plus a 5-minute buffer)
- `StaleClaimReaper` constructor drops the separate `taskTtlMs` parameter; Task and PullRequest reap queries share the same cutoff
- `docs/configuration.md` and `site/src/data/config-vars.ts` updated to remove the retired var and document the unified default
- Part of the claim-lifecycle-unification plan fixing PR app-vitals/shipwright#1321's review pile-up

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | StaleClaimReaper constructor takes one ttlMs (default from the single env var), used for both the Task and PullRequest reap queries | MET |
| 2 | claimNext()'s cutoff references the same constant/env var name | MET |
| 3 | docs/configuration.md updated to remove the retired var and reflect the new default | MET |
| 4 | stale-claim-reaper.unit.test.ts updated to cover both Task and PR reaping at the unified TTL | MET |

## Test Plan
- [x] `task-store/src/stale-claim-reaper.unit.test.ts` — 20/20 pass, covers both Task and PR reap at the unified TTL including boundary cases
- [x] `task lint` — pass
- [x] `task typecheck` — pass
- [x] `task check-strings` — pass
- [x] `task check-config-docs` — pass (31/31 env vars documented)
- [x] `task check-version-sync` — pass
- [x] `task test:coverage` — 15 pre-existing failures reproduce identically on clean `main` (kubernetes-client, installPlugins, chat-token-reporter — unrelated sandbox flakes); no task-store test regressions

Generated with [Claude Code](https://claude.com/claude-code)

PR #1050, opened 2026-07-03
feat: raise PR claim TTL to 15min and add heartbeat renewal to review.md

dmcaulay:
## Summary
- Raises the default PR claim TTL (`SHIPWRIGHT_TASK_STORE_CLAIM_TTL_MS`) from 5 minutes to 15 minutes (`300_000` → `900_000`ms) in `stale-claim-reaper.ts` and the fallback in `pull-request-service.ts`, keeping the value operator-overridable via the env var.
- Syncs the new default across `docs/configuration.md` and `site/src/data/config-vars.ts`.
- Adds an explicit heartbeat renewal step to `review.md` (new Step 5, item 8) between context-gathering and the review-writing/posting steps, so long reviews don't get their claim reaped mid-review.

Root cause: the stale-claim reaper resets `reviewState` to `pending` for any `phase='review'` claim whose heartbeat exceeds the TTL. No command file ever called the heartbeat endpoint, and real reviews often exceed the old 5-minute default — confirmed cause of PR #1016 receiving 5 duplicate review posts against an unchanged commit over 9 hours.

## Acceptance Criteria

| # | Criterion | Status | Evidence |
|---|-----------|--------|----------|
| 1 | Default TTL raised to 900_000ms in stale-claim-reaper.ts + fallback in pull-request-service.ts, still operator-overridable | MET | `DEFAULT_TTL_MS = 900_000` in stale-claim-reaper.ts; `?? 900_000` fallback in pull-request-service.ts; both still read `SHIPWRIGHT_TASK_STORE_CLAIM_TTL_MS` first |
| 2 | docs/configuration.md and config-vars.ts reflect 900000, all three sources in sync | MET | Both docs updated to `900000`, matching the reaper constant |
| 3 | review.md heartbeat step between Step 5 and review-writing/posting | MET | New step inserted after Step 5's context-gathering, before Step 6 (Classify Changes) and Step 7 (Deep Review) |
| 4 | stale-claim-reaper.unit.test.ts extended with 900_000ms boundary case, additive only | MET | 2 new tests added (just-under / just-over TTL boundary), no existing tests retired |

## Test Plan
- `bun test` in `task-store/` — 207 pass, 0 fail (16/16 in `stale-claim-reaper.unit.test.ts`, including 2 new boundary tests)
- `bun run typecheck` in `task-store/` — clean
- [x] Pre-commit checks passing

Generated with [Claude Code](https://claude.com/claude-code)

PR #1016, opened 2026-07-02
fix: scope check-patch self-review exclusion to APPROVE-verdict only

dmcaulay:
## Summary
- Narrow the self-authored-review exclusion in `hasUnaddressedFindings` and `hasMergeOnlyStaleFindings` (`plugins/shipwright/scripts/check-patch.ts`) from blanket author-based exclusion to APPROVE-verdict-body exclusion, matching `deploy.md`'s `trimStart().startsWith("APPROVE")` convention.
- A self-review with real findings (non-APPROVE body) now still counts as an unaddressed finding; only a clean self-approval is excluded.
- Extracted the shared predicate into a new `isSelfCleanApprove` helper to avoid duplicating the check across both functions.

Fixes the over-correction from CPF-1.1 (#998), which excluded ALL self-authored reviews regardless of content — silently dropping self-reviews that found real issues.

## Acceptance Criteria

| # | Criterion | Status | Evidence |
|---|-----------|--------|----------|
| 1 | Both functions replace `r.author.login !== currentUser` with an APPROVE-verdict-body check | MET | Both now use `!isSelfCleanApprove(r, currentUser)` |
| 2 | 3 existing CPF-1.1 self-review tests pass unmodified | MET | All 3 use APPROVE bodies, unchanged, still pass |
| 3 | 2 new regression tests (non-APPROVE self-review at HEAD; stale + merge-only) | MET | Added, both pass |
| 4 | Doc comments updated on both functions | MET | Rewritten to describe APPROVE-only exclusion |
| 5 | `task ci` passes (lint, check-strings, typecheck, full test suite) | MET | lint clean, check-strings clean, typecheck clean (6/6 packages), full suite 2942 pass / 14 fail — all 14 failures confirmed pre-existing on `main` (environment-specific: local CA cert, local HOME/marketplace path), unrelated to this diff |

## Test Plan
- `bun test plugins/shipwright/scripts/check-patch.test.ts` — 27 pass, 0 fail
- `bunx biome lint .` — clean
- `bun plugins/shipwright/scripts/check-banned-strings.ts` — clean
- `bun run --filter='*' typecheck` — clean across all 6 packages
- [x] Pre-commit checks passing

Generated with [Claude Code](https://claude.com/claude-code)

PR #1741, opened 2026-07-17
fix(task-store): release PR claim in the same write that posts or approves a review

EugeneTrapeznikov:
## Problem

A review session never releases the PR claim it holds. It transitions the PR to `posted` (via `POST /prs/:id/complete`) or `approved` (via `PATCH /prs/:id`) and then exits, leaving `claimedBy`/`claimedAt`/`heartbeatAt`/`phase` set. Once that claim ages past the `StaleClaimReaper` TTL, the reaper's sweep releases it — and, because the reap keyed the reset on `phase = 'review'`, it also regressed `reviewState` back to `pending`. A reviewed-but-unmerged PR then looks like fresh review work and gets a duplicate review dispatched against it.

This is the failure behind #1016 (5 duplicate reviews of one PR over ~9h). #1050 and #1369 lengthened the claim/reaper TTLs, which widens the window but does not fix the semantics: any PR that stays unmerged longer than one TTL still re-triggers the bug. On a self-hosted setup with a human merge gate, PRs routinely sit reviewed-but-unmerged for longer than a TTL, so the regression fires reliably.

## Evidence from my deployment

I run a self-hosted agent with a human merge gate. A concrete instance:

- 05:02 — review completed, `reviewState` set to `posted`; the claim was never released.
- ~05:47 — the TTL sweep reaped the stale claim and regressed `posted` -> `pending`.
- 05:48 — a duplicate review was dispatched on the same head SHA. It re-reviewed an already-reviewed PR and posted nothing new, at a cost of ~814K wasted input tokens.

After deploying this change I observed review completions clearing the claim in the same write (a `reviewCycles` increment with the claim fields going null under a single `updatedAt`), and a full TTL window elapsed with no duplicate review dispatched.

One note on scope: my initial version of the fix patched only `update()`. Production showed the review flow actually completes via `POST /prs/:id/complete` (`PullRequestService.complete()`), so a claim posted through that path still lingered until the reaper. The fix therefore covers both write paths, and the integration tests drive the real routes rather than the service methods directly.

## Fix

Three parts:

1. `PullRequestService.update()` — when a PATCH sets `reviewState` to `posted` or `approved`, clear `claimedBy`/`claimedAt`/`heartbeatAt`/`phase` in the same write. This mirrors the adjacent `state: 'merged'` block that already clears the claim, and sits alongside the existing `approved` -> `readyForDeployAt` auto-stamp.
2. `PullRequestService.complete()` — the `POST /prs/:id/complete` path review sessions actually use. Same clear in the same write that sets `reviewState = 'posted'`.
3. `StaleClaimReaper` — the PR bulk UPDATE still releases the claim fields for every stale claim, but now only regresses `reviewState` -> `pending` when it is currently `pending`/`in_progress`. A stale claim on a record already at `posted`/`approved` releases without regressing `reviewState`, so it no longer re-dispatches a duplicate review.

Precedent: the handler already auto-stamps `readyForDeployAt` on `approved` and clears the claim on `state = 'merged'`; this extends the same same-write pattern to review completion. The reaper narrowing is belt-and-suspenders — it keeps stale claims releasing while no longer regressing a completed review, which also covers claims that predate this change.

## Tests

- Unit (`pull-request-service.unit.test.ts`): `update()` clears the claim on `posted` and on `approved` (and still stamps `readyForDeployAt`), the release wins over claim fields supplied in the same PATCH body, a non-`reviewState` update leaves the claim untouched, re-asserting `posted` is idempotent; `complete()` clears the claim while preserving `reviewCycles`/`reviewState`/`reviewedAt`/`readyForPatchAt`.
- Unit (`stale-claim-reaper.unit.test.ts`): the reap SQL regresses `reviewState` only for `pending`/`in_progress` (CASE with `ELSE "reviewState"`) and releases claim fields for every stale claim regardless of `reviewState`.
- Integration, real Postgres (`stale-claim-reaper.integration.test.ts`): a stale claim seeded per `reviewState` plus a fresh control — `pending`/`in_progress` regress to `pending` and release; `posted`/`approved` keep their `reviewState` and release; the fresh control is untouched.
- Integration, route layer (`pull-request.integration.test.ts`): drives the actual `POST /prs/:id/complete` and `PATCH /prs/:id` (approved) endpoints through the Hono app with a real service, reading the row back to prove the claim was released in the same write.
- Docs: `docs/task-store.md` gains the side-effect sentence for the `posted`/`approved` release.

`bun test` passes with 0 failures; `tsc --noEmit` is clean for task-store; the real-DB integration tests pass against a throwaway `postgres:16-alpine` after `prisma migrate deploy`.

### DIFF
--- a/docs/task-store.md
+++ b/docs/task-store.md
@@ -252,6 +252,8 @@ Writable fields: `staged`, `commitSha`, `taskId`, `agentId`, `state`, `mergedAt`
 
 **Side effect:** When `state` is set to `merged` or `closed`, the claim fields (`claimedBy`, `claimedAt`, `heartbeatAt`, `phase`) are automatically cleared. This ensures that merged or closed PRs are no longer held by an agent claim.
 
+**Side effect:** When `reviewState` is set to `posted` or `approved`, the claim fields (`claimedBy`, `claimedAt`, `heartbeatAt`, `phase`) are cleared in the same write. This releases the review claim as soon as the review is done, so a PR awaiting patch/deploy is not held by a stale claim that the reaper would otherwise reap (which would regress `reviewState` and re-dispatch a duplicate review).
+
 #### PR lifecycle endpoints
 
 | Endpoint | Effect |
--- a/task-store/src/pull-request-service.ts
+++ b/task-store/src/pull-request-service.ts
@@ -181,6 +181,20 @@ export class PullRequestService implements PullRequestServiceLike {
         }
       }
 
+      // When a review session transitions reviewState to 'posted' or
+      // 'approved' it is done with the record, so release its claim in the same
+      // write — mirroring the state:'merged' block below. Review skills
+      // historically left claimedBy/claimedAt/heartbeatAt/phase set and relied
+      // on a follow-up release call that agents often skipped; a claim left
+      // dangling past the reaper TTL then got reaped and its reviewState
+      // regressed, re-dispatching a duplicate review (app-vitals/shipwright#1016).
+      if (data.reviewState === "posted" || data.reviewState === "approved") {
+        updateData.claimedBy = null;
+        updateData.claimedAt = null;
+        updateData.heartbeatAt = null;
+        updateData.phase = null;
+      }
+
       // deploy.md's merge-completion PATCH transitions state to 'merged' —
       // explicitly release the claim in the same call, mirroring patch()'s
       // always-release behavior. Defense-in-depth: claimNext()'s WHERE clause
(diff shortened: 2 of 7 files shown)


dmcaulay (approved):
Verdict: APPROVE — Root-causes a documented recurring bug (#1016): review sessions never released their PR claim, so the stale-claim reaper would later regress reviewState back to pending and trigger a duplicate review. This clears the claim in the same write in both `update()` and `complete()` (the path review sessions actually use), and narrows the reaper's regression logic to only reset pending/in_progress claims, preserving posted/approved. Verified the new claim-release block composes safely with the existing state:merged/closed block (same fields, same values, order-independent), and that the reaper's WHERE clause is unchanged — only the SET-clause CASE gate moved from phase to reviewState. Well-tested across unit, real-Postgres integration, and route-layer integration.

PR #1764, opened 2026-07-17
feat: warn when the loop dispatches the same item N times in a row (spin detection)

dmcaulay:
## Summary
- Add module-scoped spin detection to `runLoopTick()` in `agent/src/loop-orchestrator.ts`: tracks the last-dispatched itemId and a consecutive-repeat counter, mirroring the existing `busy` guard
- When the same itemId is dispatched 3+ times in a row, emit `console.warn` identifying the item and repeat count on every subsequent dispatch past the threshold (not just at the crossing), so the signal persists and escalates in Sentry via `consoleLoggingIntegration` for as long as the spin continues
- Refresh `docs/agent.md`'s `shipwright-loop` cron description to mention spin detection

## Acceptance Criteria

| 1 | Three consecutive dispatches against the same itemId produce a console.warn identifying the item and repeat count | MET | Counter increments per matching itemId; console.warn fired at threshold >=3 with itemId and count in message |
| 2 | A different itemId dispatched in between resets the counter | MET | Else branch resets consecutiveDispatchCount=1 on itemId mismatch; covered by unit test |
| 3 | Re-warn cadence design decision stated | MET | Code comment states: warn on every dispatch once threshold reached (not just at crossing) so signal persists/escalates in Sentry |
| 4 | State is module-scoped like busy flag, no new persisted storage | MET | lastDispatchedItemId/consecutiveDispatchCount declared as let in same closure as busy, in-memory only |
| 5 | Unit test added to loop-orchestrator.unit.test.ts, purely additive | MET | 3 new tests added; all 22 existing+new tests pass |

## Test Plan
- `bun test loop-orchestrator.unit.test.ts` — 22 pass, 0 fail (includes 3 new spin-detection tests: warns on 3rd consecutive dispatch, no warn on 1st/2nd, counter resets on intervening different itemId)
- `bun run typecheck` — clean


dmcaulay (commented):
Verdict: APPROVE — Spin-detection counter logic is correct, tests cover the stated acceptance criteria, and the docs/agent.md update and Sentry-escalation claim both check out against the actual code.

PR #1795, opened 2026-07-17
feat: opt-in warning if claim TTL doesn't cover claude timeout + buffer

dmcaulay:
## Summary
- Add a pure, exported `checkClaimTtlBuffer(ttlMs, claudeTimeoutMs)` function that warns when the task-store's resolved claim TTL doesn't leave enough headroom over the agent's configured `claude -p` timeout + `CLAIM_TTL_BUFFER_MS`
- Wire it into `task-store/src/main.ts` startup, next to the `StaleClaimReaper` construction, reading an optional local `SHIPWRIGHT_CLAUDE_TIMEOUT_MS` copy from task-store's own env (opt-in, zero new required config)
- Expose `StaleClaimReaper.ttlMs` publicly so startup code can validate it
- Document both `SHIPWRIGHT_TASK_STORE_CLAIM_TTL_MS` and `SHIPWRIGHT_CLAUDE_TIMEOUT_MS` in `docs/configuration.md`, including the N:1 fleet guidance (use the max timeout across the whole agent fleet)

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | `checkClaimTtlBuffer` pure function + unit tests (undefined→null, sufficient→null, insufficient→warning naming both values) | MET |
| 2 | Wired into main.ts: reads `SHIPWRIGHT_CLAUDE_TIMEOUT_MS`, calls check vs reaper's `ttlMs`, console.warns | MET |
| 3 | docs/configuration.md updated for both env vars w/ N:1 fleet guidance | MET |
| 4 | Unit-test-only layer (no I/O), main.ts wiring untested by design | MET |
| 5 | Safe to deploy standalone (additive, no renames/removals) | MET |

## Test Plan
- 5 new unit tests in `task-store/src/claim-ttl-buffer-check.unit.test.ts` covering the undefined/sufficient/insufficient cases plus both boundary sides
- `bunx biome lint .` clean
- `bun run --filter="*" --sequential typecheck` clean across all workspaces
- `NODE_ENV=test bun test` — full suite green (4075 pass / 0 fail, 327 skip — pre-existing unrelated skips)


task-store/src/main.ts
```diff
@@ -125,6 +126,27 @@ async function startServer(): Promise<void> {
...
+  // (a second copy of the same value, not new cross-service coupling), warn
+  // when the resolved claim TTL doesn't leave enough headroom over it. In an
+  // N:1 fleet (many agents sharing one task-store), use the MAX
+  // SHIPWRIGHT_CLAUDE_TIMEOUT_MS across all agents, not any single agent's
+  // value.
+  const rawClaudeTimeoutMs = process.env.SHIPWRIGHT_CLAUDE_TIMEOUT_MS;
+  const parsedClaudeTimeoutMs = rawClaudeTimeoutMs
+    ? Number(rawClaudeTimeoutMs)
+    : undefined;
+  const claudeTimeoutMs = Number.isFinite(parsedClaudeTimeoutMs)
```
dodizzle, line 142: When `SHIPWRIGHT_CLAUDE_TIMEOUT_MS` is set but non-numeric (typo), `Number.isFinite` rejects the `NaN` and this silently falls back to the same `undefined` path as "unset" — no warning that the value was present but unparseable. Given this check exists specifically to catch misconfiguration, consider a distinct warning for "set but invalid" vs. "unset". Not blocking.

dodizzle (approved):
Verdict: APPROVE — Clean, well-tested, additive opt-in safety check. One minor non-blocking suggestion left as an inline comment.

PR #1804, opened 2026-07-17
feat: patch subagent posts rebuttal comment for REJECT-only findings

dmcaulay:
## Summary
- When the patch fix subagent classifies every review finding as REJECT (no code changes, no commit), it now posts a `gh pr comment` rebuttal summarizing why each finding was rejected instead of silently leaving the review unaddressed.
- This activates the existing `isAddressedByAuthorReply` escape hatch in `check-patch.ts`, so the review stops being reflagged as unaddressed on every subsequent patch cron run.
- Content-only change to `patch.md` (Step 5b Instructions [D]/[F], Step 5c) — the existing ACCEPT/MODIFY commit+push flow is untouched.

## Acceptance Criteria
| Criterion | Status |
|-----------|--------|
| Step 5b requires posting a `gh pr comment` rebuttal when STATUS is DONE_WITH_CONCERNS with zero files changed (all findings REJECT) | MET |
| Step 5c's DONE_WITH_CONCERNS-with-no-push branch confirms the rebuttal comment was posted before proceeding | MET |
| Existing ACCEPT/MODIFY code-fix flow (Step 5c DONE path with a real push) is unchanged | MET |
| Test coverage added in `patch.content.test.ts` asserting the gated instruction text; no existing tests retired | MET |

## Test Plan
- [x] `bun test commands/patch.content.test.ts` — 25 pass (19 pre-existing + 6 new)
- [x] Full `plugins/shipwright` suite — 694 pass, 0 fail
- [x] `check-banned-strings.ts` — clean
- [x] Independent spec-compliance review — all 4 criteria MET

### DIFF
--- a/plugins/shipwright/commands/patch.md
+++ b/plugins/shipwright/commands/patch.md
@@ -642,10 +642,29 @@ INSTRUCTIONS — follow in order:
   - Re-run until both pass cleanly
 
 [D] Commit
-  - Stage only the files you changed: `git add {changed files}`
-  - Commit with a conventional commit message describing what was fixed:
-    "fix: address review findings on #{pr} — {one-line summary of changes}"
-  - Push: `git push origin {branch}`
+  - **If at least one finding was ACCEPTED or MODIFIED** (i.e. you have file changes
+    staged):
+    - Stage only the files you changed: `git add {changed files}`
+    - Commit with a conventional commit message describing what was fixed:
+      "fix: address review findings on #{pr} — {one-line summary of changes}"
+    - Push: `git push origin {branch}`
+  - **If every finding was classified REJECT in [A.5]** (premise incorrect on all of
+    them — no files changed, nothing to commit): do not commit or push. Instead, post a
+    PR-level rebuttal comment explaining why each finding was rejected, so the review is
+    not left looking unaddressed:
+    ```bash
+    gh pr comment {pr} --repo {org}/{repo} --body "$(cat <<'EOF'
+    Reviewed the finding(s) above and did not implement changes — premise did not hold:
+
+    - {finding summary}: {reason rejected}
+    - {finding summary}: {reason rejected}
+    EOF
+    )"
+    ```
+    List every rejected finding, one bullet per finding, drawn from the CONCERNS you
+    compiled in [A.5]. This comment is what allows a future patch run to recognize the
+    review was addressed (rejected with reasoning, not ignored) instead of reflagging it
+    forever.
 
 [E] Resolve addressed inline threads
   PR-level comments cannot be resolved programmatically — skip them here.
@@ -672,6 +691,10 @@ INSTRUCTIONS — follow in order:
   {bullet list of each finding addressed and how}
 
   CONCERNS: (if DONE_WITH_CONCERNS)
+  {if every finding was REJECT and no push happened, explicitly confirm here that the
+  [D] rebuttal comment was posted, e.g. "All findings rejected (premise incorrect) — no
+  code changes; posted rebuttal comment via gh pr comment summarizing why each was
+  rejected." Otherwise, describe the correctness gap as usual.}
   BLOCKER: (if BLOCKED)
 ```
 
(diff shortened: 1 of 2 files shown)


dodizzle (commented):
Verdict: COMMENT — The rebuttal-comment mechanism only stops reflagging for review-body-level findings; it doesn't help when findings arrive as unresolved inline threads (the common case), since check-patch.ts's hasUnaddressedFindings() short-circuits on any unresolved thread before isAddressedByAuthorReply is ever consulted, and Step [E] explicitly skips resolving threads for REJECTed ("not applicable") findings.

plugins/shipwright/commands/patch.md
```diff
@@ -642,10 +642,32 @@ INSTRUCTIONS — follow in order:
...
+    #   - {finding summary}: {reason rejected}
+    #   - {finding summary}: {reason rejected}
+    gh pr comment {pr} --repo {org}/{repo} --body-file /tmp/shipwright-patch-rebuttal-{pr}.txt
+    rm /tmp/shipwright-patch-rebuttal-{pr}.txt
+    ```
+    The temp file path MUST include the PR number to avoid collisions — `/tmp` is shared
+    across all worktrees. List every rejected finding, one bullet per finding, drawn from
+    the CONCERNS you compiled in [A.5]. This comment is what allows a future patch run to
+    recognize the review was addressed (rejected with reasoning, not ignored) instead of
+    reflagging it forever.
```
dodizzle: This claim is overstated for the common case. `check-patch.ts`'s `hasUnaddressedFindings()` short-circuits to `true` whenever `unresolvedThreads.length > 0` — before `isAddressedByAuthorReply` is ever consulted, and independent of it. Step [E] (right below) only resolves threads whose finding "was addressed" and explicitly skips threads that "were not applicable" — a REJECTed finding is exactly that, so its inline thread stays unresolved. This file's own pre-existing text (lines 204-206) already documents this: "an unresolved thread on the same review continues to count as a finding regardless of any reply." Since `/shipwright:review` maps any finding with a `file:line` reference to an inline comment, most real findings arrive as inline threads — so this fix won't prevent reflagging for the majority of all-REJECT cases. Consider having [E] also resolve threads for REJECTed findings (the rebuttal comment is itself the explanation), or caveat that this mechanism only covers review-body-level findings.

plugins/shipwright/commands/patch.md
```diff
@@ -659,9 +699,11 @@ INSTRUCTIONS — follow in order:
     }
   }'
   ```
-  Run this for each Thread ID that was addressed. Skip threads whose findings were not
-  applicable or were not addressed. Do not attempt to resolve PR-level comments — they
-  have no resolution mechanism.
+  Run this for each Thread ID whose finding was fixed. Threads for REJECTed findings were
+  already resolved in [D] above — do not process them again here. Skip only threads whose
```
dmcaulay: In a mixed ACCEPT+REJECT outcome, [D] branch 1 ("at least one finding ACCEPTED/MODIFIED") fires and never posts a rebuttal or resolves the REJECTed finding's thread — only branch 2 (all-REJECT) does that. This sentence's blanket claim ("already resolved in [D] above") is false in that scenario, which could mislead a future subagent run into believing the thread was handled when it wasn't. Not a regression (pre-PR, [E] already left these threads unresolved in the mixed case), but the new wording should be scoped to the all-REJECT branch, or [D] branch 1 should also handle REJECTed findings alongside the ACCEPT/MODIFY commit.

dmcaulay (commented):
Verdict: COMMENT — All-REJECT fix is correct and verified against check-patch.ts, but [E]'s new wording falsely claims REJECTed threads were "already resolved in [D] above" for the mixed ACCEPT+REJECT case, where [D] branch 1 never touches them.

dmcaulay (commented):
Verdict: COMMENT — The prior finding (mixed-case [D]/[E] binary-framing mismatch) is genuinely fixed, but the same fix leaves the mixed ACCEPT+REJECT case's rebuttal/thread-resolution step unverified by the orchestrator (Step 5c/[F] only check for it on the all-REJECT/no-push path), which can silently reproduce the original reflagging-forever bug via a different path.

plugins/shipwright/commands/patch.md
```diff
@@ -682,7 +737,17 @@ Parse the subagent's STATUS:
 - **DONE**: Record the findings addressed. Proceed to Step 5c.5 (upsert PR record).
 - **DONE_WITH_CONCERNS**: Read concerns. If they are correctness gaps, log them in the
```
dmcaulay: The mixed ACCEPT+REJECT case now requires the subagent to post a rebuttal comment and resolve REJECTed inline threads (per [D] branch 2, which fires here too), but this DONE_WITH_CONCERNS branch only verifies that happened when "the subagent did not push" (the all-REJECT sub-case). In the mixed sub-case (push already happened), the guidance is just "log them in the report but proceed... to Step 5c.5" — nothing checks that the rebuttal+resolve step for the REJECTed findings was actually performed. [F]'s report template has the same asymmetry (explicit confirmation only prompted "if every finding was REJECT and no push happened"). If a subagent forgets the secondary rebuttal/resolve step while handling a mixed run, nothing catches it, and the REJECTed findings' inline threads stay unresolved — reproducing this PR's original bug via the mixed-case path instead of the all-REJECT path. Suggest extending Step 5c's confirmation check and [F]'s report template to require rebuttal/thread-resolution confirmation whenever CONCERNS lists any REJECTed finding, not only when there was no push.

dodizzle (approved):
Verdict: APPROVE — Closes the mixed ACCEPT+REJECT-with-push gap flagged in the prior review; [D]/[E] thread-resolution split is now non-contradictory, the check-patch.ts claims check out against source, and Step 5c.5 upsert routing is correct for both the all-REJECT/no-push and mixed/with-push cases.

PR #1814, opened 2026-07-17
fix: reset reviewState to pending after an all-REJECT, no-diff patch cycle

dmcaulay:
## Summary
- Fix `patch.md` Step 5c so an all-REJECT, no-diff patch cycle no longer skips Step 5c.5 entirely — it now always proceeds, carrying forward whether the rebuttal was confirmed.
- Step 5c.5 conditionally issues `PATCH /prs/{id}` with `{"reviewState": "pending"}` when the cycle was DONE_WITH_CONCERNS + zero files changed + rebuttal comment posted, letting the PR re-qualify in `check-review.ts`'s dedup despite an unchanged `headRefOid` so it can earn a real APPROVE.
- The ACCEPT/MODIFY push-a-commit path is untouched — the reset never fires when a new commit SHA already re-qualifies the PR naturally.

## Acceptance Criteria
| 1 | Step 5c.5 conditionally PATCHes reviewState:pending on the all-REJECT/no-push/rebuttal-confirmed case; existing patchCycles bump still fires unconditionally | MET | New `if [ "$ALL_REJECT_NO_PUSH_REBUTTAL_CONFIRMED" = "true" ]` block added after the existing unconditional heartbeat + commitSha POST |
| 2 | ACCEPT/MODIFY push path unchanged; reset never fires on a path with a new commitSha | MET | Step 5b [D] untouched; new test asserts no `reviewState` mention there |
| 3 | check-review.ts needs no code change; existing test already covers reviewState:pending + unchanged commitSha eligibility | MET | Confirmed via `check-review.ts`'s dedup gate and the existing test "returns a candidate when PR record has reviewState=pending (even if commitSha matches)" — no code/test change needed there |
| 4 | Add coverage in patch.content.test.ts; no existing tests retired | MET | New "RPF-1.2" describe block added; the one pre-existing test whose assertion contradicted the new behavior was rewritten (not deleted) |

## Test Plan
- `NODE_ENV=test bun test plugins/shipwright/commands/ agent/src/check-review.unit.test.ts` — 253 pass, 0 fail
- `bunx biome lint` — clean
- `tsc --noEmit` (plugins/shipwright) — clean

### DIFF
--- a/plugins/shipwright/commands/patch.md
+++ b/plugins/shipwright/commands/patch.md
@@ -741,23 +741,27 @@ Parse the subagent's STATUS:
 - **DONE**: Record the findings addressed. Proceed to Step 5c.5 (upsert PR record).
 - **DONE_WITH_CONCERNS**: Read concerns. If any concern reports a REJECTed finding (per
   Step 5b Instructions [D], this fires whenever at least one finding was REJECTed —
-  whether that was the *only* outcome (the all-REJECT, no-diff path) or one branch of a
-  mixed ACCEPT+REJECT run where a push also happened), confirm the subagent's CONCERNS
-  text reports both that it already posted the required `gh pr comment` rebuttal AND that
-  it resolved the inline threads for the REJECTed findings. Both are needed — the rebuttal
-  activates the `isAddressedByAuthorReply` escape hatch in `check-patch.ts`, but
+  whether every finding in the run was REJECTed with no push at all, one branch of a mixed
+  ACCEPT+REJECT run where a push also happened, or a mixed run where no push happened
+  because every ACCEPTED/MODIFIED finding in that run resolved to a zero-diff no-op
+  alongside the REJECTed one(s)), confirm the subagent's CONCERNS text reports both that it
+  already posted the required `gh pr comment` rebuttal AND that it resolved the inline
+  threads for the REJECTed findings. Both are needed — the rebuttal activates the
+  `isAddressedByAuthorReply` escape hatch in `check-patch.ts`, but
   `hasUnaddressedFindings()` short-circuits to `true` on any unresolved inline thread before
   that escape hatch is even consulted, so the threads must also be resolved or the review
   stops being reflagged only for body-level findings, not inline ones (the common case).
   Do not post the comment here or resolve the threads here; Step 5c only verifies it
   already happened. If the report doesn't confirm both, treat it as a concern in the final
-  report (the reflagging loop will otherwise persist via this mixed-case path, not just the
-  all-REJECT path). For other, non-REJECT correctness-gap concerns, just log them in the
-  report. Either way, always proceed to Step 5c.5 (upsert PR record) — in the mixed case
-  there IS a new commit SHA to record, and in the all-REJECT, no-push case Step 5c.5 still
-  needs to run so it can reset `reviewState` to `pending` (see below), even though there is
-  no new commit SHA. Carry forward into Step 5c.5 whether this was the all-REJECT, no-push,
-  rebuttal-confirmed case — that's the condition that gates the `reviewState` reset there.
+  report (the reflagging loop will otherwise persist regardless of which no-push variant
+  produced it). For other, non-REJECT correctness-gap concerns, just log them in the
+  report. Either way, always proceed to Step 5c.5 (upsert PR record) — when a push happened
+  there IS a new commit SHA to record, and when no push happened Step 5c.5 still needs to
+  run so it can reset `reviewState` to `pending` (see below), even though there is no new
+  commit SHA. Carry forward into Step 5c.5 whether this cycle had no push
+  (`HEAD_SHA_POST_PATCH` unchanged from before dispatch) with at least one REJECTed finding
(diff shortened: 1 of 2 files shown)


plugins/shipwright/commands/patch.md
```diff
@@ -786,11 +788,31 @@ if [ -n "$PR_RECORD_ID" ]; then
...
+      "$SHIPWRIGHT_TASK_STORE_URL/prs/$PR_RECORD_ID" \
+      -d '{"reviewState": "pending"}' > /dev/null 2>&1 || \
+      echo "⚠ PATCH /prs/$PR_RECORD_ID reviewState reset failed — continuing"
+  fi
 else
   echo "⚠ no PR_RECORD_ID from pre-work claim — skipping PR record update"
 fi
 ```
 
+`ALL_REJECT_NO_PUSH_REBUTTAL_CONFIRMED` is the condition carried forward from Step 5c: set
```
dmcaulay: **Confidence: 80** — This condition assumes "no push" and "all-REJECT" always co-occur, but Step 5b [D]'s push condition is actually gated on "you have file changes staged", not on finding classification. A mixed run where a finding is classified ACCEPT/MODIFY but produces zero diff (e.g. the subagent determines the issue is already resolved, or a MODIFY's correct fix is a no-op) would have no push, yet wouldn't be "all-REJECT" as literally defined here — the PR could get stuck again via a different trigger than the one this PR fixes. Consider defining the condition as "no push happened this cycle" (HEAD_SHA_POST_PATCH unchanged) AND "rebuttal+thread-resolution confirmed for at least one REJECTed finding", rather than requiring literally every finding to be REJECT.

dmcaulay (commented):
Verdict: COMMENT — The all-REJECT/no-push reviewState reset is sound and tested, but the ALL_REJECT_NO_PUSH_REBUTTAL_CONFIRMED gating prose leaves a plausible mixed-outcome edge case undefined (an ACCEPT/MODIFY finding producing zero file diff), risking a recurrence of the same dedup-stuck-PR bug this PR is meant to close.

dodizzle (commented):
Verdict: COMMENT — The reviewState:pending reset is sound and well-verified against the task-store's patch() no-op semantics, but the new NO_PUSH_REBUTTAL_CONFIRMED variable is never explicitly assigned in bash anywhere in the file, unlike every other conditional-gating variable here (PR_RECORD_ID, HEAD_SHA_PRE_PATCH) — a real ambiguity gap for the orchestrating agent.

plugins/shipwright/commands/patch.md
```diff
@@ -786,11 +792,34 @@ if [ -n "$PR_RECORD_ID" ]; then
     "$SHIPWRIGHT_TASK_STORE_URL/prs/$PR_RECORD_ID/patch" \
     -d "{\"commitSha\": \"$HEAD_SHA_POST_PATCH\"}" > /dev/null 2>&1 || \
     echo "⚠ POST /prs/$PR_RECORD_ID/patch failed — continuing"
+  if [ "$NO_PUSH_REBUTTAL_CONFIRMED" = "true" ]; then
```
dodizzle, line 815: `NO_PUSH_REBUTTAL_CONFIRMED` is consumed here (`if [ "$NO_PUSH_REBUTTAL_CONFIRMED" = "true" ]`) but never explicitly assigned in any fenced bash block in this file — the only source is prose in Step 5c and the paragraph below this code block. Every other variable gating a conditional here (`PR_RECORD_ID`, `HEAD_SHA_PRE_PATCH`, `HEAD_SHA_POST_PATCH`) has a literal bash assignment preceding its use (e.g. `HEAD_SHA_PRE_PATCH=$(git ... rev-parse HEAD)`). Since this file's "code" is natural-language instructions interpreted by a future Claude orchestrator session, this is a real ambiguity gap, not a style nit — consider adding an explicit assignment (e.g. derived from comparing `$HEAD_SHA_POST_PATCH` to `$HEAD_SHA_PRE_PATCH` plus the subagent's STATUS/CONCERNS confirmation) matching the pattern used elsewhere in this file.

dmcaulay:
Addressed both review findings:

- @dmcaulay: the ACCEPT/MODIFY-zero-diff mixed-outcome edge case was already widened
  into the gating condition in the prior commit on this branch (widen reviewState
  reset gate beyond all-REJECT).
- @dodizzle: added an explicit assignment for NO_PUSH_REBUTTAL_CONFIRMED in Step 5c
  (commit d5405fb8), matching the bash-assignment convention used elsewhere in this
  file for PR_RECORD_ID/HEAD_SHA_PRE_PATCH.

dmcaulay (commented):
Verdict: APPROVE — Both prior rounds' findings (the mixed-outcome gating gap and the missing explicit NO_PUSH_REBUTTAL_CONFIRMED assignment) are resolved. Independently re-verified the underlying dedup/reset logic against check-review.ts:167 and pull-request-service.ts's patch() — no double-reset conflict with the existing ACCEPT/MODIFY self-heal path. All tests pass, CI is green. No new issues found.

PR #1818, opened 2026-07-17
docs: flag backfill live-data checks and constraint sequencing in plan-session

dmcaulay:
## Summary

Prompted by a vitals-os incident where a merged backfill migration blocked prod deploys for ~5.5 hours: its mapping was "verified" against config/API docs but missed a real data shape that only a live query of the table would have surfaced. `/plan-session` didn't have a checklist item that would have caught this at planning time.

Two gaps closed in `plugins/shipwright/commands/plan-session.md`:

1. **Step 2 (Flag complexity risks)** — a data-backfill migration that pattern-matches an existing column now must be flagged: acceptance criteria have to require a live `SELECT DISTINCT` check against the real table (not just docs/config), and the migration's own test fixtures must be seeded from that same live query.
2. **Step 2 (Breaking Change Scan) + Step 5 (Breaking Change Safety)** — adding a `NOT NULL`, foreign key, or unique constraint to a column on a table that already has rows is now flagged as a breaking change (previously only renames/removals were), and required to follow the same add → migrate → remove task-splitting pattern already used elsewhere: add nullable → backfill (verified) → constrain, as three sequential tasks.

## Test plan
- Docs-only change to a command's markdown instructions — no code path to exercise.
- Read through the full file to confirm the new bullets read consistently with the existing structure and don't duplicate other sections.

### DIFF
--- a/plugins/shipwright/commands/plan-session.md
+++ b/plugins/shipwright/commands/plan-session.md
@@ -115,9 +115,9 @@ Example flags:
 
 **Data-backfill migrations need a live-data check, not just a design review.** If a task involves a migration that assigns or attributes existing rows by pattern-matching a column (mapping an existing value to a new one), flag that the task's acceptance criteria must require querying the live table for its actual distinct values and confirming every distinct group is covered — not just checking the mapping against docs/config/API references. Test fixtures for that migration must be seeded from the same live-distinct-values check. A mapping that looks complete against documentation can still miss a real data shape that only a live query would surface.
 
-Mark this task **HITL** — see the judgment step in Step 5.5.
+This is likely **HITL** — see the conditional judgment step in Step 5.5.
 
-**Breaking Change Scan** — additions are safe to deploy at any time; renames and removals are not. For any rename or removal in the spec, grep for all current callers before proposing tasks:
+**Breaking Change Scan** — additions are safe to deploy at any time; renames and removals are not. For any rename or removal in the spec, grep for all current callers before proposing tasks; for a constraint addition (see below), verify backfill completeness against live data instead:
 
 - **DB**: dropping or renaming a table or column — who reads or writes it?
 - **DB**: adding a `NOT NULL`, foreign key, or other constraint to a column on a table that already has rows — has the existing data been backfilled to satisfy it? Adding the constraint before the backfill is complete will fail on (or corrupt) existing rows.
--- a/plugins/shipwright/commands/plan-session.md
+++ b/plugins/shipwright/commands/plan-session.md
@@ -115,7 +115,7 @@ Example flags:
 
 **Data-backfill migrations need a live-data check, not just a design review.** If a task involves a migration that assigns or attributes existing rows by pattern-matching a column (mapping an existing value to a new one), flag that the task's acceptance criteria must require querying the live table for its actual distinct values and confirming every distinct group is covered — not just checking the mapping against docs/config/API references. Test fixtures for that migration must be seeded from the same live-distinct-values check. A mapping that looks complete against documentation can still miss a real data shape that only a live query would surface.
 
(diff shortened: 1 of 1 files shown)


plugins/shipwright/commands/plan-session.md
```diff
@@ -113,16 +113,21 @@ Example flags:
 - "⚠ The spec adds X to the billing API but the billing service has no test coverage — any change here is risky without tests first"
 - "⚠ This feature requires a new abstraction that doesn't exist yet — adds ~2h of foundational work before the feature itself"
 
+**Data-backfill migrations need a live-data check, not just a design review.** If a task involves a migration that assigns or attributes existing rows by pattern-matching a column (mapping an existing value to a new one), flag that the task's acceptance criteria must require querying the live table for its actual distinct values and confirming every distinct group is covered — not just checking the mapping against docs/config/API references. Test fixtures for that migration must be seeded from the same live-distinct-values check. A mapping that looks complete against documentation can still miss a real data shape that only a live query would surface.
+
+Mark this task **HITL** — see the judgment step in Step 5.5.
```
dodizzle: This unconditional "Mark this task HITL" contradicts the conditional criteria in the Step 5.5 judgment step it references (line 282), which only flags HITL "in a repo where the dev-task agent's normal execution environment cannot reach production databases directly." An agent following this line literally will over-flag HITL in repos where it does have live DB access. Suggest: "This is likely HITL — see the conditional judgment step in Step 5.5" instead of an unconditional imperative.

plugins/shipwright/commands/plan-session.md
```diff
@@ -113,16 +113,21 @@ Example flags:
...
+
 **Breaking Change Scan** — additions are safe to deploy at any time; renames and removals are not. For any rename or removal in the spec, grep for all current callers before proposing tasks:
```
dodizzle: Minor: this intro sentence ("For any rename or removal in the spec, grep for all current callers before proposing tasks") still only describes the rename/removal verification method, but the new constraint-addition bullet added below (line 123) is neither a rename nor removal and is verified differently (backfill completeness against live data, not grepping callers). Self-correcting on a careful read since the bullet's own text is accurate, but worth broadening the umbrella sentence for consistency.

dodizzle (commented):
Verdict: COMMENT — PR title is missing a Conventional Commits prefix and is currently failing `pr-title-lint`; also flagged two internal-consistency gaps in the new instructions (Step 2's HITL line is unconditional but the Step 5.5 judgment step it references is conditional, and the Breaking Change Scan intro sentence wasn't updated to cover the new constraint-addition bullet's different verification method).

dodizzle (commented):
Verdict: COMMENT — Both prior findings (title lint, conditional HITL wording, Breaking Change Scan intro) are resolved by the latest commit, but it left one new gap: the Breaking Change Safety gate in Step 5 still only checks for renames/removals and would mark a constraint-addition task "Safe to deploy standalone: yes" despite the new guidance requiring it to be split into sequential tasks.

plugins/shipwright/commands/plan-session.md
```diff
@@ -244,6 +249,8 @@ Before finalizing the task list, check each task for renames or removals flagged
 
 A task that drops or renames something while a later task updates the consumers is not safe to deploy independently — that gap is a broken intermediate state in production.
 
+**Same pattern for new constraints on existing tables.** A task that adds a `NOT NULL`, foreign key, or unique constraint to a column on a table with existing rows must split into sequential tasks the same way: (a) add the column/relation nullable, (b) backfill existing rows (with a task depending on (a), acceptance criteria requiring the live-data check above), (c) a later task adds the constraint — and only once (b)'s backfill has actually run and verified zero gaps, not just once its code has merged. A task that adds the constraint in the same migration as the backfill, or before the backfill task, risks failing on (or silently corrupting) existing rows.
+
 If a task has no renames or removals, mark it: `Safe to deploy standalone: yes`.
```
dodizzle: This gate ("If a task has no renames or removals, mark it: `Safe to deploy standalone: yes`") wasn't updated for the new constraint-addition case added just above (line 252). A task that adds a `NOT NULL`/FK/unique constraint to a table with existing rows has no rename or removal, so it would pass this gate as "safe to deploy standalone" even though the new paragraph says it must be split into three sequential tasks and risks corrupting existing rows if done as one. Consider updating this line — and the intro at line 245 ("check each task for renames or removals flagged in Step 2") — to also cover constraint additions, mirroring the fix already applied to the Breaking Change Scan intro earlier in this PR.

dodizzle (approved):
Verdict: APPROVE — All four findings across the two prior review rounds (Conventional Commits title, conditional HITL wording, Breaking Change Scan intro, and the Breaking Change Safety gate now covering constraint additions) are resolved and internally consistent, with no new issues introduced.

PR #1823, opened 2026-07-17
fix(admin): make task and PR references clickable across admin UI

dmcaulay:
## Summary
Full sweep of the admin app for entities (tasks, PRs, agents, crons) rendered as plain text when they should be clickable.

**Tasks**
- Tasks-list blocker badges (`Blocked: <id>`) now link to the blocking task
- Task-detail Blockers section and Dependencies list now link to their task pages

**PRs**
- PR-detail `Task` field now links to the task
- PR-detail `PR Number` field now links out to GitHub — this page previously had **no outbound GitHub link at all**
- Cron-log & work-queue item cells (task or `repo#prNumber`) now link to the task or out to GitHub

**Agents**
- Tasks-list Agent column, task-detail Assignee/Agent Hint/Claimed By, PR-list and PR-detail Claimed By/Agent ID, and the tokens page's Agent column now link to `/admin/agents/:id`

**Crons**
- Cron-logs page's Cron column now links to the same page filtered to that cron's id

**Bug found + fixed along the way:** the unauthenticated public task board (`/public/tasks`, `readOnly=true`) reuses `renderTasksPage` for its rows. The new agent-cell link (and pre-existing blocker-badge task link) had no `readOnly` guard, so a claimed or blocked public task would have leaked internal `/admin/...` links to unauthenticated visitors. Both are now gated on `readOnly`, with regression tests covering the leak directly.

Out of scope: chat message bodies can mention task/PR/session ids in free text — auto-linkifying arbitrary text is a materially different feature from wiring up existing structured fields, left out of this pass.

## Test plan
- [x] `bun test admin/src/admin-ui-pages.unit.test.ts admin/src/admin-ui.smoke.test.ts` — 547 pass, new coverage for every new link and the public-board leak guard
- [x] `bun test` (full admin suite) — 1275 pass, 172 skip, 0 fail
- [x] `bun run typecheck` (admin) — clean
- [x] `bunx biome check` — clean

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -24,6 +24,27 @@ import type { ModelBreakdownEntry } from "./openapi-schemas.ts";
 
 // ─── Helpers ──────────────────────────────────────────────────────────────────
 
+/** Renders a link to a task's detail page, styled like the rest of the admin UI's inline links. */
+function taskLink(taskId: string): string {
+  return `<a href="/admin/tasks/${escapeHtml(taskId)}" style="color:#6366f1;text-decoration:none">${escapeHtml(taskId)}</a>`;
+}
+
+/**
+ * Renders a link for a work-queue/cron-log item id, which is either a task id
+ * (e.g. "WLS-2.2") or a "repo#prNumber" PR reference (e.g. "acme/x#123") per
+ * the RankedWorkItem/CronRunItem id convention. Task ids link into the admin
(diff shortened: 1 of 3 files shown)


dodizzle (approved):
Verdict: APPROVE — Well-scoped link-wiring sweep with a legitimate security fix (public task board was leaking internal /admin/ links for claimed/blocked tasks); the readOnly gating, route auth boundaries, and injection surface all checked out independently, and the fix is covered by a direct regression test. No blocking concerns.

PR #1830, opened 2026-07-17
feat: add Piper TTS synthesis and fix the speak-marker silent-failure contract

dodizzle:
## Summary
- Adds `synthesizePiper()` in `agent/src/voice.ts`, mirroring `synthesizeEdgeTTS`'s DI shape (injectable `SpawnFn`, never throws, resolves `null` on spawn error or non-zero exit). Uses `voiceConfig.piperVoice` resolved via `piper-voice.ts` to build the `--model` path, writes `.wav` output (Piper's native format — no ffmpeg transcode).
- Rewires `synthesizeSpeech` dispatch: `ELEVENLABS_API_KEY` set → ElevenLabs; else → Piper; else → text fallback.
- Fixes the silent-failure bug in `agent/src/slack.ts`: when `synthesizeSpeechFn` returns `null` (or throws) inside the `[speak:]` marker handling, the reply now posts as a normal text message via `client.chat.postMessage` instead of silently dropping, plus a log line naming the cause. Extracted the duplicated fallback-post logic into a small `postSpeakTextFallback()` helper.
- Refreshes `docs/deploy-kubernetes.md` to reflect Piper replacing edge-tts as the fallback TTS path.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | No ELEVENLABS_API_KEY + working Piper → returns non-empty `.wav` path | MET |
| 2 | ELEVENLABS_API_KEY set → ElevenLabs called, Piper's spawnFn never invoked | MET |
| 3 | Piper non-zero exit / spawn error → `null`, no throw; slack.ts posts text fallback + log | MET |
| 4 | No `[speak:]` marker / `[silent]`-only response unaffected | MET |
| 5 | New unit tests cover all 4 dispatch outcomes via injected SpawnFn; existing integration test updated for text-fallback | MET |

## Test Plan
- New tests in `agent/src/voice.unit.test.ts`: ElevenLabs dispatch, Piper success, Piper non-zero exit, Piper spawn error — all via injected `SpawnFn`, no real subprocess, no global mocking.
- Updated `agent/src/slack.integration.test.ts` — `[speak:text] in DM` and `dispatchMarkers — direct` tests now assert `client.chat.postMessage` text fallback instead of silent skip; added a throw-path test.
- Updated `agent/src/voice.integration.test.ts` for `.wav` output naming.
- [x] `bunx biome lint .` passing
- [x] `bun run --filter='*' --sequential typecheck` passing (all workspaces)
- [x] Full test suite: 4134 pass, 0 fail, 330 skip
- [x] Coverage gate: 80.15% lines / 80.79% functions (threshold 80/80)
- [x] `check-banned-strings`, `check-config-docs`, `sync-version --check` all passing

### DIFF
--- a/agent/src/slack.ts
+++ b/agent/src/slack.ts
@@ -214,6 +214,13 @@ export async function dispatchMarkers(
         );
         continue;
       }
+      const postSpeakTextFallback = () =>
+        client.chat
+          .postMessage({ channel, thread_ts: threadTs, text: marker.text })
+          .catch((err: unknown) =>
+            console.warn("[markers] speak text fallback post failed:", err),
+          );
+
       try {
         const audioPath = await synthesizeSpeechFn(marker.text, voiceConfig);
         if (audioPath) {
@@ -231,25 +238,14 @@ export async function dispatchMarkers(
           console.warn(
             "[markers] speak synthesis returned null — posting text fallback",
           );
-          await client.chat
-            .postMessage({ channel, thread_ts: threadTs, text: marker.text })
-            .catch((err: unknown) =>
-              console.warn("[markers] speak text fallback post failed:", err),
-            );
+          await postSpeakTextFallback();
         }
       } catch (err) {
         console.warn(
           "[markers] speak synthesis failed — posting text fallback:",
           err,
         );
-        await client.chat
-          .postMessage({ channel, thread_ts: threadTs, text: marker.text })
-          .catch((postErr: unknown) =>
-            console.warn(
-              "[markers] speak text fallback post failed:",
-              postErr,
-            ),
-          );
+        await postSpeakTextFallback();
       }
     }
   }
--- a/agent/src/slack.ts
+++ b/agent/src/slack.ts
@@ -529,7 +529,7 @@ export function createSlackApp(
         threadTs: msg.thread_ts ?? msg.ts,
         synthesizeSpeechFn,
         voiceConfig,
-        postedText: cleaned,
+        postedText: shouldSuppress ? undefined : cleaned,
       });
     } catch (err) {
       console.error("[slack] error:", err);
@@ -664,7 +664,7 @@ export function createSlackApp(
         threadTs: replyTs,
         synthesizeSpeechFn,
         voiceConfig,
-        postedText: cleaned,
+        postedText: isSilent ? undefined : cleaned,
       });
     } catch (err) {
       console.error("[slack] error:", err);
--- a/agent/src/slack.ts
+++ b/agent/src/slack.ts
@@ -229,11 +229,27 @@ export async function dispatchMarkers(
             );
         } else {
           console.warn(
-            "[markers] speak synthesis returned null — no audio uploaded",
+            "[markers] speak synthesis returned null — posting text fallback",
           );
+          await client.chat
+            .postMessage({ channel, thread_ts: threadTs, text: marker.text })
+            .catch((err: unknown) =>
+              console.warn("[markers] speak text fallback post failed:", err),
+            );
         }
       } catch (err) {
-        console.warn("[markers] speak synthesis failed:", err);
+        console.warn(
+          "[markers] speak synthesis failed — posting text fallback:",
+          err,
+        );
+        await client.chat
+          .postMessage({ channel, thread_ts: threadTs, text: marker.text })
+          .catch((postErr: unknown) =>
+            console.warn(
+              "[markers] speak text fallback post failed:",
+              postErr,
+            ),
+          );
       }
     }
   }
--- a/agent/src/slack.ts
+++ b/agent/src/slack.ts
@@ -146,6 +146,21 @@ export type ResolveUserFn = (
   client: any,
 ) => Promise<string>;
 
+// Normalizes text for the speak-marker fallback's duplicate-content check:
+// lowercased, trimmed, trailing punctuation stripped, internal whitespace
+// collapsed. Deliberately simple — not fuzzy matching, just enough to catch
+// the common "Here is the answer. [speak:Here is the answer]" pattern where
+// the marker text is a near-verbatim echo of the already-posted reply.
+function normalizeForComparison(text: string | undefined): string {
+  if (!text) return "";
+  return text
+    .trim()
+    .toLowerCase()
+    .replace(/[.!?]+$/, "")
+    .replace(/\s+/g, " ")
+    .trim();
+}
+
 export async function dispatchMarkers(
   markers: Marker[],
   {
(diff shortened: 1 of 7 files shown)


dmcaulay (commented):
Verdict: COMMENT — Solid implementation of the Piper TTS synthesis and the speak-marker silent-failure fix; one test-quality gap worth fixing before merge: the AC1 unit test in agent/src/voice.unit.test.ts (lines 196-206) is titled as verifying a "non-empty .wav file" but never checks file existence or content — the fake spawn's stdin is a no-op that never touches the filesystem, so the assertion would pass regardless of whether a file was actually written.

agent/src/voice.unit.test.ts
```diff
@@ -99,3 +108,133 @@ describe("makeWhisperSvcClient — onerahmet /asr contract", () => {
...
+    const outPath = await synthesizeSpeech(
+      "hello",
+      {},
+      undefined,
+      makeSuccessSpawn(),
+    );
+
+    expect(outPath).not.toBeNull();
+    expect(outPath).toContain(".wav");
+  });
```
dmcaulay: This test's title claims it verifies "a non-empty .wav file", but the assertions only check `outPath` is non-null and contains `.wav`. `synthesizePiper` never writes `outPath` itself — the fake spawn's `stdin = { write: () => {}, end: () => {} }` is a no-op that never touches the filesystem, so no file is created here (unlike the ElevenLabs test above, which writes a real file and `unlinkSync`s it). The assertion would pass identically whether or not a real, non-empty `.wav` existed at that path.

Either write a stub file to `outPath` inside `makeSuccessSpawn`'s `onImmediate` and assert `existsSync`/non-zero size, or retitle the test to reflect what it actually checks.

agent/src/slack.ts
```diff
@@ -214,6 +214,13 @@ export async function dispatchMarkers(
         );
         continue;
       }
+      const postSpeakTextFallback = () =>
```
dodizzle: `postSpeakTextFallback()` posts `marker.text` as a brand-new `chat.postMessage` whenever synthesis returns `null` or throws. This codebase's typical `[speak:text]` usage pairs a visible reply (posted via `say()`) with a near-duplicate spoken echo (e.g. `"Here is the answer. [speak:Here is the answer]"` in slack.integration.test.ts:1476). When synthesis fails in that common case, the user now sees the same content twice — once from `say()`, once from this fallback. None of the new/updated tests exercise this combination (they all use marker-only responses with no separate `cleaned` text). Consider only posting the fallback when it meaningfully differs from the already-posted `cleaned` text.

dodizzle (commented):
Verdict: COMMENT — Solid Piper TTS implementation and a genuinely-fixed silent-failure bug (the prior AC1 test-quality gap from @dmcaulay's review is verified resolved), but the new text-fallback path can double-post near-duplicate content when a `[speak:text]` marker echoes text already posted via `say()` — this codebase's most common usage pattern (e.g. `"Here is the answer. [speak:Here is the answer]"`). Worth a small guard before merge.

agent/src/voice.unit.test.ts
```diff
@@ -99,3 +108,159 @@ describe("makeWhisperSvcClient — onerahmet /asr contract", () => {
...
+    );
+
+    expect(mockFetch).toHaveBeenCalledTimes(1);
+    expect(spawnFn).not.toHaveBeenCalled();
+    expect(outPath).not.toBeNull();
+    expect(outPath).toContain(".mp3");
+    if (outPath) unlinkSync(outPath);
+  });
+
+  test("with no ELEVENLABS_API_KEY and a working Piper, synthesizeSpeech returns a path to a non-empty .wav file", async () => {
```
dodizzle: This `describe` block performs real filesystem I/O (`writeFileSync`/`existsSync`/`statSync`/`unlinkSync`) and exercises `synthesizeSpeech`, which calls `mkdirSync` and spawns subprocess-shaped fake processes. `docs/test-readiness/test-system.md` defines `unit` as "Pure logic — no I/O of any kind. No filesystem reads, no network calls, no process spawning" and calls out this exact pattern under Boundary violations: "A unit test that reads a file, opens a socket, or spawns a subprocess → move to integration." The same dispatch coverage is already correctly placed in `voice.integration.test.ts` in this PR using the same injected `SpawnFn` doubles. Suggest moving this block (and its fs-touching helpers) to `voice.integration.test.ts`, or stripping this file down to pure-logic assertions.

dodizzle (commented):
Verdict: COMMENT — Solid Piper TTS implementation and a genuinely-fixed silent-failure bug, but the new synthesizeSpeech dispatch test block in voice.unit.test.ts does real filesystem I/O, violating this repo's unit-layer no-I/O boundary rule (and duplicating coverage already correctly placed in voice.integration.test.ts).

agent/src/slack.ts
```diff
@@ -486,6 +529,7 @@ export function createSlackApp(
         threadTs: msg.thread_ts ?? msg.ts,
         synthesizeSpeechFn,
         voiceConfig,
+        postedText: cleaned,
```
dodizzle: `postedText: cleaned` is passed unconditionally here, even when `shouldSuppress` is true and `cleaned` was never actually posted via `say()`. `[silent]` and `[speak:text]` markers can coexist in the same response (independent regexes in `parseMarkers`). If the speak-marker text echoes `cleaned` and Piper synthesis then fails, `postSpeakTextFallback`'s duplicate guard will wrongly suppress the fallback post — even though nothing was ever shown to the user. Suggest `postedText: shouldSuppress ? undefined : cleaned`.

agent/src/slack.ts
```diff
@@ -620,6 +664,7 @@ export function createSlackApp(
         threadTs: replyTs,
         synthesizeSpeechFn,
         voiceConfig,
+        postedText: cleaned,
```
dodizzle: Same issue as the `message` handler: `postedText: cleaned` is passed even when `isSilent` is true and the reply was never posted. Suggest `postedText: isSilent ? undefined : cleaned`.

dodizzle (commented):
Verdict: COMMENT — The prior review round's finding (unit-layer I/O boundary violation) is genuinely resolved in this commit. New issue found: `postedText` is passed to `dispatchMarkers` unconditionally in the `message` and `app_mention` handlers, even when the text reply was suppressed (`shouldSuppress`/`isSilent` true). Since `[silent]` and `[speak:text]` markers can coexist in one response, if the speak-marker text echoes the suppressed `cleaned` text and Piper synthesis then fails, the duplicate-content guard wrongly treats it as already-posted and drops the fallback — silently losing the reply in the `[silent]+[speak:]` combination. See inline comments.

dodizzle (commented):
Verdict: APPROVE — The postedText fix (don't pass it through when the reply was suppressed) is correct and complete across all five dispatchMarkers call sites in slack.ts and cron-handler.ts, verified directly against the live source rather than just the diff. Covered by new tests for the [silent]+[speak:] edge case. No new issues found after four review/patch cycles.

PR #1843, opened 2026-07-18
feat: add computeDependencyLayout pure layered-DAG layout function

dmcaulay:
## Summary
- Add `computeDependencyLayout(nodes: DependencyNode[])`, a pure layered-DAG layout function in `admin/src/admin-ui-pages.ts` that assigns each dependency-graph node a column (depth = longest path from any root, following only in-session `dependsOn` edges) and a row within that column, converting to fixed 220x88 card pixel coordinates matching the approved intel mockup (`session-dependency-graph.html`, including its 16-node fan-in/fan-out stress-test variant).
- Cycle-safe via a `Set` tracking the current DFS recursion stack — a re-entrant node on a cycle stops that branch instead of recursing further, so every node still resolves to a finite depth and gets a position.
- Export `DependencyNode`, `DependencyLayout`, and `DependencyLayoutPosition` so this can be consumed by the renderer in a later task (DGV-1.2). No rendering changes in this PR — layout math only.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Returns position map ({x,y,depth}) + {width,height}, longest-path-from-roots depth | MET |
| 2 | No filtering — caller's job | MET |
| 3 | Cycle-safe via visited-set guard, both nodes still positioned | MET |
| 4 | Deterministic column/row ordering for given input order | MET |
| 5 | Unit tests only, covering linear/fan-out/fan-in/cycle/deterministic | MET |

## Test Plan
- Unit tests added to `admin/src/admin-ui-pages.unit.test.ts`: linear chain (with exact pixel assertions locking in mockup spacing), fan-out, fan-in (longest-path-wins), cycle safety, deterministic repeat-call equality, deterministic same-depth row ordering, empty input.
- [x] `bun run --filter="*" --sequential typecheck` clean across all workspaces
- [x] Full `NODE_ENV=test bun test`: 4229 pass / 0 fail


dodizzle (approved):
Verdict: APPROVE — Correct, cycle-safe pure layout function with no breaking API surface; tests verify their stated claims (linear/fan-out/fan-in/cycle/deterministic/empty). No blocking concerns.

dmcaulay (commented):
Verdict: APPROVE — Cycle-safe layered-DAG layout is correct and well-tested (linear/fan-out/fan-in/cycle/deterministic/empty); renderer switch to positioned SVG nodes+edges is consistent with the mockup and escaping is applied throughout. No blocking concerns. Minor: PR description still says "no rendering changes" but later commits on this branch (a7cf32f3, e371193e) do change the renderer — worth a description update, not a blocker.

PR #1849, opened 2026-07-18
feat: link legacy phase crons as shipwright-loop children in reconcileSystemCrons

dmcaulay:
## Summary
- Marks the 4 legacy pipeline crons (`shipwright-dev-task`, `shipwright-patch`, `shipwright-review`, `shipwright-deploy`) with a new `parentCron: "shipwright-loop"` field in `SYSTEM_CRONS`.
- `reconcileSystemCrons()` now resolves and sets `parentCronId` on those 4 crons via a two-pass approach inside the existing transaction, so the link self-heals fleet-wide on every agent boot regardless of array order or whether it's the agent's very first reconcile.
- Confirmed via repo-wide grep that `LOOP_SUPERSEDED_CRON_NAMES` / force-disable-on-loop-enable logic does not exist anywhere in the codebase — acceptance criterion #2 is trivially satisfied (nothing to remove).
- Refreshed `docs/agent.md` and `docs/agent-api.md` to describe the new two-pass reconcile strategy and parent/child cron linking.

## Acceptance Criteria
| 1 | The 4 legacy system cron entries declare `parentCron: "shipwright-loop"`; reconcile sets their `parentCronId` to the loop cron's id for every agent on next boot. | MET | `system-crons.ts` markers + two-pass reconcile in `agent-cron-jobs.ts` |
| 2 | `LOOP_SUPERSEDED_CRON_NAMES` and its force-disable branch are removed from `reconcileSystemCrons`. | MET | Confirmed absent repo-wide via grep — nothing to remove |
| 3 | Integration tests assert `parentCronId` linkage; obsolete force-disable test removed. | MET | Added fresh-boot linkage test + self-heal-on-reconcile test; no obsolete test existed to remove |

## Test Plan
- `admin/src/system-crons.test.ts` — 52 tests pass (no DB required).
- `admin/src/agent-cron-jobs.integration.test.ts` — 2 new tests added (parentCronId linkage on fresh boot, self-heal on subsequent reconcile); skipped in this sandbox (no `DATABASE_URL_ADMIN_TEST`), will run against real Postgres in CI.
- `task lint` — clean.
- `task typecheck` — clean across all 7 packages.

Note: this PR covers LPC-1.2 only. The bundled task LPC-1.3 (structural dispatch-exclusion guard in `cron-handler.ts`) targets the same branch and will land as a follow-up commit on this PR per the plan's bundling note.


dodizzle (approved):
Verdict: APPROVE — Two-pass reconcile correctly links the four legacy pipeline crons to shipwright-loop with no ordering/race issues; all three LPC-1.2 acceptance criteria are met, tests are meaningful, and docs are accurate.

admin/src/agent-cron-jobs.ts
```diff
@@ -177,12 +177,16 @@ export class AgentCronJobService {
...
+   * generic per-minute dispatch scheduling (fetched on startup and every
+   * 60s). Excludes rows with a non-null parentCronId regardless of their own
+   * enabled value — a parented row is a phase of another cron (config only,
+   * e.g. read by the loop orchestrator) and must never be independently
+   * scheduled, structurally rather than via an enabled-flag convention.
    */
   async listEnabled(): Promise<AgentCronJob[]> {
     return this.prisma.agentCronJob.findMany({
-      where: { enabled: true },
+      where: { enabled: true, parentCronId: null },
```
dmcaulay: `listEnabled()` has no production caller anywhere in the repo (only this file's own integration test calls it). The real runtime dispatch path is `agent/src/index.ts`'s `syncCrons()` → `runtimeClient.listAgentCronJobs()` → `GET /agents/:id/crons` → `admin/src/agents-api.ts`'s `listWithRunSummary()`, which has no `enabled`/`parentCronId` filter at all. The actual pipeline-phase exclusion at dispatch time lives in `agent/src/loop-cron-classifier.ts`'s `classifyCronJobsForScheduling()`, gated on a hardcoded `PIPELINE_PHASE_JOB_NAMES` name set — it never reads `parentCronId`, and this PR doesn't touch that file. So this change (and the new tests for it) have no effect on real cron scheduling, despite the commit message and updated docstring both claiming this wires up the dispatch guard. Either wire `classifyCronJobsForScheduling`/the `GET /agents/:id/crons` path to actually use `parentCronId`, or drop this commit and re-scope LPC-1.3 with an accurate description.

dmcaulay (commented):
Verdict: COMMENT — LPC-1.2's two-pass reconcile/parent-linking logic is solid, but the bundled third commit (LPC-1.3) modifies `listEnabled()`, a method with zero production callers — the real dispatch path (`agent/src/index.ts` syncCrons → `GET /agents/:id/crons` → `listWithRunSummary()`) never calls it, and actual scheduler exclusion is still purely name-based in `agent/src/loop-cron-classifier.ts`, untouched here. So this commit doesn't deliver the "exclude parented cron rows from generic scheduler dispatch" guard its own commit message and docstring claim. See inline comment for the specific line, and full review notes in state/reviews/PR_REVIEW_1849.md.

dmcaulay:
Addressed — reverted the LPC-1.3 commit (968f679c) via c4dab37b.

You're right: `listEnabled()` is dead in production and the commit's claim of delivering dispatch-time exclusion didn't hold up. Rather than bolt on the real wiring (parentCronId through the OpenAPI schema, serializer, agent-side type, and classifyCronJobsForScheduling) unsupervised in a patch cycle, I reverted the misleading commit and reset LPC-1.3 in the task store (status back to pending, description updated with this root-cause context, branch reassigned to `feat/lpc-1-3-scheduler-dispatch-guard`) so it gets redone properly with its own PR and review. The LPC-1.2 two-pass parent-linking logic underneath (69102555) is untouched and still lands as-is — it's real and tested, just not yet consumed by the scheduler.

admin/src/agent-cron-jobs.ts
```diff
@@ -427,10 +441,24 @@ export class AgentCronJobService {
...
             },
           });
+          idByName.set(systemCron.name, createdCron.id);
           created++;
         }
       }
 
+      // Pass 2: link declared parent/child crons now that every row in this
+      // reconcile has a known id, regardless of array order or first-boot.
+      for (const systemCron of SYSTEM_CRONS) {
```
dmcaulay, line 458: Pass 2 only *sets* `parentCronId` for entries that currently declare `parentCron` — it never resets an existing row's `parentCronId` back to `null` if a future `SYSTEM_CRONS` entry drops its `parentCron` declaration (or repoints to a name that no longer resolves). That makes the "self-heals the link on every reconcile call" claim in this docstring (and in `docs/agent-api.md`/`docs/agent.md`) only true in the null→set direction. Not a live bug today — all four entries permanently declare `parentCron: "shipwright-loop"`, which always resolves — but worth closing the gap: when an entry has no resolvable `parentCron` but the existing row still has a non-null `parentCronId`, explicitly clear it to `null`.

dmcaulay (commented):
Verdict: APPROVE — LPC-1.3's ineffective listEnabled() commit was fully reverted, resolving all three prior findings (dead-code guard, tests validating it, missing acceptance criteria). LPC-1.2's two-pass reconcileSystemCrons() is correctly transactional, order-independent, and well-tested. The parentCronId-never-cleared-to-null finding raised here was resolved in a follow-up commit — Pass 2 now clears a stale parentCronId to null when the entry has no resolvable parentCron, confirmed by a later re-review at commit c1d4fa16 with no outstanding findings.

dmcaulay (commented):
Verdict: APPROVE — The prior finding (parentCronId never cleared back to null when a parentCron declaration is later removed) is fully resolved: Pass 2 now clears a stale parentCronId to null when the entry has no resolvable parentCron, backed by a new integration test and matching doc updates in docs/agent-api.md and docs/agent.md. No outstanding findings. CI green on c1d4fa16.

dodizzle (approved):
Verdict: APPROVE — Both findings raised in @dmcaulay's review rounds are resolved: the ineffective LPC-1.3 listEnabled() guard was reverted (rescoped to its own follow-up PR), and reconcileSystemCrons()'s pass 2 now clears a stale parentCronId back to null when an entry no longer declares a resolvable parentCron, backed by a new integration test. Confirmed the current head is functionally identical to the last-reviewed commit (c1d4fa16) for every file this PR touches — the head only moved via a main merge. CI green. All three LPC-1.2 acceptance criteria remain met. Confirming to unblock merge.

PR #1858, opened 2026-07-18
feat: add consolidation-scan skill for churn-weighted pattern detection (CVG-1.1)

dmcaulay:
## Summary
- Adds `plugins/shipwright/skills/consolidation-scan/SKILL.md` — a read-only skill that discovers *emerging, unnamed* duplicate/similar code patterns via judgment-driven comparison (not a token/AST clone detector), weights candidates by git churn, and tracks them across multiple runs in a local ledger before ever proposing anything.
- Adds the thin `plugins/shipwright/commands/consolidation-scan.md` wrapper, matching `entropy-scan`'s command shape.
- Adds `plugins/shipwright/skills/consolidation-scan/references/ledger-schema.md` documenting the `state/consolidation-ledger.json` shape (fingerprint, `occurrence_count`, `consecutive_stable_runs`, `status`) and the promotion rule.
- Refreshes `docs/architecture.md`'s skill catalog to list the new skill.

This is a separate maintenance system from `entropy-scan`/`entropy-fix`: entropy-scan checks fixed, named rules from `principles.md` in a single stateless pass. `consolidation-scan` instead persists candidates to `state/consolidation-ledger.json` and only writes `consolidation-report.md` entries once a pattern has recurred with a stable shape across runs (`occurrence_count >= 3 AND consecutive_stable_runs >= 2`), suppressing anything matched by `.claude/shipwright/consolidation-decisions.md` (added by the sibling task CVG-1.2 — this skill degrades gracefully while that file doesn't yet exist).

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | SKILL.md documents full detection → churn-weighting → ledger-matching → stabilization-gating → report-writing flow | MET |
| 2 | Reads consolidation-decisions.md, suppresses matching accepted-debt entries | MET |
| 3 | Reads/writes state/consolidation-ledger.json with fingerprint/occurrence_count/consecutive_stable_runs/status | MET |
| 4 | Only ready_to_propose written to consolidation-report.md | MET |
| 5 | No new coupling outside plugins/shipwright/ and state/ | MET |

## Test Plan
- No `*.test.ts` needed for this pure prompt/skill markdown content — matches `entropy-scan`/`entropy-fix` precedent (zero test files for skill-only PRs)
- [x] `task lint` — clean
- [x] `task check-strings` — clean
- [x] `task typecheck` — clean across all 7 workspaces
- [x] `task check-config-docs` — 32/32 env vars documented, unaffected
- [x] `task check-version-sync` — passed
- [x] `NODE_ENV=test bun test` — 4241 pass, 0 fail, 329 skip
- [x] Coverage gate — 80.30% lines / 80.97% functions (threshold 80%/80%, repo-wide aggregate; this diff adds no source code)

### DIFF
--- a/plugins/shipwright/skills/consolidation-scan/SKILL.md
+++ b/plugins/shipwright/skills/consolidation-scan/SKILL.md
@@ -93,7 +93,7 @@ authoritatively before promotion (Step 7, which is the actual gate).
          "firstSeen": "<ISO8601>",
          "lastSeen": "<ISO8601>",
          "history": [
-           { "run": "<ISO8601>", "description": "<description as observed that run>", "files": ["<path>", "..."] }
+           { "run": "<ISO8601>", "description": "<description as observed that run>", "notes": "<why-they-look-like-the-same-responsibility note as observed that run>", "files": ["<path>", "..."] }
          ]
        }
      }
@@ -128,7 +128,9 @@ SonarQube CPD, or a diff-based similarity tool. Instead:
    - The **involved files** (each file/module implementing the responsibility
      independently), with line ranges where practical
    - A short note on **why they look like the same responsibility** (what's
-     structurally or behaviorally shared, even if no lines are shared verbatim)
+     structurally or behaviorally shared, even if no lines are shared verbatim) —
+     retain this note; it is persisted to the ledger's `history` in Step 6 and is
+     what future runs compare against to judge shape stability
 4. Before finalizing a candidate, loosely check it against the suppression list from
    Step 1 — if a suppression's pattern description clearly matches, you may skip
    spending further effort refining that candidate's description now (it will be
@@ -244,8 +246,8 @@ For every fingerprint match found in Step 5:
      low (it just tracks one more run), while the cost of over-promoting a still-
      shifting pattern is a report entry proposing a canonical shape that's already
      stale.
-3. Append this run's observation (description, files, timestamp) to the entry's
-   `history` array — never truncate or overwrite prior history entries.
+3. Append this run's observation (description, files, notes, timestamp) to the
+   entry's `history` array — never truncate or overwrite prior history entries.
 4. Update `lastSeen` to this run's timestamp and `files` to this run's observed file
    list (the most recent observation, for report display — `history` retains the
    full trail).
--- a/plugins/shipwright/skills/consolidation-scan/SKILL.md
+++ b/plugins/shipwright/skills/consolidation-scan/SKILL.md
@@ -216,17 +216,28 @@ For every fingerprint match found in Step 5:
    entry's most recently recorded `history` entry:
    - If the new instance's description and the "why they look like the same
      responsibility" reasoning are **consistent** with the prior recorded shape
-     (same responsibility, same general approach — e.g. still "retry with
-     exponential backoff," not now "retry with a fixed delay and no backoff") →
+     (same responsibility, same general approach — e.g. still "retry a flaky
+     network call with backoff," implemented the same way structurally) →
      **increment `consecutive_stable_runs` by 1**.
    - If the new instance's implementation approach has **diverged meaningfully**
      from what was previously recorded (a different algorithm, a materially
      different scope, or the responsibility has forked into something no longer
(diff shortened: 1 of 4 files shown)


dodizzle (commented):
Verdict: COMMENT — one important internal-consistency issue: Step 6's shape-divergence example can't actually be reached given Step 5's description-keyed fingerprint (see inline comment). Otherwise solid, well-precedented documentation.

plugins/shipwright/skills/consolidation-scan/SKILL.md
```diff
@@ -0,0 +1,415 @@
...
+1. **Always increment `occurrence_count` by 1.** A fingerprint match is by
+   definition a recurrence of the same description + file set, so the raw count
+   always goes up.
+2. **Judge shape stability** before touching `consecutive_stable_runs`. Compare this
+   run's candidate description/structure notes (from Step 3) against the ledger
+   entry's most recently recorded `history` entry:
+   - If the new instance's description and the "why they look like the same
+     responsibility" reasoning are **consistent** with the prior recorded shape
+     (same responsibility, same general approach — e.g. still "retry with
+     exponential backoff," not now "retry with a fixed delay and no backoff") →
```
dodizzle: The fingerprint (Step 5) is keyed on the pattern description's own normalized text (`sha256(normalized_description + "|" + sorted_files)`). But this worked example — description going from "retry with exponential backoff" to "retry with a fixed delay and no backoff" — describes a description change big enough that its normalized hash would no longer match the prior fingerprint. Step 5 would route that as a brand-new candidate (`occurrence_count: 1`), so Step 6's divergence-reset branch ("for every fingerprint match found in Step 5") would never actually see this case. The rule only reaches this branch when the *generic* description stays stable but the recorded structural/reasoning notes diverge — worth either fingerprinting on files-only so genuine description drift can reach this check, or rewriting the example to match the case that can actually trigger it today.

plugins/shipwright/skills/consolidation-scan/SKILL.md
```diff
@@ -0,0 +1,426 @@
...
+---
+
+## Step 6: Update Occurrence and Stability Counters
+
+For every fingerprint match found in Step 5:
+
+1. **Always increment `occurrence_count` by 1.** A fingerprint match is by
+   definition a recurrence of the same description + file set, so the raw count
+   always goes up.
+2. **Judge shape stability** before touching `consecutive_stable_runs`. Compare this
```
dmcaulay, line 216: This fix resolves the prior reachability issue (the example no longer relies on a description change that would've produced a new Step-5 fingerprint), but it introduces a new internal-consistency gap.

Step 6 here instructs comparing "this run's candidate description/structure notes (from Step 3)" against "the ledger entry's most recently recorded `history` entry," and the worked example below (lines 231-236) hinges on detecting that "this run's notes show the backoff calculation changed" relative to what was *previously recorded*.

But per this same step's append instruction (line 247: "Append this run's observation (description, files, timestamp) to the entry's `history` array") and `references/ledger-schema.md`'s documented history entry shape (`{ run, description, files }`), the `history` array never actually stores the "why they look like the same responsibility" reasoning/structural notes from Step 3 — only description, files, and timestamp.

So the persisted history entry this step says to diff against doesn't contain the data the worked example needs to detect the divergence it describes. Either add a `notes`/`reasoning` field to the history entry shape (here and in ledger-schema.md) so structural notes are actually persisted, or reword this step to make clear the comparison is a fresh judgment call against the current code rather than a diff against a stored notes field that doesn't exist.

dmcaulay (commented):
Verdict: COMMENT — Step 6's fixed divergence example still compares against ledger fields the `history` array never persists; see inline comment.

dmcaulay (commented):
Verdict: APPROVE — Both prior review findings are addressed: this commit persists the "why they look like the same responsibility" reasoning as a `notes` field in the ledger's `history` entries (SKILL.md + ledger-schema.md), so Step 6's divergence comparison now diffs against data the ledger actually stores. No new issues in the diff since the last review.

PR #1875, opened 2026-07-18
fix: isolate per-item claim() failures from aborting the whole drain tick

dmcaulay:
## Summary
- Wrap the per-item `claimTask()` pre-claim call in `runLoopTick` in a try/catch so a thrown error (e.g. a task-store 5xx) no longer aborts the entire drain tick — it now skips just the offending item and continues draining review/patch/deploy candidates in the same tick.
- Add a per-tick `failedPreClaimTaskIds` guard that filters a previously-failed item out of subsequent `getDevTaskCandidates()` re-collections within the same tick, preventing an unbounded spin on the same item.
- Log the skip via `console.warn` (Sentry-eligible) with the item id and error so repeated claim failures stay observable instead of being silently swallowed.

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Thrown claimTask error caught per-item, drain continues instead of rejecting the whole tick | MET |
| 2 | Skip is logged with item id + error | MET |
| 3 | No spin-loop on repeated failure within a tick | MET |
| 4 | Unit test: item A throws, item B (different type/phase) still dispatches | MET |
| 5 | Integration test with real claim client + recorded 5xx double | MET |
| 6 | No regression to 409-skip / success-dispatch behavior | MET |

## Test Plan
- `agent/src/loop-orchestrator.unit.test.ts`: new cases covering per-item skip-and-continue and the spin-loop bound; all pre-existing CBD-1.2/1.3 cases (409 skip, success dispatch) still pass.
- `agent/src/loop-orchestrator.integration.test.ts` (new): real `createTaskStoreClient().claim` HTTP client against a real `Bun.serve()` stub returning a 500 from `POST /tasks/:id/claim` — confirms the real throw doesn't reject `runLoopTick()` and the failed item isn't re-claimed.
- Full repo suite (`bun test`): 4313 pass, 0 fail.
- `task lint`, typecheck (all packages), and the aggregate coverage gate (80.37% lines / 81.10% functions) all pass.


agent/src/loop-orchestrator.ts
```diff
@@ -363,7 +374,22 @@ export function createLoopOrchestrator(
...
+            console.warn(
+              `Pre-claim failed for task ${itemId}, skipping for this tick: ` +
+                `${err instanceof Error ? err.message : String(err)}`,
+            );
+            failedPreClaimTaskIds.add(itemId);
+            continue;
+          }
           if (!claimed) continue;
         } else {
           const claimResult = await claimPr(item.pr);
```
dodizzle: This `claimPr` call has no try/catch, unlike the `claimTask` branch just above. `claimPr`'s production implementation (`check-helpers.ts:361-386`) throws `task-store POST /prs/claim → ${status}` on any non-409, non-ok status — the identical throw-on-5xx behavior this PR just fixed for `claimTask`. A task-store 5xx during a review/patch/deploy pre-claim still aborts the whole drain tick uncaught, undermining this PR's stated goal for 3 of the 4 loop phases. Worth wrapping in the same try/catch/skip pattern, or calling out explicitly as a scoped follow-up.

dodizzle (commented):
Verdict: COMMENT — The dev-task claimTask() fix is correct and well-tested, but the sibling claimPr() call (review/patch/deploy pre-claims) remains unguarded and can still throw a task-store 5xx uncaught out of runLoopTick, aborting the whole drain tick for 3 of the 4 loop phases — the exact bug class this PR fixes for the task path.

dmcaulay (commented):
Verdict: APPROVE — Symmetric try/catch fix on both claimTask and claimPr pre-claim paths, fully closing the gap the prior review flagged (claimPr was left unguarded). All 6 acceptance criteria verified in code and tests; 46/46 tests pass, lint/typecheck clean.

PR #1900, opened 2026-07-18
feat: read loop phase toggles from child AgentCronJob rows (LPC-2.1)

dmcaulay:
## Summary
- `resolveLoopPhaseToggles()` now reads each pipeline phase's enabled/disabled state exclusively from child `AgentCronJob` rows (`parentCronId === loopCronId`), replacing the old top-level-cron-by-name lookup
- The one call site in `loop-orchestrator.ts`'s `runLoopTick` now passes `loopCronId` through
- Relies on `reconcileSystemCrons()` (LPC-1.2) having run for a given agent to populate `parentCronId` — an unreconciled agent soft-fails to zero active phases (pauses dispatch, not an error) until its next reconcile, an accepted rollout risk per the plan

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `resolveLoopPhaseToggles()` reads phase state exclusively from child rows, not top-level lookup | MET |
| 2 | Disabling a phase child row's `enabled` flag stops dispatch; re-enabling resumes it | MET |
| 3 | Tests updated to seed child rows under a parent loop cron and assert resolution reads from them; old top-level-lookup test retired | MET |

## Test Plan
- `loop-cron-classifier.unit.test.ts`: rewrote the `resolveLoopPhaseToggles` suite to the child-row model; added cases for top-level same-named rows being ignored, child rows under a different parent being ignored, and the zero-child-rows soft-fail case
- `loop-orchestrator.integration.test.ts`: new suite composing the real `createLoopOrchestrator()` end-to-end against seeded parent+child fixtures, proving enable → dispatch, disable → no dispatch, re-enable → dispatch resumes
- `loop-orchestrator.unit.test.ts`: one-line fixture default fix (`parentCronId: "shipwright-loop"`) to keep the existing behavioral suite valid under the new child-row model
- `agent/src` has no database of its own (HTTP-only client to the admin service), so these tests use `CronJobLike` fixtures matching the admin API's real response shape rather than a literal Postgres connection — a deliberate call consistent with this repo's documented data-layer-ownership testing boundary (a literal DB test belongs in `admin/src/*.integration.test.ts`, which owns that schema)
- [x] `bun test agent/src/loop-cron-classifier.unit.test.ts agent/src/loop-orchestrator.unit.test.ts agent/src/loop-orchestrator.integration.test.ts` — 65 pass / 0 fail
- [x] Full `bun test agent/src/` — 1092 pass / 0 fail
- [x] `bun run --filter='*' --sequential typecheck` — clean across all 7 workspaces
- [x] `task lint` / `task check-strings` — clean


agent/src/loop-cron-classifier.ts
```diff
@@ -129,9 +137,11 @@ export function classifyCronJobsForScheduling<T extends CronJobLike>(
 
 export function resolveLoopPhaseToggles<T extends CronJobLike>(
   jobs: T[],
+  loopCronId: string,
 ): LoopPhaseToggles {
   const enabledByName = (name: string): boolean =>
```
dodizzle, line 157: Before this PR, an unreconciled agent's top-level phase rows (`parentCronId: null`) still resolved correctly since matching was by name only. Now requiring `parentCronId === loopCronId` means an unreconciled agent silently gets all four phases `false` — the loop just stops dispatching with zero log output. `loop-orchestrator.ts` already has an established pattern for this (pre-claim-failure and spin-detection both `console.warn` so soft-fail states stay observable). Consider a `console.warn` here (or in the caller) when all four phases resolve false and a `shipwright-loop` row is present but no matching child rows exist — otherwise operators can't tell "idle, no work" from "stuck, unreconciled."

dodizzle (commented):
Verdict: COMMENT — Solid change with good test coverage, but the new all-phases-silently-false soft-fail path for unreconciled agents has no warning, unlike this file's established console.warn convention for analogous soft-fail conditions.

agent/src/loop-orchestrator.ts
```diff
@@ -337,7 +337,29 @@ export function createLoopOrchestrator(
...
+        // false and the drain simply stops with no candidates collected. Without
+        // this warn that state is indistinguishable from a legitimately idle or
+        // all-phases-disabled tick. Logged so it stays observable (Sentry-eligible)
+        // instead of the loop silently going quiet.
+        if (
+          !toggles.devTask &&
+          !toggles.review &&
+          !toggles.patch &&
+          !toggles.deploy &&
+          jobs.some((job) => job.name === "shipwright-loop") &&
```
dmcaulay, line 370: This guard (`!jobs.some((job) => job.parentCronId === loopCronId)`) only fires when there are zero child rows under the loop at all. If reconciliation is partial — e.g. one phase's child row backfilled with `parentCronId`, the other three not yet — this condition is false (since at least one job matches), so the warn stays silent even though the all-phases-false tick is still a reconciliation gap, not a legitimate idle/disabled state. Since this warn exists specifically to disambiguate that case (per the prior review comment this commit addresses), consider checking whether all four phase names lack a same-parent child row, rather than checking for presence of any child row at all.

dmcaulay (commented):
Verdict: COMMENT — The child-row-scoped toggle migration itself is correct and well-tested, but the unreconciled-agent warn added in this PR's own fix commit only fires when zero child rows exist under the loop at all, missing the more likely partial-reconciliation case (see inline comment).

dmcaulay (commented):
Verdict: APPROVE — The fix commit closes the partial-reconciliation gap from the last review: the unreconciled-agent warn now fires whenever any of the four phase names lacks a same-parent child row (not only when all four do), backed by a new dedicated test. No blocking issues remain.

PR #1901, opened 2026-07-18
feat: nest phase toggles under shipwright-loop in admin cron UI

dmcaulay:
## Summary
- Excludes any cron row with a non-null `parentCronId` (the four `shipwright-loop` phase crons: dev-task/review/patch/deploy) from the flat top-level admin cron table.
- Nests those phase rows directly beneath the `shipwright-loop` row in a labeled "Phases of shipwright-loop" section, each with the same enable/disable toggle control as top-level crons.
- Parent-child matching is generic (`child.parentCronId === parent.id`), not hardcoded to the loop by name, so it stays correct if another cron gains children in the future.

## Acceptance Criteria

| 1 | The 4 phase cron rows no longer appear as flat top-level entries in the admin cron table | MET | `topLevelCrons = crons.filter((c) => !c.parentCronId)` feeds `systemCrons`/`customCrons` |
| 2 | The shipwright-loop row visually nests its 4 phase children beneath it, each independently toggleable | MET | `renderNestedPhasesRow()` renders each child via the shared `renderCronActions()` toggle form |
| 3 | The nested phase section is clearly labeled as belonging to shipwright-loop | MET | Renders "Phases of shipwright-loop" label above the nested child rows |
| 4 | Smoke test asserting exclusion from top-level + nested inclusion with working toggle forms | MET | New smoke test in `admin-ui.smoke.test.ts` asserts label presence and toggle-form wiring/ordering |

## Test Plan
- New smoke test: `"nests shipwright-loop phase crons under the loop row instead of listing them flat"` (Hono in-process `app.request()`)
- `bun test src/admin-ui.smoke.test.ts src/admin-ui-pages.unit.test.ts` — 579 pass, 0 fail
- `bunx biome lint .` — clean

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -149,6 +149,7 @@ export interface CronJobItem {
   enabled: boolean;
   name: string | null;
   system: boolean;
+  parentCronId: string | null;
   preCheck?: string | null;
   lastRun?: {
     startedAt: Date;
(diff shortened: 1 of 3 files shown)


dodizzle (approved):
Verdict: APPROVE — Clean, focused change: nesting/filtering logic is correct, escaping is consistent, and the new smoke test genuinely proves nesting via ordering assertion rather than just presence. Verified tests (579 pass), typecheck, and lint all clean; traced parentCronId's origin to confirm no orphan-parent risk in production.

PR #1911, opened 2026-07-18
feat: detect same-branch sibling task ordering before claiming a bundled task

dmcaulay:
## Summary
- Adds a "Same-Branch Sibling Check" to `dev-task` Step 1, run after branch validation and before Step 2's claim
- Detects an actively-worked (fresh-heartbeat) sibling task on the same `branch` via `GET /tasks?branch={branch}&status=in_progress`, using the task-store's own 35-minute claim-TTL formula (`DEFAULT_CLAIM_TTL_MS`) to judge freshness
- On a fresh sibling: releases this task's own claim and stops silently instead of proceeding, avoiding the repeated claim-then-release churn seen on prose-bundled tasks (e.g. SECP-1.2 waiting on SECP-1.1)
- Stale/orphaned same-branch siblings fall through unchanged to Step 4's existing Branch/PR Reality Check recovery path — no behavior change there

## Acceptance Criteria
| 1 | Detects a same-branch sibling in_progress with fresh heartbeat and defers (releases claim) instead of proceeding | MET | New "### Same-Branch Sibling Check" section in `dev-task.md` Step 1 |
| 2 | No change in behavior for genuinely orphaned/stale same-branch work | MET | Explicit fallthrough to Step 4's unchanged incomplete/stale recovery path |

## Test Plan
- 8 new content assertions in `dev-task.content.test.ts` covering: section placement (before claim, after branch validation), the `branch=`+`status=in_progress` query, self-exclusion, the 35-min TTL freshness formula, the release+`[silent]` defer path, and the unchanged stale fallthrough
- Narrowed one pre-existing guard test that banned the literal substring `status=in_progress` (originally guarding against the retired bare self-scan) so it no longer collides with this new, legitimately-scoped query
- `bun test plugins/shipwright/` — 896/896 pass
- `task lint`, `task check-strings`, `tsc --noEmit` — all clean

### DIFF
--- a/plugins/shipwright/commands/dev-task.content.test.ts
+++ b/plugins/shipwright/commands/dev-task.content.test.ts
@@ -36,7 +36,12 @@ describe("dev-task.md — explicit-target-only argument contract", () => {
   });
 
   it("removes the in_progress resume-check query from Step 1", () => {
-    expect(content).not.toContain("status=in_progress");
+    // The old bare self-scan (`?status=in_progress&assignee=$SHIPWRIGHT_AGENT_ID`, unscoped
+    // to any specific task) must be gone. The Same-Branch Sibling Check's targeted
+    // `?branch={branch}&status=in_progress` lookup is a different, legitimate query and is
+    // not banned by this guard.
+    expect(content).not.toContain("tasks?status=in_progress&assignee=$SHIPWRIGHT_AGENT_ID");
+    expect(content).not.toContain('"$SHIPWRIGHT_TASK_STORE_URL/tasks?status=in_progress"');
     expect(content).not.toContain("Resuming interrupted task");
   });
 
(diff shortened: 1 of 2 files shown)


plugins/shipwright/commands/dev-task.md
```diff
@@ -93,6 +93,50 @@ curl -sf -X PATCH -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
...
 Post the message above and stop. A missing branch cannot be recovered from at runtime — worktree creation will fail silently if attempted.
 
+### Same-Branch Sibling Check
+
+Some tasks are bundled onto an existing branch by prose in their `description` rather than
+a tracked `dependencies` entry (e.g. "joins branch X, land after task Y is merged"). Before
+claiming (pending) or resuming (in_progress) this task, check whether another task is
+already actively working the same `{branch}` — claiming (or continuing to hold a claim on)
+a task whose same-branch prerequisite is still mid-flight wastes a claim cycle and starves
+the prerequisite. This applies regardless of this task's own status.
```
dodizzle, line 104: This section says it applies "regardless of this task's own status," and the PR's acceptance criteria/tests assert coverage for "pending or resumed in_progress." But Step 1's pre-existing `status == "in_progress"` branch (untouched by this diff, ~line 46) says to "skip Step 2's claim and proceed directly to Step 3," and explicitly names only Step 4's Branch/PR Reality Check as the one exception that runs "unconditionally... regardless of this status." That's the same pattern that makes the older Dependency Check section scoped "(pending tasks only)." This new section was never added to that in_progress cross-reference, so a resumed task may skip straight to Step 3 and never run this check — missing exactly the resumed-session scenario it's meant to cover. Consider adding an explicit pointer from the in_progress bullet (like the one for Step 4), or moving this check above the status branch so both paths pass through it.

dodizzle (commented):
Verdict: COMMENT — Well-verified change (TTL formula and query filters check out exactly against the code), but the new Same-Branch Sibling Check may not actually be reached on the resumed `in_progress` path it explicitly claims to cover — see inline comment.

dodizzle (commented):
Verdict: COMMENT — The prior in_progress-reachability gap is correctly fixed and well-tested, but this fix commit introduces a new Biome lint violation that breaks the merge-blocking `task ci` gate (confirmed red on GitHub Actions for this exact commit).

plugins/shipwright/commands/dev-task.content.test.ts
```diff
@@ -45,7 +50,18 @@ describe("dev-task.md — explicit-target-only argument contract", () => {
...
+  });
+
+  it("in_progress path is explicitly routed through the Same-Branch Sibling Check before Step 3", () => {
+    // The in_progress bullet must not just skip to Step 3 — it must pass through the
+    // Same-Branch Sibling Check first, or a resumed task never runs that check.
+    const inProgressBulletMatch = content.match(
+      /\*\*Found, `status == "in_progress"`\*\*:[\s\S]*?(?=\n- \*\*Found)/,
+    );
+    expect(inProgressBulletMatch).not.toBeNull();
+    const bullet = inProgressBulletMatch![0];
```
dodizzle: `inProgressBulletMatch![0]` trips Biome's `noNonNullAssertion` rule — confirmed this is why the `CI` workflow run on this commit (cb145f6c) failed (`task lint` is the first stage of the merge-blocking `task ci` chain). The preceding line already does `expect(inProgressBulletMatch).not.toBeNull()`, so swapping to `inProgressBulletMatch?.[0]` fixes the lint error without changing test semantics.

dodizzle (commented):
Verdict: COMMENT — The prior Biome lint fix is correct and CI is now green, but the new Same-Branch Sibling Check query should be scoped by `repo` to avoid a false-positive deferral for agents whose token spans multiple repos.

plugins/shipwright/commands/dev-task.md
```diff
@@ -93,6 +94,50 @@ curl -sf -X PATCH -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
...
+
+```bash
+curl -sf -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
+  "$SHIPWRIGHT_TASK_STORE_URL/tasks?branch={branch}&status=in_progress" | jq '.tasks'
```
dodizzle: This query omits `repo`. `task-store`'s list handler filters repo-scoped agent tokens via `agentScope: { agentId, repos }` — an OR union across every repo in the token's scope, not a single repo. An agent scoped to multiple repos querying an unscoped `branch`+`status` filter could match an in_progress task on an identically-named branch in a *different* scoped repo, causing this task to defer against a false-positive "sibling." Suggest adding `&repo={repo}` to the query.

plugins/shipwright/commands/dev-task.md
```diff
@@ -45,11 +45,12 @@ curl -sf -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
...
-  branch/PR left behind by the prior session is discovered and handled there — no separate
-  orphan-recovery path is needed here. Record `task_started_at` (current ISO timestamp) for
-  metrics.
+  Step 2's claim, run the Same-Branch Sibling Check (below) — it applies to resumed
+  in_progress tasks too, see that section — then proceed to Step 3. Step 4's Branch/PR
+  Reality Check (below) runs unconditionally before worktree creation regardless of this
+  status, so any stale branch/PR left behind by the prior session is discovered and handled
+  there — no separate orphan-recovery path is needed here. Record `task_started_at` (current
+  ISO timestamp) for metrics.
 - **Found, `status == "pending"`**: check dependency satisfaction before claiming (see
```
dodizzle, line 54: The `in_progress` bullet above was explicitly edited this PR to say "run the Same-Branch Sibling Check (below) ... then proceed to Step 3" — a fix for exactly this reachability ambiguity, per round 1 of this review. This `pending` bullet still just says "proceed to Step 2's Mark In-Progress with this task," with no equivalent forward-reference to the new check. The Same-Branch Sibling Check section does self-declare it covers "claiming (pending) or resuming (in_progress)" and sits earlier in the document than Step 2, so a careful linear read still reaches it — but pending is the primary scenario this PR targets, and leaving this bullet without the same explicit callout the in_progress bullet just got is an inconsistency worth closing for LLM-execution reliability. Suggest: "...If satisfied, run the Same-Branch Sibling Check (below), then proceed to Step 2's Mark In-Progress."

dodizzle (commented):
Verdict: COMMENT — Well-verified change (TTL formula, self-exclusion, and the task-store endpoints all check out against the code, and both prior review rounds' findings are fixed with CI green), but the pending-task bullet doesn't get the same explicit "run the Same-Branch Sibling Check" callout that the in_progress bullet just received — see inline comment.

dodizzle (approved):
Verdict: APPROVE — Both prior findings (in_progress-path reachability, then the pending-bullet routing gap) are fixed and verified against the actual code; the repo-scoping fix for the sibling query also checks out against task-store's agentScope filter. Tests (40/40) and lint are clean.

PR #1918, opened 2026-07-19
feat: add Created column to Tasks list (ATC-1.5)

dmcaulay:
## Summary
- Adds a `Created` column to `renderTasksPage()`'s table, rendering `task.createdAt` — matching the underlying `createdAt` sort order and the sibling `renderPrsPage()` Created column pattern (ATC-1.1).
- Joins the new column to the existing `.col-session`/`.col-repo` mobile-hide set (`.col-created`) in `admin-ui-styles.ts`, given an 8th visible column on a 640px viewport warrants the same density fix.
- Adds unit test coverage in `admin-ui-pages.unit.test.ts` for the header, formatted-date rendering, missing-value fallback, invalid-date NaN guard, and the mobile-hide class.

## Acceptance Criteria
| 1 | renderTasksPage()'s table has a new 'Created' column rendering task.createdAt | MET | `<th class="col-created">Created</th>` + per-row `createdCell` rendering `t.createdAt` |
| 2 | New column joins mobile-hide class set if row density warrants | MET | `.col-session, .col-repo, .col-created { display: none; }` in admin-ui-styles.ts |
| 3 | Unit test added asserting Created column renders correct value | MET | New `describe("renderTasksPage — Created column")` block, 5 tests, all passing |

## Test Plan
- [x] `bun test admin-ui-pages.unit.test.ts` — 422 pass, 0 fail
- [x] `bun test` (full admin suite) — 1330 pass, 0 fail, 175 skip
- [x] `tsc --noEmit` — clean

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -1536,6 +1536,7 @@ export function renderTasksPage(
   opts?: { error?: string; agentFilterActive?: boolean },
   suggestions?: { sessions?: string[]; repos?: string[]; agents?: string[] },
   readOnly = false,
+  timezone = "America/Los_Angeles",
 ): string {
   const errorHtml = opts?.error
     ? `<div class="alert alert-error">${escapeHtml(opts.error)}</div>`
@@ -1621,6 +1622,7 @@ export function renderTasksPage(
                       : d.toLocaleString("en-US", {
                           dateStyle: "medium",
                           timeStyle: "short",
+                          timeZone: timezone,
                         });
                   })(),
                 )
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -1596,7 +1596,7 @@ export function renderTasksPage(
 
   const rows =
     tasks.length === 0
-      ? `<tr><td colspan="8" class="empty-state">No tasks found.</td></tr>`
+      ? `<tr><td colspan="9" class="empty-state">No tasks found.</td></tr>`
       : tasks
           .map((t) => {
             const agentId = t.claimedBy ?? t.assignee;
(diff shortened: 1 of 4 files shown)


dodizzle (commented):
Verdict: COMMENT — The new Created-column date formatting omits the `timeZone` option that every other identical `dateStyle`/`timeStyle` call site in this file passes, so the same `task.createdAt` will render at a different wall-clock time on the Tasks list page than on the Task Detail page.

admin/src/admin-ui-pages.ts
```diff
@@ -1612,6 +1612,19 @@ export function renderTasksPage(
                 : t.prUrl
                   ? `<a href="${escapeHtml(t.prUrl)}" style="color:#6366f1;text-decoration:none" title="View PR">#${t.pr ?? "PR"}</a>`
                   : '<span style="color:#9ca3af">—</span>';
+            const createdCell = t.createdAt
```
dodizzle, line 1616: This omits the `timeZone` option that every other `dateStyle`/`timeStyle` date cell in this file passes (e.g. `renderTaskDetailPage`'s `dateField` at line 1915, `renderPrsPage`'s `createdCell` at line 2612), both using a `timezone` param defaulting to `"America/Los_Angeles"`. `renderTasksPage` has no `timezone` parameter to thread through. As a result, the same `task.createdAt` will render at a different wall-clock time here than on the Task Detail page's "Added"/"Created" field for the same task. Suggest adding a `timezone = "America/Los_Angeles"` param to `renderTasksPage` and passing `timeZone: timezone` here, mirroring the sibling pages.

dodizzle (approved):
Verdict: APPROVE — Timezone fix from the prior review is verified complete and correct: `timezone` threads through both `renderTasksPage()` call sites at the right position, matches the sibling `dateField` pattern (NaN guard + escapeHtml), and all three acceptance criteria are met.

PR #1919, opened 2026-07-19
feat: wrap agent detail sub-tables and add Created columns (ATC-1.2)

dmcaulay:
## Summary
- Wrap all 8 top-level `<table class="data-table">` blocks in `renderAgentDetailPage()` (Members, Env Vars, Repos, System Crons, Custom Crons, Tools, Task Store Tokens, Plugins) in `.data-table-wrapper` so mobile content scrolls horizontally instead of clipping.
- Add a `Created` column (rendering `createdAt`) to the Tools, Plugins, System Crons, and Custom Crons tables — all four are sorted by `createdAt asc` server-side but previously showed no timestamp at all (Crons tables kept `Last run`, a distinct field, alongside the new column).
- Add `.col-created { display: none; }` to the mobile media query, scoped to the Crons tables (the densest, now at 8 visible columns) — Tools/Plugins stay visible at 4 columns.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | All 8 tables in renderAgentDetailPage() are wrapped in .data-table-wrapper | MET |
| 2 | Tools table has a new 'Created' column rendering tool.createdAt | MET |
| 3 | Plugins table has a new 'Created' column rendering plugin.createdAt | MET |
| 4 | System Crons and Custom Crons tables each have a new 'Created' column rendering cron.createdAt, alongside the existing 'Last run' column | MET |
| 5 | New mobile-hide class(es) added to admin-ui-styles.ts's media query for tables where the added column needs it | MET |
| 6 | Unit tests added to the existing Members/Tools/Plugins/Crons describe blocks in admin-ui-pages.unit.test.ts asserting wrapper presence and new Created-column values; no existing tests removed, purely additive | MET |

## Test Plan
- 20 new unit tests added across Members/Env Vars/Repos/Crons/Tools/Tokens/Plugins describe blocks in `admin-ui-pages.unit.test.ts`, asserting wrapper presence and Created-column header/value rendering.
- `bun test admin/src/admin-ui-pages.unit.test.ts` → 432 pass / 0 fail.
- `task typecheck` clean across all workspaces.
- `task ci` (lint → check-strings → typecheck → check-config-docs → check-version-sync → test:coverage → secret-scan) passes; coverage gate 80.45%/81.06% (threshold 80%).
- `docs/agent.md` refreshed in a separate commit to document the new `createdAt`/`updatedAt` fields now surfaced on `AgentCronJob`, `AgentTool`, `AgentPlugin` in the data model table.


admin/src/admin-ui-pages.unit.test.ts
```diff
@@ -852,6 +883,45 @@ describe("renderAgentDetailPage — crons", () => {
...
+    // assert there are two independent wrapper+table+Schedule occurrences.
+    const matches = html.match(
+      /<div class="data-table-wrapper">\s*<table class="data-table">\s*<thead>\s*<tr>\s*<th>Schedule<\/th>/g,
+    );
+    expect(matches?.length).toBe(2);
+  });
+
+  test("System/Custom Crons table has a 'Created' column header", () => {
+    const html = render([SYSTEM_CRON]);
+    const createdHeaderCount = (html.match(/<th>Created<\/th>/g) ?? []).length;
```
dodizzle: This assertion (and the `<th>Created</th>` check in the next test) doesn't actually test the Crons table's markup — Crons renders `<th class="col-created">Created</th>` (admin-ui-pages.ts:1092/:1133), which never matches the bare `<th>Created<\/th>` regex. The count of 2+ is satisfied by the always-rendered Tools/Tokens/Plugins headers on the same page output (their `[]` fixtures still render static `<thead>` markup), regardless of whether the Crons table has a Created column at all. Suggest matching the real markup: `expect(html).toContain('<th class="col-created">Created</th>')`.

dodizzle (commented):
Verdict: COMMENT — One important finding: two new unit tests for the Crons table's 'Created' column are tautological — they match always-rendered headers from the Tools/Tokens/Plugins tables instead of the Crons table's actual `<th class="col-created">Created</th>` markup, so acceptance criterion 4/5 for Crons is effectively unverified. Everything else (wrapper restructuring, colspan bumps, escaping, CSS scoping) checks out cleanly.

dodizzle (approved):
Verdict: APPROVE — Clean, additive change: all 8 sub-tables wrapped for mobile scroll, Created columns added to Tools/Plugins/Crons with consistent escaping, colspans correctly bumped, and the prior review's tautological-test finding is verifiably fixed in this commit.

PR #1924, opened 2026-07-19
fix: wrap dependency graph in scrollable container on mobile

dmcaulay:
## Summary
- Wrap `renderDependencyGraph()`'s output in `<div class="data-table-wrapper">` at its call site in `renderSessionDetailPage()`'s `depsSection`, reusing the existing generic scroll+shadow CSS class already applied to tables elsewhere in the file.
- Remove the dead `overflow-x:auto` from the graph's own inner div — that div is sized exactly to its content (via `computeDependencyLayout`) so it never overflows itself; the real overflow happens one level up, where the graph sits inside a plain `.card` with no scroll handling.

## Acceptance Criteria
| 1 | Graph wrapped in `<div class="data-table-wrapper">` at depsSection call site | MET | `admin-ui-pages.ts` — wrapper div added around `renderDependencyGraph(dependencyNodes)` |
| 2 | Dead `overflow-x:auto` removed from inner exactly-sized div | MET | `admin-ui-pages.ts` — style trimmed to `position:relative;width:...px;height:...px` |
| 3 | Unit test asserting wrapper present when dependencyNodes.length > 0 | MET | new test "graph is wrapped in .data-table-wrapper for horizontal scroll on mobile" |
| 4 | Unit test covering 2+ depth columns (width > 375px), wrapper carries overflow-x:auto not inner div | MET | new test "2-depth-column graph..." — wrapper class present, inner div's inline style lacks overflow-x:auto, layout.width (620px) > 375px |
| 5 | No changes to `computeDependencyLayout()`'s layout math | MET | diff touches only `renderDependencyGraph()` and `renderSessionDetailPage()` |

## Test Plan
- `bun test admin/src/admin-ui-pages.unit.test.ts` — 445 pass, 0 fail (includes 2 new tests, TDD red→green confirmed)
- `bunx biome lint admin/src` — clean
- `tsc --noEmit` (admin) — clean

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -2450,7 +2450,7 @@ function renderDependencyGraph(nodes: DependencyNode[]): string {
       ${edgePaths.join("\n      ")}
     </svg>`;
 
-  return `<div style="position:relative;width:${layout.width}px;height:${layout.height}px;overflow-x:auto">
+  return `<div style="position:relative;width:${layout.width}px;height:${layout.height}px">
       ${svgHtml}
       ${nodes.map(nodeCard).join("\n      ")}
     </div>`;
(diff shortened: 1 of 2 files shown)


dodizzle (approved):
Verdict: APPROVE — Minimal, well-scoped fix reusing the existing `.data-table-wrapper` pattern for the dependency graph's mobile overflow; new tests verify both wrapper placement and that the tested scenario genuinely needs the scroll. Tests (445 pass) and lint verified locally.

PR #1586, opened 2026-07-15
feat: port pr_open reconciliation and taskId backfill into pr-state-reconciler

dodizzle:
## Summary
- Extends `agent/src/pr-state-reconciler.ts`'s existing sweep with a new pass that reconciles `pr_open` tasks whose linked PR has actually merged on GitHub, via both the direct-pr-number path and a branch-based fallback (mirroring the legacy `check-deploy.ts`'s `reconcileStalePrOpenTasks`)
- Backfills `PullRequest.taskId` when a Task↔PR match is resolved and the PR record's `taskId` is currently null
- Per-task error isolation matches the existing sweep's per-record pattern — one lookup failure doesn't abort the batch

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | New pass reconciles pr_open tasks via both direct-pr-number and branch-fallback paths | MET |
| 2 | taskId backfilled only when currently null; existing taskId left untouched | MET |
| 3 | Per-task lookup failure doesn't abort the batch | MET |
| 4 | pr-state-reconciler.unit.test.ts extended with injected fakes covering all specified cases | MET |

## Test Plan
- `bun test agent/src/pr-state-reconciler.unit.test.ts` — 17/17 passing (7 pre-existing + 10 new)
- `bunx biome lint .` — clean
- `tsc --noEmit` (agent package) — clean
- [x] Pre-commit checks passing

Generated with [Claude Code](https://claude.com/claude-code)

PR #1939, opened 2026-07-19
docs(planning): spec the deploying→deployed reconciliation pass (DSR-2.1)

dodizzle:
Specs a fourth pass for `agent/src/pr-state-reconciler.ts`, extending the same file `DSR-1.1` (#1586) extended for `pr_open → merged`.

## Problem
Tasks are stranded in `deploying` forever when their deploy actually succeeded. The `deploying → deployed` PATCH is the last step of the `/shipwright:deploy` command — an agent action. `promote.yml` has no task-store step, so an interrupted agent session strands the task while the code ships normally.

Observed 2026-07-19: a consuming repo had **ten** tasks stuck in `deploying`, oldest for two days. Every one had its merge commit on `main` and a successful *Promote to Prod* run 9–13 minutes after merge. Corrected by hand; nothing prevents recurrence.

Same class of bug as `DSR-1.1` one status earlier — an agent-owned status write that silently doesn't happen — so it belongs in the same reconciler rather than a new mechanism.

## Design
`reconcileDeployingTasks(deps)` using the existing `makeListTasksByStatus` helper. For each `deploying` task, find the latest successful *Promote to Prod* run for `task.repo`; if its `createdAt` post-dates `task.mergedAt`, PATCH to `deployed` with that timestamp.

Adds one injected dep, `ghListWorkflowRuns`, keeping the file's DI shape so the pass unit-tests with no network.

**Deliberate constraints**
- `deployedAt` is the promote run's `createdAt`, **never `now()`** — a reconcile-time stamp inflates cycle-time metrics and is worse than the null it replaces because it looks plausible.
- The only transition is `deploying → deployed`. A reconciler that can move tasks out of `pending`/`blocked`/`cancelled` is a queue-corruption vector.
- Requires `conclusion === "success"` and non-null `mergedAt`; skips and logs anything ambiguous.
- Per-task error isolation, matching `reconcilePrOpenTasks`.

## Blast radius (stated so it isn't oversold)
Task **dependencies are unaffected** — they're satisfied at `merged`/`deploying` already, so stranded tasks never block downstream work. What actually breaks: `error-resolve` (requires `deployed`/`done` before resolving a linked issue), planning-gate conventions keyed on `deployed`/`done`, and cycle-time metrics.

## Out of scope
Moving the `deployed` PATCH into `promote.yml` — the more complete fix, but it changes the pipeline contract for every consumer and belongs in its own change.

Queued as `DSR-2.1`.

https://claude.ai/code/session_015rcdzRLZtnvkxgq3pr1wTk

### DIFF
--- a/planning/deploy-status-reconciliation/PRODUCT-SPEC-deploying-pass.md
+++ b/planning/deploy-status-reconciliation/PRODUCT-SPEC-deploying-pass.md
@@ -1,5 +1,14 @@
 # Deploy-Status Reconciliation — `deploying → deployed` pass
 
+> **STATUS: Implemented.** This pass shipped in PR #1941 (merged 2026-07-19T16:20:52Z), in
+> `agent/src/pr-state-reconciler.ts`'s `reconcileDeployingTasks(deps)` — same function name,
+> same injected `ghListWorkflowRuns` dependency shape, same per-task error isolation, and the
+> same "`deployedAt` is the promote run's `createdAt`, never `now()`" constraint this spec
+> describes below. Full test coverage for every case in this doc's Testing section already
+> exists in `agent/src/pr-state-reconciler.unit.test.ts` (`describe("reconcileDeployingTasks", …)`).
+> Do not queue a new DSR-2.1 implementation task off this spec — the rest of the document is
+> kept as the historical design record for what was built and why.
+
 **Date**: 2026-07-19
 **Session**: `deploy-status-reconciliation`
 **Task**: `DSR-2.1`
--- a/planning/deploy-status-reconciliation/PRODUCT-SPEC-deploying-pass.md
+++ b/planning/deploy-status-reconciliation/PRODUCT-SPEC-deploying-pass.md
@@ -1,0 +1,107 @@
+# Deploy-Status Reconciliation — `deploying → deployed` pass
+
+**Date**: 2026-07-19
+**Session**: `deploy-status-reconciliation`
+**Task**: `DSR-2.1`
+**Extends**: `agent/src/pr-state-reconciler.ts` (the same file `DSR-1.1` / PR #1586 extended for the
+`pr_open → merged` pass)
+
+## Problem
+
+Tasks are stranded in `deploying` forever when their deploy actually succeeded.
+
+The `deploying → deployed` transition is **not performed by CI**. It is the final step of the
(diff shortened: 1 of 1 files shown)


planning/deploy-status-reconciliation/PRODUCT-SPEC-deploying-pass.md
```diff
@@ -0,0 +1,107 @@
...
+  `merged`/`deploying`/`deployed`/`done`/`cancelled`, so stranded tasks never block downstream work.
+- `error-resolve` **is** affected — it requires `deployed`/`done` before resolving a linked issue, so
+  a stranded task silently keeps issues open.
+- Any planning-gate convention that treats only `deployed`/`done` as met will report a doc blocked
+  when its gate actually shipped.
+- `deployedAt` drives cycle-time metrics; a null on a shipped task corrupts that reporting.
+
+## Design
+
+A fourth pass in `agent/src/pr-state-reconciler.ts`, alongside `reconcilePrState`,
```
dodizzle, line 49: `agent/src/pr-state-reconciler.ts` on `main` already exports `reconcileDeployingTasks(deps)` with exactly this shape — same `ghListWorkflowRuns` dependency, same per-task error isolation, same "never `now()`" constraint, same `DSR-2.1` references in its JSDoc — and `pr-state-reconciler.unit.test.ts` already covers every case this spec's Testing section lists. It landed via PR #1941, merged 2026-07-19T16:20:52Z. This spec presents the pass as not-yet-built; worth confirming it isn't now stale before merging, to avoid a duplicate DSR-2.1 task getting queued for already-shipped work.

dodizzle (commented):
Verdict: COMMENT — This spec describes a `reconcileDeployingTasks` pass that already shipped in PR #1941 (merged 2026-07-19T16:20:52Z, same function name/shape/constraints, full test coverage already on main). Worth a rebase-and-recheck before merging so this doesn't queue a duplicate DSR-2.1 task for work that's already done.

dmcaulay (approved):
Verdict: APPROVE — Doc-only PR; every claim in the "STATUS: Implemented" callout independently checks out against main (PR #1941 merged, `reconcileDeployingTasks` shape/constraints/tests all match), and the one prior human review comment (already-shipped duplicate concern) was addressed by this PR's second commit before I reviewed. Minor cosmetic nit: the Acceptance Criteria checkboxes are still unchecked despite the doc claiming full implementation — not blocking.

PR #1940, opened 2026-07-19
feat(mcp-server): mount MCP Streamable HTTP transport and add server entry point

dodizzle:
## Summary
- Mount the MCP protocol over Streamable HTTP on the mcp-server Hono app (`POST`/`GET`/`DELETE /mcp`), bridging Hono's Web-Standard `Request`/`Response` directly to the SDK's `WebStandardStreamableHTTPServerTransport` — stateful sessions, keyed by `mcp-session-id`.
- Add `mcp-server/src/main.ts`, a minimal `Bun.serve` entry point that binds `PORT` (default `3010`), plus a `"start"` script in `mcp-server/package.json`.
- Keep the existing stdio entry point (`src/serve.ts`) untouched — both transports share `createMcpServer()`.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | POST /mcp completes an MCP initialize handshake and returns a session id | ✅ |
| 2 | tools/list over HTTP returns the same allowlisted tools as the stdio transport | ✅ |
| 3 | tools/call over HTTP proxies to the task store and returns the tool result | ✅ |
| 4 | mcp-server has a start script and a Bun.serve entry point that binds PORT | ✅ |
| 5 | src/serve.ts (stdio) still works unchanged | ✅ |
| 6 | Smoke test covers the HTTP transport end to end via app.request() | ✅ |

## Test Plan
- `mcp-server/src/http-transport.smoke.test.ts` drives a real `initialize` → `tools/list` → `tools/call` handshake via `app.request()` (no real socket), plus a rejected-request edge case.
- `bun test mcp-server/` — 37 pass, 0 fail.
- `tsc --noEmit` and `bunx biome lint` clean.
- Coverage 94.65% lines / 95.24% functions for the touched package (repo gate is 80/80).

### DIFF
--- a/mcp-server/src/http-transport.ts
+++ b/mcp-server/src/http-transport.ts
@@ -1,0 +1,102 @@
+/**
+ * http-transport.ts
+ * Mounts the MCP protocol over Streamable HTTP on a Hono app.
+ *
+ * Bridges Hono's Web-Standard Request/Response directly to the SDK's
+ * `WebStandardStreamableHTTPServerTransport` — no Node-stream shim required
+ * (see @modelcontextprotocol/sdk/server/webStandardStreamableHttp.js).
+ *
+ * Runs in STATEFUL mode: each new session (an `initialize` request with no
+ * `mcp-session-id` header) gets its own transport + `createMcpServer()`
+ * instance, keyed by the session id the transport generates. Follow-up
+ * requests carrying a known `mcp-session-id` header reuse that same
+ * transport instance, which validates the session internally on
+ * GET/DELETE/subsequent POSTs.
+ *
+ * The request body is read once (`await c.req.json()`) and threaded through
+ * via `options.parsedBody` so the transport doesn't re-read the stream.
+ */
+
+import { randomUUID } from "node:crypto";
+import type { Hono } from "hono";
+import { WebStandardStreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/webStandardStreamableHttp.js";
+import { isInitializeRequest } from "@modelcontextprotocol/sdk/types.js";
+import { type CreateMcpServerOptions, createMcpServer } from "./mcp-server.ts";
+
+const MCP_PATH = "/mcp";
+const SESSION_ID_HEADER = "mcp-session-id";
+
+export interface MountMcpHttpTransportOptions extends CreateMcpServerOptions {}
+
+/**
+ * Mount the MCP Streamable HTTP endpoint (`/mcp`) on `app`, handling
+ * POST (JSON-RPC messages, including `initialize`), GET (SSE stream for
+ * server-initiated messages), and DELETE (session termination).
+ */
+export function mountMcpHttpTransport(
+  app: Hono,
+  options: MountMcpHttpTransportOptions = {},
+): void {
+  const transports = new Map<
+    string,
+    WebStandardStreamableHTTPServerTransport
+  >();
+
+  app.post(MCP_PATH, async (c) => {
+    const sessionId = c.req.header(SESSION_ID_HEADER);
+    const existing = sessionId ? transports.get(sessionId) : undefined;
+
+    if (existing) {
+      return existing.handleRequest(c.req.raw);
+    }
+
+    // No known session: only a fresh `initialize` request may start one.
+    const parsedBody = await c.req.json();
+    if (sessionId || !isInitializeRequest(parsedBody)) {
+      return Response.json(
+        {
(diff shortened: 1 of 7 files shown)


dodizzle (commented):
Verdict: COMMENT — the module-level `transports` Map in `http-transport.ts` only evicts entries via an explicit client `DELETE /mcp`; there's no idle-timeout or reaper, so any remote client that disconnects, crashes, or never sends DELETE leaks its transport instance forever. Everything else (all 6 acceptance criteria, PORT parsing, the smoke test, docs) checks out.

mcp-server/src/http-transport.ts
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Mount the MCP Streamable HTTP endpoint (`/mcp`) on `app`, handling
+ * POST (JSON-RPC messages, including `initialize`), GET (SSE stream for
+ * server-initiated messages), and DELETE (session termination).
+ */
+export function mountMcpHttpTransport(
+  app: Hono,
+  options: MountMcpHttpTransportOptions = {},
+): void {
+  const transports = new Map<
```
dodizzle, line 80: The `transports` Map only evicts an entry via `transport.onclose`, which the SDK only fires from an explicit `DELETE /mcp`. There's no idle-timeout, TTL, or reaper anywhere in this file. Since this endpoint is meant for remote clients (Claude Desktop connectors, external applications per the PR description), any client that disconnects, crashes, or simply never sends DELETE leaves its transport + `createMcpServer()` instance in this Map forever — unbounded growth for the life of the process. Consider an idle-timeout sweep (track last-activity per session, periodically evict stale entries) before this runs as a long-lived deployed service.

mcp-server/src/http-transport.ts
```diff
@@ -0,0 +1,159 @@
...
+    return transport.handleRequest(c.req.raw, { parsedBody });
+  });
+
+  const handleSessionRequest = async (c: Context): Promise<Response> => {
+    const sessionId = c.req.header(SESSION_ID_HEADER);
+    const transport = sessionId ? transports.get(sessionId) : undefined;
+    if (!transport) {
+      return new Response("Invalid or missing session ID", { status: 400 });
+    }
+    touch(sessionId);
```
dodizzle, line 172: `touch()` only fires when a request opens (POST/GET/DELETE), not per server-pushed SSE message. The SDK's standalone GET SSE stream (`_standaloneSseStreamId`) stays open indefinitely for server-initiated pushes with no further client request — a client that opens the stream and only receives pushes for longer than `idleTimeoutMs` (30 min default) will have that still-active connection forcibly evicted and closed by the reaper. Consider touching the session on each SSE event sent, not just at stream-open time.

mcp-server/src/http-transport.ts
```diff
@@ -0,0 +1,159 @@
...
+    ...serverOptions
+  } = options;
+
+  const transports = new Map<
+    string,
+    WebStandardStreamableHTTPServerTransport
+  >();
+  const lastActivity = new Map<string, number>();
+
+  const touch = (sessionId: string | undefined): void => {
```
dodizzle, line 86: This reaper calls `Date.now()` directly with no injectable `Clock`, and the new test drives it with real `setTimeout` waits rather than a fixed clock. `task-store/src/stale-claim-reaper.ts` is a near-identical TTL-based reaper in this repo that takes an injected `Clock` (`SystemClock()`/`FixedClock()`) and is tested without real time delays — the pattern root CLAUDE.md's Test isolation section calls for ("inject time via a Clock"). Worth threading a `Clock` through here to match.

dodizzle (commented):
Verdict: COMMENT — the session-leak fix is structurally correct (transports/lastActivity stay in sync, all 5 tests pass, typecheck/lint clean), but the reaper can force-close a genuinely active push-only SSE stream since `touch()` isn't refreshed per pushed message, and its time logic bypasses this repo's Clock-injection convention used by the near-identical `task-store/src/stale-claim-reaper.ts`.

dodizzle (commented):
Verdict: APPROVE — both prior findings (SSE stream force-closed by the reaper, missing Clock-injection) are genuinely fixed: `transport.send` is now wrapped to touch activity on every server push, and the reaper takes an injectable `Clock` mirroring `task-store/src/clock.ts`. Both fixes are covered by new regression tests using a manually-advanced `FixedClock`, no real timer waits. All 6 acceptance criteria remain satisfied and no new issues were introduced.

PR #1941, opened 2026-07-19
feat: add deploying-to-deployed reconciliation pass to pr-state-reconciler (DSR-2.1)

dmcaulay:
## Summary
- Adds `reconcileDeployingTasks(deps)`, a fifth reconciliation pass in `agent/src/pr-state-reconciler.ts`, self-healing task-store Tasks left stuck at `status:"deploying"` when the deploy actually succeeded but the agent session crashed before the completion PATCH.
- For each `status:"deploying"` task, finds the latest successful `"Promote to Prod"` GitHub Actions run for the task's repo and, only when its `createdAt` is strictly after the task's `mergedAt`, PATCHes the task to `status:"deployed"` with `deployedAt` set to that run's real creation time (never `now()`, to avoid inflating cycle-time metrics).
- Registered as its own independent try/catch on the same reconciler tick in `agent/src/index.ts` (mirrors the `reconcileReviewState` precedent), reusing the existing `reconcilerDeps` object.
- Refreshes `docs/agent.md` and `docs/configuration.md` to document the new Pass 5.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Deploying task transitions to deployed with deployedAt = run's createdAt, not reconcile time | MET |
| 2 | Deploying task with no successful promote is left untouched | MET |
| 3 | Task in any other status never modified | MET |
| 4 | Null mergedAt skipped and logged | MET |
| 5 | Single task's lookup failure doesn't abort batch | MET |
| 6 | Nothing-to-reconcile tick performs zero writes; re-run is no-op | MET |
| 7 | Tests extend existing unit file with injected fakes, no existing tests retired | MET |

## Test Plan
- Extended `agent/src/pr-state-reconciler.unit.test.ts` with a new `describe("reconcileDeployingTasks", ...)` block: 11 new tests covering all acceptance criteria plus edge cases (non-`success` conclusion ignored, `createdAt === mergedAt` not reconciled, pagination beyond default page limit, per-task error isolation).
- `NODE_ENV=test bun test agent/src/pr-state-reconciler.unit.test.ts` — 60 pass, 0 fail (51 pre-existing + new, none retired)
- `NODE_ENV=test bun test agent/` — 1108 pass, 0 fail (no regressions)
- `cd agent && bunx tsc --noEmit` — clean
- `bunx biome lint .` — clean


agent/src/pr-state-reconciler.ts
```diff
@@ -888,6 +1032,34 @@ export function buildProductionDeps(opts: {
...
+    ) => {
+      const [org, repoName] = splitOrgRepo(repo);
+      const data = await ghJson<{
+        workflow_runs: Array<{
+          name: string;
+          status: string;
+          conclusion: string | null;
+          created_at: string;
+          id?: number;
+        }>;
```
dmcaulay, line 1068: This fetches the 20 most recent workflow runs of *any* workflow in the repo (no `head_sha`/`status`/workflow scoping), then filters client-side for `r.name === "Promote to Prod"`. The comment above claims this mirrors `check-deploy.ts`'s `fetchActiveDeployRuns`/`fetchCiRuns` precedent "exactly," but those filter server-side (`status=in_progress`/`queued`, or `head_sha=`) — neither does an unscoped repo-wide fetch. On `vitals-os` (the only configured repo with a `Promote to Prod` workflow), I saw live bursts of unrelated activity (9 "Bump Shipwright Chart" runs in 5 minutes) that would push a real successful promote run outside this top-20 window well within the 45-minute reconcile interval — leaving the task stuck at `deploying`, the exact failure mode this PR fixes. Consider using the workflow-scoped runs endpoint (`/repos/{owner}/{repo}/actions/workflows/{workflow_id_or_filename}/runs`) instead, which filters server-side.

dmcaulay (commented):
Verdict: COMMENT — `ghListWorkflowRuns` fetches the last 20 repo-wide workflow runs unscoped and filters client-side by name, so a burst of unrelated workflow activity (confirmed live on vitals-os, e.g. 9 "Bump Shipwright Chart" runs in 5 minutes) can push the real "Promote to Prod" run out of the window and leave the task stuck at deploying — the exact bug this pass is meant to fix. Everything else (pagination, per-task error isolation, deployedAt-never-now() design, test coverage) looks solid.

dmcaulay (commented):
Verdict: COMMENT — The `ghListWorkflowRuns` scoping fix from the prior review is correct and complete (workflow id resolved server-side, runs queried against that workflow's own endpoint). But the fix commit's own new comments (in both the source and test file) write out the target product's name in plain text where the rest of this codebase always uses the split-string obfuscation convention for that exact word — and this repo's `check-strings` CI gate (a banned-string scan backing the "this repo is going public" scrub rule in CLAUDE.md) is currently failing on this commit because of it. Needs a quick reword of both comments before merge.

agent/src/pr-state-reconciler.ts
```diff
@@ -888,6 +1032,50 @@ export function buildProductionDeps(opts: {
       fetchFn: opts.fetchFn,
     }),
     now: () => new Date().toISOString(),
+    listDeployingTasks: (limit: number, offset: number) =>
+      listTasksByStatus("deploying", limit, offset),
+    // Server-side scoped by workflow (fixes a live bug: a client-side-filtered
+    // `actions/runs?per_page=N` repo-wide fetch can miss the target workflow
+    // entirely when unrelated workflows are noisy — confirmed live on
+    // vitals-os, e.g. bursts of "Bump Shipwright Chart" runs pushing a real
```
dmcaulay: This line writes the target product's name in plain text (not the split-string form used elsewhere in this codebase for that exact word), and `check-banned-strings.ts`'s `BANNED_PATTERNS` bans it — confirmed this is why `lint / typecheck / test` CI is currently failing on this commit (check-strings exit 201). Same issue at `pr-state-reconciler.unit.test.ts:1891`. Reword both to avoid the literal name, e.g. "confirmed live in production".

dmcaulay (commented):
Verdict: APPROVE — Both review-round findings are now resolved: the `ghListWorkflowRuns` server-side scoping fix is correct (verified against `agent/src/pr-state-reconciler.ts:1034-1078`), and the banned-string violation from the prior fix commit is reworded away in both the source and test file (`grep -rn "vitals-os"` over both now returns nothing). Ran `NODE_ENV=test bun test agent/src/pr-state-reconciler.unit.test.ts` locally — 62 pass, 0 fail — and `bunx tsc --noEmit` clean. No new issues.

PR #1944, opened 2026-07-19
feat(mcp-server): add inbound bearer auth middleware (TSM-2.6)

dodizzle:
## Summary
- Add `createBearerAuthMiddleware` (`mcp-server/src/auth.ts`) requiring `Authorization: Bearer <token>` on every request except `/health`, using constant-time comparison (`node:crypto`'s `timingSafeEqual`).
- Refactor `mcp-server/src/index.ts` from a module-level singleton to a `createApp({ token })` factory, so tests can inject a fixed token without touching `process.env`.
- `mcp-server/src/main.ts` is now the sole reader of `SHIPWRIGHT_MCP_SERVER_TOKEN`; it throws synchronously and refuses to start if the token is unset.
- `/mcp/tools` (the discovery listing) is gated behind the same middleware — documented inline as a deliberate choice to avoid unauthenticated capability enumeration.
- Refreshed `docs/architecture.md` and `docs/configuration.md` to document the new env var and auth behavior.

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Requests to /mcp without a valid bearer token are rejected 401 with a WWW-Authenticate header | MET |
| 2 | Requests with the correct token are served normally | MET |
| 3 | Token comes from env per SHIPWRIGHT_* convention; fails closed if unset | MET |
| 4 | Token comparison is constant-time | MET |
| 5 | /health remains unauthenticated | MET |
| 6 | Tests cover valid/missing/wrong/malformed token cases | MET |

## Test Plan
- `mcp-server/src/auth.smoke.test.ts` (new): missing header, non-Bearer scheme, malformed header, wrong token, valid token.
- `mcp-server/src/index.smoke.test.ts` updated to build the app via `createApp({ token })` and pass a valid bearer token to the `/mcp/tools` test.
- `NODE_ENV=test bun test mcp-server/src` — 44 pass, 0 fail.
- `bunx biome lint mcp-server` and `tsc --noEmit` clean.


dodizzle (commented):
Verdict: APPROVE — Clean, well-scoped inbound bearer-auth implementation for mcp-server's HTTP entry point. Mirrors task-store's existing auth.ts pattern (constant-time comparison, identical 401 shape), fails closed if the token env var is unset, correctly updates every importer of the refactored createApp() factory (stdio path is unaffected), and has solid smoke-test coverage. No blocking concerns.

PR #1946, opened 2026-07-19
feat: escalate to HITL instead of looping on a re-reviewed rebuttal (RPF-1.3)

dmcaulay:
## Summary
- Adds a new `Step 5a.7: Second-Round Escalation Check (RPF-1.3)` to `patch.md`, run between the pre-work claim (Step 5a.6) and the fix-subagent dispatch (Step 5b)
- Detects a genuine human-judgment deadlock: a qualifying review whose finding has an author-reply comment (the RPF-1.1 rebuttal) dated *before* that review's `submittedAt` — i.e. we already rebutted once and the reviewer looked at it and still disagreed
- On detection: PATCHes the linked task-store task (resolved via the PR record's `taskId`) with `hitl: true`, posts a single PR comment explaining the deadlock, releases the pre-work claim, and skips this PR's patch cycle for the run — no further rebuttal, no `reviewState` reset
- First-round rebuttals (no prior author reply before the current review) are unaffected — RPF-1.1/1.2 behavior proceeds normally

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Skip rebuttal/reviewState-reset path on second-round disagreement | MET |
| 2 | PATCH task hitl:true + one PR comment + stop | MET |
| 3 | First-round rebuttals unaffected | MET |
| 4 | Content-only test coverage in patch.content.test.ts, no existing tests retired | MET |

## Test Plan
- Added a new `describe` block (9 tests) to `plugins/shipwright/commands/patch.content.test.ts` asserting the second-round detection logic, the `hitl: true` PATCH, the single PR comment, the claim release, and that first-round rebuttals still proceed to Step 5b
- `bun test plugins/shipwright/commands/patch.content.test.ts` — 52/52 pass (43 pre-existing + 9 new)
- `bun run --filter='@shipwright/plugin' typecheck` — passes
- `task check-strings`, `task check-config-docs`, `task check-version-sync` — all pass

### DIFF
--- a/plugins/shipwright/commands/patch.md
+++ b/plugins/shipwright/commands/patch.md
@@ -575,8 +575,8 @@ headRefOid=$(gh pr view {pr} --repo {org}/{repo} --json headRefOid -q '.headRefO
 - **`headRefOid == PRECLAIM_COMMIT_SHA`** (marker is current): trust it. Set
   `PR_RECORD_ID = PRECLAIM_RECORD_ID` and **skip this site's own `/prs/claim` call below**
   — the orchestrator's `/prs/claim` already holds this PR under `phase: "patch"`. Proceed
-  directly to Step 5b (`PR_RECORD_ID` is reused by the post-fix update in Step 5c.5, same
-  as the self-claim path).
+  to Step 5a.7 (`PR_RECORD_ID` is reused by the post-fix update in Step 5c.5, same as the
+  self-claim path).
 - **`headRefOid != PRECLAIM_COMMIT_SHA`** (stale marker — new commits landed between the
   orchestrator's claim and now) **or no marker present**: fall back to self-claiming
   exactly as today — run the claim below unchanged.
--- a/plugins/shipwright/commands/patch.md
+++ b/plugins/shipwright/commands/patch.md
@@ -657,11 +657,18 @@ do not reset `reviewState`.
    ```
    The temp file path MUST include the PR number to avoid collisions — `/tmp` is shared
    across all worktrees.
-4. Resolve the unresolved inline threads (from Step 3a) belonging to each qualifying
-   second-round review — escalating without resolving them would leave those threads
-   `isResolved == false`, so Step 3a's List A criteria would re-flag this same PR next
-   cycle and re-fire this same escalation indefinitely. Use the same mutation as Step 5b
-   [D]/[E]:
+4. Resolve **all** currently-unresolved inline threads on this PR (from Step 3a's
+   `reviewThreads.nodes[]`) — not just threads tied to the qualifying second-round review.
+   Step 3a's query carries no field linking a thread back to the review that raised it
+   (only `id`, `isResolved`, and the first comment's `author.login`/`body`/`path`/`line`),
+   so scoping resolution to "threads belonging to" a specific review isn't something this
+   step can actually determine. Escalating already means giving up on automated resolution
+   for this cycle — the PR comment posted in step 3 above tells the human reader that
+   everything was escalated for manual review, not silently fixed, so resolving every
+   open thread here carries no silent-dismissal risk. Leaving any thread unresolved,
+   however, would leave it `isResolved == false`, so Step 3a's List A criteria would
+   re-flag this same PR next cycle and re-fire this same escalation indefinitely — the
+   exact loop this step exists to close. Use the same mutation as Step 5b [D]/[E]:
    ```bash
    gh api graphql -f query='
    mutation {
@@ -670,11 +677,8 @@ do not reset `reviewState`.
      }
    }'
    ```
-   Run this for the Thread ID of every unresolved inline thread whose comment belongs to a
-   qualifying second-round review — the PR comment posted in step 3 above is the record of
-   why these are being resolved without a further code change. If a qualifying review has
-   no unresolved inline threads (its finding was a bare review-body comment with no inline
-   thread), there is nothing to resolve for that review — move on.
+   Run this for the Thread ID of every thread in Step 3a's `reviewThreads.nodes[]` with
+   `isResolved == false`. If there are none, there is nothing to resolve — move on.
 5. Release the pre-work claim from Step 5a.6 — no fix is in flight, this cycle intentionally
    stops short of dispatching one:
    ```bash
--- a/plugins/shipwright/commands/patch.md
+++ b/plugins/shipwright/commands/patch.md
@@ -614,7 +614,7 @@ lives, in Step 5b Instructions [D]), check whether this PR's List A finding is a
 round of the same disagreement.
 
 For each qualifying review from Step 3a (`state == "COMMENTED"` or `"CHANGES_REQUESTED"`,
-`commit.oid == headRefOid`, contributing to this PR's List A membership), check the PR-level
+contributing to this PR's List A membership), check the PR-level
 comments already fetched in Step 3a (`comments.nodes`) for an author reply: a comment whose
 `author.login == CURRENT_USER` with `createdAt` **before** that review's `submittedAt`. This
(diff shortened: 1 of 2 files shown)


dmcaulay (commented):
Verdict: COMMENT — the pre-claim fast path (the primary automated dispatch route) still routes straight to Step 5b and bypasses the new Step 5a.7 escalation check entirely; only this self-claim branch was updated. See inline comment.

plugins/shipwright/commands/patch.md
```diff
@@ -600,7 +600,79 @@ Skip the rest of Step 5 for this PR. Move to the next qualifying PR in List A. I
 candidates remain, continue to Step 6.
 
 **Otherwise** (`200` or `201`): the claim succeeded. `PR_RECORD_ID` is reused by the
-post-fix update in Step 5c.5 — no second claim call is needed. Proceed to Step 5b.
+post-fix update in Step 5c.5 — no second claim call is needed. Proceed to Step 5a.7.
```
dmcaulay, line 603: This self-claim ("Otherwise") branch now correctly routes to Step 5a.7, but the sibling **Pre-Claim Fast Path** branch a few lines up (Step 5a.6, `headRefOid == PRECLAIM_COMMIT_SHA`) still says "Proceed directly to Step 5b" — untouched by this diff. That fast path is taken whenever the loop-orchestrator has already claimed the PR before dispatch, which per the Arguments section is the standard automated invocation route (this review itself was dispatched with a `[preclaim:...]` marker). As written, any PR reached via the pre-claim fast path skips Step 5a.7 entirely, so the second-round escalation check never runs in the common case — only in this self-claim fallback (stale/missing marker). Suggest updating that bullet's "Proceed directly to Step 5b" to "Proceed to Step 5a.7" to match.

dmcaulay (commented):
Verdict: COMMENT — the pre-claim fast path fix is correctly applied (both branches now route to Step 5a.7), but the new Step 5a.7 has two logic bugs: it references a `commit.oid` field that Step 3a never fetches/filters on, and it never resolves inline threads, so the escalation will likely re-fire every patch cycle instead of settling into a terminal state. See inline comments.

plugins/shipwright/commands/patch.md
```diff
@@ -600,7 +600,79 @@ Skip the rest of Step 5 for this PR. Move to the next qualifying PR in List A. I
...
+`gh pr comment`) and `reviewState` reset to `pending` so a fresh review can re-evaluate the
+rebuttal. If that fresh review still flags the same (or an equivalent) issue, another
+rebuttal+reset cycle would repeat indefinitely — the reviewer and the fix subagent disagree,
+and that is a genuine human-judgment deadlock, not something another automated pass will
+resolve. Before dispatching the fix subagent (which is where RPF-1.1's rebuttal-comment step
+lives, in Step 5b Instructions [D]), check whether this PR's List A finding is a *second*
+round of the same disagreement.
+
+For each qualifying review from Step 3a (`state == "COMMENTED"` or `"CHANGES_REQUESTED"`,
+`commit.oid == headRefOid`, contributing to this PR's List A membership), check the PR-level
```
dmcaulay: Step 3a's GraphQL query for `reviews` only selects `author { login }, state, submittedAt, body` — it never fetches `commit { oid }` — and Step 3a's actual List A/unaddressed-findings criteria has no commit-oid filter at all (I grepped the whole file; `commit.oid` appears nowhere else in `patch.md`). This `commit.oid == headRefOid` filter matches `check-patch.ts`'s `hasUnaddressedFindings` (a separate TS candidate provider with its own GraphQL query that does fetch `commit { oid }`), but that data isn't available here. Since this file is executed literally by an LLM agent at runtime, this leaves the agent checking a field it never fetched — recommend dropping the clause or adding the field to Step 3a's query if per-commit filtering is genuinely needed.

plugins/shipwright/commands/patch.md
```diff
@@ -600,7 +600,79 @@ Skip the rest of Step 5 for this PR. Move to the next qualifying PR in List A. I
...
+`author.login == CURRENT_USER` with `createdAt` **before** that review's `submittedAt`. This
+mirrors `check-patch.ts`'s `isAddressedByAuthorReply` (an author reply *after* a review marks
+that review addressed) but checks the opposite direction — a reply dated *before* the
+current review means we already rebutted once, the reviewer looked at that rebuttal, and
+still raised a finding this round.
+
+**If any qualifying review has an author-reply comment dated before its `submittedAt`**
+(second round on the same disagreement): escalate instead of looping. Skip the rest of Step
+5 for this PR entirely — do not dispatch the fix subagent, do not post another rebuttal, and
+do not reset `reviewState`.
```
dmcaulay, line 628: This branch never resolves inline review threads (unlike Step 5b [D]'s REJECT handling) and leaves `reviewState` untouched. Per this file's own commentary at Step 5b [D] (`hasUnaddressedFindings()` short-circuits to `true` on any unresolved inline thread), if the qualifying review's finding was an inline thread (the common case), the PR will keep re-qualifying for List A on every subsequent patch cycle and this same escalation will re-fire — reposting the PR comment and re-PATCHing `hitl: true` indefinitely. Recommend resolving the affected threads as part of escalation, or otherwise recording that this PR was already escalated so a later cycle short-circuits instead of repeating.

plugins/shipwright/commands/patch.md
```diff
@@ -600,7 +600,97 @@ Skip the rest of Step 5 for this PR. Move to the next qualifying PR in List A. I
...
+   #   This finding was already rebutted once and the review still disagrees after
+   #   re-evaluating that rebuttal — this looks like a genuine disagreement between the
+   #   reviewer and the automated fix, not something another automated pass will resolve.
+   #   Flagging for a human decision instead of rebutting again.
+   gh pr comment {pr} --repo {org}/{repo} --body-file /tmp/shipwright-patch-escalation-{pr}.txt
+   rm /tmp/shipwright-patch-escalation-{pr}.txt
+   ```
+   The temp file path MUST include the PR number to avoid collisions — `/tmp` is shared
+   across all worktrees.
+4. Resolve the unresolved inline threads (from Step 3a) belonging to each qualifying
```
dmcaulay: This resolves the trivial single-review re-fire case, but on a PR with multiple reviews there's no mechanism to bind an unresolved thread to "each qualifying second-round review" specifically. Step 3a's `reviewThreads.nodes[]` (lines 136-149) carries only `id`, `isResolved`, and the first comment's `author.login`/`body`/`path`/`line` — no field linking a thread back to the review that raised it — and `check-patch.ts`'s `hasUnaddressedFindings` (the TS logic this step explicitly mirrors) doesn't correlate threads to reviews either. Step 5b's use of the same mutation is safe only because it runs inside a dispatched subagent doing the finding↔thread correlation as an explicit per-item reasoning step with full context; Step 5a.7 is a bare orchestrator-level instruction with no such mechanism. Under-resolving re-triggers the exact re-escalation bug this commit fixes; over-resolving would silently mark an unrelated review's real findings resolved without a fix or rebuttal. Since escalating already means giving up on this cycle's findings, consider explicitly scoping this to "all currently-unresolved threads on this PR" instead, with a documented rationale.

dmcaulay (commented):
Verdict: COMMENT — the commit.oid fix is solid and the trivial single-review re-fire case is now closed, but the thread-resolution fix has a remaining gap: nothing binds an unresolved thread to the specific qualifying review, so on a multi-review PR the escalation can still under- or over-resolve threads. See inline comment.

dmcaulay (commented):
Verdict: APPROVE — the multi-review thread-resolution gap from the last round is closed: Step 5a.7 now scopes resolution to all currently-unresolved threads on the PR, with the documented rationale this review previously suggested as the defensible option. No remaining blocking concerns.

PR #1956, opened 2026-07-19
fix: match phase cron schedules to shipwright-loop parent (LPS-1.1)

dmcaulay:
## Summary
- The four pipeline-phase system crons (`shipwright-dev-task`, `shipwright-patch`, `shipwright-review`, `shipwright-deploy`) are children of `shipwright-loop` (`parentCronId` set) and are structurally excluded from independent node-cron scheduling — dispatch is driven entirely by `shipwright-loop`'s own `* * * * *` tick plus readiness logic. Their staggered `schedule` strings in `admin/src/system-crons.ts` were vestigial and implied independent dispatch that doesn't actually happen.
- Changed `schedule` on the four phase entries to `"* * * * *"`, matching `shipwright-loop`'s own schedule — accurate, since these rows are genuinely evaluated on the same every-minute tick as part of the loop dispatch.
- Updated `system-crons.unit.test.ts`: replaced the 3-test "pipeline schedule staggering" describe block with a single assertion that each phase cron's schedule equals `shipwright-loop`'s schedule, and removed the now-inapplicable "polling crons keep a 30-minute cadence" test (plus its now-dead `expandMinuteField` helper).
- Refreshed `docs/agent.md`'s cron schedule table to reflect the new every-minute schedules.

No Prisma migration needed (`schedule` stays a required non-null String, `"* * * * *"` passes the existing `isValidCron` check). No admin UI change needed — `topLevelCrons` filters out any cron with a `parentCronId`, so phase rows only ever render as read-only text via `renderNestedPhasesRow`, not the editable top-level-only form.

## Acceptance Criteria
| 1 | Four phase-cron schedules changed to "* * * * *" | MET | `admin/src/system-crons.ts` |
| 2 | Stagger describe block replaced with single parent-match assertion | MET | `admin/src/system-crons.unit.test.ts` |
| 3 | "polling crons keep a 30-minute cadence" test removed | MET | dead code + test removed |
| 4 | Unit-only test decision, net -3 tests | MET | no I/O introduced |
| 5 | `admin/src/system-crons.unit.test.ts` passes | MET | 61 pass, 0 fail |

## Test Plan
- `bun test admin/src/system-crons.unit.test.ts` — 61 pass, 0 fail, 100% line/func coverage on `system-crons.ts`
- `task lint` — clean (514 files)
- `task typecheck` — all 7 workspaces pass


dodizzle (approved):
Verdict: APPROVE — Vestigial staggered schedule strings on the four loop-child phase crons replaced with the parent's schedule; verified the classifier structurally excludes parentCronId rows from independent dispatch, so this is a pure cosmetic/config cleanup with no functional impact. Tests updated to match (61 pass, 0 fail).

PR #1983, opened 2026-07-20
docs: insert pre-filing check into test-fix, fix source field, close HITL gap

dmcaulay:
## Summary
- Wire the pre-filing-verification checklist into `test-fix/SKILL.md` Step 2, run per parsed `T-NNN` row before it proceeds to repo detection / dedup / task creation. Explicitly calls out the T-130 grep-vs-`path.join()` false-negative pattern: unconfirmable indirect/constructed-path references must route to HITL rather than being treated as a confirmed pass or absence.
- Fix the task-creation JSON template's hardcoded `"source": "shipwright"` to `"source": "test-fix"`.
- Close a real HITL gap: add rule (d) — a general "is the fix approach obvious" judgment — to Step 5.2's HITL classification, modeled on `consolidation-fix`'s existing pattern, applied on top of the existing 3 hardcoded keyword rules. Includes a worked example (clear-precedent unit test vs. novel-integration-client test).

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Step 2 references `pre-filing-verification.md`, calling out indirect/constructed-path references must route to HITL | MET |
| 2 | Task-creation JSON `source` field reads `"test-fix"` | MET |
| 3 | HITL section documents new general ambiguous-fix judgment criterion on top of existing 3 rules, with worked example | MET |
| 4 | No test change needed (prose-only edit) | MET |

## Test Plan
- Prose-only `SKILL.md` edit — no executable code changed.
- Confirmed no test files reference `test-fix` (`grep -rl "test-fix" --include="*.test.ts" plugins/` — none found).
- `task lint` — pass.
- `task check-strings` — pass (fixed a banned-string hit from an internal repo-slug embedded in the T-130 example; scrubbed to reference the task ID alone).
- `task check-config-docs` — pass.

### DIFF
--- a/plugins/shipwright/skills/test-fix/SKILL.md
+++ b/plugins/shipwright/skills/test-fix/SKILL.md
@@ -79,8 +79,8 @@ Before starting, check for flags:
      produce).
    - **Route rows whose target-file or fixture existence can't be confirmed by a literal
      check to HITL, rather than assuming they're safe to drop or safe to file clean**
-     (Checklist Item 3). This is the exact failure mode that caused the `test-t-130-vitals-os`
-     false negative: a grep-based existence check missed a fixture that was referenced via
+     (Checklist Item 3). This is the exact failure mode that caused task T-130's false
+     negative: a grep-based existence check missed a fixture that was referenced via
      `path.join()` construction rather than a literal import or string match, so the row was
      treated as a confirmed pass when the fixture's real status couldn't actually be
      determined that way. When a row's target file or fixture is referenced indirectly —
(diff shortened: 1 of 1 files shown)


dmcaulay (commented):
Verdict: COMMENT — the `source` field fix (`"shipwright"` → `"test-fix"` in Step 5's task JSON) is correct in isolation, but Step 4's exact-ID dedup check still filters on `source == "shipwright"` (SKILL.md:133, also :145-146/:156), so tasks this skill creates going forward will never satisfy its own dedup filter. Worth updating those three references to `"test-fix"` before merge — everything else (Step 2's pre-filing-verification wiring, new HITL rule d) checks out cleanly against the referenced files.

dodizzle (approved):
Verdict: APPROVE — clean, well-scoped doc fix. Source-field dedup bug fully resolved (verified zero remaining `source == "shipwright"` references, addressing the prior reviewer's concern), new HITL rule (d) and pre-filing-verification wiring are logically complete and follow the same convention already used in entropy-fix and error-fix.

PR #1984, opened 2026-07-20
feat: insert pre-filing check into security-fix, fix source field, close HITL gap

dmcaulay:
## Summary
- Insert a pre-filing verification call into `security-fix`'s Step 3, mirroring the pattern already used by `entropy-fix`/`error-fix`, so findings are re-verified against current repo state before becoming tasks
- Fix the task-creation JSON template's hardcoded `"source": "shipwright"` to `"source": "security-fix"` (the Step 6q.1 dedup check's `source == "shipwright"` query, which reads historical tasks, is left untouched)
- Close a real HITL gap: add a general ambiguous-fix judgment criterion (single canonical shape + precedent -> `hitl:false`; multiple shapes/cross-boundary/no precedent -> `hitl:true`) modeled on `consolidation-fix`'s Step 7 pattern, applied on top of the existing fixed lookup table — including for rule categories not yet in the Step 2 table — with a worked example

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Step 3 of security-fix/SKILL.md references plugins/shipwright/references/pre-filing-verification.md and instructs the check to run before a finding becomes a task | MET |
| 2 | Task-creation JSON template's source field reads "security-fix" instead of "shipwright" | MET |
| 3 | HITL section documents a new general ambiguous-fix judgment criterion (multi-shape / no clear precedent -> hitl:true) that applies even for rule categories not in the existing fixed lookup table, with a worked example | MET |
| 4 | No test change needed — SKILL.md prose edit, not executable code; existing security-fix static content checks still pass unchanged | MET |

## Test Plan
- `bun test plugins/shipwright/skills/security-fix/SKILL.content.test.ts` — 38 pass, 0 fail
- `task lint`, `task check-strings`, `task check-config-docs`, `task check-version-sync` — all pass
- Independent spec-compliance subagent review confirmed all 4 acceptance criteria MET

### DIFF
--- a/plugins/shipwright/skills/security-fix/SKILL.md
+++ b/plugins/shipwright/skills/security-fix/SKILL.md
@@ -256,19 +256,18 @@ Look up the rule in the Step 2 classification table and route:
     not — a judgment call, not a mechanical fix.
   - No default lean either way; judge each group on its own facts.
 
-**General ambiguous-fix criterion (applies on top of the lookup table above, including to
-rule categories not yet in the Step 2 table):** the three bullets above cover the Step 2
-table's fixed classifications, but the Step 2 table is not guaranteed to be exhaustive
-forever — `security-scan` may add a new rule before this skill's Step 2 table is updated to
-classify it. For any finding — whether its rule has a Step 2 entry or not — apply this
-additional judgment, mirroring `consolidation-fix`'s Step 7 per-finding `hitl` heuristic:
+**General ambiguous-fix criterion (structures the `HITL: per-finding` judgment call above):**
+the `HITL: per-finding` bullet above says to "use judgment" per finding group but doesn't
+say what that judgment should weigh. For any finding whose rule is classified
+`HITL: per-finding` in the Step 2 table (`authz-missing-check` today), apply this additional
+structure, mirroring `consolidation-fix`'s Step 7 per-finding `hitl` heuristic:
 
 - `hitl: false` (autonomous) when there is a **single, obvious, canonical fix shape** with
   **existing precedent elsewhere in the codebase** — e.g. the rule's own fix guidance in the
   Step 2 table names one clear, mechanical remediation (as `osv-cve`'s "bump to the patched
-  version" or `secret-weak-compare`'s "swap in `crypto.timingSafeEqual`" already do), or an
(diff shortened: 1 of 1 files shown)


plugins/shipwright/skills/security-fix/SKILL.md
```diff
@@ -239,6 +256,43 @@ Look up the rule in the Step 2 classification table and route:
     not — a judgment call, not a mechanical fix.
   - No default lean either way; judge each group on its own facts.
 
+**General ambiguous-fix criterion (applies on top of the lookup table above, including to
```
dmcaulay: This new criterion is framed to apply "including to rule categories not yet in the Step 2 table," and the worked example below is built around a hypothetical unclassified rule (`insecure-deserialization`). But Step 3.3 (unchanged by this PR) filters findings to "only groups whose rule has `PR-worthy: true` in the Step 2 table" before any finding reaches this HITL-computation step — and the Constraints section says this skill "enforces the Step 2 classification, it does not modify it." A finding for a rule with no Step 2 entry has no `PR-worthy` value, so it can't pass Step 3.3's filter to ever reach this criterion. Unlike `consolidation-fix`'s Step 7 (which this is modeled on), that skill has no equivalent upstream table-gate, so the pattern doesn't transplant cleanly here. Either amend Step 3.3 to let unclassified-rule findings through, or drop the "not yet in the table" framing/example and scope this to ambiguity within already-classified rules.

dmcaulay (commented):
Verdict: COMMENT — The new general ambiguous-fix HITL criterion (SKILL.md:259) is framed around rules absent from the Step 2 table, but Step 3.3's unchanged filter ("only groups whose rule has PR-worthy: true in the Step 2 table") means such findings never reach that criterion in the first place — the scenario and worked example are unreachable as written.

dmcaulay (commented):
Verdict: APPROVE — Follow-up commit rescopes the general ambiguous-fix HITL criterion to rules classified `HITL: per-finding` in the Step 2 table (`authz-missing-check`), which are always `PR-worthy: true` and therefore reachable through Step 3's filter — this genuinely resolves the prior COMMENT finding (unreachable "unclassified rule" scenario). No vestige of the old framing remains. CI green at a085302a; no new issues.

PR #1988, opened 2026-07-20
feat: check-review.ts: use shared helpers, gate on PR-level hitl + task-blocked

dodizzle:
## Summary
- Adds `hitl?: boolean | null` to `check-review.ts`'s local `PrRecord` interface
- Replaces the inline `if (linkedTask?.hitl === true) continue;` with `isTaskBlockedForDispatch(linkedTask)`, so a linked task with `status:"blocked"` is now also excluded from review candidacy (not just `hitl:true`)
- Adds `isPrRecordBlockedForDispatch(record)` alongside the existing `claimedBy` check, so a PR-record with `hitl:true` is excluded even when there is no linked task at all
- Same shared-helper pattern already applied to `review.ts` in PRB-2.2

## Acceptance Criteria

| 1 | getReviewCandidates() excludes a PR whose PR-record has hitl:true, even when there is no linked task at all | MET | `isPrRecordBlockedForDispatch(record)` check added independent of `linkedTask`; new unit test covers this exact case |
| 2 | getReviewCandidates() excludes a PR whose linked task has status:"blocked", in addition to hitl:true | MET | `isTaskBlockedForDispatch(linkedTask)` replaces the old inline hitl-only check; new unit test covers the blocked-status case |
| 3 | A PR with hitl:false/no PR record and a linked task in any other status remains an eligible candidate (no regression) | MET | Existing regression test updated and passing |
| 4 | Safe to deploy standalone | MET | More-conservative exclusion only — no removal of existing checks |
| 5 | Test coverage: PR-record hitl:true exclusion + linked-task status:blocked exclusion, existing inline hitl-only test replaced | MET | `check-review.unit.test.ts` extended with both new cases |

## Test Plan
- `bun test agent/src/check-review.unit.test.ts` — 34 pass, 0 fail
- `bun test` (full repo) — 4595 pass, 0 fail
- `bun run --filter='*' --sequential typecheck` — clean across all packages
- `bunx biome lint .` — clean


dodizzle (commented):
Verdict: APPROVE — Minimal, correct wiring of pre-existing isTaskBlockedForDispatch/isPrRecordBlockedForDispatch helpers into check-review.ts's getReviewCandidates(); gate ordering is sound, no breaking API surface, and the new unit tests genuinely exercise the added branches.

PR #2004, opened 2026-07-20
feat: stream-parse claude -p output, accumulate per-model usage

dmcaulay:
## Summary
- Switch `agent/src/claude.ts`'s `_buildArgs` from `--output-format json` to `--output-format stream-json --verbose`, and rewrite `_spawn()` to consume stdout incrementally as NDJSON instead of buffering a single JSON blob.
- Accumulate per-model token usage across `assistant` stream lines, deduped by `message.id` (a turn can repeat the same id across multiple lines with identical usage — counted once). The terminal `result` event's `modelUsage`/`total_cost_usd`/`usage` remain the byte-identical authoritative source on a clean exit; a stream that ends without a `result` event now returns the accumulated partial usage (`{ result: "", modelUsage, streamIncomplete: true }`) instead of throwing everything away, and `ClaudeRunError`/`ClaudeTimeoutError` now carry partial `modelUsage` too.
- Add an optional `onProgress(modelUsage)` callback fired as each new distinct message id is observed, for downstream tasks (CSU-2.1/CSU-3.x) to wire into mid-run progress reporting.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `_buildArgs` emits `--output-format stream-json --verbose` | MET |
| 2 | Per-model usage aggregated across turns/models, deduped by message id | MET |
| 3 | Clean-exit totals byte-identical to today's `result` event | MET |
| 4 | Partial usage surfaced when stream ends without a `result` event | MET |
| 5 | `onProgress` fires with running total as turns complete | MET |
| 6 | Fake spawner reworked to emit NDJSON via ReadableStream; new fixtures; existing test updated | MET |

## Test Plan
- `bun test agent/src` — 1150 pass, 0 fail
- `bunx biome check` on changed files — clean
- `bun run --filter='*' --sequential typecheck` — all packages pass
- New fixtures under `agent/src/fixtures/stream-json/` (`clean-multi-turn.ts`, `truncated-no-result.ts`) are **hand-authored synthetic fixtures** built to the documented public Claude Code CLI `stream-json --verbose` schema — not real captures. This was an explicitly authorized stand-in for the original task's human-capture step. @Dan: please validate against a real `claude -p --output-format stream-json --verbose` transcript when you get a chance, and we'll drop the `hitl` flag on CSU-1.1 once confirmed.

### DIFF
--- a/agent/src/claude.ts
+++ b/agent/src/claude.ts
@@ -61,6 +61,17 @@ export interface ClaudeRunResult {
    * terminal `result` event. In that case `result` is empty and `modelUsage`
    * carries whatever per-model usage was accumulated from the `assistant`
    * lines (costUSD unknown → 0), rather than throwing all of it away.
+   *
+   * Caveat: this accumulated `modelUsage` (and the partial usage attached to
+   * `ClaudeTimeoutError`/`ClaudeRunError`) undercounts `outputTokens`. Each
+   * `assistant` line's `usage.output_tokens` only reflects *visible* output —
+   * extended-thinking tokens are billed as output but often come back with an
+   * empty/redacted `thinking` field, so they never show up in that per-line
+   * total. Only the terminal `result` event's `modelUsage` is authoritative;
+   * treat any partial/incomplete usage as a lower bound, not a true count.
+   * Confirmed against real `claude -p --output-format stream-json --verbose`
+   * captures during CSU-1.1 validation — see anthropics/claude-code#27361 and
+   * #64153.
    */
   streamIncomplete?: boolean;
 }
@@ -295,6 +306,11 @@ export function createRunClaude(
         accumulated[model] = entry;
       }
       entry.inputTokens += message.usage.input_tokens ?? 0;
+      // NOTE: message.usage.output_tokens excludes extended-thinking tokens
+      // (billed as output, but the thinking content is often redacted before
+      // it reaches this line) — this accumulation undercounts true output
+      // tokens whenever the `result` event isn't reached. See the
+      // `streamIncomplete` doc comment above for the full explanation.
       entry.outputTokens += message.usage.output_tokens ?? 0;
       entry.cacheReadInputTokens += message.usage.cache_read_input_tokens ?? 0;
       entry.cacheCreationInputTokens +=
--- a/agent/src/claude.ts
+++ b/agent/src/claude.ts
@@ -56,7 +56,21 @@ export interface ClaudeRunResult {
   totalCostUsd?: number;
   modelUsage?: ModelUsage;
   recoveredFromError?: boolean;
+  /**
+   * True when the stream ended cleanly (no process error) but never emitted a
+   * terminal `result` event. In that case `result` is empty and `modelUsage`
+   * carries whatever per-model usage was accumulated from the `assistant`
+   * lines (costUSD unknown → 0), rather than throwing all of it away.
+   */
+  streamIncomplete?: boolean;
 }
+
+/**
+ * Callback fired as each new assistant turn/message completes (a new distinct
+ * `message.id` is observed), receiving the running accumulated per-model total.
+ * The passed map is a fresh snapshot — safe to retain without it mutating.
+ */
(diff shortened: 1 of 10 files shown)


agent/src/claude.ts
```diff
@@ -213,51 +342,68 @@ export function createRunClaude(
...
+          `claude exited ${exitCode}: api_error_status=${result.api_error_status ?? "unknown"} ${result.result}`,
+          result.api_error_status,
+          result.result,
+          result.session_id,
+          result.modelUsage ?? modelUsage,
+        );
+      }
+      // A truncated stream that still carried some usage: surface it on the
+      // error rather than discarding it.
+      if (Object.keys(modelUsage).length > 0) {
```
dodizzle, line 383: This new branch (non-zero exit, no `is_error` result, but some `modelUsage` was accumulated before the crash) throws `ClaudeRunError` with `resultMessage` set to `result?.result ?? ""` — empty here, since no `result` event exists in this branch. The actual diagnostic text (`stderr.trim() || "stream truncated"`) only lands in the outer `Error.message`.

`agent/src/slack.ts`'s `_detailLine()` reads `err.resultMessage` for `ClaudeRunError` instances (not `.message`), and `_classifyError` falls through to a generic `"Claude returned an error."` lead since `apiErrorStatus` is `undefined` here. Net effect: a user hitting this new code path in Slack sees a generic error with an empty detail line — losing the stderr/crash diagnostic that would have been shown before this PR (previously this scenario always threw a plain `Error` whose `.message` was surfaced directly).

Suggest passing the diagnostic text as `resultMessage` too: `new ClaudeRunError(message, undefined, stderr.trim() || "stream truncated", result?.session_id, modelUsage)`.

dodizzle (commented):
Verdict: COMMENT — Solid PR, all six acceptance criteria genuinely met and well-tested, but one real regression: the new non-zero-exit + partial-usage error branch in `_spawn()` drops the actual diagnostic text from the Slack-facing error message (see inline comment).

agent/src/claude.ts
```diff
@@ -213,51 +358,69 @@ export function createRunClaude(
...
-        `claude exited ${exitCode}: ${stderr.trim() || stdout.trim()}`,
+        `claude exited ${exitCode}: ${stderr.trim() || raw.trim()}`,
       );
     }
 
-    if (!structured) {
-      throw new Error(
-        `claude returned non-JSON stdout: ${stdout.slice(0, 200)}`,
-      );
+    if (!result) {
```
dmcaulay, line 398: `streamIncomplete: true` is new here, but none of the three existing production callers check it before consuming `result`/`modelUsage`: `slack.ts:478` treats `result: ""` as an ordinary empty response (no post, no Sentry capture — just a "markers-only response" log line), and `cron-handler.ts:301-364` is worse — it records the run as `"completed"` and, for channel-routed jobs, posts the empty formatted text to Slack, making a truncated run indistinguishable from a real success in cron monitoring. `loop-orchestrator.ts` has the same unguarded pattern. Likely intentional groundwork for a later task to consume, but as shipped it's a live behavior regression for current callers in the interim — worth either wiring the check into the three callers now or explicitly tracking the follow-up so it isn't silently live in production.

dmcaulay (commented):
Verdict: COMMENT — Streaming NDJSON parser itself is solid and well-tested, but the new `streamIncomplete: true` return shape isn't checked by any existing caller (slack.ts, cron-handler.ts, loop-orchestrator.ts) — a truncated-but-clean-exit stream currently looks like a normal success everywhere it's consumed. See inline comment.

dmcaulay (commented):
Verdict: APPROVE — Both prior review findings are resolved: `streamIncomplete: true` is now checked at all three call sites (slack.ts, cron-handler.ts, loop-orchestrator.ts) and correctly converted into each file's existing failure path, and the earlier diagnostic-swallowing fix on the truncated-exit branch remains correct. New regression tests cover all four sites; 326/326 pass in the affected files. No blocking concerns remain.

PR #2017, opened 2026-07-20
fix: filter pr-state-reconciler queries to records updated in the last hour

dmcaulay:
## Summary
- Thread an `updatedSince` (now - 1h, computed once per pass via the injected `Clock`/`now()`) through `pr-state-reconciler.ts`'s production deps (`listOpenPrRecords`, `listPendingReviewRecords`, `listPrOpenTasks`, `listOrphanCandidateTasks`, `listDeployingTasks`) so a reconcile pass only re-fetches/gh-calls records updated within the last hour, instead of unconditionally re-scanning every record every 45 minutes.
- `listPostedReviewRecords` is deliberately excluded from the filter — CHU-2.4's healing pass must keep scanning stale (untouched) records regardless of age.
- Extracted a shared `computeUpdatedSinceCutoff(nowMs)` helper to avoid repeating the cutoff computation across the 5 reconcile entry points.

## Acceptance Criteria

| 1 | listOpenPrRecords, listPendingReviewRecords, and shared listTasksByStatus helper pass updatedSince via injected Clock/now(), never Date.now() directly | MET | `computeUpdatedSinceCutoff()` computes cutoff from `deps.now()`/`deps.clock.now()`; no direct `Date.now()` calls in reconcile logic |
| 2 | Records/tasks not updated within window never fetched or gh-called | MET | `updatedSince` threaded server-side into all `GET /prs` and `GET /tasks` query params (task-store already supports `gte` filtering) |
| 3 | Window boundary correct (gte semantics, not off-by-one) | MET | gte semantics confirmed server-side; tests assert exact cutoff = now - 1h |
| 4 | Unit tests added mirroring TCR-1.2 pagination-param tests, no existing tests retired | MET | New `updatedSince filtering (PSR-1.1)` test suites for both `buildProductionDeps` and `buildReviewStateProductionDeps`; pre-existing TCR-1.2 pagination tests updated to compile, not removed |

## Test Plan
- `bun test agent/src/pr-state-reconciler.unit.test.ts` — 79/79 pass
- `tsc --noEmit` — clean
- `bunx biome check` — clean
- Coverage for `pr-state-reconciler.ts` improved vs. main (78.00%→83.20% lines, 85.71%→90.54% funcs); remaining gaps are pre-existing error-handling catch blocks


agent/src/pr-state-reconciler.ts
```diff
@@ -630,13 +696,18 @@ export async function reconcileDeployingTasks(
 export async function reconcilePrState(
   deps: PrStateReconcilerDeps,
 ): Promise<void> {
+  // Computed once per reconcile pass (PSR-1.1) — see reconcilePrOpenTasks above.
+  const updatedSince = computeUpdatedSinceCutoff(
```
dodizzle: This `updatedSince` cutoff (and the matching ones in `reconcilePrOpenTasks`, `reconcileOrphanedTasks`, and `reconcileDeployingTasks`) reintroduces the failure mode this file's `listPostedReviewRecords` was deliberately exempted from a few lines up: a record this reconciler needs to heal is, by definition, one that nothing else will ever touch again after the crash that stranded it. Once such a record's `updatedAt` falls outside `now - 1h`, it drops out of every future scan permanently — before this PR, a missed tick just meant "healed one cycle later," now it can mean "never healed." With the default 45-minute tick interval against a 60-minute window, a single missed/delayed tick (agent restart, deploy) is enough to trigger this. Worth applying the same `listPostedReviewRecords` reasoning here — either exempt these four passes from the filter, widen the safety margin significantly, or add a periodic unfiltered sweep as a backstop.

dodizzle (commented):
Verdict: COMMENT — The new `updatedSince` recency filter silently reintroduces the exact crash-recovery blind spot this reconciler exists to close for four of its five passes (`reconcilePrOpenTasks`, `reconcileOrphanedTasks`, `reconcileDeployingTasks`, `reconcilePrState`): a record whose last patch happened before the crash never gets touched again, so if it ages out of the 1-hour window before the next tick catches it, it's stuck forever with no self-correction. The PR's own `listPostedReviewRecords` exemption (CHU-2.4: "must keep scanning stale (untouched) records regardless of age") shows this exact reasoning was already applied once in this diff — it just wasn't extended to the other four passes with the identical characteristic. See inline comment for detail.

dodizzle (commented):
Verdict: COMMENT — Reaffirming my prior review on this PR (db2acbe1, 2026-07-20T22:12:26Z): the `updatedSince` recency filter is correctly implemented and tested for the passes it's safe on, but `reconcilePrOpenTasks`, `reconcileOrphanedTasks`, `reconcilePrState`, and the pending-review scan in `reconcileReviewState` exist specifically to heal records whose `updatedAt` will never change again once truly stuck — the exact case `listPostedReviewRecords` is deliberately exempted from this same filter for. With the default 45-minute tick interval against this PR's 60-minute window, a single missed/delayed tick (restart, deploy) permanently drops such a record out of every future scan, with no backstop sweep anywhere. The only commit since my last review merges `main` (removing an unrelated, already-superseded `reconcileDeployingTasks` pass) and doesn't address this. Suggest exempting these passes the same way `listPostedReviewRecords` is, widening the window well past the max tick-interval gap, or adding a periodic unfiltered sweep as a backstop.

agent/src/pr-state-reconciler.ts
```diff
@@ -753,16 +815,21 @@ async function reconcilePostedReviewStateRecord(
 export async function reconcileReviewState(
   deps: PrReviewStateReconcilerDeps,
 ): Promise<void> {
+  // Computed once per reconcile pass (PSR-1.1) — never freshly inside the
+  // page loop, so a single pass uses one consistent cutoff across all pages.
+  // Deliberately only threaded into the pending-scan below —
+  // listPostedReviewRecords (CHU-2.4's healing pass) must keep scanning
+  // regardless of record age; see its dep-interface doc comment above.
+  const updatedSince = computeUpdatedSinceCutoff(deps.clock.now().getTime());
```
dodizzle: Still unresolved from my prior two reviews on this PR (db2acbe1, c8a05323): this pending-scan (and `reconcilePrOpenTasks`, `reconcileOrphanedTasks`, `reconcilePrState`) filters by `updatedSince` the same way `listPostedReviewRecords` deliberately does *not* — but these four passes exist to heal records that are stuck precisely because nothing touches `updatedAt` again once they crash/hang. With a 45-minute default tick interval against this 60-minute window, one missed tick permanently drops a stuck record out of every future scan. No commit since my last review addresses this.

dodizzle (commented):
Verdict: COMMENT — Reaffirming my prior reviews on this PR (db2acbe1, 2026-07-20T22:12:26Z; c8a05323, 2026-07-20T23:28:45Z): the `updatedSince` recency filter is correctly implemented and tested for the passes it's safe on, but `reconcilePrOpenTasks`, `reconcileOrphanedTasks`, `reconcilePrState`, and the pending-review scan in `reconcileReviewState` exist specifically to heal records whose `updatedAt` will never change again once truly stuck — the exact case `listPostedReviewRecords` is deliberately exempted from this same filter for. With the default 45-minute tick interval against this PR's 60-minute window, a single missed/delayed tick (restart, deploy) permanently drops such a record out of every future scan, with no backstop sweep anywhere. The three commits since my last review (#2018's repo-scoping fix, two version bumps, and a main merge) don't touch this filtering logic and don't address the concern. Suggest exempting these passes the same way `listPostedReviewRecords` is, widening the window well past the max tick-interval gap, or adding a periodic unfiltered sweep as a backstop.

dmcaulay:
Addressed dodizzle's recurring finding (db2acbe1, c8a05323, and the latest review): applied the same `updatedSince` exemption reasoning already used for `listPostedReviewRecords` to `reconcilePrOpenTasks`, `reconcileOrphanedTasks`, `reconcilePrState`'s open-record scan, and `reconcileReviewState`'s pending-scan.

Verified the premise first: all four passes exist specifically to heal task-store records that are stuck precisely because nothing else will ever touch their `updatedAt` again after the crash/hang that stranded them — the exact case `listPostedReviewRecords` was already exempted for (CHU-2.4). Recency-filtering them meant a record that aged out of the 1-hour window before the next tick caught it would never be healed again, worse than pre-PSR-1.1 behavior.

Chose full exemption (option 1) over widening the window or adding a backstop sweep, for consistency with the `listPostedReviewRecords` precedent already in this diff and to avoid leaving a partially-mitigated risk. Concretely: removed the `updatedSince` parameter entirely from `listOpenPrRecords`, `listPrOpenTasks`, `listOrphanCandidateTasks`, and `listPendingReviewRecords`'s dep signatures and production implementations (rather than leaving it wired but unused), and removed `computeUpdatedSinceCutoff`/`ONE_HOUR_MS` as dead code since no consumer needs them anymore. All four passes now scan unconditionally on every tick, same as `listPostedReviewRecords` always has.

Updated/added unit test coverage asserting each pass calls its list dep with no `updatedSince` filter, and the full agent test suite (1142 tests) plus `bunx biome lint .` pass clean.

dmcaulay:
Flagging a design tradeoff for human review before this merges/deploys: the fix above resolves dodizzle's correctness concern by fully exempting all four passes from the `updatedSince` recency filter — but that also fully reverts PSR-1.1's original motivation (per the module's own doc comment: "a single unconditional full-table gh-backed scan every 45 minutes was exhausting the shared GitHub GraphQL rate limit on a long tail of untouched fixture/smoke-test records").

After this fix, no recency filtering remains anywhere in this file — every pass scans unconditionally on every tick, same as before PSR-1.1 existed. The reviewer offered two other options (widen the window well past the tick interval, or add a periodic unfiltered backstop sweep) that would have preserved some rate-limit mitigation while still healing permanently-stuck records; full exemption was chosen for consistency with the `listPostedReviewRecords` precedent, but it means the original GraphQL rate-limit problem this PR was filed to solve is now back, unaddressed.

If the rate-limit pressure that motivated PSR-1.1 is still a live concern, this may need a follow-up task (e.g. a slower/independent backstop-sweep interval, or a wider window) rather than closing this out as fully resolved.

dmcaulay:
Resolved per human decision (option B — widen the window): `reconcilePrOpenTasks`, `reconcileOrphanedTasks`, `reconcilePrState`'s open-record scan, and `reconcileReviewState`'s pending-scan now filter on `updatedSince` again, but with a 6h window (`SHIPWRIGHT_RECONCILER_UPDATED_SINCE_WINDOW_MS`, default `21600000`) instead of the original 1h — several multiples of the reconciler's 30-60 minute tick interval, so a missed/delayed tick (restart, deploy) still self-corrects on the next tick or two without losing the ability to heal a truly-stuck record, while still bounding worst-case `gh` API usage against ancient fixture/smoke-test records. `listPostedReviewRecords` remains permanently exempt (CHU-2.4).

Also rebased onto main (84 commits, including the `reconcileDeployingTasks` removal) and restored the `updatedSince`-filtering unit test coverage for `buildProductionDeps`/`buildReviewStateProductionDeps` that had been dropped from this branch's history along the way.

`task ci` passes clean (lint, typecheck, 80/80 coverage gate, 75/75 reconciler unit tests).

dmcaulay (commented):
Verdict: APPROVE — Widened-window fix correctly resolves the escalated correctness gap (permanently-stuck records no longer age out); independently verified server-side `updatedSince` gte semantics, ran the unit suite (75/75 pass), tsc, and biome clean, plus a blind second-opinion subagent reached the same conclusion. No findings ≥75 confidence.

PR #2021, opened 2026-07-20
fix(security): patch uuid, astro, and esbuild lockfile CVEs

dodizzle:
## Summary
- Overrides `uuid` to `11.1.1` in the root workspace (transitive dep of `node-cron@3.0.3`, which pins `8.3.2`) to resolve **GHSA-w5hq-g745-h8pq** (missing buffer bounds check in v3/v5/v6)
- Bumps `site`'s `astro` from `5.18.2` → `6.4.8` — all `5.x` releases are vulnerable to four GHSAs below, only fixed starting in the `6.x` line — and bumps `@astrojs/mdx`/`@astrojs/sitemap` to their astro-6-compatible releases:
  - **GHSA-2pvr-wf23-7pc7** — Host header SSRF in prerendered error page fetch
  - **GHSA-8hv8-536x-4wqp** — reflected XSS via unescaped slot name
  - **GHSA-j687-52p2-xcff** — XSS in `define:vars` via incomplete `</script>` tag sanitization
  - **GHSA-jrpj-wcv7-9fh9** — XSS via unescaped attribute names in spread props
  - **GHSA-xr5h-phrj-8vxv** — server island encrypted parameters vulnerable to cross-component replay
- Overrides `esbuild` to `0.28.1` in `site` (astro's own declared dependency range is still `^0.27.3`) to resolve **GHSA-g7r4-m6w7-qqqr** (dev server arbitrary file read)
- Migrates `site/src/content.config.ts` off the legacy `type: "content"` collection API (removed entirely in Astro 6) to the `glob()` loader — **the astro 6 bump silently broke this**: the `docs` collection built empty (`getStaticPaths()` returned zero entries), which would have shipped with all `/docs/*` pages missing from the static build

## Notes
- `astro@6.4.8` triggered a new, unrelated finding via `osv-scanner`: **GHSA-4g3v-8h47-v7g6** (reflected XSS via unescaped View Transition animation properties, published today, fixed only at `astro@7.1.0`). This site uses `output: "static"` and doesn't use View Transitions (`astro:transitions`) anywhere in `src/`, so it isn't exploitable here — the vulnerable code path requires server-rendered routes with attacker-controlled animation values. Left out of scope for this PR (not part of the original 7 findings, and a 6→7 jump warrants its own review); filing separately.
- Also found but out of scope: `js-yaml@4.2.0` (GHSA-52cp-r559-cp3m, fixed at 4.3.0) — not one of uuid/astro/esbuild, left for a future patrol pass.
- Playwright site tests could not be run in this sandbox (missing shared libs for headless Chromium, no sudo to install) — verified instead via `astro build` (all 17 pages generate, including all 11 `/docs/*` routes) and `brand-lint` (all pages/CSS on-brand). CI will run the Playwright suite.

## Acceptance Criteria
| 1 | uuid, astro, esbuild bumped to patched versions | MET | `bun.lock`/`site/bun.lock` show uuid@11.1.1, astro@6.4.8, esbuild@0.28.1 |
| 2 | osv-scanner reports zero findings for these three packages | MET | `osv-scanner scan source` on both lockfiles shows no uuid/astro/esbuild findings (only unrelated js-yaml + the newly-published, inapplicable astro 7.1.0 finding remain) |

## Test Plan
- [x] `bun test` (root) — 4633 pass, 0 fail
- [x] `bun run --filter='*' --sequential typecheck` — clean across all workspaces
- [x] `astro build` in `site/` — all 17 pages build, including all `/docs/*` routes
- [x] `npm run brand-lint` in `site/` — all pages/CSS on-brand
- [x] `osv-scanner` against both lockfiles — uuid/astro/esbuild findings resolved

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -44,7 +44,7 @@
     },
     "agent": {
       "name": "@shipwright/agent",
-      "version": "1.42.0",
+      "version": "1.43.0",
       "dependencies": {
         "@octokit/auth-app": "^8.2.0",
         "@sentry/bun": "^10.63.0",
@@ -98,7 +98,7 @@
     },
     "metrics": {
       "name": "@shipwright/metrics",
-      "version": "1.42.0",
+      "version": "1.43.0",
       "dependencies": {
         "@hono/zod-openapi": "^0.18.3",
         "@sentry/bun": "^10.63.0",
@@ -116,7 +116,7 @@
     },
     "plugins/shipwright": {
       "name": "@shipwright/plugin",
-      "version": "1.42.0",
+      "version": "1.43.0",
       "devDependencies": {
         "bun-types": "*",
         "typescript": "*",
@@ -140,6 +140,9 @@
         "typescript": "*",
       },
     },
+  },
+  "overrides": {
+    "uuid": "11.1.1",
   },
   "packages": {
     "@actions/core": ["@actions/core@3.0.1", "", { "dependencies": { "@actions/exec": "^3.0.0", "@actions/http-client": "^4.0.0" } }, "sha512-a6d/Nwahm9fliVGRhdhofo40HjHQasUPusmc7vBfyky+7Z+P2A1J68zyFVaNcEclc/Se+eO595oAr5nwEIoIUA=="],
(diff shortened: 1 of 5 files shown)


dodizzle (commented):
Verdict: APPROVE — Correctly scoped uuid/esbuild overrides and a verified-complete Astro 6 content-collection migration (all downstream astro:content call sites checked, all doc routes still build). One non-blocking note: `@astrojs/tailwind@6.0.2`'s peer range doesn't list astro 6 yet — resolves silently under bun today, worth a follow-up.

PR #2022, opened 2026-07-20
fix(security): remediate grype CVE findings for vendored esbuild + uuid

dodizzle:
## Summary
- Grype flagged 45 filesystem-scan matches: go1.23.12 stdlib (19 distinct Go CVEs, most severe GO-2026-4337 critical) embedded in the vendored `esbuild` binary shipped inside `site/node_modules/vite/node_modules/esbuild`, plus cross-scanner confirmation of `uuid@8.3.2` (GHSA-w5hq-g745-h8pq) and `esbuild@0.27.7`/`astro@5.18.2` findings already known from the osv-cve scan.
- Root cause for the esbuild finding: vite `6.4.3`'s own `esbuild` dependency range (`^0.25.0`) pins it to the `0.25.x` line regardless of astro/vite version, so the vendored binary never picked up a newer Go-built release. Added a `site/package.json` `overrides.esbuild` entry to force resolution to `0.28.1` for both astro's and vite's `esbuild` copies — verified with `astro build` after reinstall.
- `uuid@8.3.2` was pulled in transitively by `node-cron@3.0.3` (unused directly anywhere in the codebase). Bumped `node-cron` to `4.6.0`, which has zero dependencies and drops `uuid` entirely. `@types/node-cron` devDependency removed since 4.x ships its own types.

## Acceptance Criteria
None specified on the task (mechanical dependency-bump security fix).

## Test Plan
- [x] `bun install` (root) — confirmed `uuid` and old `node-cron@3.0.3` gone from `bun.lock`
- [x] `bun install` (site/) — confirmed `esbuild` resolves to a single `0.28.1` entry (both astro's and vite's copies) in `site/bun.lock`
- [x] `cd site && npm run build` — Astro site builds cleanly with the overridden esbuild
- [x] `agent`: `bun run typecheck` and `bun test` pass with `node-cron@4.6.0` (API used — `schedule()`/`.stop()` — is unchanged)
- [x] `task ci` (lint, check-strings, typecheck, check-config-docs, check-version-sync, test:coverage, secret-scan placeholder) passes at repo root


dodizzle (commented):
Verdict: APPROVE — Clean, well-verified mechanical dependency bump (node-cron 3→4 drops the transitive uuid CVE; esbuild pinned to 0.28.1 via override to clear vendored Go-stdlib CVEs). Only call site for node-cron (schedule/stop, no options) is unaffected by the major bump, and the esbuild override was verified with an actual astro build. No blocking concerns.

dodizzle (commented):
Verdict: APPROVE — Re-review after a no-op merge from main (which absorbed #2021's identical esbuild fix, so this PR's net diff against base is now just the node-cron 3→4 bump). Independently re-verified: node-cron's only call site (schedule/stop, no options) is unaffected by the major bump, and the bun.lock entry is internally consistent with no dangling references. No blocking concerns.

PR #2036, opened 2026-07-21
feat(investigate-cron): add --item mode and admin-API run lookup

dodizzle:
## Summary
- Rewrites `plugins/shipwright/skills/investigate-cron/SKILL.md` to resolve cron runs via the admin cron-runs API (`GET /agents/{id}/crons`, `GET /agents/{id}/crons/{loopCronId}/runs`) instead of guessing from transcript file mtimes.
- Adds a new `--item <org/repo#N|taskId>` entry mode that returns the full cross-phase dispatch history for a given PR or task, alongside the existing `<name> <time>` mode.
- Retains the old ±90min mtime-window + `[Cron job:]` string-matching logic as a documented fallback for history predating admin-API run tracking.

## Acceptance Criteria

| 1 | Step 0 parses `--item` and routes; absence falls back to `<name> <time>` | MET | `SKILL.md` Step 0 branches on `[ "$1" = "--item" ]` |
| 2 | Name+time mode resolves `loopCronId`/`phaseId` via `GET /agents/{id}/crons`, then filters `GET .../runs` by `phaseId`, picks closest `startedAt` | MET | Steps 1a/1b resolve ids, fetch+filter runs client-side, sort by distance to target |
| 3 | Item mode filters same endpoint by `itemId`, returns all 4-phase dispatches in order | MET | Step 1b item-mode filters by `itemId`, sorts by `startedAt` |
| 4 | Old fuzzy ±90min/mtime + `[Cron job:]` retained as documented fallback, not deleted | MET | "Fallback: pre-admin-API history" section retains original logic verbatim |
| 5 | Both test files extended with new assertions; existing Step 1-5 assertions unmodified | MET | 3 new `describe` blocks added identically to both test files; 64/64 tests pass |

Note: `GET /agents/{id}/crons/{cronId}/runs` only supports `limit`/`offset` server-side — `phaseId`/`itemId` filtering is implemented client-side after paginated fetch, per the real API contract (verified against `admin/src/agents-api.ts` and `docs/agent-api.md`).

## Test Plan
- `bun test skills/investigate-cron/` — 64/64 pass (content-assertion tests against SKILL.md)
- `tsc --noEmit` — clean
- `bunx biome check` — clean

### DIFF
--- a/plugins/shipwright/skills/investigate-cron/SKILL.md
+++ b/plugins/shipwright/skills/investigate-cron/SKILL.md
@@ -281,7 +281,13 @@ the agent workspace root.
 **With an admin-API run resolved (preferred path):** you have an exact
 `startedAt` for the run (Step 1c). Use a tight window (e.g. ±5 minutes) around
 that timestamp against the `.jsonl` file mtimes to find the matching transcript
-— this is a precision narrowing step now, not the primary matching mechanism:
+— this is a precision narrowing step now, not the primary matching mechanism.
+`startedAt` is ISO 8601 (e.g. `2026-07-21T01:11:46.391Z`) — convert it to epoch
+seconds first (in item mode, repeat this per entry in `ITEM_RUNS`):
+
+```bash
+RUN_STARTED_EPOCH=$(date -d "$RUN_STARTED_AT" +%s 2>/dev/null || date -j -f "%Y-%m-%dT%H:%M:%S" "${RUN_STARTED_AT%%.*}" +%s)
+```
 
 ```bash
 python3 -c "
--- a/plugins/shipwright/skills/investigate-cron/SKILL.md
+++ b/plugins/shipwright/skills/investigate-cron/SKILL.md
@@ -97,7 +97,7 @@ filtered to just that phase:
 ```bash
 # Only needed in name+time mode.
 PHASE_ID=$(echo "$CRONS_JSON" | jq -r --arg loop "$LOOP_CRON_ID" --arg name "$CRON_NAME" \
-  '.crons[] | select(.parentCronId == $loop and .name == $name) | .id' | head -1)
+  '.crons[] | select(.parentCronId == $loop and (.name | sub("^shipwright-"; "")) == $name) | .id' | head -1)
 echo "phaseId: $PHASE_ID"
 ```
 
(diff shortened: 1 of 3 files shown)


plugins/shipwright/skills/investigate-cron/SKILL.md
```diff
@@ -1,80 +1,195 @@
...
 
-Do not proceed to Step 1 until both variables are set.
+Name+time mode additionally needs the phase cron's id (`phaseId`) so runs can be
+filtered to just that phase:
 
----
+```bash
+# Only needed in name+time mode.
+PHASE_ID=$(echo "$CRONS_JSON" | jq -r --arg loop "$LOOP_CRON_ID" --arg name "$CRON_NAME" \
+  '.crons[] | select(.parentCronId == $loop and .name == $name) | .id' | head -1)
```
dodizzle: `PHASE_ID` is resolved via `select(.parentCronId == $loop and .name == $name)`, but `$name` is the bare `CRON_NAME` from the invocation (e.g. `deploy`, per Step 0's own example). The real system cron names are prefixed — `shipwright-dev-task`, `shipwright-patch`, `shipwright-review`, `shipwright-deploy` (verified in `admin/src/system-crons.ts`). This equality never matches for a real invocation, so `PHASE_ID` always resolves empty and every name+time-mode call silently falls back to the old fuzzy path instead of using the new admin-API lookup this PR adds.

Fix: strip the prefix before comparing, e.g. `select(.parentCronId == $loop and (.name | sub("^shipwright-"; "")) == $name)`.

dodizzle (commented):
Verdict: COMMENT — Two new bugs in the admin-API lookup logic mean the headline feature (exact admin-API resolution for name+time mode) will never actually engage: the phase-name lookup compares a bare name against prefixed system cron names (always empty), and the jq date parser can't handle the API's millisecond-precision timestamps. Both silently degrade to the old fuzzy fallback on every invocation. See inline comments.

plugins/shipwright/skills/investigate-cron/SKILL.md
```diff
@@ -1,80 +1,195 @@
...
 ```
-~/.claude/projects/<encoded-cwd>/
+
+**Name+time mode** — filter client-side to `phaseId === <PHASE_ID>`, then pick
+the run whose `startedAt` is closest to `TARGET_EPOCH` (computed in Step 2):
+
+```bash
+BEST_RUN=$(jq -r --arg phase "$PHASE_ID" --argjson target "$TARGET_EPOCH" '
+  map(select(.phaseId == $phase))
+  | map(. + {distance: (((.startedAt | fromdateiso8601) - $target) | if . < 0 then -. else . end)})
```
dodizzle: `.startedAt | fromdateiso8601` will fail on every real run record. jq's `fromdateiso8601` only accepts `%Y-%m-%dT%H:%M:%SZ` (no fractional seconds), but the admin API always serializes `startedAt` via `.toISOString()` (`admin/src/agents-api.ts:1621`), which includes milliseconds — the same format this SKILL.md's own Step 3 example uses (`2026-07-21T01:11:46.391Z`). Verified: `echo '"2026-07-21T01:11:46.391Z"' | jq 'fromdateiso8601'` errors with `date ... does not match format "%Y-%m-%dT%H:%M:%SZ"`.

Fix: strip fractional seconds first, e.g. `(.startedAt | sub("\\.[0-9]+Z$"; "Z") | fromdateiso8601)`.

dodizzle (commented):
Verdict: APPROVE — Both prior findings (phase-name prefix mismatch, jq millisecond-timestamp parse failure) are fixed correctly and verified independently. Tests pass, CI is green.

PR #2039, opened 2026-07-21
fix: SHA-pin actions, harden permissions/credentials, fix template injection in workflows

dodizzle:
## Summary
- SHA-pin every third-party GitHub Actions reference across all 14 `.github/workflows/*.yml` files (was tag-pinned, e.g. `@v4` — mutable and spoofable) — resolves all `unpinned-uses` findings.
- Add `persist-credentials: false` to every `actions/checkout` step that doesn't need to push (resolves `artipacked`); checkouts that genuinely need to push (chart-release.yml's gh-pages publish, deploy-site.yml's gh-pages deploy, sync-plugin-version.yml's branch push) keep the default and are annotated with a comment explaining why.
- Narrow `permissions:` blocks to least privilege at workflow- and job-level across `ci.yml`, the six `build-*.yml` files, and `release.yml` (which previously had no `permissions:` block at all) — resolves `excessive-permissions`.
- Move untrusted/expression values out of direct `${{ }}` interpolation in `run:` shell blocks into `env:` vars (High-severity `template-injection` findings only): the "Resolve base SHA" and "Tag release" steps in all six `build-*.yml` files, `release.yml`'s "Assert actor is admin" step, and `sync-plugin-version.yml`'s "Commit and push sync branch" / "Open sync PR" steps.
- Scope `release.yml`'s `actions/create-github-app-token` step down from the App's full blanket installation permissions to exactly `contents: write`, `issues: write`, `pull-requests: write` (what `@semantic-release/git` and `@semantic-release/github` actually need) — resolves the `github-app` finding.

## Not addressed in this PR (judgment-dependent, left as follow-ups)
- **dangerous-triggers**: `workflow_run` trigger on all six `build-*.yml` files. Zizmor flags this trigger type as "almost always used insecurely" in general, but these workflows already gate on `github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.head_branch == 'main'` and only build/push images + tag — changing the trigger model is a bigger architectural call than this PR's scope.
- **cache-poisoning**: `sync-plugin-version.yml`'s runtime-generated artifact publish pattern. Needs a closer look at the actual publish flow before a safe fix can be scoped.
- **template-injection (Informational)**: the 47 informational-severity findings (concentrated in `auto-bump-chart.yml` and `sync-plugin-version.yml`) were left untouched per the security patrol's own fix guidance — lower priority than the High-severity ones fixed here.

## Test Plan
- [x] All 14 modified workflow YAML files verified to parse correctly (`Bun.YAML.parse`)
- [x] Verified `actions/create-github-app-token@v1.12.0`'s actual `action.yml` input names (`permission-contents`, `permission-issues`, `permission-pull-requests`) against the upstream repo before using them
- [x] `bunx biome lint .` passes clean (no application code touched — YAML-only change)
- [x] Manually diffed every workflow file against `main` to confirm each fix matches its target finding without changing workflow behavior (triggers, job outcomes, and publish/tag/release mechanics are unchanged)
- Real validation happens once these workflows themselves run in CI on this PR

### DIFF
--- a/.github/workflows/auto-bump-chart.yml
+++ b/.github/workflows/auto-bump-chart.yml
@@ -83,6 +83,9 @@ jobs:
   debounce:
     name: debounce burst of release tags
     runs-on: ubuntu-latest
+    # This job only sleeps — it never touches the repo, so it needs none of
+    # the workflow-level contents:write / pull-requests:write permissions.
+    permissions: {}
     concurrency:
       group: chart-bump-debounce
       cancel-in-progress: true
--- a/.github/workflows/auto-bump-chart.yml
+++ b/.github/workflows/auto-bump-chart.yml
@@ -111,7 +111,7 @@ jobs:
       # Check out main (not the tag's commit) so we always bump from the
       # latest chart version on main rather than whatever state the tag points at.
       - name: Checkout main
-        uses: actions/checkout@v4
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           ref: main
--- a/.github/workflows/build-admin.yml
+++ b/.github/workflows/build-admin.yml
@@ -17,7 +17,7 @@ jobs:
     outputs:
       changed: ${{ steps.filter.outputs.changed }}
     steps:
-      - uses: actions/checkout@v4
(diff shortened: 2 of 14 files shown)


dodizzle (commented):
Verdict: APPROVE — Clean, well-scoped security hardening (SHA-pinning, persist-credentials, least-privilege permissions, template-injection env-var refactor) with no behavioral regressions found. Posted as COMMENT since this is a self-authored PR (GitHub disallows self-APPROVE via the API).

PR #2075, opened 2026-07-21
feat: serialize dashboard client-side metric fetches

dodizzle:
## Summary
- Change `fetchAll()` in `metrics/src/dashboard/app.js` to issue its 5 section requests (summary, trends, features, queue, tokens) sequentially instead of via `Promise.all`, reducing concurrent downstream DB connection pressure on task-store per dashboard load/refresh.
- Extract the ordered fetch-and-parse sequence into a pure, DOM-free `fetchSequential(entries, fetchImpl)` helper that takes an injectable fetch implementation, making it unit-testable without browser globals.
- Add `metrics/src/dashboard/app.unit.test.js` (net-new — this file previously had zero test coverage) covering strict sequential ordering, per-endpoint error fallback, and uncaught-rejection propagation.
- Exclude `metrics/src/dashboard/app.js` from the aggregate coverage gate (`scripts/check-coverage.ts`): reaching the new test-only export requires importing the whole file, which would otherwise drag ~1000 lines of untested DOM/chart-rendering code into the aggregate and fail CI.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `fetchAll(range)` issues its 5 section requests sequentially; existing per-endpoint error handling (catch-and-null for features/queue/tokens, uncaught for summary/trends) preserved exactly | MET |
| 2 | Ordered fetch-and-parse sequence extracted into a small pure helper accepting an injectable fetch implementation | MET |
| 3 | New `app.unit.test.js` validates sequential (non-concurrent) issuance and that a per-endpoint failure doesn't abort subsequent requests | MET |
| 4 | Conditional cost-efficiency fetch (guarded by `#cost-efficiency-section`) unchanged, still runs after the main sequence | MET |
| 5 | No existing tests retired — net-new coverage | MET |

## Test Plan
- `bun test` — 4748 pass / 0 fail (full repo suite)
- `bunx biome lint .` — clean
- `bun run scripts/check-coverage.ts` — aggregate gate passes (80.71% lines / 81.11% functions, matching main)
- `bun scripts/check-config-docs.ts` — clean
- `bun run scripts/sync-version.ts --check` — clean

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -151,6 +151,8 @@ The `/public/*` endpoints require **no authentication**. Access is controlled by
 
 Unit + integration + smoke layers (`bun test --filter metrics`). Integration tests for the task-store backend inject `RecordedTaskStoreClient` + `RecordedAdminMetricsClient` doubles (from `metrics/src/providers/task-store-recorded.ts`) over canned cassettes and a fixed clock — pre-recorded sample results, no network calls; smoke tests drive the Hono app via `app.request()`. The suite stays fully offline with no external service URLs configured.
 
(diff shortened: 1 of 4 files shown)


dmcaulay (approved):
Verdict: APPROVE — Clean, narrowly-scoped refactor. fetchSequential's per-entry error handling faithfully matches the original Promise.all semantics, the window-guard for the test-only export has no SSR/third-context risk, the coverage-gate exclusion is an exact single-file path, and the 6 new unit tests exercise real ordering/error-propagation behavior — this is app.js's first-ever test coverage.

PR #2104, opened 2026-07-22
feat: seed default AgentTool rows at agent creation

dodizzle:
## Summary
- Seed 4 default `AgentTool` rows (`Bash`, `WebSearch`, `WebFetch`, `Agent`, each `enabled: true`) in `createAgentRoute` right after `agentService.create()` succeeds and before provisioning.
- If provisioning subsequently fails and the agent row is rolled back, the seeded rows are cascade-deleted via the existing `AgentTool` FK — no new rollback code needed.
- Refreshed `docs/agent-api.md` and `docs/agent.md` to describe the new seeding behavior.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | POST /agents results in exactly 4 AgentTool rows (Bash, WebSearch, WebFetch, Agent), each enabled: true | MET |
| 2 | Integration test creates an agent via the route and asserts the 4 expected patterns exist | MET |
| 3 | Integration test asserts provisioning-rollback cleanly cascade-deletes seeded AgentTool rows | MET |

## Test Plan
- `admin/src/agents-api.integration.test.ts`: new test asserts a freshly created agent has exactly 4 AgentTool rows (Bash/WebSearch/WebFetch/Agent, all enabled) — verified against a real Postgres DB.
- `admin/src/agents-api.integration.test.ts`: new test wires a throwing provisioner, asserts the create request 500s, and asserts zero orphaned AgentTool rows remain after the agent-row rollback (cascade delete).
- Refactored duplicated `AdminDeps` construction in the integration test file into a shared `makeDeps()` helper.
- [x] `bun test admin` — 1602 pass, 0 fail
- [x] `bun run typecheck` (all workspaces)
- [x] `bun run lint` (Biome, repo-wide)
- [x] Coverage on changed files: `agent-tools.ts` 100%/100%, `agents-api.ts` 98.91%/98.41%

### DIFF
--- a/admin/src/agents-api.integration.test.ts
+++ b/admin/src/agents-api.integration.test.ts
@@ -331,4 +331,96 @@ describeOrSkip("admin CRUD API (integration)", () => {
     });
     expect(rows).toHaveLength(0);
   });
+
+  // ─── Seeded default AgentTool rows ─────────────────────────────────────────
+
+  it("POST /agents creates exactly 4 default AgentTool rows: Bash, WebSearch, WebFetch, Agent", async () => {
+    const res = await app.request("/agents", {
+      method: "POST",
+      body: JSON.stringify({ name: "New Agent With Tools" }),
+      headers: {
+        "Content-Type": "application/json",
+        Cookie: `admin_session=${cookie}`,
+      },
+    });
+    expect(res.status).toBe(201);
+    const body = await res.json();
+    const newAgentId: string = body.id;
+
+    // Query the AgentTool rows for this agent
+    const tools = await prisma.agentTool.findMany({
+      where: { agentId: newAgentId },
+      orderBy: { pattern: "asc" },
+    });
+
+    // Should have exactly 4 rows
+    expect(tools).toHaveLength(4);
+
+    // Extract patterns and verify all are enabled
+    const patterns = tools.map((t) => t.pattern).sort();
+    expect(patterns).toEqual(["Agent", "Bash", "WebFetch", "WebSearch"]);
+
+    // All should be enabled: true
+    for (const tool of tools) {
+      expect(tool.enabled).toBe(true);
+    }
+  });
+
+  it("POST /agents with provisioner failure rolls back seeded AgentTool rows via cascade delete", async () => {
+    // Create a new app with a throwing provisioner
+    const throwingProvisioner = {
+      provision: async () => {
+        throw new Error("provisioner failure");
+      },
+      deprovision: async () => {},
+      reconcile: async () => ({ recreated: [], orphans: [], failed: [], updated: [] }),
+    };
+
+    const depsWithFailingProvisioner: AdminDeps = {
+      agentService: new AgentService(prisma),
+      agentEnvService: new AgentEnvService(prisma, makeTokenCrypto()),
+      agentCronJobService: new AgentCronJobService(prisma),
+      agentCronRunService: new AgentCronRunService(prisma),
+      agentCronRunStatsService: new AgentCronRunStatsService(prisma),
+      agentToolService: new AgentToolService(prisma),
+      agentTokenService: new AgentTokenService(prisma),
+      agentPluginService: new AgentPluginService(prisma),
+      agentChatTokenService: new AgentChatTokenService(prisma),
+      agentWorkQueueService: new AgentWorkQueueService(prisma),
+      prisma,
+      provisioner: throwingProvisioner,
+      taskStore: new NoopTaskStoreProvisioningClient(),
+      chatService: new NoopChatServiceProvisioningClient(),
+      slack: {
+        deleteApp: async () => {},
+      },
+      decrypt: (value: string) => makeTokenCrypto().decrypt(value),
+      sessionSecret: SESSION_SECRET,
+    };
+
+    const appWithFailingProvisioner = createAdminApp(
+      depsWithFailingProvisioner,
+    );
+
+    // POST should fail because provisioner throws
+    const res = await appWithFailingProvisioner.request("/agents", {
+      method: "POST",
+      body: JSON.stringify({ name: "Doomed Agent With Tools" }),
+      headers: {
+        "Content-Type": "application/json",
+        Cookie: `admin_session=${cookie}`,
+      },
+    });
+    expect(res.status).toBe(500);
+
+    // After rollback, NO AgentTool rows should remain for any agent created during this test.
+    // Because the agent itself was deleted on rollback, we can't query by agentId.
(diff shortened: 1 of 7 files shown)


admin/src/agents-api.ts
```diff
@@ -955,6 +955,14 @@ export function createAdminApp(deps: AdminDeps): OpenAPIHono<AdminAuthEnv> {
       selfHosted: body.selfHosted ?? false,
     });
 
+    // Seed default AgentTool rows for the revocable tool set, before
+    // provisioning below — a provisioning failure rolls the agent back and
+    // cascade-deletes these rows via the AgentTool FK.
+    const defaultTools = ["Bash", "WebSearch", "WebFetch", "Agent"];
+    for (const pattern of defaultTools) {
+      await agentToolService.add(agent.id, pattern);
```
dodizzle: The 4 sequential `agentToolService.add()` calls run before and outside the try/catch that wraps `provisioner.provision()`. That try/catch is the only rollback path (`agentService.delete(agent.id)` on failure). If any `add()` call throws, the exception propagates straight to `onError` with a 500 — but the just-created Agent row (with 0-3 seeded AgentTool rows) is never deleted. This is a new orphan-risk window; the PR's tests only cover a provisioner failure, not a seed-loop failure. Suggest wrapping the seeding loop in the same try/catch (or extending it to start before the loop) so a seeding failure also rolls back the agent.

admin/src/agents-api.ts
```diff
@@ -955,6 +955,14 @@ export function createAdminApp(deps: AdminDeps): OpenAPIHono<AdminAuthEnv> {
       selfHosted: body.selfHosted ?? false,
     });
 
+    // Seed default AgentTool rows for the revocable tool set, before
+    // provisioning below — a provisioning failure rolls the agent back and
+    // cascade-deletes these rows via the AgentTool FK.
+    const defaultTools = ["Bash", "WebSearch", "WebFetch", "Agent"];
```
dodizzle: `scripts/seed-dev-agent.ts` already exports a canonical `DEFAULT_TOOLS` constant (Read, Write, Edit, Glob, Grep, Bash, WebSearch, WebFetch, Skill, Agent — a 10-item superset) used to seed AgentTool rows for the dev agent. This introduces a second, un-exported, divergent 4-item list for the same concept. Two sources of truth means a future change to one silently drifts from the other — consider extracting a single shared constant both call sites reference.

dodizzle (commented):
Verdict: COMMENT — the AgentTool seeding loop sits outside the provisioning try/catch, so a mid-loop failure leaks an orphaned Agent row with no cleanup; also duplicates the existing DEFAULT_TOOLS constant in scripts/seed-dev-agent.ts with a divergent list.

lib/agent-default-tools.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * The default set of tool patterns seeded as AgentTool rows for a newly
+ * created agent. Single source of truth — referenced by both the admin
+ * POST /agents route (admin/src/agents-api.ts) and the local dev-agent seed
+ * script (scripts/seed-dev-agent.ts) so the two paths can't silently drift
+ * apart.
+ */
+export const DEFAULT_AGENT_TOOLS = [
```
dodizzle, line 14: This exports the full 10-item dev-seed tool list (Read, Write, Edit, Glob, Grep, Bash, WebSearch, WebFetch, Skill, Agent) — but `admin/src/agents-api.ts:966` now loops over this same constant to seed AgentTool rows on `POST /agents`, where the PR body, both updated docs, and the PR's own new integration test all expect exactly 4 rows (Bash, WebSearch, WebFetch, Agent). This is a confirmed CI failure on this exact commit: `admin/src/agents-api.integration.test.ts` — "POST /agents creates exactly 4 default AgentTool rows" fails with `Expected length: 4, Received length: 10`.

Looks like the intended dedup collapsed two different constants (a 4-item admin-seed list and this 10-item dev-seed list) into one shared source that both call sites now consume incorrectly. Suggest splitting into a separate 4-item constant (e.g. `DEFAULT_ADMIN_AGENT_TOOLS`) for the admin route, keeping this 10-item list scoped to `scripts/seed-dev-agent.ts`.

dodizzle (commented):
Verdict: COMMENT — `DEFAULT_AGENT_TOOLS` in `lib/agent-default-tools.ts` is the 10-item dev-seed list, but `admin/src/agents-api.ts`'s `createAgentRoute` now loops over it to seed AgentTool rows on `POST /agents`, seeding 10 tools instead of the documented/tested 4 (Bash, WebSearch, WebFetch, Agent). This is a confirmed, reproducing CI failure on this head commit (the PR's own new test expects length 4, gets 10) — see inline comment for details.

dmcaulay (commented):
Verdict: COMMENT — the earlier feedback (try/catch scope, 10-item vs 4-item tools list) is correctly addressed and CI is green, but the rollback-verification test's relation-filter query can't independently confirm the seeded AgentTool rows were actually cascade-deleted — see inline comment.

admin/src/agents-api.integration.test.ts
```diff
@@ -331,4 +336,73 @@ describeOrSkip("admin CRUD API (integration)", () => {
...
+    });
+    expect(res.status).toBe(500);
+
+    // After rollback, NO AgentTool rows should remain for any agent created during this test.
+    // Because the agent itself was deleted on rollback, we can't query by agentId.
+    // Instead, we verify that NO orphaned AgentTool rows exist from this creation attempt
+    // by checking the total count remains at 0 for all agents created in this test scope.
+    // Since the agent row was deleted, its tools should be cascade-deleted too.
+    const allTools = await prisma.agentTool.findMany({
+      where: { agent: { name: "Doomed Agent With Tools" } },
```
dmcaulay: This relation filter (`where: { agent: { name: ... } }`) joins through the `Agent` row, which has already been deleted by the rollback. The query returns `[]` purely because there's no `Agent` row left to join against — it can't distinguish "AgentTool rows were cascade-deleted" from "AgentTool rows are orphaned but unreachable via this filter." It's effectively re-testing that `agentService.delete()` succeeded, not that the tool rows are actually gone.

Suggest querying unscoped instead, e.g. `prisma.agentTool.count()` (valid here since the table is truncated per-test and this test's doomed-agent tools are the only rows ever inserted), or capturing the seeded rows' `agentId` before the failing request and querying by that id afterward.

dodizzle (commented):
Verdict: APPROVE — All three prior review findings (try/catch scope, 4-vs-10 tools list, cascade-delete test verification) are correctly and completely addressed. No new issues.

PR #2112, opened 2026-07-22
fix: correct repo-config's merge-gating recommendation for shipwright-deploy repos

dmcaulay:
## Summary
- `repo-config/SKILL.md` blanket-recommended `required_approving_review_count >= 1`, `enforce_admins: true`, `require branches up to date`, and `require conversation resolution` for every repo the test-readiness pipeline touches — the same defaults regardless of who actually merges PRs there.
- Shipwright repos are moving toward `shipwright-deploy` (agent self-merges via `gh pr merge --admin`) across the board, so the skill now recommends one flat set of values for every shipwright repo:
  - **Required PR, required status checks** — unconditional. `--admin` doesn't bypass "must have a PR at all," and required status checks stay a real gate for any merge path outside `deploy.md` (e.g. a human merging manually via the GitHub UI).
  - **Require branches up to date** — no. `--admin` bypasses it, and it has a real, separate cost: GitHub's strict status checks go stale whenever main advances, so with multiple agents merging PRs in parallel, every merge invalidates every other open PR's "up to date" status — forcing a rebase + full CI re-run on each just to become mergeable again.
  - **Require conversation resolution** — no. Initially recommended "yes" since it's zero-cost, but dropped after finding a real gap in `check-patch.ts`'s `hasUnaddressedFindings()`: it only revisits unresolved review threads when a qualifying review exists at the current HEAD, so a later commit that lands without a fresh review (e.g. a CI-only fix) can leave stale threads permanently unresolved. Combined with `--admin` bypassing the setting anyway, the "free safety net for manual merges" framing wasn't worth carrying that gap for.
  - **Required approving reviews** — 0. GitHub rejects self-APPROVE via the API, so `review.md` Step 10 already force-downgrades an agent's review of its own PR to `COMMENT` + a `Verdict: APPROVE` marker — this can never be satisfied by the pipeline itself.
  - **Enforce on admins** — no. `deploy.md` Step 4b already merges every PR with `--admin` by design, not as an emergency exception; `enforce_admins: true` would either block every automated merge or require the merging account to be a permanent `bypass_actor` — enforcement in name only.
- Corrected `docs/test-readiness/test-system.md`'s own copy of this recommendation for shipwright's own repo, which didn't match live `main` branch protection settings.
- Added an anti-pattern entry in `repo-config/SKILL.md` warning against recommending any of the four for a `shipwright-deploy` repo.

## Test plan
- [x] `task check-strings` passes
- [x] Prose-only change to two Markdown files; no code paths affected, no existing tests reference the old wording
- [ ] Confirm test-readiness tasks generated after this lands recommend the corrected values


dodizzle (approved):
Verdict: APPROVE — Prose-only branch-protection recommendation change; verified the cited deploy.md/check-patch.ts/review.md behavioral claims against source and the two updated docs are internally consistent. No blocking concerns.

PR #2139, opened 2026-07-22
feat: wire progress push + partial-usage-on-failure into cron-handler

dodizzle:
## Summary
- Ports CSU-3.1's progress-push + partial-usage-on-failure fix from `loop-orchestrator.ts` to `cron-handler.ts`'s generic (non-loop) cron dispatch path, so mid-run progress pushes and correct token attribution on timeout apply to all cron jobs, not just shipwright-loop dispatches.
- Wires an `onProgress` callback (debounced to 5s via the injected `Clock`) to `cronRunReporter.recordProgress()`, so token totals survive an agent-process kill mid-run.
- Extends the runner catch path: when the thrown error is a `ClaudeTimeoutError`, its `partialModelUsage` is converted via `buildTokenPayload` and attached to `completeRun`'s failed-outcome payload — the "unknown model" on timeout fix, now fleet-wide.
- Adds a compatibility adapter at `index.ts`'s `cronDeps.runner` (mirrors the existing `getLoopOrchestrator` adapter) since `createRunClaude`'s runner only binds `onProgress` at construction time, not per-call.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | onProgress wired to recordProgress, debounced (skip a push if one fired <5s ago) | MET |
| 2 | Failure path includes buildTokenPayload-shaped data derived from err.partialModelUsage when available | MET |
| 3 | cron-handler.integration.test.ts created, mirroring CSU-3.1's two integration test cases (mid-dispatch progress push, failed-run partial usage attribution) — additive only | MET |

## Test Plan
- New `agent/src/cron-handler.integration.test.ts`: mid-dispatch progress push ordering (before `completeRun`), `ClaudeTimeoutError` partial-usage attribution on failure, and a debounce regression test (two <5s-apart pushes → one `recordProgress` call).
- `agent/src/cron-handler.smoke.test.ts` unchanged, still green — no regressions (75/75 tests passing across both files).
- [x] `task ci` passes (lint, check-strings, typecheck, check-config-docs, check-version-sync, coverage 81%/81%, secret-scan)


dmcaulay (approved):
Verdict: APPROVE — Clean, faithful port of CSU-3.1's progress-push + partial-usage-on-failure fix to the generic cron-dispatch path; debounce scoping, fire-and-forget error handling, and the index.ts adapter's documented limitation all verified against source. No blocking findings.

PR #2142, opened 2026-07-22
feat: configurable empty-queue backoff in loop orchestrator (SKT-2.2)

dodizzle:
## Summary
- Adds a configurable empty-queue backoff to `runLoopTick` in `agent/src/loop-orchestrator.ts`: after `SHIPWRIGHT_LOOP_EMPTY_BACKOFF_ATTEMPTS` (default 3) consecutive ticks that collect zero candidates, the loop skips all candidate collection (no GitHub calls, no task-store queries) for `SHIPWRIGHT_LOOP_EMPTY_BACKOFF_MS` (default 5 min) before resuming normal polling.
- A tick that dispatches at least one item resets the empty-tick counter and clears any active backoff immediately, so real work interrupts a backoff on the very next tick.
- Both env vars are read fresh on every tick (never cached), and the backoff deadline is computed via the injected `Clock`, never a raw `Date.now()`/`new Date()`.
- Refreshes `docs/configuration.md` (new "Loop orchestrator" env var section) and `docs/agent.md`'s `loop-orchestrator.ts` description to cover the new behavior.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | After N consecutive empty ticks (configurable), skips all candidate collection for a configurable duration | MET |
| 2 | A tick that dispatches ≥1 item resets the empty-tick counter and clears backoff immediately | MET |
| 3 | Both env vars read fresh each tick, not cached at orchestrator construction time | MET |
| 4 | Uses the injected Clock (`clock.now()`), never raw `Date.now()`/`new Date()` | MET |
| 5 | Safe to deploy standalone — no existing dispatch behavior changed when the queue isn't empty | MET |
| 6 | Unit tests in `loop-orchestrator.unit.test.ts` using the existing FixedClock/injected-deps pattern, covering all 3 specified cases, no existing tests retired | MET |

## Test Plan
- [x] 3 new unit tests: backoff triggers after N empty ticks and suppresses candidate collection within the window; a dispatch resets the counter; advancing the clock past the backoff window resumes normal collection
- [x] Full `loop-orchestrator.unit.test.ts` suite (71 tests) passes
- [x] Full repo test suite passes (4868 pass, 0 fail) with coverage gate (81.07%/81.42% aggregate, 90.5% for the changed file)
- [x] `lint`, `check-strings`, `typecheck`, `check-config-docs`, `check-version-sync` all pass


agent/src/loop-orchestrator.ts
```diff
@@ -718,6 +781,7 @@ export function createLoopOrchestrator(
           recordId,
           preClaimMarker,
         );
+        dispatchedThisTick = true;
```
dmcaulay: `dispatchedThisTick` is only set `true` after a successful `dispatch()`. Every claim-loss path (409 conflict or thrown 5xx pre-claim, both a few lines above) does `continue` without touching this flag, so the post-loop bookkeeping below treats "candidates existed but every claim lost a race to a sibling replica" identically to "zero candidates existed." Since this codebase runs multiple agent replicas racing for the same candidates, a busy contended queue (lots of real candidates, most claims lost to siblings) will still climb `consecutiveEmptyTicks` and eventually arm the backoff — skipping all candidate collection for the whole window even though the queue is non-empty. This contradicts Acceptance Criterion #1 ("consecutive ticks that **collect** zero candidates") and #5 ("no existing dispatch behavior changed when the queue isn't empty"). Suggest tracking candidate presence (e.g. whether `tasks.length + prs.length` was ever nonzero this tick) separately from dispatch success.

agent/src/loop-orchestrator.unit.test.ts
```diff
@@ -1434,6 +1434,147 @@ describe("createLoopOrchestrator", () => {
...
+        if (original === undefined) {
+          delete process.env[name];
+        } else {
+          process.env[name] = original;
+        }
+      }
+    }
+  }
+
+  test("after SHIPWRIGHT_LOOP_EMPTY_BACKOFF_ATTEMPTS consecutive empty ticks, a subsequent tick within the backoff window performs zero candidate-collection calls", async () => {
```
dmcaulay, line 1479: None of the 3 new backoff tests cover a tick where candidates are present but every claim call returns false (409) or throws (5xx) — the scenario that would surface the `dispatchedThisTick`/`consecutiveEmptyTicks` conflation flagged on `loop-orchestrator.ts:784`. The suite already has fixtures elsewhere in this file for a false-returning/throwing claim that could be composed with this feature's tests. Worth adding a test asserting the empty-tick counter/backoff does NOT trigger when candidates exist but all claims are lost.

dmcaulay (commented):
Verdict: COMMENT — the empty-tick counter conflates "no candidates" with "all claims lost to a sibling replica or failed", which under this repo's multi-replica contention can false-trigger the backoff even though the queue isn't actually empty.

dodizzle (commented):
Verdict: APPROVE — Self-review (own PR; GitHub blocks self-APPROVE via API, posted as COMMENT). The candidatesSeenThisTick fix in 049e86b4 correctly and completely resolves dmcaulay's contention false-positive concern from the prior review round — it decouples "candidates collected" from "dispatch succeeded," is properly scoped per-tick, and is backed by new regression tests for both the 409 and thrown-pre-claim contention paths. No new issues found.

dodizzle:
Patch check: verified this PR's outstanding review findings are already resolved at current HEAD (049e86b4).

- dmcaulay's contention false-positive finding (empty-tick counter conflating "no candidates" with "every claim lost to a sibling replica") is fixed — `candidatesSeenThisTick` now tracks candidate presence independently of dispatch success, exactly as suggested.
- The follow-up finding about missing test coverage for the 409/thrown-pre-claim contention path is also fixed — new regression tests cover both cases.

Both corresponding inline threads are already resolved. No code changes needed this cycle — leaving this comment so future patch/review cycles recognize the finding as addressed (the fix landed as a commit rather than a reply comment, which the automated classifier doesn't otherwise pick up on).

PR #2148, opened 2026-07-22
feat(agent): bundle Piper TTS binary + default voice in agent image

dodizzle:
## Summary
Bundles the **Piper** TTS binary + the default `en_US-hfc_female-medium` voice into the Shipwright agent image so voice-out works with **no external TTS key** — making agent voice a pure Helm toggle (`agent.voice.enabled`), symmetric with the self-hosted Whisper STT pod.

## Why
The agent already implements `synthesizeSpeech()` → `synthesizePiper()` (`agent/src/voice.ts`), which spawns a bare `piper` on PATH when voice is enabled and no ElevenLabs key is set. But the image deliberately omitted voice deps ("voice is not deployed on the harness yet"), so `piper` ENOENT'd and voice replies silently fell back to text. This completes the feature.

## Changes (`agent/Dockerfile`)
- Runtime stage: install the pinned Piper release (`2023.11.14-2`, linux x86_64) to `/opt/piper` (added to PATH) and download the default voice to `/app/agent/voices/` — the exact path `piper-voice.ts` hardcodes. Installed as root, voice dir chowned to uid 1000; placed **after** the `/app/agent` COPY so it isn't clobbered.
- Updated the stale base-stage comment (Piper is now bundled; ffmpeg/edge-tts stay omitted — Piper writes WAV natively).

## Validation
- Piper synthesizes a valid WAV from the exact binary + voice on the Debian/glibc base (matches `oven/bun:1-slim`).
- On the real bun base: build succeeds, runtime user `uid=1000`, `piper` resolves on PATH, voices present + readable. Native-amd64 end-to-end synth is exercised by `build-agent` CI (local arm64 can't cleanly emulate nested x86_64 exec).

## Notes / follow-ups
- Adds ~80 MB (mostly the medium voice) to every agent image. If that's unwanted for non-voice deployments, a build ARG could gate it later.
- amd64 only (matches the single-arch harness build); an arm64 image would need the `aarch64` asset.
- Downstream: once released, `app-vitals/vitals-os` picks up the new `agent-v*` base via its base-tag bump chain, enabling zero-key voice there via values only.

### DIFF
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -32,8 +32,8 @@ FROM oven/bun:1-slim AS base
 #   - jq           — bash JSON parsing in the time-tracking Claude Code hooks
 #   - build-essential / libssl-dev / pkg-config — native + Rust toolchain so the
 #     agent can build client code at runtime (e.g. Rust projects + OpenSSL crates).
-#     (voice deps — python/edge-tts/ffmpeg — deliberately omitted; voice is not
-#     deployed on the Shipwright harness yet.)
+#     (Piper TTS for voice-out is bundled in the runtime stage below; ffmpeg /
+#     python-edge-tts stay omitted — Piper emits WAV natively.)
 RUN apt-get update && apt-get install -y --no-install-recommends \
     git \
     curl \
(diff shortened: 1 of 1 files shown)


dmcaulay (approved):
Verdict: APPROVE — Single-file Dockerfile change, correct build-stage/user handling, and the PR's claims (single-arch build, no-ffmpeg-needed, hardcoded voice path) all check out against the actual code and CI config.

PR #2153, opened 2026-07-22
feat: validate Step 5b workflow names against live repo workflows before polling

dodizzle:
## Summary
- Step 5b now performs a one-time `gh api repos/{org}/{repo}/actions/workflows` check to confirm the resolved pipeline stage names (custom from CLAUDE.md or the literal Deploy/Canary/Promote to Prod defaults) actually correspond to real workflows in the target repo before committing to the 30-minute named-stage watch.
- On mismatch, prints a warning naming the missing stage(s) and actual available workflow names, then falls back to a SHA-only (unnamed) watch for the remainder of the 30-minute budget — mirroring Step 5c's existing no-pipeline fallback pattern rather than blindly polling for a name that will never appear.
- A name mismatch alone never sets `blocked` by itself — only real run outcomes or a genuine timeout in the SHA-only fallback path do.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | One-time `gh api .../actions/workflows` check, positioned after Stage-names resolution and before the poll | MET |
| 2 | Full match → unchanged behavior | MET |
| 3 | Mismatch → warning + SHA-only fallback for remainder of 30-min budget | MET |
| 4 | Mismatch alone never sets `blocked` | MET |
| 5 | Additive content-test coverage mirroring AEW-1.1/DPS-2.1 structure | MET |

## Test Plan
- New `describe("deploy.md — validate workflow names before Step 5b watch (DWV-1.1)", ...)` block in `deploy.content.test.ts` asserting presence, positioning, mismatch wording, SHA-only fallback wording, and the never-blocked-alone guarantee.
- `bun test plugins/shipwright/commands/deploy.content.test.ts` — 78 pass / 0 fail.
- `bunx biome lint` clean on both changed files.


plugins/shipwright/commands/deploy.md
```diff
@@ -513,6 +513,33 @@ Track three stages by workflow name (`.name`):
...
+  and the actual available workflow names:
+  ```
+  ⚠ Stage name mismatch — {missing names} not found among live workflows: {actual workflow names}
+    Falling back to SHA-only watch (unnamed) for the remainder of the 30-minute budget.
+  ```
+  Then fall back to watching all workflow runs matching `$SQUASH_SHA` by SHA only
+  (unnamed, no `.name` filter) for the remainder of the 30-minute budget — same query
+  shape as Step 5c's `head_sha`-only filter, and the same success/failure/timeout
+  terminal-condition handling, but continuing within this step's own 30-minute budget
+  and elapsed time rather than Step 5c's separate 10-minute window. A name mismatch
```
dmcaulay, line 548: The reference to "Step 5c's ... terminal-condition handling" isn't concretely actionable from here: the `### Terminal Conditions` section that immediately follows 5b (after `### ARC Desync Detection`, a sibling heading, not nested under 5b) is written entirely in named-stage terms — "Deploy stage failed (...on the Deploy run)", "Canary failed (...on the Canary run)", "Promote succeeded (...on the Promote run)", etc. None of those five bullets branches for the SHA-only unnamed fallback introduced here. An agent in fallback mode has no run `.name` to match against those checks, and nothing routes it back to 5c's generic all-success/any-failure/budget-exhausted scheme — risking exactly the full 30-minute blind poll this fix set out to avoid.

Suggest either inlining the three generic terminal conditions directly in this fallback bullet, or adding an explicit conditional at the top of `### Terminal Conditions` routing to 5c's scheme when in SHA-only fallback mode.

dmcaulay (commented):
Verdict: COMMENT — The SHA-only fallback (deploy.md:535-539) points to "Step 5c's terminal-condition handling" but the section an executing agent reads next, `### Terminal Conditions`, is entirely named-stage-based with no branch for the fallback mode — so the fallback isn't concretely actionable as written.

dodizzle (commented):
Verdict: APPROVE — Second commit concretely resolves the prior finding by adding an explicit SHA_ONLY_FALLBACK branch to Terminal Conditions; no new issues found.

dodizzle:
Reviewed this finding against the current state of the branch — it was already
addressed by the prior commit ("fix: route SHA-only fallback mode to generic terminal
conditions in Step 5b"), which added an explicit routing conditional for SHA-only
fallback mode in the Terminal Conditions section. No further code changes needed.

- dmcaulay's finding (deploy.md:539, SHA-only fallback not concretely actionable in
  Terminal Conditions): already fixed — the `### Terminal Conditions` section (deploy.md:584)
  now opens with an explicit `**If `SHA_ONLY_FALLBACK=true`**` conditional, before any of the
  named-stage bullets. That branch inlines three concrete, self-contained terminal checks
  (all-success — `conclusion == "success"` for every run seen; any-failure —
  `conclusion == "failure"` on any run; budget-exhausted-with-runs-pending), mirroring Step
  5c's generic name-agnostic scheme, complete with the matching task-store/PR-record update
  bash blocks and handoff-block print formats. It then explicitly instructs skipping the
  named-stage bullets below in this mode, and only falls through to
  `**Otherwise (named-stage mode, the default)**` when `SHA_ONLY_FALLBACK` isn't set. An
  agent in SHA-only fallback mode therefore has a concrete, self-contained branch to follow
  with no `.name` matching required — closing the gap the original finding identified.

plugins/shipwright/commands/deploy.content.test.ts
```diff
@@ -346,6 +346,84 @@ describe("deploy.md — chained in-Bash polling for Step 5b (AEW-1.1)", () => {
   });
 });
 
+describe("deploy.md — validate workflow names before Step 5b watch (DWV-1.1)", () => {
```
dmcaulay, line 349: All 5 new tests in this describe block scope through `extractStep5bSection()` (line 21-26), whose regex (`/### 5b\. Monitor Pipeline[\s\S]*?(?=\n#{2,3} |\n---)/`) stops at the first subsequent `##`/`###` heading — `### ARC Desync Detection`, well before `### Terminal Conditions` where the new `SHA_ONLY_FALLBACK` branch (the actual fix for the prior review finding) lives. Confirmed via grep: zero occurrences of `SHA_ONLY_FALLBACK`, `Pipeline monitoring passed/failed/still pending` in this test file.

Consider adding a helper scoped to `### Terminal Conditions` (through the next `---`/`##`) and asserting the `SHA_ONLY_FALLBACK` branch's three terminal outcomes, the 30-minute budget references, and the Step 9 handoff labels — otherwise the safety-critical part of this change has no regression protection.

dmcaulay (commented):
Verdict: COMMENT — The follow-up commit resolves the prior finding (SHA_ONLY_FALLBACK now has concrete, self-contained terminal conditions mirroring Step 5c), but the new DWV-1.1 tests scope through `extractStep5bSection()`, which stops before `### Terminal Conditions` — so the exact content that fixes the prior finding has zero test coverage.

dodizzle:
This finding was already rebutted once and the review still disagrees after
re-evaluating that rebuttal — this looks like a genuine disagreement between the
reviewer and the automated fix, not something another automated pass will resolve.
Flagging for a human decision instead of rebutting again.

plugins/shipwright/commands/deploy.content.test.ts
dodizzle, line 349: `extractStep5bSection` uses `/### 5b\. Monitor Pipeline[\s\S]*?(?=\n#{2,3} |\n---)/`, which stops at the next `##`/`###` heading — in the current `deploy.md` that's `### ARC Desync Detection` (line 576), which comes before `### Terminal Conditions` (line 593). So none of these 5 new tests can see or assert on the new `SHA_ONLY_FALLBACK` Terminal Conditions bullets added later in this same diff (the "All runs completed successfully" / "Any run fails" / "Budget exhausted" cases). This is the exact gap a prior review round on this PR flagged (test coverage doesn't reach the fix), and it's still unaddressed — the only commit since then is a no-op merge from main. Contradicts the PR body's claim that AC5 ("Additive content-test coverage mirroring AEW-1.1/DPS-2.1 structure") is MET. Suggest adding a Terminal-Conditions-scoped extraction helper and asserting the SHA_ONLY_FALLBACK bullets directly.

dodizzle (commented):
Verdict: COMMENT — The new SHA_ONLY_FALLBACK Terminal Conditions logic (the code that resolves the prior review round's finding) has zero test coverage, because `extractStep5bSection()` stops before reaching `### Terminal Conditions` — this contradicts the PR body's claim that AC5 is fully met.

dodizzle:
This finding was already rebutted once and the review still disagrees after
re-evaluating that rebuttal — this looks like a genuine disagreement between the
reviewer and the automated fix, not something another automated pass will resolve.
Flagging for a human decision instead of rebutting again.

plugins/shipwright/commands/deploy.md
```diff
@@ -522,6 +522,35 @@ Track three stages by workflow name (`.name`):
...
+  alone never sets `blocked` by itself — only an actual run outcome (success/failure)
+  or a genuine timeout in this SHA-only fallback does. **Set `SHA_ONLY_FALLBACK=true`**
+  for the remainder of this step — the poll loop and the Terminal Conditions section below
```
dodizzle: This overstates the change: the poll loop's bash block (lines 490-499) never references `SHA_ONLY_FALLBACK` and doesn't filter by `.name` in either mode — only the Terminal Conditions section below actually branches on this flag. Since this file is literal instructions an executing agent follows, suggest rewording to "the Terminal Conditions section below branches on this flag" to avoid implying poll-loop logic that doesn't exist.

dodizzle (commented):
Verdict: COMMENT — Prior finding (SHA_ONLY_FALLBACK Terminal Conditions had zero test coverage) is now resolved: 89/89 tests pass and the new `extractTerminalConditionsSection` coverage genuinely exercises the branch. Two new lower-severity issues: `### ARC Desync Detection` (deploy.md:576, untouched by this diff) is written entirely in named-stage terms with no defined behavior under the new SHA_ONLY_FALLBACK mode, and deploy.md:550-551 overstates that "the poll loop ... branch[es] on this flag" when only Terminal Conditions does. Neither breaks a met acceptance criterion or regresses the default named-stage path.

dodizzle:
Reviewed the ARC Desync Detection finding and did not implement changes:

- The `### ARC Desync Detection` section (deploy.md:576) is pre-existing and untouched by
  this diff — it predates the SHA_ONLY_FALLBACK validation logic this PR introduces.
- The reviewer's own comment explicitly confirms it "doesn't break a met acceptance
  criterion or regress the default named-stage path" — i.e. non-blocking.
- The section's detection logic (`status: "queued"` for 15+ minutes with no `in_progress`
  runs) is generic enough to still function under SHA_ONLY_FALLBACK — it just won't have a
  named stage label in the warning message, which is a cosmetic gap, not a functional one.
- Extending this section for SHA_ONLY_FALLBACK-specific wording is a reasonable follow-up,
  but it's scope creep for this PR (which is about validating workflow names before
  polling, not about reworking the pre-existing desync-detection messaging). Filing it here
  risks conflating an unrelated doc enhancement with this PR's actual change set.

Happy to pick this up as a small follow-up task if the team wants the ARC Desync message to
explicitly branch on SHA_ONLY_FALLBACK.

dodizzle (commented):
Verdict: APPROVE — Both outstanding findings from prior rounds (Terminal Conditions SHA_ONLY_FALLBACK branch had zero test coverage; deploy.md:550-551 overstated that the poll loop also branches on the flag) are verifiably and correctly resolved in this commit (89/89 tests pass). A fresh line-by-line check against the live file content found no new correctness, wording, or CLAUDE.md-conformance issues.

PR #2154, opened 2026-07-22
fix: isolate a thrown dispatch from aborting the whole shipwright-loop tick

dmcaulay:
## Summary
- In `agent/src/loop-orchestrator.ts`, when `runner()` throws inside `dispatch()`, the error was reported via `cronRunReporter.completeRun` and then rethrown — but nothing caught that rethrow inside the drain-until-dry `while` loop, so it propagated out of the whole tick and aborted every remaining candidate in that tick's queue (unrelated dev-task/review/patch/deploy items included).
- Wraps the `await dispatch(...)` call in the drain loop in a try/catch, following the same throw-isolation pattern already used for pre-claim failures (CBD-2.1): log a warning, then `continue` the drain instead of letting the tick die. `dispatch()` itself is unchanged — it still reports the failed run and rethrows, so the isolation lives entirely at the call site.
- The already-pre-claimed item stays claimed in the task store, so it naturally drops out of the next iteration's candidate collection rather than spin-looping on the same failure within the tick.

## Test plan
- [x] Updated the 5 existing unit/integration tests that asserted the old "runner throw rejects the whole tick" behavior to assert the tick now resolves cleanly while still recording a `failed` run
- [x] Added a new regression test: a dev-task dispatch throw does not block a younger review candidate from dispatching in the same tick
- [x] `bun test agent/src/loop-orchestrator.unit.test.ts agent/src/loop-orchestrator.integration.test.ts` — 78 pass
- [x] `bun test agent/` — full agent suite passes (1 pre-existing unrelated flake in `extraAllowedTools`, reproduces identically on clean `main`)
- [x] `tsc --noEmit` (agent workspace) — clean
- [x] `task lint` — clean


dmcaulay (commented):
Verdict: APPROVE — Correctly-scoped throw-isolation fix that mirrors the existing CBD-2.1 pre-claim pattern in the same function, preserves failure observability via the existing completeRun("failed") row, and is backed by tests that verify the failure-reporting behavior rather than just the absence of a rejection. No blocking findings.

agent/src/loop-orchestrator.ts
```diff
@@ -33,11 +33,10 @@
  *      /shipwright:patch | /shipwright:deploy) via the injected claude
  *      runner, and report the run.
  *   5. Repeat immediately (not waiting for the next cron tick) while work
- *      remains. Stop when nothing is selected. A thrown dispatch for one item
- *      (e.g. the injected runner throwing or timing out) is caught and
- *      isolated per-item (CBD-2.3) — logged via console.warn and skipped —
- *      so the drain continues on to the next candidate instead of aborting;
- *      it no longer stops the tick the way it used to.
+ *      remains. Stop when nothing is selected. A dispatch that throws (e.g.
```
dmcaulay, line 36: This whole PR's functional fix — the try/catch around `dispatch()` isolating a thrown dispatch from aborting the tick — already landed on `main` via #2157 (commit `0e40c083`, tracked there as CBD-2.3), which merged before `main` was merged into this branch (producing this commit). Diffing this branch against current `main` now shows only comment rewording, a log-message tweak (`item.type` added), and a few redundant `consumed.add()` calls in test mocks that duplicate what `claimTask: consumingClaimTask(consumed)` already does. Worth confirming this PR should be closed as superseded rather than merged as a no-op.

dmcaulay (commented):
Verdict: COMMENT — This PR's core fix (try/catch isolating a thrown dispatch in the drain loop) already landed on main via PR #2157 (commit 0e40c083, referenced there as CBD-2.3), merged before main was merged into this branch. Diffing against current main now shows only cosmetic changes (comment rewording, one log-message tweak, redundant test bookkeeping) — verified clean by a follow-up deep review. Confirm whether this PR should be closed as superseded, or trimmed to just the still-useful cosmetic delta before merging.

dmcaulay:
Reviewed and decided: trim+merge, not close.

Diffed this branch against current main directly — confirmed the core throw-isolation fix
did land via #2157/CBD-2.3, and everything else here is either already-redundant test
description/comment cleanup, or one still-useful improvement not on main: the dispatch
failure log message gained `item.type` (`Dispatch failed for ${itemId}` -> `Dispatch failed
for ${item.type} ${itemId}`), which main still lacks.

Given how small the remaining diff is, keeping it as one merge rather than cherry-picking
just the log line. Rebased onto main (resolved a conflict in
loop-orchestrator.unit.test.ts from two DBV-1.1 tests added on main since this branch
diverged — kept both main's new tests and this PR's updated test description) and
`bun.lock` (pure version drift, regenerated via `bun install`). All 87
loop-orchestrator tests pass post-merge.

dodizzle (approved):
Verdict: APPROVE — Core throw-isolation fix already landed on main via #2157/CBD-2.3 (confirmed independently, consistent with the prior review/patch round's trim-and-merge decision); remaining diff is a correct item.type log-message improvement plus harmless comment/test cleanup and inert bun.lock version drift. All 87 loop-orchestrator tests pass, CI green.

PR #2157, opened 2026-07-22
fix: isolate a thrown dispatch (runner failure) per-item instead of aborting the whole drain tick

dmcaulay:
## Summary
- Wraps the drain loop's `await dispatch(...)` call in `agent/src/loop-orchestrator.ts` in a try/catch that logs via `console.warn` and continues to the next candidate, instead of letting a thrown runner (e.g. a timed-out dev-task) abort the entire tick's drain.
- Updates the two existing tests asserting the old abort-on-throw behavior, plus 4 more tests (found during verification) that exercised the same dispatch-throw code path without modeling real claim semantics — under the new catch+continue, those would have spun forever re-selecting the same always-failing item. Added a shared `consumingClaimTask` test helper.
- Adds a new test proving a runner throw for one item doesn't block a younger candidate of a different type/phase from dispatching in the same tick.
- Corrects the two stale doc comments describing the old abort behavior.
- Refreshes `docs/agent.md`'s `loop-orchestrator.ts` reference entry to describe the new per-item isolation.

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Thrown dispatch() caught, logged via console.warn, loop continues to next candidate | MET |
| 2 | dispatch()'s existing failed-run reporting unchanged | MET |
| 3 | Two existing tests updated to assert isolation instead of abort | MET |
| 4 | New test proves isolation + younger candidate still dispatches same tick | MET |
| 5 | Stale doc comments corrected | MET |
| 6 | task ci passes | MET |

## Test Plan
- `bun test agent/src/loop-orchestrator.unit.test.ts agent/src/loop-orchestrator.integration.test.ts` — 78 pass, 0 fail
- `bunx biome lint` on all changed files — clean
- `bun run --filter='@shipwright/agent' typecheck` — clean
- Full `bun test agent/` — 1227/1228 pass (1 pre-existing flake unrelated to this change, confirmed reproducing on clean `main`)


dodizzle (approved):
Verdict: APPROVE — Clean per-item dispatch-throw isolation mirroring the CBD-2.1 pre-claim pattern; all six acceptance-criteria claims verified against the diff and dispatch()'s existing failed-run reporting is untouched.

PR #2158, opened 2026-07-23
feat: pilot Monitor tool for deploy Step 5a poll (MTP-1.1)

dodizzle:
## Summary
- Replace deploy.md Step 5a's chained in-Bash sleep loop with a single native Monitor tool call, as a narrow pilot before considering the same swap for Step 5b's 30-minute pipeline watch
- Update `deploy.content.test.ts`'s Step 5a describe block to assert on the Monitor-based implementation, while keeping the existing ScheduleWakeup-avoidance assertion (Monitor is the replacement, not a return to ScheduleWakeup)
- Step 5b and Step 5c sections/tests are untouched

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Step 5a's implementation section instructs using the Monitor tool (not a chained in-Bash sleep loop) | MET |
| 2 | 5-minute budget and 30-second poll interval wording preserved in Step 5a | MET |
| 3 | Break-early-on-Deploy-workflow-appearing and fallback-to-Step-5c behavior preserved | MET |
| 4 | deploy.content.test.ts's Step 5a describe block updated to assert on the Monitor-based implementation and still asserts no ScheduleWakeup usage | MET |
| 5 | Step 5b and Step 5c sections and tests are unchanged | MET |
| 6 | task ci passes | MET |

## Test Plan
- `bun test plugins/shipwright/commands/deploy.content.test.ts` — 72/72 passing
- `bunx biome lint .` — clean
- `bun scripts/check-config-docs.ts` — clean
- `bun run scripts/sync-version.ts --check` — clean
- `bun run --filter='*' --sequential typecheck` — clean across all packages

### DIFF
--- a/plugins/shipwright/commands/deploy.md
+++ b/plugins/shipwright/commands/deploy.md
@@ -353,26 +353,29 @@ read or tool call is needed. Look for a `## Deploy model` section:
 
 Before starting the full 30-minute pipeline watch, poll for a Deploy workflow run matching `SQUASH_SHA` for up to **5 minutes** (poll every 30 seconds, up to 10 polls).
 
-**Implementation: inline in-Bash sleep loop.** Do not wait between polls via a
-scheduled wakeup mechanism that resumes the session later (each resumption is a
-brand-new process invocation — costly and unnecessary here). Instead, run the
-30-second-interval checks as a shell-level loop inside a single Bash tool call. The
-full 5-minute/10-poll budget fits comfortably within one Bash call, well under Bash's
-practical ~10-minute ceiling, so — unlike Step 5b's 30-minute watch — no chaining
-across multiple Bash calls is needed here:
+**Implementation: Monitor tool.** A single Monitor call replaces the chained Bash
+sleep loop: the poll script below runs as a backgrounded shell command, and Monitor
+surfaces one event per stdout line. The script stays silent and emits nothing until a
+Deploy workflow run matching `SQUASH_SHA` appears — only then does it `echo` the match
+and `break`, ending the watch. Silence for the full budget is a valid terminal state
+here (it means "no Deploy workflow", not a crash): Monitor's own `timeout_ms` (default
+300000ms = 5 minutes, matching this budget exactly, so no override is needed) kills the
+process if nothing was ever emitted — that's the expected shape of the "no pipeline"
+outcome, not a failure.
 
-```bash
-for i in $(seq 1 10); do
-  REPO="{org}/{repo}"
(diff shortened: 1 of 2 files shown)


plugins/shipwright/commands/deploy.md
```diff
@@ -353,26 +353,29 @@ read or tool call is needed. Look for a `## Deploy model` section:
...
+and `break`, ending the watch. Silence for the full budget is a valid terminal state
+here (it means "no Deploy workflow", not a crash): Monitor's own `timeout_ms` (default
+300000ms = 5 minutes, matching this budget exactly, so no override is needed) kills the
+process if nothing was ever emitted — that's the expected shape of the "no pipeline"
+outcome, not a failure.
+
+```
+Monitor({
+  description: "Poll for Deploy workflow run matching SQUASH_SHA",
+  command: "REPO=\"{org}/{repo}\"; while true; do RESULT=$(gh api \"repos/$REPO/actions/runs?per_page=50\" --jq \"[.workflow_runs[] | select(.head_sha == \\\"$SQUASH_SHA\\\" and .name == \\\"Deploy\\\") | {id, name, status, conclusion}]\"); if [ \"$RESULT\" != \"[]\" ]; then echo \"$RESULT\"; break; fi; sleep 30; done",
```
dodizzle: A transient `gh api` failure (network blip, GitHub 5xx, secondary rate limit, auth hiccup) exits non-zero with empty stdout, so `RESULT=$(...)` captures `""`, not `"[]"`. `[ "$RESULT" != "[]" ]` is then true, so the script echoes the empty string and `break`s — exactly matching the doc's own "Deploy workflow appears" branch, sending a live deploy straight to Step 5b based on an API failure, not a real match. No `set -e`/`pipefail`/exit-code check guards this anywhere in the script, and it's new logic — the old chained-Bash-loop version had no conditional break at all.

Suggested fix: `RESULT=$(gh api ... --jq ... 2>/dev/null) || RESULT="[]"` so a fetch failure falls through to the next poll instead of matching.

dodizzle (commented):
Verdict: COMMENT — The Monitor-tool rewrite and matching content tests are solid, but the new poll script has a silent-failure bug: a transient `gh api` error is indistinguishable from a genuine Deploy-workflow match and would short-circuit straight into Step 5b during a live deploy.

dodizzle (commented):
Verdict: APPROVE — Follow-up fix correctly resolves the previously-flagged silent-failure bug in the Step 5a Monitor poll script (gh api failures now route to "keep polling" instead of "match found"). All 6 acceptance criteria met, CI green.

PR #2166, opened 2026-07-23
feat: log busy-skip and backoff-skip early returns in runLoopTick (LPF-7.1)

dodizzle:
## Summary
- Busy-flag early return in `runLoopTick` now emits `console.warn` with a distinguishable "busy-skip" message including elapsed time since `busy` last flipped true (tracked via a new `busySince: Date | null` closure variable, driven by the injected clock).
- `busySince` is set on every busy-true transition and cleared everywhere `busy` resets to false (the backoff-skip early return and the tick's `finally` block), so it never leaks stale state into a later tick.
- Backoff-skip's existing cheap `console.log` (message + `backoffUntil` timestamp) is left untouched — no added computation, per the "as cheap as possible" invariant for in-window ticks.
- The "genuinely nothing to do" drain path is unchanged — still relies solely on `workQueueReporter`'s per-iteration snapshot.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | busy-skip: `console.warn`, distinct message, elapsed time from `busySince` | MET |
| 2 | backoff-skip: `console.log`, distinct message, includes `backoffUntil` | MET |
| 3 | `busySince` set via clock on busy=true, cleared everywhere busy resets to false | MET |
| 4 | genuinely-empty path unchanged | MET |
| 5 | Tests: extend busy-flag test w/ `withCapturedWarnings`; extend backoff test w/ `backoffUntil` assertion | MET |

## Test Plan
- Extended `agent/src/loop-orchestrator.unit.test.ts`'s existing busy-flag no-op test to assert the `console.warn` busy-skip message includes elapsed busy time (via an advanceable injected clock).
- Extended the existing backoff-active-skip test to assert the message includes the exact `backoffUntil` ISO timestamp.
- `bun test agent/src/loop-orchestrator.unit.test.ts` — 78 pass, 0 fail.
- `bun run lint`, `bun run typecheck` — clean.
- Coverage on `loop-orchestrator.ts`: 96% lines / 91.22% functions.

### DIFF
--- a/agent/src/loop-orchestrator.ts
+++ b/agent/src/loop-orchestrator.ts
@@ -338,6 +338,12 @@ export function createLoopOrchestrator(
 
   // Persisted across ticks: guards against a second concurrent drain.
   let busy = false;
+  // LPF-7.1: set (via the injected clock) whenever `busy` flips to true and
+  // cleared everywhere `busy` resets to false (the backoff-skip early
+  // return and the tick's `finally` block) — lets the busy-skip warn below
+  // report how long the in-flight tick has been draining, with no stale
+  // leakage into a later tick.
+  let busySince: Date | null = null;
 
   // Spin detection state: tracks the last dispatched itemId and consecutive
   // repeat count to warn when the same item is dispatched repeatedly.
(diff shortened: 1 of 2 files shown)


dodizzle (commented):
Verdict: APPROVE — Clean, narrowly-scoped logging change (busySince tracking + busy-skip warn with elapsed time). All 5 acceptance criteria verified, busy/busySince set-clear sites exhaustively checked with no leak path, tests exercise the actual new content via the injected clock, biome lint clean, and 78/78 local tests pass. The red CI run is unrelated pr-state-reconciler flakiness (Postgres unique-constraint errors in a file this diff doesn't touch), not a regression from this change.

PR #2171, opened 2026-07-23
feat: wire real per-call onProgress into index.ts's cron and loop-orchestrator adapters

dodizzle:
## Summary
- Forward the per-call `onProgress` argument into the shared `runner`'s new third param at both `cronDeps.runner` and `getLoopOrchestrator`'s runner adapter, instead of silently dropping it
- Delete the two stale doc comments that explained why `onProgress` couldn't be routed per-call — now actively wrong since CSU-4.1 added the slot
- This makes `cronRunReporter.recordProgress` (CSU-2.1) and the CSU-3.1/CSU-3.2 `onProgress`-consuming closures in `loop-orchestrator.ts`/`cron-handler.ts` actually receive live turn-by-turn progress in production, instead of only ever seeing the timeout/crash partial-usage fallback

## Acceptance Criteria
| 1 | cronDeps.runner forwards onProgress into shared runner's third param | MET | `runner: (message, onProgress) => runner(message, undefined, onProgress)` |
| 2 | getLoopOrchestrator's runner adapter forwards onProgress the same way | MET | Identical pattern applied |
| 3 | Two stale doc comments deleted | MET | Both comment blocks removed |
| 4 | No test added (documented rationale) | MET | index.ts is the untested composition-root/bootstrap file; correctness already covered by CSU-4.1's claude.ts tests plus existing cron-handler.ts/loop-orchestrator.ts tests |

## Test Plan
- [x] `tsc --noEmit` passes across all workspaces
- [x] `bunx biome lint .` passes
- [x] Full test suite passes (1 pre-existing unrelated failure in `claude.unit.test.ts`'s `extraAllowedTools` test, confirmed present on `main` before this change)

### DIFF
--- a/agent/src/index.ts
+++ b/agent/src/index.ts
@@ -159,23 +159,7 @@ const workQueueReporter =
 
 const cronDeps: CronHandlerDeps = {
   slack,
-  // `runner` (createRunClaude's return value) takes `sessionKey` as its
-  // second positional param, not the per-call `onProgress` (CSU-3.2) that
-  // CronHandlerDeps' `runner` type now accepts — an incompatible use of the
-  // same arg slot. createRunClaude only supports binding a single onProgress
-  // callback at construction time (shared across every call this one runner
-  // instance ever makes, for cron/Slack/chat too), not a distinct callback
-  // per invocation, so it cannot route a per-dispatch callback to the right
-  // runId here. Adapting by dropping onProgress means this particular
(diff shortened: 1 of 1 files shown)


dodizzle (commented):
Verdict: APPROVE — Correctly wires per-call onProgress into createRunClaude's third param at both adapters; verified against the actual runner signature and both consumer call sites, no breaking changes, no test gap.

PR #2176, opened 2026-07-23
feat(hitl): wire review/patch candidates into HITL loop

dmcaulay:
## Summary
- Extends `scripts/hitl.ts` to collect review and patch PR candidates alongside task candidates on each poll iteration
- Merges all candidates via `selectNextWorkItem` (strict age-based FIFO across tasks, reviews, and patches)
- Pre-claims winners before dispatch: tasks via `client.claim(id)`, PRs via `client.claimPr(...)` with `[preclaim:sha]` marker appended to the command
- Routes dispatch correctly: tasks to `/shipwright:dev-task` or `/shipwright:hitl`; review PRs to `/shipwright:review`; patch PRs to `/shipwright:patch`

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | HITL poll loop collects task, review, and patch candidates | MET |
| 2 | Candidates merged via selectNextWorkItem — strict age-based FIFO | MET |
| 3 | Task pre-claimed via client.claim(id); PR via client.claimPr with [preclaim:...] | MET |
| 4 | Dispatch routes correctly per type/phase | MET |
| 5 | Warning when workspace/repos/ empty, review/patch return [] | MET |
| 6 | 409 on pre-claim skips item and continues polling | MET |
| 7 | No cron-run/work-queue reporting, spin detection, or empty-queue backoff | MET |
| 8 | No new test files | MET |

## Test Plan
- [x] Existing hitl unit tests pass (14/14)
- [x] Lint clean (biome lint)
- [x] Format clean (biome format)
- [x] No new type errors introduced

### DIFF
--- a/scripts/hitl.ts
+++ b/scripts/hitl.ts
@@ -501,7 +501,7 @@ async function runLoop(): Promise<void> {
         continue;
       }
 
-      const preclaimMarker = `[preclaim:${claimResult.commitSha}]`;
+      const preclaimMarker = `[preclaim:${claimResult.id}:${claimResult.commitSha}]`;
       command =
         next.pr.phase === "review"
           ? `/shipwright:review ${next.pr.id} ${preclaimMarker}`
--- a/scripts/hitl.ts
+++ b/scripts/hitl.ts
@@ -471,7 +471,15 @@ async function runLoop(): Promise<void> {
     let label: string;
 
     if (next.type === "task") {
-      const claimed = await client.claim(next.task.id);
+      let claimed: boolean;
+      try {
+        claimed = await client.claim(next.task.id);
+      } catch (err) {
+        log(
+          `task ${next.task.id} claim failed: ${err instanceof Error ? err.message : err}`,
+        );
+        continue;
+      }
       if (!claimed) {
         log(`task ${next.task.id} already claimed (409) — skipping`);
         continue;
@@ -490,12 +498,20 @@ async function runLoop(): Promise<void> {
         continue;
       }
 
-      const claimResult = await client.claimPr({
-        repo: parsed.repo,
-        prNumber: parsed.prNumber,
-        commitSha: next.pr.commitSha,
-        phase: next.pr.phase,
-      });
+      let claimResult: Awaited<ReturnType<typeof client.claimPr>>;
+      try {
+        claimResult = await client.claimPr({
+          repo: parsed.repo,
+          prNumber: parsed.prNumber,
+          commitSha: next.pr.commitSha,
+          phase: next.pr.phase,
+        });
+      } catch (err) {
+        log(
+          `PR ${next.pr.id} claim failed: ${err instanceof Error ? err.message : err}`,
+        );
+        continue;
+      }
       if (!claimResult) {
         log(`PR ${next.pr.id} already claimed (409) — skipping`);
         continue;
--- a/scripts/hitl.ts
+++ b/scripts/hitl.ts
@@ -22,9 +22,39 @@
  *   SHIPWRIGHT_HITL_POLL_INTERVAL — seconds between empty-queue polls (default: 15)
  */
 
-import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
+import {
+  existsSync,
+  mkdirSync,
+  readdirSync,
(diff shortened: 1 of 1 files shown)


scripts/hitl.ts
```diff
@@ -367,22 +398,121 @@ async function runLoop(): Promise<void> {
...
+        prNumber: parsed.prNumber,
+        commitSha: next.pr.commitSha,
+        phase: next.pr.phase,
+      });
+      if (!claimResult) {
+        log(`PR ${next.pr.id} already claimed (409) — skipping`);
+        continue;
+      }
+
+      const preclaimMarker = `[preclaim:${claimResult.commitSha}]`;
```
dodizzle: This marker is missing the record id — it should be `[preclaim:{recordId}:{commitSha}]` to match `agent/src/loop-orchestrator.ts`'s `formatPreClaimMarker` (the format `/shipwright:review`/`/shipwright:patch` actually parse). `claimResult` already has `.id` available from `client.claimPr(...)`. As written, the downstream command can't extract a valid record id or commit sha from this marker, so its fast-path validation fails and it falls back to self-claiming — which races against the claim this loop already holds server-side for the PR, defeating the pre-claim optimization AC #3 claims as met.

Suggested fix:
```ts
const preclaimMarker = `[preclaim:${claimResult.id}:${claimResult.commitSha}]`;
```

dodizzle (commented):
Verdict: COMMENT — The pre-claim marker built at scripts/hitl.ts:504 (`[preclaim:${claimResult.commitSha}]`) drops the record id, so it doesn't match the two-field `[preclaim:{recordId}:{commitSha}]` contract that `/shipwright:review` and `/shipwright:patch` parse (see `agent/src/loop-orchestrator.ts`'s `formatPreClaimMarker`). That silently defeats the pre-claim fast path this PR's AC #3 claims as met — worth a one-line fix (`claimResult.id` + `claimResult.commitSha`) before merge.

scripts/hitl.ts
```diff
@@ -367,22 +398,121 @@ async function runLoop(): Promise<void> {
...
       continue;
     }
 
-    const task = tasks[0];
-    const command = buildTaskCommand(task);
+    let command: string;
+    let label: string;
+
+    if (next.type === "task") {
+      const claimed = await client.claim(next.task.id);
```
dodizzle: `client.claim()` throws on any non-200/409 response (see `check-helpers.ts:369-381`). Unlike `fetchReadyTasks()` above and the `getReviewCandidates`/`getPatchCandidates` call just above this, this isn't wrapped in try/catch — a transient task-store error here propagates out of `runLoop()` and hits the top-level `catch` in `import.meta.main`, which kills the task-store/admin services this script just booted and exits. Worth catching and `continue`-ing, same as the sibling I/O calls.

scripts/hitl.ts
```diff
@@ -367,22 +398,121 @@ async function runLoop(): Promise<void> {
...
+      if (!parsed) {
+        log(`malformed PR candidate id: ${next.pr.id} — skipping`);
+        continue;
+      }
+      if (!next.pr.phase) {
+        log(`PR candidate ${next.pr.id} missing phase — skipping`);
+        continue;
+      }
+
+      const claimResult = await client.claimPr({
```
dodizzle: Same issue as the task-claim call above: `client.claimPr()` throws on any non-200/409 response and isn't wrapped in try/catch, so a transient claim failure crashes the whole loop instead of skipping this PR candidate.

dodizzle (commented):
Verdict: COMMENT — Preclaim marker fix from the last round landed correctly, but the new client.claim()/claimPr() calls aren't wrapped in try/catch like their sibling I/O calls in the same loop iteration, so a transient task-store error during a claim would crash the whole local dev stack instead of just skipping the item.

dodizzle (approved):
Verdict: APPROVE — Both previously-flagged issues are verifiably fixed: the pre-claim marker now includes the record id (`scripts/hitl.ts:520`, matches `formatPreClaimMarker`'s two-field contract), and `client.claim()`/`client.claimPr()` are now wrapped in try/catch that logs and continues rather than crashing the loop. No new issues found.

PR #2182, opened 2026-07-23
feat: escalate shipwright-loop's busy-skip log to console.error once busy is stuck past a safety margin (LPF-7.2)

dodizzle:
## Summary
- Adds `BUSY_STALL_THRESHOLD_MS` (35 minutes) constant, set clearly above `claude.ts`'s 30-minute runner ceiling, so it can never fire during normal drain overlap.
- Escalates the busy-skip log from `console.warn` to `console.error` once `busySince`'s elapsed time exceeds that threshold — the message explicitly calls out a stuck/wedged tick that cannot self-recover without a process restart, matching the existing spin-detection escalation style (both Sentry-eligible via `consoleLoggingIntegration`).
- Below the threshold, LPF-7.1's existing `console.warn` behavior is unchanged (byte-for-byte identical message).

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | busySince ≤ threshold keeps console.warn unchanged | MET |
| 2 | busySince > threshold escalates to console.error with stuck/wedged wording | MET |
| 3 | Named constant near SPIN_DETECTION_THRESHOLD/DEFAULT_EMPTY_BACKOFF_*, clearly above 30-min ceiling | MET |
| 4 | Two new unit tests using the Clock fixture, reusing withCapturedWarnings + new withCapturedErrors helper, no existing tests retired | MET |

## Test Plan
- `agent/src/loop-orchestrator.unit.test.ts`: two new tests — one at the 35-minute threshold boundary (asserts console.warn fires, console.error does not), one at 40 minutes (asserts console.error fires with "stuck"/"wedged" wording and console.warn does not).
- `bun test agent/src/loop-orchestrator.unit.test.ts` → 82 pass, 0 fail (80 pre-existing + 2 new).
- `bunx biome lint .` → clean.
- `bun run --filter="*" --sequential typecheck` → all packages pass.
- Full `bun test` → 1249 pass, 1 pre-existing unrelated failure in `claude.unit.test.ts` (confirmed pre-existing on main, not touched by this diff).


agent/src/loop-orchestrator.ts
```diff
@@ -582,13 +595,25 @@ export function createLoopOrchestrator(
...
+    // console.error (Sentry-eligible, matching the spin-detection warn's
+    // style) instead of the routine console.warn.
     if (busy) {
       const elapsedMs = busySince
         ? clock.now().getTime() - busySince.getTime()
         : 0;
-      console.warn(
-        `[loop-orchestrator] tick skipped: busy — a prior tick has been draining for ${elapsedMs}ms (still running)`,
-      );
+      if (elapsedMs > BUSY_STALL_THRESHOLD_MS) {
```
dodizzle, line 621: `busy`/`busySince` are set once per tick (line 619-620), before the `while (true)` drain-until-dry loop starts, and only cleared after that loop fully exhausts the queue (line 940-941). The 35-minute threshold is justified against `claude.ts`'s single-`runner()`-call 30-minute cap, but a tick can sequentially dispatch many candidates within the same busy window — three ~15-20 minute dispatches back-to-back would exceed 35 minutes of cumulative busy time with nothing actually wedged, incorrectly firing `console.error` (Sentry-eligible) and contradicting the PR's own claim that this "can never fire during normal drain overlap." Consider timing each dispatch individually rather than the whole tick, or explicitly re-derive/raise the threshold for cumulative multi-item drain time.

dodizzle (commented):
Verdict: COMMENT — pr-title-lint is currently failing (no Conventional Commits prefix on the PR title), and the new busy-stall threshold measures whole-tick drain time rather than a single dispatch, so it can likely false-positive as "stuck/wedged" during a normal multi-item drain past 35 minutes cumulative.

dodizzle:
Reviewed the finding(s) above:

- `agent/src/loop-orchestrator.ts:608` (busy-stall threshold measures whole-tick drain time, not a single dispatch): confirmed — `busy`/`busySince` were set once per tick (before the `while (true)` drain loop) and only cleared after the loop fully exhausted the queue, so cumulative multi-item drain time could exceed the 35-minute safety margin with nothing actually wedged. Fixed by resetting `busySince` at the top of every drain iteration instead of once at tick-start, so the elapsed-time check now reflects only the current iteration (candidate collection through one dispatch) — bounded the same way a single `runner()` call is. Added a regression test (`LPF-7.2: busySince resets per drain iteration — cumulative multi-item drain time does not falsely escalate`) exercising two sequential 20-minute dispatches (40 min cumulative) to confirm no false escalation.
- pr-title-lint failing on the PR title: this was already resolved before this patch cycle — the title was updated and the latest `pr-title-lint` run against the current head commit is green (run #3708, succeeded), superseding the earlier failing run #3707. No action needed.

dodizzle (commented):
Verdict: APPROVE — Both prior findings (PR title Conventional Commits prefix, busySince measuring whole-tick drain time instead of a single dispatch iteration) are verifiably fixed. busySince now resets at the top of every drain-loop iteration, traced the control flow to confirm no race window, and the new regression test reconstructs the exact false-positive scenario and confirms it no longer escalates. No new issues found.

PR #2227, opened 2026-07-27
test: add fail-open catch-path coverage for isBundleComplete (T-077)

dodizzle:
## Summary
- Adds a unit test proving the fail-open catch path in `getReviewCandidates`'s `isBundleComplete` check (`agent/src/check-review.ts:173-177`): when `isBundleComplete` rejects, the PR proceeds as a candidate rather than being excluded.
- Modeled directly on the two adjacent bundle-gate test cases, using a distinct fixture branch name to avoid collisions.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | New test(...) block added in agent/src/check-review.unit.test.ts, modeled on the two adjacent bundle-gate cases | MET |
| 2 | Injects `async () => { throw new Error(...) }` as isBundleComplete | MET |
| 3 | Asserts getReviewCandidates's result still includes the PR when isBundleComplete rejects | MET |
| 4 | Verification command `bun test agent/src/check-review.unit.test.ts` passes | MET |

## Test Plan
- `bun test agent/src/check-review.unit.test.ts` — 45 pass, 0 fail
- `bunx biome lint .` — clean
- `tsc --noEmit` — clean
- Verified fail-open assertion actually distinguishes behavior: temporarily flipped production `.catch(() => true)` to `.catch(() => false)`, confirmed the new test fails, then reverted (production code unchanged in final diff)

### DIFF
--- a/agent/src/check-review.unit.test.ts
+++ b/agent/src/check-review.unit.test.ts
@@ -662,6 +662,26 @@ describe("getReviewCandidates", () => {
     expect(result).toHaveLength(1);
     expect(result[0].id).toBe("example-org/example-repo#42");
   });
+
+  test("includes a PR when isBundleComplete rejects for its branch (fail-open)", async () => {
+    const pr = makePr({ headRefName: "feat/bundle-check-throws" });
+    const result = await getReviewCandidates(
+      makeDeps(
+        [pr],
+        async () => null,
+        "bodhi-agent",
+        false,
+        async () => null,
+        undefined,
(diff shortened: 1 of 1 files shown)


dodizzle (commented):
Verdict: APPROVE — Single, correctly-scoped unit test covering the isBundleComplete fail-open catch path; verified non-vacuous, no isolation/scrubbing issues, 45/45 tests pass.

PR #2232, opened 2026-07-27
fix: PATCH task blocked status on dev-task Step 5c dead-end BLOCKED

dodizzle:
## Summary
- `dev-task.md` Step 5c's BLOCKED handling now PATCHes the task to `status:'blocked'` with a `blockedReason` when the model-upgrade ladder (haiku → sonnet → opus) is exhausted and the blocker is a genuine dead end — mirroring the exact PATCH shape already used at Steps 1/7/9/9b.5 in the same file.
- Previously this dead-end case just said "escalate to the user" with no task-store write, so the task stayed `in_progress` until the claim TTL reaper reset it back to `pending`, where it re-qualified as ready with no memory of the prior failure and looped indefinitely.
- The PATCH is scoped to only fire when the blocker is *not* a context/size/plan problem the step already knows how to resolve (those branches still self-correct via re-dispatch).
- Refreshed `docs/task-store.md` to document the new `blockedReason` write path.

## Acceptance Criteria

| 1 | Step 5c: when still BLOCKED after the model-upgrade re-dispatch (or already at opus) and the blocker isn't context/size/plan-resolvable, PATCH `status:'blocked'` + `blockedReason`, same curl shape as Steps 1/7/9/9b.5 | MET | `dev-task.md` Step 5c now includes the PATCH with `blockedReason: "implementation_blocked_after_model_escalation"` |
| 2 | Fix does not fire for the context/size/plan cases the step already resolves | MET | PATCH is gated behind "Otherwise —" after the three self-correcting branches; test asserts ordering |
| 3 | `dev-task.content.test.ts` updated to assert Step 5c documents the blocked-status PATCH (content-layer test only, matching BHE-1.1 rationale) | MET | 4 new tests added |

## Test Plan
- `bun test plugins/shipwright/commands/dev-task.content.test.ts` — 44/44 pass
- `bun test plugins/shipwright/commands/` — 353/353 pass
- `bun test` (full suite) — 5138 pass, 2 pre-existing failures unrelated to this change (confirmed failing on `main` too: `agent/src/claude.unit.test.ts`, `scripts/quickstart.smoke.test.ts`)
- `bunx biome lint .` — clean

### DIFF
--- a/docs/task-store.md
+++ b/docs/task-store.md
@@ -99,12 +99,14 @@ Returns `404` if the task doesn't exist or is outside the agent's scope.
 PATCH /tasks/:id
 ```
 
-Body: partial task fields. Agent tokens can only update their own tasks (by `assignee` or `claimedBy`). **Agent tokens cannot set the following fields via PATCH** — these are managed exclusively by their lifecycle endpoints:
+Body: partial task fields. Agent tokens can only update their own tasks (by `assignee` or `claimedBy`). Writable fields (agents): `status` (except `'pending'`), `blockedReason`, `description`, `note`, `model`, and any other fields not listed below. **Agent tokens cannot set the following fields via PATCH** — these are managed exclusively by their lifecycle endpoints:
 
 - `claimedBy`, `claimedAt`, `heartbeatAt` — use `/claim` or `/release`
 - `status: 'pending'` — use `/release` to unclaim, or `/claim` to reclaim
 
 Admin tokens (`agentId === null`) may set any field. Returns the updated task.
+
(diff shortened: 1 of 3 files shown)


dodizzle (commented):
Verdict: APPROVE — Small, precise fix closing a real infinite-loop bug (blocked tasks with no durable give-up signal kept re-qualifying as ready forever). Verified `status:'blocked'`/`blockedReason` are genuinely agent-writable against the actual task-store schema and route guard, the PATCH is correctly gated behind the three self-resolving branches, and the new content tests are substantive. 44/44 tests pass locally, CI green. No blocking issues.

PR #2038, opened 2026-07-21
chore: remove unreferenced chat-types.ts and its generation script

dmcaulay:
## Summary
- Deletes `lib/chat-types.ts` — a generated OpenAPI-types file for the chat service with zero import sites anywhere in the codebase (unlike its siblings `lib/admin-types.ts` and `lib/task-store-types.ts`, which are actively consumed by `admin-spec.smoke.test.ts`, `agent/src/shipwright-runtime-client.ts`, and `lib/task-store-types.unit.test.ts`).
- Removes the now-unused `generate:chat-types` script from root `package.json`. `chat/openapi.json` and its `generate:chat-spec` script are still in use and were left untouched, as was the `openapi-typescript` devDependency (still needed by the admin/task-store generators).
- Fixes a stale doc reference in `docs/test-readiness/test-inventory.md` that still listed the removed file.

The file and its generation script appear to have been scaffolded ahead of an actual chat API client in #1339, but that client was never built — grepping for `openapi-fetch` usage confirms nothing in the `chat/` service (or elsewhere) consumes generated chat types.

## Acceptance Criteria

| # | Criterion | Status | Evidence |
|---|-----------|--------|----------|
| 1 | Confirm lib/chat-types.ts has zero import sites codebase-wide before deleting | MET | grep verified across .ts/.tsx/.json — zero import sites; only reference was the generate script that produced it |
| 2 | Delete lib/chat-types.ts once confirmed dead | MET | 1066-line file removed |
| 3 | Remove generate:chat-types script if nothing consumes its output | MET | package.json script line removed |
| 4 | Do not remove chat/openapi.json or generate:chat-spec (still in use) | MET | Both left untouched; openapi-typescript devDependency kept |
| 5 | Codebase builds/typechecks/tests clean after removal | MET | task typecheck (7/7 workspaces) and bun test (4650 pass, 0 fail) both pass |

## Test Plan
- `bun test` — 4650 pass, 0 fail, 337 skip
- `task typecheck` — all 7 workspaces pass
- `task ci` — lint, check-strings, typecheck, check-config-docs, coverage gate (80.91% lines / 81.30% functions, threshold 80%) all pass
- [x] Pre-commit checks passing

Generated with [Claude Code](https://claude.com/claude-code)

PR #2241, opened 2026-07-27
chore: remove unreferenced task-store-types.ts and its generation script

dodizzle:
## Summary
- Remove `lib/task-store-types.ts` (generated OpenAPI types for the task-store service) — zero import sites anywhere in the codebase
- Both real task-store clients (`metrics/src/lib/task-store-client.ts`, `admin/src/task-store-provisioning-client.ts`) use hand-written inline types instead, never this generated file
- Remove the existence-only test (`lib/task-store-types.unit.test.ts`) and the `generate:task-store-types` package.json script that produced the dead file
- `task-store/openapi.json` (still consumed by mcp-server tool generation) and the `openapi-typescript` devDependency (still used by `generate:admin-types`) are left in place
- Refresh `docs/test-readiness/test-inventory.md` and `docs/test-readiness/test-migration.md` to drop stale references to the removed file

## Precedent
#2038 removed the analogous unreferenced `lib/chat-types.ts` following the same pattern.

## Test Plan
- [x] `bun test` — full suite passes (2 pre-existing unrelated failures on `main`, confirmed unaffected by this change)
- [x] `bun run --filter='*' --sequential typecheck` — all packages pass
- [x] `bun scripts/check-config-docs.ts` — passes
- [x] `bun run scripts/sync-version.ts --check` — passes
- [x] `bun plugins/shipwright/scripts/check-banned-strings.ts` — clean
- [x] Verified no remaining references to `task-store-types` anywhere in the repo


dmcaulay:
This is actually backwards. I think we want a single task store client built on generated types.

dmcaulay:
Closing — after review, this is backwards. lib/task-store-types.ts is a real, actively-maintained generated-OpenAPI-types file (10+ years of real commits keeping it in sync with the task-store API), not stale scaffolding. It has zero references because both existing clients (metrics/src/lib/task-store-client.ts, admin/src/task-store-provisioning-client.ts) independently hand-rolled their own inline types instead of adopting it — that duplication is the actual entropy here, not this file. Deleting it locks in the drift risk permanently. Filing a follow-up plan session to consolidate both clients onto the generated types instead.

PR #2242, opened 2026-07-27
feat(toolchain): docs-first discovery + cross-run cache

dmcaulay:
## Summary
- dev-task's Step 0b and patch's three duplicated toolchain-detection sites (4a.5/5a.5/6a.5) now read `CLAUDE.md` + `docs/*.md`/`ai-docs/*.md` first for explicit build/test/lint/typecheck commands, treating what they find as authoritative — config-file scanning (package.json/Cargo.toml/etc.) now only fills gaps the docs didn't cover. Many projects wrap raw tool invocations (custom scripts, task runners, mise-managed runtimes) that config-file sniffing alone can't see.
- Detection results are cached in `state/toolchain-cache.json` (same tier as `state/error-patrol-ledger.json` — one level up from the repo checkout), keyed by repo, so dev-task and patch share one cache instead of each redetecting from scratch on every task/PR.
- Cache invalidation is a single `git log -1 --format=%H -- CLAUDE.md docs ai-docs package.json ...` call scoped only to toolchain-relevant paths — cheap, and self-healing whenever those files actually change on a branch. A stale/missing cache just costs what detection would cost with no cache at all.

Raised by Dan — prompted by looking at dev-task's Step 0b and noticing it never consulted the project's own docs before guessing from config files.

## Test plan
- [x] `bun test plugins/shipwright/commands/dev-task.content.test.ts plugins/shipwright/commands/dev-task.unit.test.ts plugins/shipwright/commands/patch.content.test.ts` — 145 pass, 0 fail (pre-existing suites, unaffected by these doc-only edits)
- [x] `bun plugins/shipwright/scripts/check-banned-strings.ts` — clean
- [ ] No functional test possible for prompt-content changes beyond the above — behavior will be validated by the next live dev-task/patch run picking up the new cache file

### DIFF
--- a/plugins/shipwright/references/toolchain-patterns.md
+++ b/plugins/shipwright/references/toolchain-patterns.md
@@ -36,7 +36,7 @@ One file per repo (not one shared file keyed by repo) — a shared file read-mod
 **Fingerprint** — a cheap staleness check scoped only to the paths that can change the toolchain, so unrelated commits elsewhere in the repo don't force a redundant re-detection:
 
 ```bash
-git -C {repo-dir} log -1 --format=%H -- CLAUDE.md docs ai-docs package.json package-lock.json yarn.lock pnpm-lock.yaml bun.lock bun.lockb Cargo.toml go.mod pyproject.toml setup.py Gemfile Makefile pom.xml build.gradle build.gradle.kts
+git -C {repo-dir} log -1 --format=%H -- CLAUDE.md docs ai-docs package.json package-lock.json yarn.lock pnpm-lock.yaml bun.lock bun.lockb Cargo.toml go.mod pyproject.toml setup.py Gemfile Makefile Taskfile.yml justfile Justfile mise.toml .mise.toml pom.xml build.gradle build.gradle.kts
 ```
 
 `{repo-dir}` is whichever checkout is live at the point detection runs — `${SHIPWRIGHT_REPO_DIR:-$HOME/src}/{repo}` for dev-task's pre-worktree detection (Step 1/0b runs before the worktree exists); the active `{worktree-path}` for patch, which always operates on an already-existing branch.
--- a/plugins/shipwright/references/toolchain-patterns.md
+++ b/plugins/shipwright/references/toolchain-patterns.md
@@ -1,6 +1,50 @@
 # Toolchain Detection Patterns
 
-Lookup table for auto-detecting project toolchains from config files. Used by all Shipwright commands at startup.
+Lookup table for auto-detecting project toolchains from config files. Used by all Shipwright commands at startup, after the docs-first discovery and cache check below.
+
+## Docs-First Discovery
+
+Before scanning config files, check whether the project's own docs already say how to build/test/lint it. Many projects wrap the raw tool commands — a custom script, a task runner (`Taskfile.yml`, `Makefile` targets that shell out further), a mise-managed runtime, a monorepo command that fans out per-package — and the config-file detection below can miss or misrepresent those wrappers entirely (e.g. detecting bare `jest` when the project actually requires `./scripts/test.sh` for DB setup first).
+
+1. Read `CLAUDE.md` (repo root, plus any nested `CLAUDE.md` the root one `@`-references) for an explicit commands section.
+2. Read `docs/*.md` and `ai-docs/*.md` — whichever directory exists — for a quickstart/setup/contributing doc naming build/test/lint/typecheck commands.
+3. Treat any explicit command, wrapper script, or "use X, not Y" rule found this way as authoritative — it overrides the config-file tables below for that command.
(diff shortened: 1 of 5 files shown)


plugins/shipwright/references/toolchain-patterns.md
```diff
@@ -1,6 +1,52 @@
...
+  }
+}
+```
+
+One file per repo (not one shared file keyed by repo) — a shared file read-modify-written from multiple concurrent processes is not atomic: two agents updating *different* repos' entries at the same time can each read the whole file and clobber the other's addition on write, even though they touched different keys. Splitting by repo removes that cross-repo collision entirely. A same-repo collision (two runs racing on the same repo) can still happen, but it's benign — both would compute the same commands from the same repo state, so a lost update just costs a redundant re-detection next time, not data loss.
+
+**Fingerprint** — a cheap staleness check scoped only to the paths that can change the toolchain, so unrelated commits elsewhere in the repo don't force a redundant re-detection:
+
+```bash
+git -C {repo-dir} log -1 --format=%H -- CLAUDE.md docs ai-docs package.json package-lock.json yarn.lock pnpm-lock.yaml bun.lock bun.lockb Cargo.toml go.mod pyproject.toml setup.py Gemfile Makefile pom.xml build.gradle build.gradle.kts
```
dodizzle: The "Docs-First Discovery" section above motivates this feature by citing task runners (`Taskfile.yml`) and mise-managed runtimes as commands config-file scanning can't see. But this fingerprint's path list doesn't include `Taskfile.yml`, `justfile`, or `mise.toml` — for a project whose build/test/lint commands live in one of those and aren't also mentioned in `CLAUDE.md`/`docs/*.md`/`ai-docs/*.md`, a change to that file after the first cache write won't invalidate the fingerprint. The cache will silently keep serving the stale command indefinitely, which contradicts `plugins/shipwright/CLAUDE.md`'s Principle 1 that a stale local-state cache should cost at most an extra detection pass, never wrong/stale behavior. This repo's own `CLAUDE.md` uses `Taskfile.yml` as its primary entrypoint. Suggest adding `Taskfile.yml`, `justfile`/`Justfile`, and `mise.toml`/`.mise.toml` to this path list.

dodizzle (commented):
Verdict: COMMENT — Fingerprint path list omits Taskfile.yml/justfile/mise.toml, so a project whose commands live in one of those (exactly the wrapper case this PR's docs-first discovery is meant to catch) won't invalidate the cache when it changes.

dmcaulay (commented):
Verdict: APPROVE — Docs-first toolchain discovery + cross-run cache is internally consistent across all four call sites (dev-task 0b, patch 4a.5/5a.5/6a.5), and the prior review's fingerprint-path-list gap (missing Taskfile.yml/justfile/mise.toml) is fully resolved — verified directly against toolchain-patterns.md on disk. No blocking findings.

dmcaulay:
Reviewed the finding above and did not implement changes — premise no longer holds:

- @dodizzle's fingerprint-path-list gap (missing `Taskfile.yml`/`justfile`/`mise.toml`): already fixed in this PR's current HEAD (`038ce85`). The fingerprint command in `plugins/shipwright/references/toolchain-patterns.md` now includes `Taskfile.yml`, `justfile`, `Justfile`, `mise.toml`, and `.mise.toml` alongside the other fingerprinted paths, and a regression test (`CPF-review-2242` in `patch.content.test.ts`) locks this in. This matches the fix @dmcaulay's own follow-up review already confirmed ("the prior review's fingerprint-path-list gap ... is fully resolved — verified directly against toolchain-patterns.md on disk").

dmcaulay:
CI (`agent docker build`) is failing on this PR — but not because of anything in this diff. The `agent/Dockerfile` step `curl -fsSL https://deb.nodesource.com/setup_22.x | bash -` is returning `curl: (22) ... 403` right now, and the identical failure reproduces on `main` itself as of the latest push (run #5174, 2026-07-27T23:33Z) and on at least two other unrelated open PRs (`fix/security-osv-cve-lockfile-bumps`, `docs/test-readiness-refresh-20260726`) in the same time window. This is a repo-wide infra break in the `agent` Docker image build, not something this PR introduced.

Skipping automated CI-fix dispatch here — patching `agent/Dockerfile` from this branch would only fix it for this one PR while every other open PR (and `main`) stays broken, and the real fix (repin the Node install method, since nodesource's `setup_22.x` script is returning 403 for everyone) belongs in its own PR against `main`. Flagging for human triage.

dodizzle (approved):
Verdict: APPROVE — Docs-first toolchain discovery + cross-run cache is internally consistent across all four call sites (dev-task 0b, patch 4a.5/5a.5/6a.5); the prior review's fingerprint-path-list gap (Taskfile.yml/justfile/mise.toml) is verified fixed and locked in by a regression test. No blocking findings.

PR #2243, opened 2026-07-27
feat: mark PR reviewed-at-commit instead of releasing to pending in review.md Step 5

dodizzle:
## Summary
- Replaces the `POST /prs/{id}/release` call in review.md Step 5's unresolved-feedback skip path with a `PATCH /prs/{id}` call that sets `reviewState: "posted"` and `commitSha: {headRefOid}`, so the same PR no longer re-qualifies as a review candidate on every loop tick while a human thread stays unresolved.
- Documents explicitly that `staged` is not set by this call, so it does not interact with `/shipwright:review-staged`'s staged-record flow.
- Adds a "Commit-level dedup pattern" note to docs/task-store.md explaining how `reviewState=posted` + `commitSha` differs from `POST /prs/:id/release`.

## Acceptance Criteria
| 1 | Replace POST /prs/{id}/release with PATCH /prs/{id} -d '{"reviewState":"posted","commitSha":"{headRefOid}"}' | MET | review.md Step 5 now issues the PATCH call with exact body shape; old release call removed |
| 2 | Confirm text notes staged:true is not set / no interaction with review-staged | MET | review.md explicit note: "staged is NOT set here... does not interact with /shipwright:review-staged" |
| 3 | Update review.content.test.ts to assert PATCH-to-posted docs, no separate unit/integration test | MET | 5 new tests in review.content.test.ts BHE-1.3 describe block, all passing (25/25 total) |

## Test Plan
- `bun test plugins/shipwright/commands/review.content.test.ts` — 25 pass, 0 fail (5 new BHE-1.3 tests)
- `bun test` (full suite) — 5154 pass, 2 pre-existing/env-dependent failures in `agent/src/claude.unit.test.ts` reproduced identically on `main`, unrelated to this diff
- `bun run --filter='*' --sequential typecheck` — all 7 workspaces clean
- `bunx biome lint .` — clean


plugins/shipwright/commands/review.md
```diff
@@ -256,8 +256,20 @@ If **any** of the following are true, this PR has substantive unresolved feedbac
 
 If substantive unresolved feedback is found: print
 `Skipping #{pr} — unresolved feedback from @{login} ({type} on {date}). No commits since.`,
-release the claim so the record returns to `pending`
-(`POST $SHIPWRIGHT_TASK_STORE_URL/prs/{PR_RECORD_ID}/release`), respond `[silent]`, and stop.
+mark the PR as reviewed-at-this-commit (without staging) so the record is not re-evaluated at the
```
dodizzle, line 259: This new dedup marker (reviewState=posted + commitSha) is silently reverted by the background reconciler for the plain-unresolved-comment trigger case (condition b above, sourced from `comments` not `reviews`).

`agent/src/pr-state-reconciler.ts`'s `reconcilePostedReviewStateRecord()` (line 897, run every 30-60 min via `agent/src/index.ts:335`) calls `hasAnyReviewAtHead()` (line 740), which only inspects `PrReviewData.reviews.nodes` — formal GitHub review objects. It never looks at issue-level PR comments. For trigger case (b) (a plain unresolved comment, no formal review), there's no review object at head, so `hasAnyReviewAtHead()` returns false and the very next reconcile pass PATCHes `reviewState` back to `pending` (line 908) — undoing this PR's dedup marker and reproducing the same re-qualify-every-tick churn this PR sets out to fix.

This only reliably holds for trigger case (a) (a formal CHANGES_REQUESTED review at head). Not a regression vs. `main` (old `POST /release` produced the same end state, just immediately instead of after a reconcile delay), but worth either fixing `hasAnyReviewAtHead()` to also account for a qualifying unresolved comment at head, or caveating this note and the new docs/task-store.md "Commit-level dedup pattern" text (which currently states unconditionally that this "prevent[s] re-review of the same head until new commits land").

dodizzle (commented):
Verdict: COMMENT — the new PATCH-to-posted dedup only reliably holds for the CHANGES_REQUESTED-review trigger case; for the plain-unresolved-comment trigger case, agent/src/pr-state-reconciler.ts's reconcilePostedReviewStateRecord() silently PATCHes reviewState back to pending on its next pass (every 30-60 min) since hasAnyReviewAtHead() only checks formal review objects, not issue comments — undoing the dedup this PR adds for that case. See inline comment.

dodizzle (commented):
Verdict: APPROVE — Docs/prompt-only change; every factual claim about task-store and reconciler behavior verified accurate against the actual TypeScript source, no breaking surface, passing tests, green CI, and the prior review round's finding was directly addressed in the latest commit.

PR #2269, opened 2026-07-28
fix: raise dedup query limit and add pagination guard to fix-skill dedup checks

dmcaulay:
## Summary
- Verified `GET /tasks`'s plain list path already returns `{ tasks, total, limit, offset }` (PR #670, pre-dates this bug) — acceptance criterion 1 was already satisfied at the API layer, so no API code changed; `docs/task-store.md` now documents that contract and how to detect truncation.
- The real gap was on the consumer side: `test-fix`, `entropy-fix`, `error-fix`, and `consolidation-fix`'s "Dedup Check" steps queried `GET /tasks?status=pending`/`status=in_progress` unbounded, with no `limit` and no check of `total` — silently undercounting once a repo's active-task count exceeded the API's default `limit=50`. This is exactly what let 3 already-queued tasks slip past dedup on 2026-07-23.
- Updated all four skills' dedup queries to pass `&limit=1000` and added a documented pagination guard: compare the response's `total` against `tasks.length`, and page via `&offset=` until every active task is collected — never build the dedup set from a partial page.

## Acceptance Criteria

| 1 | GET /tasks either returns all results by default or clearly signals truncation via `total`/`hasMore` | MET | Already implemented in `task-store/src/task-service.ts` (`list()` returns `{tasks, total, limit, offset}`); now documented in `docs/task-store.md`. |
| 2 | Dedup-style consumers (test-fix/entropy-fix/error-fix/consolidation-fix) can detect truncation and page through | MET | All four `SKILL.md` dedup sections raise `limit` and document the `total`-vs-fetched pagination guard, backed by new/extended `*.content.test.ts` assertions. |

## Test Plan
- [x] New `SKILL.content.test.ts` added for `test-fix`, `entropy-fix`, `error-fix` (mirrors existing `security-scan/SKILL.content.test.ts` pattern) — TDD RED confirmed before the `SKILL.md` edits, GREEN after.
- [x] `consolidation-fix.content.test.ts`'s existing dedup-mechanism block extended with the same pagination-safe assertions.
- [x] `bun test plugins/shipwright/` — 1072 pass, 0 fail.
- [x] `task lint`, `task check-strings`, `task check-config-docs` — all clean.
- [ ] `task typecheck` / full `task ci` — not runnable in this sandbox (`tsc` binary absent, no local Postgres for task-store integration/smoke suite); confirmed as a pre-existing environment gap, not introduced by this change. This PR only touches Markdown and `bun:test`-only test files (no new TS types), so the risk is low — CI will run the full gate.

### DIFF
--- a/docs/task-store.md
+++ b/docs/task-store.md
@@ -52,12 +52,26 @@ Query params:
 | `pr` | number | Filter by PR number |
 | `branch` | string | Filter by branch name |
 | `hitl` | `true` or `false` | Filter by HITL (human-in-the-loop) flag: return tasks with or without the flag set |
-| `limit` | number | Page size |
-| `offset` | number | Page offset |
+| `limit` | number | Page size. Defaults to `50` when omitted. |
+| `offset` | number | Page offset. Defaults to `0` when omitted. |
 | `sort` | string | `asc` (default) or `desc` — orders results by `createdAt`. Default preserves existing ascending order for all callers. |
 | `updatedSince` | string | ISO timestamp. Only return tasks with `updatedAt >= this value`. A conservative pre-filter (not a precise sync anchor). Omitting it preserves current (unfiltered) behavior. |
 
-Returns `{ tasks: Task[], total: number }`.
+Returns `{ tasks: Task[], total: number, limit: number, offset: number }`. `total` is the
+count of *all* tasks matching the filters, independent of `limit`/`offset` — compare it
+against `tasks.length` to detect truncation. A caller that queries without an explicit
+`&limit=` and gets back 50 tasks with `total: 137` has only seen the first page; re-issue the
+query with `&limit=` raised (or page via `&offset=`, accumulating pages until the summed
+`tasks.length >= total`) rather than treating the 50-task response as the full result set.
+This matters most for any dedup-style check (e.g. `entropy-fix`/`error-fix`/`test-fix`/
+`consolidation-fix`'s "Dedup Check" steps) that queries `?status=pending`/`?status=in_progress`
+before filing new tasks — an unbounded default-limit query on a repo with more than 50 active
+tasks will silently miss the tail of the list.
+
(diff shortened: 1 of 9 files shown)


docs/task-store.md
```diff
@@ -52,12 +52,26 @@ Query params:
...
 | `updatedSince` | string | ISO timestamp. Only return tasks with `updatedAt >= this value`. A conservative pre-filter (not a precise sync anchor). Omitting it preserves current (unfiltered) behavior. |
 
-Returns `{ tasks: Task[], total: number }`.
+Returns `{ tasks: Task[], total: number, limit: number, offset: number }`. `total` is the
+count of *all* tasks matching the filters, independent of `limit`/`offset` — compare it
+against `tasks.length` to detect truncation. A caller that queries without an explicit
+`&limit=` and gets back 50 tasks with `total: 137` has only seen the first page; re-issue the
+query with `&limit=` raised (or page via `&offset=`, accumulating pages until the summed
+`tasks.length >= total`) rather than treating the 50-task response as the full result set.
+This matters most for any dedup-style check (e.g. `entropy-fix`/`error-fix`/`test-fix`/
```
dmcaulay, line 66: `security-fix/SKILL.md`'s Step 6q.1 has the identical unbounded `?status=pending`/`?status=in_progress` dedup query (no `&limit=`, verified — no `limit=1000` anywhere in that file) and isn't touched by this PR or listed here. It's a fifth dedup-style consumer of the same endpoint with the same silent-undercount failure mode past 50 active tasks. Not blocking this PR (out of its stated acceptance criteria), but worth a follow-up to bring it in line with the other four.

dmcaulay (commented):
Verdict: APPROVE — Pagination fix and guard logic verified correct against task-service.ts; one non-blocking suggestion inline (security-fix/SKILL.md has the identical unpatched dedup query, worth a quick follow-up).

PR #2059, opened 2026-07-21
feat: idle-reset timeout with 1hr hard ceiling backstop

dmcaulay:
## Summary
- Replace the flat `setTimeout(timeoutMs)` in `_spawn` with a dual-timer scheme: an idle-reset timer (cleared/restarted on every stdout line, new `SHIPWRIGHT_CLAUDE_IDLE_TIMEOUT_MS`, default 25min) and a hard ceiling (bumped `DEFAULT_CLAUDE_TIMEOUT_MS` 30min → 1hr) that backstops a continuously-active-but-never-converging session.
- `ClaudeTimeoutError` gains `reason: "idle" | "ceiling"` and renames `modelUsage` → `partialModelUsage` (always partial by definition on a timeout path).
- Since `DEFAULT_CLAIM_TTL_MS` derives from `DEFAULT_CLAUDE_TIMEOUT_MS + CLAIM_TTL_BUFFER_MS` unchanged, the claim TTL grows in lockstep (35min → 65min) — updated all doc/comment/hardcoded-literal references that assumed the old values (task-store's own claim-freshness fallback, which had drifted to a hardcoded `2_100_000` instead of importing the shared constant — now fixed to import it).

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | New env var `SHIPWRIGHT_CLAUDE_IDLE_TIMEOUT_MS` read in `agent/src/config.ts`, default 1,500,000ms | MET |
| 2 | `DEFAULT_CLAUDE_TIMEOUT_MS` bumped to 3,600,000; `DEFAULT_CLAIM_TTL_MS` derivation unchanged | MET |
| 3 | Idle timer resets on each stdout line; hard ceiling still fires if idle alone never lapses but total exceeds the ceiling | MET |
| 4 | `ClaudeTimeoutError` carries partial `ModelUsage` and which limit fired (idle vs ceiling) | MET |
| 5 | Unit tests: idle-reset-on-activity, idle-fires-despite-ceiling-headroom, ceiling-fires-despite-continuous-activity; existing `ClaudeTimeoutError` tests updated in place | MET |

## Test Plan
- `agent/src/claude.unit.test.ts` — 3 new scenarios (short real-timeout convention, consistent with existing file style) plus 2 existing timeout tests + 1 partial-usage test updated for the new constructor shape
- `agent/src/config.unit.test.ts` — new env var default/override/fallback coverage
- `lib/claim-ttl.unit.test.ts` — bumped constant + derived TTL
[...]

PR #2275, opened 2026-07-28
feat: replace timestamp-only second-round escalation with content-correlation judgment

dodizzle:
## Summary
- Rewrites `patch.md` Step 5a.7 so the escalate-vs-fix decision requires an explicit `SAME_FINDING`/`DIFFERENT_FINDING` content-correlation judgment (with justification) instead of relying on reply-timestamp ordering alone.
- Keeps the timestamp comparison as a cheap pre-filter for candidate prior replies, states the anti-pattern explicitly ("a reply predating the review is not sufficient by itself to escalate"), and prefers inline-thread anchors (stable `path`/`line`) over freeform PR-comment text matching when the finding came from a thread.
- Root-caused via a real incident on PR #2153 (app-vitals/shipwright): an unrelated earlier rebuttal (about a different, already-fixed finding) happened to predate a brand-new finding's review, causing a false escalation to HITL that repeated after the flag was cleared.

## Acceptance Criteria

| 1 | SAME_FINDING/DIFFERENT_FINDING judgment with justification printed before escalating | MET | patch.md Step 5a.7 Step 2 block |
| 2 | Anti-pattern stated: reply predating review not sufficient alone | MET | patch.md "Anti-pattern" callout with PR #2153 rationale |
| 3 | Escalation gated on ≥1 SAME_FINDING; otherwise proceed to Step 5b as first-round | MET | patch.md escalation gate + "Otherwise" branch |
| 4 | Test updates in existing RPF-1.3 describe block, no regression, no runtime tests added | MET | patch.content.test.ts, 5 new content assertions, 93/93 pass |
| 5 | check-patch.ts unchanged | MET | zero diff on agent/src/check-patch.ts |

## Test Plan
- `bun test plugins/shipwright/commands/patch.content.test.ts` — 93/93 pass
- `task lint` — clean

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -46,7 +46,7 @@
     },
     "agent": {
       "name": "@shipwright/agent",
-      "version": "1.88.0",
+      "version": "1.88.1",
       "dependencies": {
         "@octokit/auth-app": "^8.2.0",
         "@sentry/bun": "^10.63.0",
@@ -99,7 +99,7 @@
     },
     "metrics": {
       "name": "@shipwright/metrics",
-      "version": "1.88.0",
+      "version": "1.88.1",
       "dependencies": {
         "@hono/zod-openapi": "^0.18.3",
         "@sentry/bun": "^10.63.0",
@@ -117,7 +117,7 @@
     },
     "plugins/shipwright": {
       "name": "@shipwright/plugin",
-      "version": "1.88.0",
+      "version": "1.88.1",
       "devDependencies": {
         "bun-types": "*",
         "typescript": "*",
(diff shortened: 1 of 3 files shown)


dmcaulay (approved):
Verdict: APPROVE — Closes the real PR #2153 false-escalation gap (same class as #2059/#2082) by requiring an explicit SAME_FINDING/DIFFERENT_FINDING content-correlation judgment instead of timestamp precedence alone. Escalation gate is internally consistent, inline-thread-anchor guidance is actionable, new content-assertion tests verify structural properties and guard against regressing to the old framing (93/93 pass), and check-patch.ts is confirmed untouched. No blocking findings.

PR #2287, opened 2026-07-28
feat: add scopeDegraded signal to auth middleware and GET /tasks

dodizzle:
## Summary
- Adds a `scopeDegraded: boolean` context var to `createBearerAuthMiddleware` (task-store/src/auth.ts), set `true` only when the agent token's repo-scope resolver call itself fails (network error, timeout, non-2xx status, malformed JSON) — distinct from a legitimate empty repo scope.
- `createScopeResolver` now re-throws on genuine infra failures instead of silently swallowing them to `[]`, while still returning `[]` for a successful-but-wrong-shaped response body (not an infra failure).
- Threads `scopeDegraded` into GET /tasks's JSON response envelope (all three branches: `?ready=true`, `?state=blocked`, main list), alongside `tasks`/`total`.
- Purely additive/observability — no change to existing authz behavior, `repos:[]` fallback, or `useAgentScope` logic.

## Acceptance Criteria

| 1 | scopeDegraded:true only on resolver throw/reject (network/timeout/non-2xx/malformed JSON); false on success (incl. legit []) and always false for admin tokens | MET | auth.ts try/catch sets `scopeDegraded=true` only in catch branch; `createScopeResolver` throws for fetch failure, non-2xx, and JSON parse failure, while still returning `[]` for shape mismatches |
| 2 | GET /tasks response includes scopeDegraded alongside tasks/total | MET | `routes/tasks.ts` reads `c.get("scopeDegraded")` and includes it in all three response branches; `TaskListResponseSchema` updated |
| 3 | No change to existing authz behavior | MET | `repos` fallback logic and `useAgentScope`/ownership logic untouched |
| 4 | Unit tests in auth.unit.test.ts (throw→true, resolve-empty→false, admin→false); smoke test extension; no existing tests retired | MET | `auth.unit.test.ts` (new, 3 tests); `tasks.openapi.smoke.test.ts` extended; 4 `auth.integration.test.ts` tests had expectations updated (not deleted) to match the deliberate throw-contract change |

## Test Plan
- `task-store/src/auth.unit.test.ts` (new): resolver throws → scopeDegraded true; resolver resolves to `[]` → scopeDegraded false; admin token → scopeDegraded false, resolver not called.
- `task-store/src/auth.integration.test.ts`: updated 4 cases (404, 500, malformed JSON, network-unreachable) to assert `createScopeResolver` rejects; body-shape-mismatch cases still assert `[]`.
- `task-store/src/routes/tasks.openapi.smoke.test.ts`: new case asserting `scopeDegraded: true` appears in the GET /tasks response body when the resolver fails.
- `docs/task-store.md` refreshed to document the new field.
- [x] Pre-commit checks passing (`task lint`, `task typecheck`, coverage: auth.ts 100%, tasks.ts 95.6% — both well above the 80% gate)
- Note: `task ci`'s `test:coverage` step reports 1 fail in `metrics/src/lib/errors.unit.test.ts` — verified this is a pre-existing test-isolation flake that reproduces identically on unmodified `main` under the full-suite run; unrelated to this change.


task-store/src/auth.unit.test.ts
```diff
@@ -0,0 +1,122 @@
...
+      return raw === AGENT_TOKEN
+        ? { id: "tok-agent", agentId: AGENT_ID }
+        : null;
+    },
+  };
+}
+
+/** Build a minimal Hono app that mounts the bearer auth middleware and exposes
+ *  the resolved `scopeDegraded` (and `repos`) on GET /whoami for inspection. */
+function makeAuthApp(
```
dodizzle: CLAUDE.md's Test conventions table defines `*.unit.test.ts` as "pure logic, no I/O" and `*.smoke.test.ts` as "HTTP route contracts — Hono endpoints via in-process `app.request()`". This file's `makeAuthApp` builds a `new Hono()` app and exercises the middleware via `app.request()` — structurally identical to the pre-existing `auth.smoke.test.ts`, which already covers this same middleware with the same `fakeAdminTokenService`/`fakeAgentTokenService`/`makeAuthApp` pattern. Consider moving these 3 `scopeDegraded` cases into `auth.smoke.test.ts`'s existing "scope resolver" describe block instead of a new file at a different layer.

dodizzle (commented):
Verdict: COMMENT — scopeDegraded semantics are implemented correctly end-to-end, but the new auth.unit.test.ts duplicates the existing auth.smoke.test.ts pattern for the same middleware and should live at the smoke layer per this repo's own CLAUDE.md test-layer naming convention.

dodizzle (commented):
Verdict: APPROVE — Prior finding (auth.unit.test.ts duplicating the smoke-layer pattern) is fully resolved: the tests were relocated into auth.smoke.test.ts's existing scope-resolver describe block and the duplicate file removed. Core logic re-verified correct across all auth.ts paths and all three GET /tasks response branches. No new issues found.

PR #2299, opened 2026-07-28
fix: strip SENTRY_DSN from Claude Code session spawn env

dodizzle:
## Summary
- `_spawn()` in `agent/src/claude.ts` now builds the child process env by destructuring `SENTRY_DSN` out of `process.env` instead of passing `process.env` straight through, so spawned Claude Code sessions (and any `bun test` run inside them) can never construct a real Sentry client from ambient env.
- This is the root-cause fix for shipwright agent containers leaking real error events to production Sentry from test runs. The agent's own operational Sentry reporting (`reportCronFailure`, session-failure capture in `claude.ts:539`) is untouched — it runs in the parent process against its own already-initialized client, not inside the spawned child's env.

## Acceptance Criteria

| 1 | `_spawn()` constructs its env as `{ ...process.env }` with `SENTRY_DSN` deleted/omitted, instead of passing `process.env` directly, before calling `spawner()`. | MET | `claude.ts:403` destructures `SENTRY_DSN` out via rest spread; `claude.ts:406` passes `env: spawnEnv` instead of `env: process.env` |
| 2 | New unit test in `claude.unit.test.ts` using the existing injected-spawner pattern; sets `SENTRY_DSN`, asserts it's omitted from the spawned env while an unrelated key (`PATH`) is retained; net-new coverage, no existing tests retired. | MET | `claude.unit.test.ts:312-331` — new test added (64 tests total, 63 pre-existing + 1 new); asserts `opts.env.SENTRY_DSN` is `undefined` and `opts.env.PATH` matches `process.env.PATH`; restores/deletes `SENTRY_DSN` in a `finally` block to avoid leaking into sibling tests |

## Test Plan
- `NODE_ENV=test bun test agent/src/claude.unit.test.ts` — 64/64 pass
- New test verified RED (failed against unfixed `env: process.env`) before the fix, GREEN after
- `bun run --filter='@shipwright/agent' typecheck` — clean
- Coverage on `agent/src/claude.ts`: 100% lines

### DIFF
--- a/agent/src/claude.ts
+++ b/agent/src/claude.ts
@@ -394,9 +394,16 @@ export function createRunClaude(
     args: string[],
     perCallOnProgress?: ProgressCallback,
   ): Promise<ClaudeRunResult> {
+    // Strip SENTRY_DSN so a spawned Claude Code session (and any `bun test`
+    // it runs internally) can never construct a real Sentry client from
+    // ambient env — this is not the agent's own operational Sentry
+    // reporting (see the reportCronFailure / captureException call sites),
+    // which runs in this parent process against its own client, unaffected
+    // by the child's env.
+    const { SENTRY_DSN: _sentryDsn, ...spawnEnv } = process.env;
     const proc = spawner(["claude", ...args], {
       cwd: workspace,
-      env: process.env,
+      env: spawnEnv,
       stdout: "pipe",
       stderr: "pipe",
     });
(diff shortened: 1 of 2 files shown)


dodizzle (commented):
Verdict: APPROVE — Correct, minimal, well-tested fix; strips SENTRY_DSN from the spawned Claude Code child env so test runs inside it can't leak to production Sentry. No breaking changes, no CLAUDE.md violations. (Posted as COMMENT — GitHub blocks self-APPROVE on own PRs.)

PR #2308, opened 2026-07-28
fix: centralize NODE_ENV=test guard via bunfig.toml preload, retire piecemeal prefixing

dodizzle:
## Summary
- Adds `scripts/test-env-preload.ts`, wired into `bunfig.toml`'s `[test]` preload array, so `NODE_ENV` is unconditionally forced to `"test"` before any test file loads — covering every `bun test` invocation path (root suite, subpackage scripts, single-file runs), not just the two sites SEN-3.1 prefixed.
- Removes the now-redundant `NODE_ENV=test` prefixes from `Taskfile.yml`'s `test`/`test:coverage` tasks and root `package.json`'s `test` script.
- Updates `docs/observability.md` to describe the preload mechanism instead of the per-site prefix approach.
- Deletes `scripts/test-env-guard.unit.test.ts` (only asserted the literal prefix string existed — never tested the actual invariant) and replaces it with `scripts/test-env-preload.unit.test.ts`, which pre-sets `NODE_ENV=production`, invokes the preload's exported logic directly, and asserts `NODE_ENV` becomes `"test"`.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Preload script sets NODE_ENV=test via bunfig.toml `[test]` preload | MET |
| 2 | Taskfile.yml/package.json no longer carry NODE_ENV=test prefix | MET |
| 3 | docs/observability.md reflects preload mechanism | MET |
| 4 | Old string-match test deleted, replaced with invariant-testing unit test | MET |

## Test Plan
- `bun test scripts/test-env-preload.unit.test.ts` with `NODE_ENV=production` preset — verified the preload correctly forces it back to `test`.
- `task lint` and `task typecheck` pass clean.
- `task test` — full suite passes except one pre-existing, unrelated failure (`agent/src/claude.unit.test.ts` › `extraAllowedTools`) confirmed to fail identically on `main` and to pass in the actual GitHub Actions CI run for `main`'s current HEAD — an environment-specific local flake, not a regression from this change.


bunfig.toml
```diff
@@ -16,5 +16,12 @@
...
 # See docs/test-readiness/naming.md for the full naming + exclusion reference.
+#
+# preload runs scripts/test-env-preload.ts before any test file loads,
+# unconditionally forcing NODE_ENV=test for every `bun test` invocation path
+# in this repo (root suite, subpackage test scripts, single-file runs) —
+# centralizing the guard here instead of prefixing individual invocation
+# sites. See docs/observability.md for why this matters (Sentry test guard).
 [test]
 pathIgnorePatterns = ["**/site/**", "**/metrics/e2e/**", "**/admin/e2e/**", "**/*.canary.test.ts"]
+preload = ["./scripts/test-env-preload.ts"]
```
dodizzle, line 30: Verified empirically against this commit: this preload does not fire when `bun test` runs with cwd inside a subpackage (no local bunfig.toml there — only one exists, at the root). That includes `bun run --filter="@shipwright/agent" test`, the same invocation shape Taskfile's `typecheck` task already uses (`bun run --filter="*" --sequential typecheck`). With NODE_ENV pre-exported as something other than `"test"` by a parent shell, a subpackage/filtered test run leaves NODE_ENV unchanged, so `buildSentryInitOptions()` would NOT treat it as a test run — Sentry could actually init and leak test error-path assertions as real events, the exact incident this guard exists to prevent. This contradicts the comment above ("every `bun test` invocation path ... subpackage test scripts") and the PR body's coverage claim. Consider a bunfig.toml in each subpackage (or keeping an explicit NODE_ENV=test prefix for any script that invokes `bun test` with cwd inside a subpackage) until Bun supports preload inheritance for nested/filtered runs.

dodizzle (commented):
Verdict: COMMENT — The centralized preload doesn't actually cover subpackage/filtered-workspace `bun test` invocations (e.g. `bun run --filter=<pkg> test`), contradicting the PR's own coverage claim and leaving the exact Sentry test-leak gap this guard exists to prevent — see inline comments.

scripts/test-env-preload.unit.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    process.env.NODE_ENV = "production";
+  });
+
+  afterEach(() => {
+    // Restore so this test doesn't leak NODE_ENV state to sibling suites
+    // sharing the Bun test process.
+    process.env.NODE_ENV = ORIGINAL_NODE_ENV;
+  });
+
+  test("forces NODE_ENV to 'test' even when already set to something else", () => {
```
dodizzle, line 35: This only calls `setTestEnv()` directly — it doesn't spawn `bun test` from a subpackage/filtered cwd to verify the preload actually fires end-to-end for the invocation shapes the PR claims are now covered ("subpackage test scripts"). That's the same class of gap the PR's own rationale cites for deleting the old test ("never tested the actual invariant") — an integration-style test spawning `bun test` with cwd set to a subpackage and a conflicting pre-exported NODE_ENV would have caught the bunfig.toml gap noted above before merge.

docs/observability.md
```diff
@@ -6,7 +6,7 @@ Everything below this line through [Configuration reference](#configuration-refe
 
 ## Overview
 
-Every Shipwright service (`admin`, `metrics`, `task-store`, `agent`) can report to Sentry when `SENTRY_DSN` is set in its own environment. Each service reads its own `SENTRY_DSN` — there is no shared/global toggle. When `SENTRY_DSN` is unset (the default), or when `NODE_ENV` is `"test"` (explicitly set via `NODE_ENV=test` prefix in test commands), Sentry is fully inert: no init call, zero telemetry, zero overhead. The test guard prevents test error-path assertions from leaking to production Sentry as real events.
+Every Shipwright service (`admin`, `metrics`, `task-store`, `agent`) can report to Sentry when `SENTRY_DSN` is set in its own environment. Each service reads its own `SENTRY_DSN` — there is no shared/global toggle. When `SENTRY_DSN` is unset (the default), or when `NODE_ENV` is `"test"` (unconditionally set via a `bunfig.toml` `[test]` preload script, `scripts/test-env-preload.ts`, before any test file loads), Sentry is fully inert: no init call, zero telemetry, zero overhead. The test guard prevents test error-path assertions from leaking to production Sentry as real events.
```
dodizzle: This states NODE_ENV=test is "unconditionally set" by the preload with no caveat — per the bunfig.toml finding, that's not true for subpackage/filtered `bun test` invocations. Worth qualifying so a future debugger of an unexpected Sentry event from such a run doesn't assume this guard is airtight.

dmcaulay (approved):
Verdict: APPROVE — Patch commit verifiably closes all three subpackage-cwd/preload gaps flagged in the author's own self-review, with a real end-to-end integration test proving it. No new issues found.

PR #2313, opened 2026-07-28
feat: remove local worktree when pr-state-reconciler confirms PR merged/closed

dmcaulay:
## Summary
- Extended `GhPrView` and its `gh pr view --json` fields with `headRefName`, piggybacking on the reconciler's existing per-record gh call (no new API cost, avoids the PSR-1.2 rate-limit pattern).
- Extended `reconcileRecord()` to remove the PR's local worktree after a successful merged/closed state PATCH, gated on the `cleanup_merged_worktrees` policy field (`state/agent-policy.md`), via a new injected `removeWorktree` dep.
- Added `readCleanupMergedWorktrees()` to `check-helpers.ts`, mirroring the existing `readAllowSelfReview()` pattern exactly.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | MERGED + `cleanup_merged_worktrees` true → `removeWorktree` called with repo+branch-slug path, in addition to the state PATCH | MET |
| 2 | Still OPEN → `removeWorktree` never called | MET |
| 3 | `cleanup_merged_worktrees` false → PATCH still happens, `removeWorktree` not called | MET |
| 4 | `removeWorktree` failure caught and logged, doesn't affect the already-succeeded PATCH | MET |
| 5 | Additive test coverage in `pr-state-reconciler.unit.test.ts`, no existing test retired | MET |

## Test Plan
- 5 new test cases in `agent/src/pr-state-reconciler.unit.test.ts` covering all acceptance criteria (merged+policy-true, closed+policy-true, still-open no-op, policy-false, removeWorktree-throws isolation).
- New `readCleanupMergedWorktrees` coverage in `agent/src/check-helpers.unit.test.ts`.
- All 80 pre-existing pr-state-reconciler tests + 144 check-helpers tests still pass, unmodified.
- [x] Pre-commit checks passing (`task typecheck`, `task lint` clean; `task ci` fails only on the known pre-existing `claude.unit.test.ts` `extraAllowedTools` flake, confirmed unrelated to this diff)

Generated with [Claude Code](https://claude.com/claude-code)

PR #2318, opened 2026-07-28
feat: wire worktree reaper pass into agent reconciler interval

dmcaulay:
## Summary
- Wires WTR-1.2's `reconcileStaleWorktrees` (stale-worktree age sweep) into `agent/src/index.ts`'s existing PR-state-reconciler `setInterval` as a third, independent try/catch pass — same tick, no new timer, no new env var.
- Deps are built lazily on first tick via `worktree-reaper.ts`'s own `buildProductionDeps`, matching the existing `reconcilerDeps`/`reviewStateReconcilerDeps` pattern.
- Updates the Step 5b doc comment to document the new pass and to note that WTR-1.3's `removeWorktree`/policy-content wiring needed **no separate pass**: it already flows through the existing PR-state-reconciler pass automatically, because `buildPrStateReconcilerDeps` (`buildProductionDeps` in `pr-state-reconciler.ts`) constructs `removeWorktree`/`isCleanupMergedWorktreesEnabled` internally — that half of the task's original two-pass description was already satisfied by WTR-1.3 landing (PR #2313) before this task ran.
- `docs/agent.md` refreshed to reflect the new pass.

## Deviation from task description (documented, deliberate)
The task description asked for "two more independent try/catch passes." Investigation showed WTR-1.2 and WTR-1.3 (this task's dependencies) had already merged, and WTR-1.3's implementation approach bakes `removeWorktree`/policy-content deps directly into `buildProductionDeps()` rather than exposing them as separate injectable params — meaning the existing first pass (`reconcilePrState`) already gets that wiring for free with zero code changes. Adding a second, duplicate `buildPrStateReconcilerDeps`/`reconcilePrState` call would have doubled GitHub API and task-store PATCH traffic for no behavioral gain. Only one genuinely new pass (`reconcileStaleWorktrees`) was added; the reasoning is documented inline in the Step 5b comment.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | New pass invoked once per tick from the same setInterval as existing passes (removeWorktree wiring confirmed active via existing pass) | MET |
| 2 | New pass wrapped in its own try/catch, isolated from the other passes | MET |
| 3 | Deps built lazily on first tick, matching existing pattern | MET |
| 4 | No automated test (index.ts has no test harness) — manual/smoke verification via `task stack`; no existing tests retired | MET |

## Test Plan
- [x] `task typecheck` — clean across all workspaces
- [x] `task lint` — clean (568 files, no fixes)
- [x] `bun test agent/` — 1343 pass / 1 fail; the 1 failure (`cron-handler.smoke.test.ts`) is a pre-existing full-suite-only flake, confirmed to also fail identically on clean `main` and to pass 72/72 in isolation — unrelated to this change
- [x] `worktree-reaper.unit.test.ts` and `pr-state-reconciler.unit.test.ts` pass in full (99/99)
- [ ] Manual/smoke verification via `task stack` (per AC #4's explicit test decision — not run in this sandboxed environment; new `[agent] PR state reconciler started` / worktree-reaper log lines should be confirmed on a live tick)


agent/src/index.ts
```diff
@@ -339,6 +355,18 @@ if (runtimeClient && agentId) {
         err instanceof Error ? err.message : String(err),
       );
     }
+
+    try {
+      worktreeReaperDeps ??= buildWorktreeReaperDeps({
+        scopedRepos: agentReposRef.get(),
```
dodizzle: The two sibling deps builders above (`buildPrStateReconcilerDeps`/`buildReviewStateReconcilerDeps`, lines 336/349) pass `getScopedRepos: agentReposRef.get` — the function reference, re-invoked live on every tick. This calls `.get()` immediately instead, capturing a one-time snapshot. Since `worktreeReaperDeps ??= buildWorktreeReaperDeps(...)` only runs once (first tick of the 45-min interval, for the process lifetime), that snapshot is frozen forever — any change to the agent's scoped repos after the reaper's first tick (a repo added or removed from the agent's config via `syncConfig()`, which calls `agentReposRef.set()` every 60s) is invisible to `reconcileStaleWorktrees` until the process restarts. Newly-scoped repos' worktree dirnames will fail `resolveOwningRepo`'s prefix match and get silently skipped indefinitely.

Suggestion: store `getScopedRepos: agentReposRef.get` and change `WorktreeReaperDeps.scopedRepos` to a `() => string[]` getter invoked inside `reconcileStaleWorktrees`, matching the sibling passes' pattern.

dodizzle (commented):
Verdict: COMMENT — one real bug: `scopedRepos: agentReposRef.get()` at agent/src/index.ts:361 is captured once at first-tick and frozen forever by the `??=` memoization, unlike the sibling passes which pass the live `agentReposRef.get` getter. Scope changes after the reaper's first tick will silently never be picked up until the process restarts.

dodizzle (approved):
Verdict: APPROVE — Prior review's scopedRepos-frozen-at-first-tick bug is correctly and completely fixed (live getScopedRepos() getter, matching the sibling passes), the new unit test meaningfully covers the fix, and no other issues found.

PR #2319, opened 2026-07-28
fix: prevent duplicate review posts when response parsing fails

dmcaulay:
## Summary
- Write the GitHub API response to a temp file instead of capturing in a shell variable (embedded newlines in the response JSON corrupt `echo "$var" | jq` parsing)
- Parse `html_url` from the temp file with `jq` directly
- Add explicit "never re-execute this POST" warnings to both `review.md` Step 11 and `post-review-guide.md`

## Context
This has caused duplicate COMMENT reviews on multiple PRs. When the shell-variable-based parse fails, the natural "fix" is to retry the `gh api -X POST` call — which submits another review. GitHub's API does not allow deleting or dismissing COMMENTED reviews, so the duplicates are permanent.

## Test plan
- [x] All 33 content tests pass
- [x] All 21 unit tests pass
- [x] Pattern matches the existing temp-file approach used by Step 4's `/prs/claim` call

### DIFF
--- a/plugins/shipwright/commands/review.md
+++ b/plugins/shipwright/commands/review.md
@@ -548,14 +548,14 @@ should be held; the inline comments convey the specific feedback to the author.
    POST_EXIT=0
    gh api -X POST /repos/{org}/{repo}/pulls/{pr}/reviews \
      --input $WORKSPACE_ROOT/state/reviews/pr_review_{pr}.json \
-     > "$TMPDIR/pr_post_{pr}.json" 2>&1 || POST_EXIT=$?
+     > "/tmp/pr_post_{pr}.json" 2>&1 || POST_EXIT=$?
    ```
    **Never re-execute this POST.** If parsing fails, re-parse the temp file — do not
    re-run `gh api -X POST`, which submits a duplicate review that cannot be deleted or
    dismissed (GitHub does not allow deleting/dismissing COMMENTED reviews via the API).
 2. Capture `html_url` from the temp file:
    ```bash
-   REVIEW_URL=$(jq -r '.html_url // empty' "$TMPDIR/pr_post_{pr}.json")
+   REVIEW_URL=$(jq -r '.html_url // empty' "/tmp/pr_post_{pr}.json")
    ```
 3. **Check the post succeeded** before doing anything else — non-zero exit and/or a missing/empty `REVIEW_URL` both count as failure:
    - **Success** (`POST_EXIT == 0` and `REVIEW_URL` present): continue to steps 4-6 below.
--- a/plugins/shipwright/commands/review.md
+++ b/plugins/shipwright/commands/review.md
@@ -542,15 +542,23 @@ should be held; the inline comments convey the specific feedback to the author.
 
 ### If `auto_post_reviews` is true (policy):
 
-1. Submit via GitHub API, capturing both the exit status and the response body:
+1. Submit via GitHub API, writing the response to a temp file (NOT a shell variable —
+   the response JSON contains embedded newlines that corrupt `echo "$var" | jq` parsing):
    ```bash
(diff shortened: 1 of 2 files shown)


plugins/shipwright/commands/review.md
```diff
@@ -542,15 +542,23 @@ should be held; the inline comments convey the specific feedback to the author.
...
+1. Submit via GitHub API, writing the response to a temp file (NOT a shell variable —
+   the response JSON contains embedded newlines that corrupt `echo "$var" | jq` parsing):
    ```bash
-   POST_RESPONSE=$(gh api -X POST /repos/{org}/{repo}/pulls/{pr}/reviews \
-     --input $WORKSPACE_ROOT/state/reviews/pr_review_{pr}.json)
-   POST_EXIT=$?
+   POST_EXIT=0
+   gh api -X POST /repos/{org}/{repo}/pulls/{pr}/reviews \
+     --input $WORKSPACE_ROOT/state/reviews/pr_review_{pr}.json \
+     > "$TMPDIR/pr_post_{pr}.json" 2>&1 || POST_EXIT=$?
```
dodizzle: `$TMPDIR` is unset in this repo's actual runtime (not set in `agent/Dockerfile`, `docs/configuration.md`, or any env file — verified live: `TMPDIR` is empty in the agent shell that executes this script). `"$TMPDIR/pr_post_{pr}.json"` therefore expands to `/pr_post_{pr}.json`, a root-level path a non-root process can't write (reproduced: `Permission denied`). Bash sets up redirections before running the piped command, so `gh api -X POST` never executes — `POST_EXIT` is always non-zero, the claim is released every time, and the next pass hits the same failure. This deterministically breaks all auto-posted reviews rather than fixing an intermittent bug. Step 4 (line 165, same file) uses a hardcoded `/tmp/pr_claim.json` for the identical problem — use that pattern (or `mktemp`) instead of `$TMPDIR`.

plugins/shipwright/references/post-review-guide.md
```diff
@@ -86,12 +86,22 @@ review JSON body is addressed to the PR author. Be direct -- no filler language
...
 ## Submitting the Review
 
+Write the response to a temp file — never capture in a shell variable (the response JSON
+contains embedded newlines that break `echo "$var" | jq` parsing):
+
 ```bash
 gh api -X POST /repos/{org}/{repo}/pulls/{number}/reviews \
-  --input state/reviews/pr_review_{number}.json
+  --input state/reviews/pr_review_{number}.json \
+  > "$TMPDIR/pr_post_{number}.json" 2>&1
```
dodizzle: Same unset-`$TMPDIR` issue as `review.md:551` — this guide is what Steps 9-10 of `review.md` point to, so the same infinite-retry-without-posting failure applies here too. Suggest `/tmp/pr_post_{number}.json` or `mktemp`.

dodizzle (commented):
Verdict: COMMENT — The duplicate-post fix is sound in principle, but the `$TMPDIR`-based temp file path isn't set anywhere in this repo's runtime (not in agent/Dockerfile, docs/configuration.md, or any env file) and would break 100% of auto-posted reviews rather than fixing an intermittent edge case — swap it for a hardcoded `/tmp/` path or `mktemp`, matching this same file's own existing precedent at Step 4 (`/tmp/pr_claim.json`).

dodizzle (approved):
Verdict: APPROVE — Both previously-flagged critical findings are resolved: the unset-`$TMPDIR` path in `commands/review.md:551` and `references/post-review-guide.md:95` is now a hardcoded `/tmp/pr_post_{pr}.json`, matching Step 4's real, working `/tmp/pr_claim.json` precedent. The `POST_RESPONSE`/`html_url` → `REVIEW_URL` rename is applied consistently across both call sites and the print statement, and no new issues were found.

PR #2327, opened 2026-07-28
feat: require startedAt before branch-based PR healing in pr-state-reconciler

dmcaulay:
## Summary
- Adds a hard precondition to `pr-state-reconciler.ts`'s two branch-based PR heal paths (`reconcileOrphanedTask()`, and `reconcilePrOpenTask()`'s branch-fallback path): skip healing when `task.startedAt` is null/missing, since a branch-only PR match can't confirm the specific task's own code exists in that PR — only that some task on the shared branch does.
- Adds `startedAt?: string | null` to `PrOpenTaskRecord`; it flows through automatically at all three `buildProductionDeps` task-mapping call sites (`listPrOpenTasks`, `listOrphanCandidateTasks`, `listAllTasksForBranch`) since they raw-cast the task-store JSON response.
- The `startedAt`-null check runs before any GitHub API call, and logs a message distinct from the existing BBR-1.1 "N tasks share this branch" message.

## Acceptance Criteria
| 1 | `PrOpenTaskRecord` gains `startedAt?: string \| null`; all 3 `buildProductionDeps` mapping call sites populate it | MET | Field added; the shared raw-cast helpers (`makeListTasksByStatus`/`makeListTasksByBranch`) pass it through automatically |
| 2 | `reconcileOrphanedTask()` skips (no PATCH, no GitHub call) + distinct log when `startedAt` null, checked before `ghListOpenPrsForBranch` | MET | Guard added before the GitHub call; covered by a new test |
| 3 | `reconcilePrOpenTask()` branch-fallback path applies the same guard before `ghListMergedPrsForBranch`; `task.pr`-set direct path untouched | MET | Guard added; direct path unmodified; covered by a new test |
| 4 | Both existing BBR-1.1 describe blocks extended with a single-task + `startedAt` null case; no existing test retired | MET | Two new tests added; 9 pre-existing tests updated to set `startedAt` so they keep exercising their original scenario |

## Test Plan
- `bun test agent/src/pr-state-reconciler.unit.test.ts` — 90/90 pass
- `task typecheck` — clean
- `task lint` — clean
- Full `task test` suite — clean except one documented, pre-existing, unrelated flake (`agent/src/claude.unit.test.ts`'s `extraAllowedTools` test, reproduces on clean main)


dmcaulay (commented):
Verdict: APPROVE — startedAt precondition is correctly placed before both branch-based heal paths' GitHub calls, field flow-through verified at all three buildProductionDeps call sites, tests cover both guards without retiring existing coverage. No blocking issues.

dodizzle (approved):
Verdict: APPROVE — startedAt precondition is correctly placed before both branch-based heal paths' GitHub calls; field flow-through, guard placement, new tests, and the 9 backfilled tests were all independently verified against source. No blocking issues.

PR #2328, opened 2026-07-28
feat: guard skips branch-only heal on degraded scope resolution

dodizzle:
## Summary
- `listAllTasksForBranch` now reads the `scopeDegraded` field (RSG-1.1) from the task-store's `GET /tasks` response and retries once (bounded delay) when it's `true`
- If still degraded after the retry, both BBR-1.1 guard call sites (`reconcileOrphanedTask` and `reconcilePrOpenTask`'s branch-fallback path) skip the branch-only heal and log a distinct "degraded scope resolution" message, instead of trusting a possibly under-counted sibling result
- A non-degraded response (including legitimate 0- or 1-task results) behaves identically to before — no regression to existing BBR-1.1 cases

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | listAllTasksForBranch reads the scopeDegraded field from the GET /tasks response | MET |
| 2 | On scopeDegraded:true, the call is retried once after a short bounded delay before giving up | MET |
| 3 | If still degraded after the retry, both BBR-1.1 guard call sites skip the heal and log a distinct message | MET |
| 4 | A non-degraded response produces identical behavior to the current guard | MET |
| 5 | Both BBR-1.1 test suites extended with degraded-then-skip and retry-then-succeed cases; no existing tests retired | MET |

## Test Plan
- 5 new unit tests: 3 in the `buildProductionDeps` retry/pagination suite, 1 in the branch-fallback-merge describe block, 1 in the orphan-heal describe block (plus a companion "recovers on retry" test in each of those two blocks)
- `task test`: 5374 pass (pre-existing, environment-specific flake in `metrics/src/lib/errors.unit.test.ts` confirmed reproducing identically on a clean `origin/main` clone — unrelated to this change)
- `task lint`, `task typecheck`: clean
- Coverage gate: 80.83% lines / 81.74% functions (threshold 80/80) — passed


dodizzle (commented):
Verdict: APPROVE — Bounded-retry guard on listAllTasksForBranch's scopeDegraded response, both BBR-1.1 call sites correctly skip the heal with a distinct log message on persistent degradation, retry is correctly bounded, and coverage matches all 5 stated acceptance criteria. Posted as COMMENT since this is a self-authored PR (GitHub blocks self-APPROVE via the API).

PR #2336, opened 2026-07-29
fix: use origin/$base in review/patch diff commands to avoid stale local refs

dmcaulay:
## Summary

- The review and patch commands diff against `$base` (the bare branch name from GitHub's `baseRefName`), but worktrees only run `git fetch origin` — the local `main` branch is never updated. When it falls behind, `git merge-base main HEAD` picks a stale merge-base, producing a diff that includes unrelated commits from other merged PRs.
- Observed in production: a PR with 4 changed files produced a 68-file, 6.5k-line diff, causing the reviewer to approve based on the wrong diff.
- All `git diff`/`git show` commands now use `origin/$base` so the merge-base is always computed against the freshly-fetched remote-tracking ref.
- Added a sanity check in Step 5.3 that compares the diff file count against the PR metadata's `changedFiles` and stops if they diverge significantly.

## Test plan

- [x] `bun test commands/review.content.test.ts` — 33/33 pass
- [ ] Run `/shipwright:review` against a PR and verify the diff file count matches the PR metadata

### DIFF
--- a/plugins/shipwright/commands/patch.md
+++ b/plugins/shipwright/commands/patch.md
@@ -608,7 +608,7 @@ From inside the worktree, collect the full picture of what needs fixing:
 1. **PR diff against base**:
    ```bash
    base=$(gh pr view {pr} --repo {org}/{repo} --json baseRefName -q '.baseRefName')
-   git diff "$base"...HEAD
+   git diff "origin/$base"...HEAD
    ```
 
 2. **Unresolved inline threads** (from Step 3a — already fetched, reuse):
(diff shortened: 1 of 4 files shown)


dodizzle (approved):
Verdict: APPROVE — Correctly fixes the stale-local-ref diff bug (origin/$base instead of bare $base); internally consistent across review.md, patch.md, and post-review-guide.md, with a useful changedFiles sanity check added and the content test kept in sync. No blocking findings.

PR #2352, opened 2026-07-29
fix: retry bump-chart on branch collision instead of silent skip (CBR-2.1)

dmcaulay:
## Summary
- Rewrites the `bump-chart` job's branch-collision handling: on a `chore/chart-v{version}` branch-name collision, the job now fetches fresh `main`, re-derives its tag batch and version, and retries — bounded to 5 attempts with a 15s backoff — instead of silently skipping and dropping the whole batch of release tags.
- If every retry attempt still collides, the job now fails loudly (non-zero exit) with an explicit error naming the release tag(s) that could not be pinned.
- Consolidates the "Collect release tags" / "Read current chart version" / "Compute new chart version" / "Check for existing bump branch" steps into a single `Resolve chart version bump (retry on branch collision)` step, and removes the `if: steps.idempotency.outputs.skip != 'true'` gating from all downstream steps (now always run once resolution succeeds).
- `test-auto-bump-chart.sh` gains `compute_batch_and_version`, `branch_exists_on_remote`, and `resolve_batch_with_retry` bash functions plus fixture-backed tests for: (a) a retry that recomputes a non-colliding version and correctly re-derives its tag batch from a fresh since-timestamp (excluding already-landed tags), and (b) the exhausted-retries path exiting non-zero with a greppable error naming the un-pinned tags. Existing `collect_batch_tags`/`select_pinned_tags`/`update_values_tag` tests and contracts are unchanged.
- `docs/helm-repo.md` refreshed to document the new collision-retry behavior.

## Acceptance Criteria

| 1 | Retry-on-collision with fresh main, bounded attempts + backoff | MET | New "Resolve chart version bump (retry on branch collision)" step, MAX_ATTEMPTS=5, BACKOFF_SECONDS=15, re-fetches/resets to origin/main and recomputes tags/version/branch each attempt |
| 2 | Fail loudly on exhaustion, naming un-pinned tags | MET | `exit 1` with `ERROR: exhausted 5 attempts... Release tag(s) that could NOT be pinned: ${TAGS_CSV}` |
| 3 | test-auto-bump-chart.sh gains coverage for retry-success and exhaustion; existing tests unchanged | MET | New `compute_batch_and_version`/`branch_exists_on_remote`/`resolve_batch_with_retry` + fixtures; 69/69 pass (61 pre-existing + 8 new) |
| 4 | PR description documents manual test plan for collision-retry | MET | See below |

## Manual Test Plan — Collision Retry

This scenario cannot be fully exercised in CI without real concurrent tag pushes + secrets — same documented limitation as the workflow's existing debounce manual-test-plan.

1. Note the current chart version in `charts/shipwright/Chart.yaml` (e.g. `X.Y.Z`), and manually pre-create + push the branch the *next* bump would use, simulating a concurrent run that already landed:
   ```
   git checkout -b chore/chart-vX.Y.(Z+1)
   git push origin chore/chart-vX.Y.(Z+1)
   git checkout main
   ```
2. Push a dummy tag to trigger a real run: `git tag agent-v9.9.7 && git push origin agent-v9.9.7`
3. In the "Resolve chart version bump (retry on branch collision)" step's logs, confirm it detects the collision on attempt 1 ("Branch ... already exists on remote"), fetches fresh main, and recomputes.
4. If all 5 attempts exhaust: confirm the job fails (non-zero exit) and the log contains a line starting with `ERROR: exhausted 5 attempts` naming the release tag(s) (e.g. `agent-v9.9.7`) that could not be pinned.
5. If a later attempt succeeds (branch deleted mid-retry): confirm the PR that opens targets the *recomputed* version/branch and still credits `agent-v9.9.7`.
6. Clean up: `git push origin --delete agent-v9.9.7 chore/chart-vX.Y.(Z+1)`

## Test Plan
- [x] `bash .github/workflows/test-auto-bump-chart.sh` — 69/69 passing (61 pre-existing + 8 new)
- [x] `bun run --filter='*' --sequential typecheck` — clean (no TS files touched by this PR, ran as a sanity check)
- [x] YAML validity of `auto-bump-chart.yml` confirmed via `python3 -c "import yaml; yaml.safe_load(...)"`
- [x] One pre-existing, unrelated full-suite flake (`task-store/src/routes/prs.openapi.smoke.test.ts`) confirmed reproducing on clean `main` — not caused by this diff

### DIFF
--- a/.github/workflows/test-auto-bump-chart.sh
+++ b/.github/workflows/test-auto-bump-chart.sh
@@ -345,12 +345,16 @@ collect_batch_tags() {
 # instead of returning success having dropped them.
 # ---------------------------------------------------------------------------
 
-# Computes the {tags_csv, new_version, branch} triple from the current
-# working-directory repo state. Mirrors chaining the "Collect release tags
-# in this batch" + "Read current chart version" + "Compute new chart
-# version" steps into one reusable unit, so the retry loop below can re-run
-# it wholesale after re-syncing to fresh main. Emits "tags_csv|version|branch"
-# on stdout (single line — tag values and versions never contain '|').
+# Computes the {tags_csv, current_version, new_version, branch} quadruple
+# from the current working-directory repo state. Mirrors chaining the
+# "Collect release tags in this batch" + "Read current chart version" +
+# "Compute new chart version" steps into one reusable unit, so the retry
+# loop below can re-run it wholesale after re-syncing to fresh main. Emits
+# "tags_csv|current_version|new_version|branch" on stdout (single line — tag
+# values and versions never contain '|'). Matches the real workflow's
+# compute_batch_and_version() 4-field output shape exactly (auto-bump-chart.yml)
+# — steps.resolve.outputs.current_version is consumed downstream by the
+# "Open bump PR" step's PR body.
 compute_batch_and_version() {
   local trigger_tag="$1"
   local chart_yaml="${2:-charts/shipwright/Chart.yaml}"
@@ -363,7 +367,7 @@ compute_batch_and_version() {
   new_version=$(bump_patch "$current_version")
   branch="chore/chart-v${new_version}"
 
-  printf '%s|%s|%s\n' "$tags_csv" "$new_version" "$branch"
+  printf '%s|%s|%s|%s\n' "$tags_csv" "$current_version" "$new_version" "$branch"
 }
 
 # Checks whether `branch` exists as a head ref in `remote_dir` (a bare or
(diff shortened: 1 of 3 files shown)


.github/workflows/test-auto-bump-chart.sh
```diff
@@ -325,6 +325,108 @@ collect_batch_tags() {
...
+
+  local tags_csv
+  tags_csv=$(collect_batch_tags "$trigger_tag")
+
+  local current_version new_version branch
+  current_version=$(read_chart_version "$chart_yaml")
+  new_version=$(bump_patch "$current_version")
+  branch="chore/chart-v${new_version}"
+
+  printf '%s|%s|%s\n' "$tags_csv" "$new_version" "$branch"
```
dmcaulay: This mirror of `compute_batch_and_version()` returns a 3-field tuple (`tags_csv|new_version|branch`), but the workflow's inline version (`auto-bump-chart.yml:265`) returns 4 fields (`tags_csv|current_version|new_version|branch`) — `current_version` is computed here (line 362) but never included in the output, and `resolve_batch_with_retry()` never threads or asserts on it. `steps.resolve.outputs.current_version` is consumed directly by the "Open bump PR" step's PR body, so this field is a real, unexercised divergence from the workflow it's meant to mirror. Worth adding to the tuple + an assertion, to match the 4-field shape and keep the two copies genuinely in sync as the comments here claim.

dmcaulay (commented):
Verdict: COMMENT — Core retry/collision/fail-loud logic is sound and well-tested; one gap: the test script's `compute_batch_and_version()` mirror drops the `current_version` field from its output tuple, even though the workflow's real step surfaces it as `steps.resolve.outputs.current_version` and the "Open bump PR" step depends on it. Not a live bug (the underlying `read_chart_version` logic is independently unit-tested), but it's an unexercised divergence in exactly the field a downstream step relies on, despite this PR's own comments claiming the two copies mirror each other exactly.

dmcaulay (commented):
Verdict: APPROVE — Prior finding (test mirror's `compute_batch_and_version()` dropping the `current_version` field) is fully and correctly resolved by 78ef644b: the 4-field tuple now matches the real workflow's shape, threading and assertions are correct against their fixtures. No other issues found.

PR #2354, opened 2026-07-29
docs: refresh stale references

dmcaulay:
## Summary

Auto-mode doc-refresh pass over the shipwright docs candidates listed for this cycle. Verified every extracted reference (file paths, symbols, generated content) against current source in this worktree; three docs had genuine drift, the rest checked out current.

**Updated:**

- **`docs/architecture.md`** — the References bullet listed `task-store.md` as living under `plugins/shipwright/references/`, but that file doesn't exist there (task-store docs are actually at `docs/task-store.md`, a different top-level doc). Swapped the example for `principles.md`, which does exist under `plugins/shipwright/references/`.
- **`docs/mcp-tools.md`** — this is a generated file (`bun run generate:mcp-docs`, sourced from `mcp-server/src/generated-tools.ts` ← `task-store/openapi.json`). The committed version still listed a `hitl` query param on `tasks_list` that `generated-tools.ts` no longer has (already regenerated on `main` at some point without the doc following). Regenerated it via the real generator script — the diff is exactly the stale `hitl` entry being dropped.
- **`docs/observability.md`** — pointed readers at `SECRET_ENV_VARS` "in `lib/sentry.ts`", but that constant is actually defined in `lib/secret-env-vars.ts` and only imported into `sentry.ts`. Fixed the file pointer.

**Checked, current, left untouched:** `docs/agent-types.md`, `docs/agent.md`, `docs/chat.md`, `docs/configuration.md`, `docs/helm-repo.md`, `docs/metrics.md`, `docs/migration.md`, `docs/task-store.md`, `docs/testing.md` (test-system.md mtime special-case does not trigger — testing.md is the newer of the two).

`docs/test-readiness/` was not touched (read-only, owned by a different plugin). `CLAUDE.md`'s Reference section already has accurate entries for every doc touched here — no scope changes warranted an update there.

## Test plan

- [x] Every changed line verified against live source in this worktree (file existence via `Glob`/`test -f`, symbol/const location via `Grep`)
- [x] `docs/mcp-tools.md` diff produced by running the actual generator (`bun run generate:mcp-docs`), not hand-edited
- [x] No changes under `docs/test-readiness/`
- [x] `git diff --stat` shows only the 3 intended files, 3 lines total


dodizzle (approved):
Verdict: APPROVE — Three verified-accurate stale-reference fixes in docs/architecture.md, docs/mcp-tools.md, and docs/observability.md. No issues found.

PR #2355, opened 2026-07-29
feat: add periodic chart-drift-check workflow

dodizzle:
## Summary
- Adds `.github/workflows/check-chart-drift.yml`, a daily-scheduled (+ `workflow_dispatch`) workflow that catches chart-pin drift: for each service (admin, metrics, agent, task-store, chat) it finds the highest-semver `{service}-v*` git tag and compares it against the tag currently pinned at that service's path(s) in `charts/shipwright/values.yaml`, failing the job (with a `MISMATCH: service=... path=... expected=... actual=...` line per drifted path) if anything is out of sync.
- Extracts the `service_for_tag` / `values_paths_for_service` / `highest_semver_tag` logic — previously inlined separately in `auto-bump-chart.yml` and mirrored in `test-auto-bump-chart.sh` — verbatim into a new shared, sourceable file: `.github/workflows/lib/chart-tag-utils.sh`.
- Adds `.github/workflows/lib/chart-drift-check.sh` with the new, read-only drift-comparison logic (`read_pinned_tag`, `compute_drift`), and `.github/workflows/test-check-chart-drift.sh` — a new companion pure-bash test file (40 assertions, no network/no yq/no python3) following the exact same pattern as `test-auto-bump-chart.sh`.

**Why a new shared-lib file instead of sourcing `test-auto-bump-chart.sh` directly:** `test-auto-bump-chart.sh` has no `if [[ "${BASH_SOURCE[0]}" == "$0" ]]` main-guard — it runs its entire assertion suite unconditionally top-to-bottom and calls `exit 1` at the bottom on failure. Sourcing it from a production workflow step would (a) run the whole existing test suite as an unwanted side effect and (b) its trailing `exit` would kill the parent shell outright, even on success. `lib/chart-tag-utils.sh` instead holds only the three function definitions, verbatim, with no side effects when sourced. `auto-bump-chart.yml` and `test-auto-bump-chart.sh` are left completely untouched — this PR only adds new files.

## Acceptance Criteria

| 1 | New workflow runs daily + on `workflow_dispatch`; compares highest-semver `{service}-v*` tag vs. values.yaml pin per service | MET | `check-chart-drift.yml`: `schedule` (daily 09:00 UTC) + `workflow_dispatch`, loops over all 5 services |
| 2 | Job fails non-zero with service/expected/actual message; no new notification infra | MET | `exit 1` with one `MISMATCH:` line per drifted path; failure is visible natively in the Actions tab |
| 3 | Reuses existing tag-parsing logic rather than re-implementing | MET | `service_for_tag`/`values_paths_for_service`/`highest_semver_tag` extracted verbatim into `lib/chart-tag-utils.sh`, sourced by the new workflow |
| 4 | Test coverage for the new comparison logic, pure-bash pattern | MET | `test-check-chart-drift.sh`, 40 assertions, all passing |
| 5 | No new `${{ secrets.* }}` reference; default `GITHUB_TOKEN` read scope only | MET | `permissions: contents: read` only; no `secrets.*` anywhere in the new workflow |

## Test Plan
- `bash .github/workflows/test-check-chart-drift.sh` — 40/40 assertions pass
- `task check-strings` — no banned strings
- `task ci` — passes except one **pre-existing, unrelated** failure in `agent/src/claude.unit.test.ts` ("empty extraAllowedTools produces identical args") caused by an `AGENT_ALLOWED_TOOLS` env var leaking into that unit test in this sandbox; reproduced identically against an unrelated worktree on `main` with no changes from this PR. This PR touches no bun-managed package, so it has no effect on the coverage gate.
- Manually verified the new workflow's YAML parses correctly and its comparison logic correctly flags real, pre-existing drift already present on `main` today (several services' `values.yaml` pins currently lag their latest release tags) — confirming the detection logic works, though fixing that pre-existing drift is out of scope for this task.

### DIFF
--- a/.github/workflows/check-chart-drift.yml
+++ b/.github/workflows/check-chart-drift.yml
@@ -1,0 +1,99 @@
+name: Chart Drift Check
+
+# Periodic drift-check: catches a future recurrence of chart-pin drift (from a
+# race condition in auto-bump-chart.yml's debounce/burst handling, or any
+# other cause) between the highest-semver release tag for each Shipwright
+# service and what's actually pinned in charts/shipwright/values.yaml, without
+# requiring manual git archaeology to discover it.
+#
+# For each service (admin, metrics, agent, task-store, chat) this workflow:
+#   1. Finds the highest-semver `{service}-v*` git tag.
+#   2. Reads the tag currently pinned at that service's values.yaml path(s).
+#   3. Compares them. Any mismatch fails the job — the job failure IS the
+#      alert (visible in the Actions tab / a red X on the default branch).
+#      No new notification infra, no new secrets: only the default
+#      GITHUB_TOKEN read scope (contents: read) already used elsewhere in
+#      this repo's workflows.
+#
+# Reuses the SAME service_for_tag / values_paths_for_service /
+# highest_semver_tag logic that auto-bump-chart.yml uses to pin tags in the
+# first place — extracted to .github/workflows/lib/chart-tag-utils.sh so both
+# workflows share one implementation instead of drifting apart themselves.
+#
+# Deliberately does NOT source .github/workflows/test-auto-bump-chart.sh
+# directly: that file has no `if [[ "${BASH_SOURCE[0]}" == "$0" ]]`
+# main-guard, so sourcing it would run its entire existing assertion suite as
+# a side effect and its trailing `exit 1` (on failure) would kill this step's
+# shell outright even on unrelated failures. lib/chart-tag-utils.sh instead
+# holds ONLY the function definitions, extracted verbatim, safe to source.
+#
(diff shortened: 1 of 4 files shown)


dodizzle (commented):
Verdict: APPROVE — Clean, well-tested ops/CI PR. Tag-parsing extraction into `chart-tag-utils.sh` was verified byte-for-byte identical to the still-inlined copies in `auto-bump-chart.yml`, and the awk-based `read_pinned_tag` parser was hand-traced against the real `charts/shipwright/values.yaml` structure (including the nested `agent.provisioning.image.tag` vs top-level `agent.image.tag` disambiguation). No secrets, no breaking changes, CI green. All five stated acceptance criteria met.

PR #2358, opened 2026-07-29
fix: write reviewedCommitSha from review.md

dmcaulay:
## Summary
- `review.md` now writes `reviewedCommitSha` (in addition to the shared claim-lock `commitSha`) at every point it records "the commit review just processed": Step 11's two staging PATCH calls and the unresolved-comment-skip PATCH.
- Step 14.2's stale-staged-review detection and defense-in-depth dedup now read/compare `record.reviewedCommitSha` instead of `record.commitSha`, so review dedup no longer collides with the claim-lock field shared across review/patch/deploy.
- The initial `/prs/claim` call's `commitSha` argument (Step 4) and the re-claim call inside the stale-refresh path are untouched — both use the shared claim-lock endpoint, which only accepts `commitSha`.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Step 11's staging PATCH calls set reviewedCommitSha alongside staged/reviewState | MET |
| 2 | The unresolved-comment-skip PATCH sets reviewedCommitSha alongside commitSha | MET |
| 3 | Step 14.2's stale-staged-review detection reads/compares record.reviewedCommitSha instead of record.commitSha throughout | MET |
| 4 | The initial /prs/claim call's commitSha argument is unchanged | MET |
| 5 | Test decision: extend existing *.content.test.ts coverage; no existing tests retired | MET |

## Test Plan
- Extended `review.content.test.ts` with a new describe block (6 tests) asserting `reviewedCommitSha` appears in the Step 11/Step 5 PATCH payloads and that Step 14's comparisons use `record.reviewedCommitSha`, not `record.commitSha`; confirmed RED against the pre-fix file, GREEN after.
- Updated one pre-existing assertion in `review.unit.test.ts` that expected `record.commitSha` in the same dedup section this task changed.
- [x] `task ci` — all suites pass except one documented pre-existing sandbox flake (`extraAllowedTools` in `agent/src/claude.unit.test.ts`), reproduced independently of this diff on a clean checkout.

### DIFF
--- a/plugins/shipwright/CLAUDE.md
+++ b/plugins/shipwright/CLAUDE.md
@@ -33,19 +33,22 @@ Applies to: **review**, **deploy**, **patch**, **dev-task**
 ### 2. Task Store PR Record Is the Dedup Source
 
 The task store `PullRequest` record (accessible via `GET /prs?repo=X`) is the source of
-truth for review deduplication. `commitSha` on the record is compared to the current
+truth for review deduplication. `reviewedCommitSha` on the record is compared to the current
 `headRefOid` from GitHub to detect "new commits since last review" without re-fetching
-history. Skills must not fail or behave incorrectly when no PR record exists — a missing
-record means the PR has not been reviewed yet and should be treated as eligible.
+history. `commitSha` is a separate claim-lock field — it is overwritten on every claim
+(including by later `patch`/`deploy` claims on the same PR) and is not safe to use for
+review dedup. Skills must not fail or behave incorrectly when no PR record exists — a
+missing record means the PR has not been reviewed yet and should be treated as eligible.
 
 Local files (`state/reviews/PR_REVIEW_{pr}.md`, `state/reviews/pr_review_{pr}.json`) hold
 the review narrative (findings, verdict, inline comments) for posting to GitHub. They are
 written by `/shipwright:review` and consumed by `/shipwright:review-staged`. They are not
 dedup state — the task store record owns that.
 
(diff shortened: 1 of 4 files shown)


dmcaulay (commented):
Verdict: COMMENT — The diff itself is correct and well-tested (all 4 reviewedCommitSha PATCH sites and both Step 14 comparison paths verified consistent, no leftover record.commitSha refs). One important gap: plugins/shipwright/CLAUDE.md is this plugin's declared Design Constitution and requires "a corresponding update to the principle itself" for any change that contradicts a principle. Principle #2 still says review dedups by `commitSha`, but this PR moves review.md's dedup to `record.reviewedCommitSha` throughout — the constitution text wasn't updated to match. Worth a quick follow-up to Principle #2's text. Non-blocking note: `review-staged/SKILL.md` (~line 92) and `agent/src/check-review.ts` still read the old `commitSha` field for their own staleness/dedup checks — same class of bug, pre-existing on main, out of scope here, but worth a follow-up task since `commitSha` is confirmed mutable by later patch/deploy claims on the same PR.

dodizzle (approved):
Verdict: APPROVE — Complete, internally consistent migration of review-dedup from the shared `commitSha` claim-lock field to the dedicated `reviewedCommitSha` field across all required call sites; the constitution text (Principle #2) is updated to match, and the prior blocking finding on that gap is now fixed. New content tests target the specific replacements. No blocking issues.

PR #2383, opened 2026-07-31
fix: clear hitl flag and guard against silent PR in Step 6b

dmcaulay:
## Summary
- Fix `/shipwright:hitl` Step 6b to always PATCH `hitl: false` alongside `status: "done"` when marking a task done, so completed HITL tasks stop showing `hitl: true` forever (proven live: RBV-V.1 is `merged` and fully shipped but still `hitl: true`).
- Add a guard: if `task.pr` is already set when Step 6 runs (a PR resulted from the HITL session), warn the human and skip the done-marking PATCH entirely — only clear `hitl` so the task re-enters normal review/patch/deploy candidacy instead of being silently marked done with an unmerged PR attached.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Step 6b PATCHes hitl:false whenever it marks the task done | ✅ MET |
| 2 | If task.pr is set, warn and don't mark status:'done' — only clear hitl | ✅ MET |
| 3 | Extend hitl.content.test.ts to assert the hitl:false PATCH and task.pr guard language | ✅ MET |

## Test Plan
- Extended `hitl.content.test.ts` with 5 new tests (2 describe blocks) asserting the Step 6b markdown includes `hitl: false` in the done-marking PATCH, and Step 6 documents the `task.pr` guard/warning/status-unchanged behavior.
- `bun test plugins/shipwright/commands/hitl.content.test.ts` — 22/22 pass.
- `task typecheck` — all workspaces pass.
- `task ci` — full suite passes except one pre-existing, diff-unrelated sandbox flake (`agent/src/claude.unit.test.ts` extraAllowedTools), confirmed to also fail on clean main.

### DIFF
--- a/plugins/shipwright/commands/hitl.content.test.ts
+++ b/plugins/shipwright/commands/hitl.content.test.ts
@@ -24,6 +24,12 @@ function extractSuppressionSubStep(md: string): string {
   return match?.[0] ?? "";
 }
 
+function extractMarkTaskDoneSubStep(md: string): string {
+  const match = md.match(/### 6b\. Mark Task Done[\s\S]*$/);
+  expect(match).not.toBeNull();
+  return match?.[0] ?? "";
+}
+
 describe("hitl.md — gitleaksignore suppression sub-step exists (GLB-2.2 AC1)", () => {
   it("contains a Step 6a suppression sub-step section", () => {
     expect(content).toContain("### 6a. Offer Gitleaksignore Suppression");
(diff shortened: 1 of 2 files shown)


dmcaulay (commented):
Verdict: APPROVE — hitl.md now always clears `hitl` on done-marking and correctly guards against silently marking a task done when a PR resulted from the session. Verified `task.pr` is a real task-store field and PATCH is a genuine partial update, so scoping the PR-branch PATCH to `{hitl: false}` alone is safe. Test coverage matches all three acceptance criteria (22/22 pass). No blocking issues.

PR #2393, opened 2026-07-31
feat: extend TaskService list and distinct for multi-repo and org filters

dmcaulay:
## Summary
- Generalize `TaskService.list()`'s `repo` filter from a single exact-match string to `string | string[]`, keeping the original exact-match shape for the single-string case (back-compat), and add a new `org: string | string[]` filter using the `buildRepoOrgWhere` helper from ORF-1.1.
- Extend `TaskService.distinct()` to return a new `orgs: string[]` field, derived by splitting each distinct repo string on `/`, deduping, and sorting.
- Add a unit test covering an edge case introduced by the org/agentScope interaction: when both `agentScope` and `org` produce an `OR` clause, they're merged under `where.AND` instead of one silently clobbering the other.
- **Docs note**: this task only extends `TaskService.list()`/`distinct()` at the internal layer — the public `GET /tasks` HTTP route is unchanged and does not yet parse an `?org=` query param or accept `?repo=` as a list. `docs/task-store.md` and `docs/migration.md` are updated to reflect this scope precisely (the `GET /tasks/distinct` endpoint's new `orgs` field IS live over HTTP, since that route passes the service result straight through).

## Acceptance Criteria
| 1 | Existing single-value repo exact-match unchanged, still covered | MET | `where.repo = filters.repo` preserved for string input; original test at task-service.unit.test.ts:585 still passes |
| 2 | New unit test: `list({ repo: ['org/a','org/b'] })` → `where.repo.in` matching both | MET | task-service.unit.test.ts |
| 3 | New unit test: `list({ org: 'app-vitals' })` → `where.repo.startsWith('app-vitals/')` | MET | task-service.unit.test.ts |
| 4 | New integration test seeding multi-repo/org rows, filtering repo-list + org intersecting | MET | tasks.integration.test.ts |
| 5 | New unit test: `distinct()` includes deduped/sorted `orgs` field | MET | task-service.unit.test.ts |
| 6 | Test decision: additive, no existing tests retired | MET | Only additions; 40/40 pre-existing unit tests pass |

## Test Plan
- `bun test task-store/src/task-service.unit.test.ts` — 40/40 pass
- Full monorepo `task ci`: 5520 pass / 1 pre-existing unrelated flake (`agent/src/claude.unit.test.ts` `extraAllowedTools` — untouched by this diff, documented sandbox flake) / 381 skip
- Coverage gate: 81.14% lines / 81.97% functions (threshold 80/80) — passed

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -4,19 +4,23 @@ Durable notes for breaking changes and the steps needed to migrate across versio
 
 ---
 
-## New: `GET /tasks` `?org` filter and distinct `orgs` field
+## New: `TaskService.list()` multi-repo/org filters and distinct `orgs` field
 
 **Version**: next (ORF-1.2)
 
-The task-store API now supports filtering tasks by organization prefix and returns distinct organizations in the distinct endpoint:
+`TaskService.list()` now accepts `repo` as `string | string[]` and a new `org: string | string[]`
+filter (repo prefix match, e.g. `"app-vitals"` matches any `repo` starting with `"app-vitals/"`).
+`TaskService.distinct()` now also returns a third field, `orgs: string[]` — the distinct
+organization prefixes extracted from the `repos` it already returns.
 
-- **`GET /tasks?org=<org>`** — filters tasks to only those whose `repo` field starts with `"<org>/"`. Combines with an existing `?repo=` filter as an AND condition (both narrow the result).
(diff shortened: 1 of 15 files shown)


dmcaulay (commented):
Verdict: APPROVE — Clean, well-tested extension of TaskService.list()/distinct() for multi-repo/org filters; where-clause construction verified correct across all filter combinations, no breaking changes, docs accurately scope HTTP vs in-process behavior.

PR #2404, opened 2026-07-31
feat: persist session id on failure so Slack threads can resume after a timeout/error

dodizzle:
## Summary
- Add `_saveSessionFromError(sessionKey, err)` to `agent/src/claude.ts`, which extracts `.sessionId` from `ClaudeRunError`/`ClaudeTimeoutError` (added in CSI-1.1) and persists it via `sessions.set()`
- Call it in `_runClaude`'s outer catch for the initial attempt's failure (before the retry-condition check), and again in the retry's own catch if the retry also fails — so the mapping always points at the most-recently-known session id, and the originally-thrown error is still rethrown unchanged
- Fixes a bug where a first-message Slack thread timeout/error never persisted the session, so replies in the same thread started a brand-new, context-free session instead of resuming — `chat-poller.ts` and `slack.ts` share this runner and get the fix for free

## Acceptance Criteria

| 1 | `_saveSessionFromError` extracts `.sessionId` from `ClaudeRunError`/`ClaudeTimeoutError` and calls `sessions.set` when both present | MET | `claude.ts` `_saveSessionFromError()` |
| 2 | Called in outer catch before retry-condition check | MET | `claude.ts` outer `catch (err)` block |
| 3 | Called again if retry attempt also fails | MET | inner `catch (retryErr)` block |
| 4 | Original error still rethrown, only side effect added | MET | `throw err` preserved in both paths |
| 5 | Test decision: unit cases (a)(b)(c) + integration test, no existing tests retired | MET | see Test Plan below |

## Test Plan
- `agent/src/claude.unit.test.ts` — new `describe("session persistence on failure")`: (a) immediate timeout with no existing session still saves the captured id, (b) resumed-session retry that also times out saves the retry's session id while the original error is still rethrown, (c) no sessionId ever captured (crash before stdout) does not call `sessions.set`
- One pre-existing test (`does not save session when is_error=true`) updated in place — not retired — to assert the new, intentionally-changed behavior (a `ClaudeRunError` on `is_error: true` is exactly the case this feature fixes)
- `agent/src/slack.integration.test.ts` — new case wiring the *real* `createRunClaude` (fake spawner + real in-memory session store): first DM message times out, second message in the same thread asserts `-r <sessionId>` is passed on the second spawn call
- [x] `task ci` passes — lint, typecheck, check-config-docs, check-version-sync, secret-scan all green
- [x] Coverage gate passes — `agent/src/claude.ts` at 100% lines
- Note: `task test:coverage` has one pre-existing failure (`extraAllowedTools > empty extraAllowedTools produces identical args to no extraAllowedTools`) unrelated to this change — confirmed by reproducing the identical failure on a fresh `origin/main` checkout. It's a test-order-dependent global-state pollution issue (the exact `Bun shares the test process` risk called out in this repo's own `CLAUDE.md`), not something introduced here.


agent/src/claude.ts
```diff
@@ -579,7 +605,10 @@ export function createRunClaude(
           const output = await _spawn(args, perCallOnProgress);
           await _saveSession(sessionKey, output);
           return { ...output, recoveredFromError: true };
-        } catch {
+        } catch (retryErr) {
+          // The retry also failed — overwrite with its own most-recently-known
+          // session id (if any), then still rethrow the ORIGINAL error.
+          await _saveSessionFromError(sessionKey, retryErr);
```
dodizzle: If `sessions.set()` throws here (the real `createFileSessionStore` does an unguarded `writeFile()` — disk-full, permissions, etc.), this exception replaces `err` and `sentryClient?.captureException(err)` on the next line never runs — silently dropping telemetry for the real failure and breaking the PR's own stated invariant that the original error is "rethrown unchanged." Suggest wrapping `_saveSessionFromError` in its own try/catch that swallows persistence failures.

agent/src/claude.ts
```diff
@@ -566,6 +587,11 @@ export function createRunClaude(
       await _saveSession(sessionKey, output);
       return output;
     } catch (err) {
+      // The initial attempt failed — save whatever session id it captured
+      // before deciding whether to retry, so even a non-retried (or
+      // retry-ineligible) failure still leaves a resumable mapping behind.
+      await _saveSessionFromError(sessionKey, err);
```
dodizzle: Same masking risk here: if `_saveSessionFromError` throws, the guaranteed `throw err` a few lines down never executes, and a persistence failure silently replaces the original Claude error. Worth the same try/catch guard as the retry-catch call site.

dodizzle (commented):
Verdict: COMMENT — Awaiting sessions.set() inside the new error-handling paths can mask the original Claude error and drop Sentry reporting if the write fails; see inline comments. Core feature logic and test coverage are otherwise solid.

dodizzle (commented):
Verdict: APPROVE — Correctly and completely fixes the previously-flagged persistence-masking issue: both `_saveSessionFromError` call sites are wrapped in try/catch so a `sessions.set()` failure can never mask the original Claude error or skip the Sentry capture/rethrow. No regressions, no breaking changes, tests follow the repo's injected-doubles isolation convention.

PR #2407, opened 2026-07-31
feat: render sessionId in the admin cron-logs table

dmcaulay:
## Summary
- Add `sessionId?: string | null` to the `CronRunItem` interface in `admin/src/admin-ui-pages.ts`
- Render a truncated (first 8 chars), monospace session id in the per-agent cron-logs table, with the full id in a `title=` tooltip — em-dash when null/undefined/empty, matching the existing started-at cell convention
- No changes required in `admin-ui.ts` — `sessionId` flows through automatically once the interface and row renderer are updated

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | CronRunItem interface gains sessionId?: string \| null | MET |
| 2 | row() renders truncated monospace session id with title= tooltip, em-dash when null | MET |
| 3 | No changes required in admin-ui.ts | MET |
| 4 | Extend existing test to cover run with/without sessionId | MET |

## Test Plan
- Extended `admin/src/admin-ui-pages.unit.test.ts` with new cases: Session column header, truncated value + tooltip, em-dash for null/undefined/empty sessionId, column count, XSS escaping in the title attribute
- `bun test src/admin-ui-pages.unit.test.ts` — 540 pass, 0 fail
- `task lint` / `task typecheck` — clean

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -202,6 +202,11 @@ export interface CronRunItem {
   itemType?: string | null;
   itemId?: string | null;
   /**
+   * The Claude session id for this cron run. Present when the run was
+   * dispatched with a session context; null/absent for runs with no session.
+   */
+  sessionId?: string | null;
+  /**
    * The owning cron's id/name/schedule — present on cross-cron listings (e.g.
    * renderCronLogsPage's per-agent table) so the Cron column can be rendered
    * without an N+1 lookup. Absent on single-cron listings.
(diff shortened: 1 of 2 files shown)


dmcaulay (commented):
Verdict: APPROVE — Small, well-scoped addition: sessionId flows through automatically via the existing `include:` query, escaping is correct on the title tooltip, and the colspan/header bump is applied consistently. No blocking concerns.

PR #2419, opened 2026-07-31
feat: fetch inline review threads and gate APPROVE on unresolved feedback

dodizzle:
## Summary
- Step 5.5 of `review.md` now fetches reviews, comments, **and inline review threads** (with `isResolved` state) via a single GraphQL call, replacing the REST `gh pr view --json comments,reviews` call that was blind to inline diff-line feedback.
- Step 5's Unresolved Comment Check now treats an unresolved inline thread (non-bot, not `CURRENT_USER`) the same as substantive unresolved top-level feedback.
- New **Step 9.5: Unaddressed-Findings Hard Gate**, placed immediately before Step 10, mirrors `patch.md`'s Step 3a "unaddressed findings" definition (qualifying COMMENTED/CHANGES_REQUESTED review at head, excluding clean self-APPROVE and author-reply-addressed reviews, OR any unresolved inline thread) and forces `event`/`Verdict: ...` to `COMMENT` when true — overriding the code-reviewer subagent's recommendation and independent of self-review status.
- Review Quality Rules' "Check for unresolved feedback first" bullet updated to describe this as mechanically enforced, not just a soft guideline.

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Step 5.5 fetches reviews/comments/reviewThreads via single GraphQL call, replacing REST call | MET |
| 2 | Unresolved Comment Check treats unresolved inline threads (non-bot, not CURRENT_USER) same as substantive feedback | MET |
| 3 | Hard gate mirrors patch.md Step 3a, forces verdict=COMMENT for both `event` and body label, overrides subagent + self-review status | MET |
| 4 | Review Quality Rules bullet updated to describe mechanical enforcement | MET |
| 5 | Content tests added confirming (a)-(d), no existing tests retired | MET |

## Test Plan
- Added 14 new content-test assertions to `review.content.test.ts` (TDD: written failing first, then implementation made them pass)
- `task test` — full suite passes except one pre-existing, unrelated failure in `agent/src/claude.unit.test.ts` (`extraAllowedTools`) — confirmed unrelated: that file was last touched by PR #2404, and this branch's diff never touches `agent/src/`
- `task lint`, `task typecheck` — clean
- Independent spec-compliance subagent review: all 5 criteria MET

### DIFF
--- a/plugins/shipwright/commands/review.md
+++ b/plugins/shipwright/commands/review.md
@@ -209,10 +209,56 @@ back to `'sonnet'`.
      -q '.workflow_runs[] | "\(.name): \(.status) \(.conclusion)"'
    ```
 
-5. **Existing reviews and comments**:
+5. **Existing reviews, comments, and inline review threads** — a single GraphQL call,
+   reusing the same query `/shipwright:patch`'s Step 3a issues (`### Step 3a: Check for
+   Unaddressed Review Findings` in `commands/patch.md`), so this command sees the same
+   inline-thread resolution state patch.md's unaddressed-findings check does, rather than
+   the REST `gh pr view --json comments,reviews` call (issue-level comments and top-level
+   review objects only — blind to inline diff-line review comments and their `isResolved`
+   state):
    ```bash
-   gh pr view {pr} --repo {org}/{repo} --json comments,reviews
+   gh api graphql -f query='
+   {
+     repository(owner: "{org}", name: "{repo}") {
+       pullRequest(number: {pr}) {
+         headRefOid
+         reviews(first: 50) {
+           nodes {
+             author { login }
+             state
+             submittedAt
+             body
+           }
+         }
+         reviewThreads(first: 100) {
+           nodes {
+             id
+             isResolved
+             comments(first: 1) {
+               nodes {
+                 author { login }
+                 body
+                 path
+                 line
+               }
+             }
+           }
+         }
+         comments(first: 50) {
+           nodes {
+             author { login }
+             body
+             createdAt
+           }
+         }
+       }
+     }
+   }'
    ```
+   Extract `reviews.nodes[]`, `reviewThreads.nodes[]` (with `isResolved` and the first
+   comment's `author.login`, `body`, `path`, `line`), and `comments.nodes[]` — the same
+   shape patch.md's Step 3a extracts. Used below by the Unresolved Comment Check and by
+   the unaddressed-findings gate before Step 10.
 
 6. **CLAUDE.md files**: read root CLAUDE.md + CLAUDE.md files in directories containing changed files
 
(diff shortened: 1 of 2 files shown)


plugins/shipwright/commands/review.md
```diff
@@ -461,10 +513,61 @@ local file is the authoritative signal that *this* agent has a prior review to a
...
+findings**, using the exact definition `/shipwright:patch`'s `### Step 3a: Check for
+Unaddressed Review Findings` (`commands/patch.md`) already applies — reuse that definition
+rather than re-deriving it in different language here (a fourth divergent copy of this logic
+is exactly the drift PRB-2.1 previously fixed between `check-deploy.ts` and
+`check-patch.ts`/`check-review.ts`). Using the `reviews`, `reviewThreads`, and `comments`
+fetched in Step 5.5:
+
+A PR has **unaddressed findings** when ANY of the following are true:
+- At least one review threads node has `isResolved == false` (any unresolved inline thread)
+- At least one review with `state == "COMMENTED"` or `state == "CHANGES_REQUESTED"` at the
```
dmcaulay: This "at the current `headRefOid`" qualifier has no supporting data — the Step 5.5 GraphQL query above (`review.md:198-234`) only fetches `author { login }`, `state`, `submittedAt`, `body` on `reviews`, with no `commit { oid }` field (unlike Step 14's existing live-review precheck query, which does fetch it for exactly this kind of filtering).

More importantly, `patch.md`'s actual Step 3a has no headRefOid restriction on reviews at all — old reviews from earlier commits count as findings until explicitly excluded via clean-APPROVE or reply-addressed rules. This PR's description says Step 9.5 "mirrors" and "reuse[s]" that definition, but this line is a real divergence from it.

This matters because the Unresolved Comment Check above unconditionally re-reviews whenever `headRefOid != lastReviewedCommit` — exactly the scenario where Step 9.5 needs to catch an earlier reviewer's still-unresolved COMMENTED/CHANGES_REQUESTED finding that new commits didn't address (no reply, thread unresolved). Under this "at head" filter, that older-but-real finding would never match and would silently fail to force COMMENT.

Suggest either adding `commit { oid }` to the reviews node in Step 5.5's query and using it correctly, or dropping the "at the current headRefOid" qualifier here to actually match patch.md's Step 3a.

dmcaulay (commented):
Verdict: COMMENT — Step 9.5's "at the current headRefOid" filter isn't supported by Step 5.5's GraphQL query (no `commit { oid }` fetched on reviews) and diverges from patch.md's Step 3a, which the PR description says it mirrors — this undermines the gate's core purpose of catching stale-but-real findings after new commits land. See inline comment for detail. PR title also appears to fail the required pr-title-lint check (no Conventional Commits type prefix).

dodizzle (commented):
Verdict: APPROVE — Clean, well-tested spec change: reuses patch.md's Step 3a GraphQL query and unaddressed-findings definition instead of duplicating the logic, already addressed prior review feedback (dropped the unsupported headRefOid filter, verified independently), and adds 14 passing content-test assertions. Investigated one prose divergence (an extra 'all inline threads resolved' qualifier on the CPF-2.3 reply exclusion vs. patch.md's literal wording) and confirmed it's a no-op given the gate's top-level OR structure — cross-checked against agent/src/check-patch.ts's hasUnaddressedFindings, which short-circuits the same way. No blocking concerns. (Posted as COMMENT — GitHub disallows self-approval via the API.)

PR #2445, opened 2026-08-03
fix: use constant-time comparison for dashboardToken in metrics auth

dodizzle:
## Summary
- Security patrol finding (`secret-weak-compare`, medium, tier 2): `createCombinedAuthMiddleware`'s bearer-token check compared `token === dashboardToken` directly, leaking the secret's content via response-time side channels.
- Swaps the `===` comparison for a length-checked `crypto.timingSafeEqual`, mirroring the existing `constantTimeEqual` precedent in `mcp-server/src/auth.ts` (Buffer.from both sides, length-check short-circuit, then `timingSafeEqual`).
- Adds `constantTimeEqual` to `metrics/src/lib/api-auth.ts` alongside the other shared auth helpers, with new unit tests covering match/mismatch/length-difference cases.

## Acceptance Criteria
| 1 | `metrics/src/api.ts`'s dashboardToken comparison uses a constant-time compare instead of `===` | MET | `metrics/src/api.ts` now calls `constantTimeEqual(token, dashboardToken)` |
| 2 | Mirrors the `mcp-server/src/auth.ts` `constantTimeEqual` pattern exactly | MET | Same Buffer.from + length-check + `timingSafeEqual` shape |

## Test Plan
- Added unit tests for `constantTimeEqual` (match, same-length mismatch, shorter/longer length mismatch, empty vs non-empty) — confirms no exception is thrown on length mismatch and behavior parity with the old `===` check for matching/non-matching tokens.
- Existing `METRICS_DASHBOARD_TOKEN gate` integration tests (correct/wrong/unset token) still pass unchanged.
- Full metrics suite: 354 pass, 0 fail.
- [x] Lint, typecheck, and metrics test suite passing

### DIFF
--- a/metrics/src/api.ts
+++ b/metrics/src/api.ts
@@ -22,7 +22,12 @@ import {
   wrapResponse,
 } from "./formatters.ts";
 import type { AccountsClient } from "./lib/accounts-client.ts";
-import type { AppHandler, AuthEnv, Caller } from "./lib/api-auth.ts";
+import {
+  type AppHandler,
+  type AuthEnv,
+  type Caller,
+  constantTimeEqual,
+} from "./lib/api-auth.ts";
 import { ErrorSchema } from "./lib/api-schemas.ts";
 import { registerWithAuthz } from "./lib/api-utils.ts";
 import { makeOnError } from "./lib/errors.ts";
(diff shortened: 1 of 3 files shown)


dodizzle (commented):
Verdict: APPROVE — Clean, well-scoped security fix. `constantTimeEqual` mirrors the established `mcp-server/src/auth.ts` precedent exactly, the length-check short-circuit is correct, and repo-wide grep confirms no other direct `===` secret comparisons were missed in `metrics/src`. Tests appropriately cover match/mismatch/length-difference cases.

PR #2461, opened 2026-08-05
fix: apply RVG-1.1 fresh-author-reply exception to RVD-1.1 live-review dedup check

dodizzle:
## Summary
- Hoists the `hasFreshAuthorReply` computation in `agent/src/check-review.ts`'s `getReviewCandidates` to run once, immediately after `reviewData` is fetched, instead of only inline at the later RVG-1.1 check
- Applies the resulting boolean as an exception to BOTH the RVD-1.1 live-review dedup `continue` and the RVG-1.1 record-based `continue`, closing a control-flow gap where RVD-1.1's unconditional skip fired first and made RVG-1.1's fresh-reply exception unreachable
- Adds unit test coverage for the real-world trigger shape (a clean/no-finding COMMENT review, unlike the existing RVG-1.1 fixtures which deliberately use an unresolved thread to dodge this exact interaction) plus a regression case for the no-fresh-reply path

## Acceptance Criteria
| # | Criterion | Status | Evidence |
|---|-----------|--------|----------|
| 1 | A PR with a terminal live-GitHub review at head, whose author has posted a comment after the record's `reviewedAt`, is NOT skipped by RVD-1.1 | MET | New test: clean COMMENT review + fresh author reply → stays a candidate |
| 2 | Existing RVG-1.1 record-based exception behavior unchanged | MET | All pre-existing RVG-1.1 tests (unresolved-thread fixtures) still pass unmodified |
| 3 | Terminal review at head + no fresh reply (or stale reply) still skipped by both checks | MET | New regression test: reply predates `reviewedAt` → PR excluded |
| 4 | Unit test added covering the real clean-COMMENT trigger shape + regression case, no existing tests retired | MET | Two new tests added to `check-review.unit.test.ts`; all 71 tests in the file pass |

## Test Plan
- `bun test agent/src/check-review.unit.test.ts` — 71 pass, 0 fail
- `task typecheck` — passes across all workspaces
- `task lint` — clean
- `task ci` — full suite run; 1 pre-existing unrelated failure in `metrics/src/lib/errors.unit.test.ts` (confirmed present on `main` prior to this branch, untouched by this diff)
- [x] Pre-commit checks passing

Generated with [Claude Code](https://claude.com/claude-code)

PR #2466, opened 2026-08-06
feat: review.md: add fresh-author-reply exception to the RVD-1.2 live-review precheck

dodizzle:
## Summary
- Extends RVD-1.2's `gh api graphql` query in `plugins/shipwright/commands/review.md` to fetch `submittedAt` on reviews, the PR's `comments` (author + `createdAt`), and the PR author's login, all within the same self-contained query.
- Updates the jq terminal-computation to override `terminal: true` with `false` when the PR author has posted a comment after the latest terminal-labeled review's `submittedAt` — porting the RFR-1.1 fresh-author-reply exception (`agent/src/check-review.ts`, PR #2461) to this separate bash/jq precheck.
- Extends `review.content.test.ts` to assert the new graphql fields and the fresh-reply-exception prose are present in the RVD-1.2 section.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `terminal` is `false` when a terminal-labeled review exists at head but the PR author replied after it | MET |
| 2 | Unchanged behavior otherwise (no fresh reply, or reply predates the review) — `terminal: true` | MET |
| 3 | Query/jq logic stays self-contained, no new task-store fetch introduced | MET |
| 4 | `review.content.test.ts` extended per the content-test convention | MET |

## Test Plan
- Verified the jq expression against 6 synthetic scenarios (terminal+no-reply, terminal+fresh-reply, no-terminal-review, terminal+no-comments, terminal+non-author-reply, multiple-terminal-reviews-max-timestamp) — all behave as expected.
- `NODE_ENV=test bun test plugins/shipwright/commands/review.content.test.ts` — 63/63 pass.
- `bun run --filter='*' --sequential typecheck` — clean across all workspaces.


plugins/shipwright/commands/review.md
```diff
@@ -801,33 +801,53 @@ precheck=$(gh api graphql -f query='
 {
   repository(owner: "{org}", name: "{repo}") {
     pullRequest(number: {pr}) {
+      author {
+        login
+      }
       headRefOid
+      comments(first: 50) {
```
dodizzle: `comments(first: 50)` fetches the *oldest* 50 comments. `agent/src/check-review.ts`'s `fetchPrReviews` (the reference implementation this PR explicitly ports) uses `comments(last: 50)` for the identical fresh-author-reply-detection purpose — verified on `main`. On any PR with more than 50 total comments, `first: 50` silently excludes the most recent comments (including a fresh author reply), so the newly-added exception never fires on exactly the busy/long-running PRs where it matters most. Suggest changing to `comments(last: 50)`.

dodizzle (commented):
Verdict: COMMENT — `comments(first: 50)` in the new GraphQL query uses the wrong pagination direction (should be `last: 50`, matching `agent/src/check-review.ts`'s `fetchPrReviews`), which silently defeats the fresh-author-reply exception on PRs with more than 50 comments.

dodizzle (commented):
Verdict: COMMENT — Prior finding (`comments(first: 50)` wrong pagination direction) is fixed in this diff — `review.md:812` now uses `comments(last: 50)`, with a regression assertion added in `review.content.test.ts`. No new issues found on re-review of the jq pipeline and GraphQL query structure. Posting as COMMENT rather than APPROVE since the prior review's body was never rewritten to a clean approve and no author reply closed it out, per this repo's unaddressed-findings dedup convention.

dodizzle:
Both review findings on this PR concern the same issue: whether the new GraphQL query in `review.md`'s RVD-1.2 live-review precheck used `comments(first: 50)` (wrong pagination direction) instead of `comments(last: 50)` (matching `agent/src/check-review.ts`'s `fetchPrReviews`).

Verified at current HEAD (`fef2a405`):
- `plugins/shipwright/commands/review.md:808` uses `comments(last: 50)` — the correct pagination direction. There is no `comments(first: 50)` anywhere in the RVD-1.2 precheck section (the only `comments(first: 50)` in the file is unrelated code at line 247).
- `plugins/shipwright/commands/review.content.test.ts` has a regression test (`the graphql query paginates comments with last: 50, not first: 50, matching check-review.ts's fetchPrReviews`) asserting `comments(last: 50)` is present and `comments(first: 50)` is absent from that section.
- Full test suite passes: 64/64 in `review.content.test.ts`.

The second review's own body confirms this: "Prior finding ... is fixed in this diff." No further code change is needed. Closing out both findings as addressed at head.

dmcaulay (approved):
Verdict: APPROVE — jq logic verified against 6 synthetic scenarios (all correct), fixes a real pagination-direction bug from an earlier review round, content tests match the existing convention, CI green. No blocking findings.

dodizzle (commented):
Verdict: APPROVE — Prior pagination-direction finding (`comments(first: 50)` → `comments(last: 50)`) is confirmed fixed at head and independently approved by @dmcaulay. Re-reviewed the jq watermark logic (null-safe `max` on empty array, ISO8601 comparison, GraphQL query structure) against the `check-review.ts` reference implementation — no new issues found. Head only moved via merge-from-main commits since the last pass, no new PR content to review. Posting as COMMENT event since this is a self-review (GitHub blocks self-APPROVE via the API).

PR #2483, opened 2026-08-07
feat(chart): let task-store and chat use the bundled PostgreSQL

dodizzle:
First of four PRs toward a full `helm install` of the whole Shipwright stack on Minikube with zero hand-created Secrets. This one closes the database gap.

## Why

`task-store` and `chat` could only get `DATABASE_URL_*` from a pre-existing Secret (default `shipwright-secrets`), and no `shipwright_task_store` / `shipwright_chat` database was ever created. Enabling either service on a bundled-PostgreSQL install was impossible without manual setup.

## Changes

- **Bundled-DB path for task-store and chat.** Set `database.existingSecret: ""` and the chart creates a dedicated database and assembles the DSN into its own Secret — the same idiom admin and metrics already used. Adds `taskStore.database.name` / `chat.database.name`.
- **`dbBootstrap` hook Job** (`post-install,post-upgrade`). PostgreSQL runs `/docker-entrypoint-initdb.d` scripts *only* on the first boot of an empty volume, so the init path alone can never provision databases on an existing release. The Job applies the same guarded SQL from the same ConfigMap against a live database.
- **`wait-for-postgres` initContainer** for both services, bundled path only.
- Password/DSN assembly consolidated into `shipwright.postgresql.password` / `.dsn` helpers.

## Bug fix

`admin-secret.yaml` and `metrics-secret.yaml` looked the Postgres password up under key `postgresql-password`. Bitnami postgresql 16.7.27 emits `postgres-password`, `password`, `replication-password`, `ldap-password` — **never** `postgresql-password` (verified against the vendored subchart). That branch could never fire, so an install relying on an auto-generated password rendered an **empty-password DSN**. This is why every file in `ci/` hardcodes a password.

*Upgrade note:* on a release relying on the broken fallback, the DSN changes from empty-password to the real one, so admin and metrics restart once. That is the fix landing.

## Compatibility

- **No defaults changed.** `database.existingSecret` still defaults to `"shipwright-secrets"`; the bundled path is strictly opt-in.
- `extraEnv` still renders last, so hand-wired production overrides still win (asserted by test).
- The initdb ConfigMap is renamed as a *file* but its **rendered name stays `<release>-metrics-initdb`** — `values.yaml` references it statically as `postgresql.primary.initdb.scriptsConfigMap`, and renaming it would leave a dangling volume and a Postgres pod that never starts. Guarded by a regression assertion.

## Tests

293 passing, up from 247. New suites: `task_store_secret_test`, `chat_secret_test`, `db_bootstrap_job_test`. Extended: `postgres_initdb_configmap_test`, both workload suites (bundled path, initContainer presence/absence, `extraEnv`-wins, fail-fast, and backward-compat regressions on the external path).

`task helm:check` green (lint + 293 unittests + kubeconform). Full-stack render also kubeconform-validated: 30/30 resources valid.

⚠️ Not yet validated against a real cluster — that lands with the Minikube profile PR.


dodizzle (commented):
Verdict: APPROVE — Well-documented, well-tested bundled-PostgreSQL support for task-store and chat. Traced the new _helpers.tpl helpers, the dbBootstrap hook Job, and the fail-fast branches against the actual source and values.yaml defaults; the backward-compatibility claims (no defaults changed, extraEnv wins last, initdb ConfigMap rendered name unchanged) all hold, and the new helm-unittest coverage targets the right gating branches. No blocking issues found.

dodizzle (commented):
Verdict: APPROVE — Re-reviewed after a merge-from-main sync; diff content is byte-identical to the previously-approved commit (7f13cca0), no new feature code landed. Independent re-verification of the password/DSN helper consolidation, opt-in bundled-database gating, ConfigMap-rename identity, bootstrap Job correctness, and helm-unittest coverage found no issues. No blocking issues found.

PR #2484, opened 2026-08-07
feat(chart): generate the inter-service token mesh so a fresh install needs no Secrets

dodizzle:
Second of four PRs toward a full `helm install` of the whole Shipwright stack on Minikube.

**Stacked on #2483** — review that first; this PR targets its branch and will be retargeted to `main` once it merges.

## Why

The chart rendered *no* inter-service wiring at all. Admin never received `SHIPWRIGHT_TASK_STORE_URL`, `SHIPWRIGHT_TASK_STORE_ADMIN_TOKEN`, or `SHIPWRIGHT_ADMIN_API_KEYS`; task-store and chat never received their seed tokens or scope-resolver credentials. Production hand-wires all of it through `extraEnv`. A fresh install could bring up pods that couldn't talk to each other.

## The three tokens

| Token | Produced by | Consumed by |
|---|---|---|
| `TASK_STORE_ADMIN_TOKEN` | task-store (`TASK_STORE_SEED_ADMIN_TOKEN`) | admin, metrics |
| `CHAT_ADMIN_TOKEN` | chat (`CHAT_SEED_ADMIN_TOKEN`) | admin |
| `INTERNAL_API_KEY` | admin (inside `SHIPWRIGHT_ADMIN_API_KEYS`) | metrics, task-store, chat |

Generated in exactly one template (`internal-secret.yaml`), referenced everywhere else by `secretKeyRef`. That's forced by Helm, not style: templates evaluate independently, so `randAlphaNum` in a producer and again in a consumer yields two different strings. Reused across upgrades via the same `lookup` idiom as the admin session keys.

`SHIPWRIGHT_ADMIN_API_KEYS` is composed in that same Secret because the format embeds the raw key — verified against `parseAdminApiKeys` in `admin/src/api-auth.ts` (`name:token:scope`, first segment name, **last** scope, middle is the token). Scope is `*` because no agent exists at install time.

## Also fixes a crash-on-boot

`METRICS_ADMIN_URL` / `METRICS_TASK_STORE_URL` now auto-default to in-cluster Service DNS. Without `METRICS_TASK_STORE_URL` the provider selector falls through to SQLite, which has no writable path in the published image and crashes with `SQLITE_CANTOPEN`.

## Compatibility

- Every generated value is a **fallback**. Service-specific `existingSecret`s always win; `extraEnv` still renders **last** and wins over everything (asserted).
- `internal.enabled=false` restores pre-1.9 behavior exactly, for operators who hand-wire the mesh.
- ⚠️ **If you hand-wire `SHIPWRIGHT_ADMIN_API_KEYS` via `admin.extraEnv`**, it shadows the composed value and the chart-generated key won't be accepted — consumers 401 on fallback. `admin.apiKeys.extra` is the migration path. Called out in CHANGELOG.
- ⚠️ `helm template | kubectl apply` doesn't execute `lookup` and would rotate all three tokens per apply. `internal.existingSecret` is the escape hatch.
- One intentional behavior change: enabling chat no longer requires `chat.adminToken.existingSecret`. The old test asserting "injects nothing" is updated, and a new case pins the old behavior under `internal.enabled=false`.

## Tests

332 passing, up from 293. New: `internal_secret_test`, `service_mesh_env_test` — the latter asserts **both sides** of every producer/consumer pair resolve to the same Secret *and* key, so a rename on one side fails here instead of 401ing at runtime.

Mutation-checked: pointing admin's chat token at a wrong key fails 2 suites; reverting restores green. (Four assertions using ambiguous `any: true` regex semantics were rewritten as explicit `contains`/`notContains` after I couldn't confirm they weren't passing vacuously.)

`helm lint` + 332 unittests + kubeconform green (default 19/19, full stack 31/31).

⚠️ Not yet validated against a real cluster — that lands with the Minikube profile PR, including a `helm upgrade`-twice check that the tokens stay byte-identical.

### DIFF
--- a/charts/shipwright/Chart.yaml
+++ b/charts/shipwright/Chart.yaml
@@ -9,7 +9,7 @@ type: application
 # Chart version (SemVer). Bump on every chart change, independent of appVersion.
 # See CHANGELOG.md and the README "Versioning" section — CI enforces this bump
 # via `ct lint --check-version-increment`.
-version: 1.9.0
+version: 1.10.0
 # Version of the Shipwright application this chart deploys by default.
 appVersion: "0.1.0"
 home: https://shipwrightharness.com
@@ -30,11 +30,9 @@ annotations:
   # entry in CHANGELOG.md (keep the two in sync on every bump).
   artifacthub.io/changes: |
     - kind: added
-      description: "inter-service token mesh: the chart now generates the task-store, chat and internal API tokens so a fresh install needs zero hand-created Secrets"
-    - kind: added
-      description: "metrics auto-defaults METRICS_ADMIN_URL and METRICS_TASK_STORE_URL to the in-cluster Services, avoiding the SQLITE_CANTOPEN crash-on-boot"
-    - kind: added
-      description: "admin.apiKeys.extra merges caller-supplied entries into SHIPWRIGHT_ADMIN_API_KEYS (migration path off hand-wiring it via extraEnv)"
+      description: "agent.credentials.* supplies deployment-wide Claude credentials to every provisioned agent, merged under each agent's own env rows"
+    - kind: fixed
+      description: "agent.provisioning.pvc.size is now injected as SHIPWRIGHT_AGENT_PVC_STORAGE_GI; it was never rendered, so every provisioned agent silently requested the provisioner's 40Gi default"
 dependencies:
   # Bitnami PostgreSQL subchart. Pinned to a chart version whose default image tag
   # is concrete (NOT `latest`), because Bitnami's 2025 catalog change moved newer
--- a/charts/shipwright/Chart.yaml
+++ b/charts/shipwright/Chart.yaml
@@ -9,7 +9,7 @@ type: application
 # Chart version (SemVer). Bump on every chart change, independent of appVersion.
 # See CHANGELOG.md and the README "Versioning" section — CI enforces this bump
 # via `ct lint --check-version-increment`.
-version: 1.8.3
+version: 1.9.0
 # Version of the Shipwright application this chart deploys by default.
 appVersion: "0.1.0"
 home: https://shipwrightharness.com
@@ -29,8 +29,12 @@ annotations:
   # Artifact Hub change log for this release. Must mirror the matching version
   # entry in CHANGELOG.md (keep the two in sync on every bump).
   artifacthub.io/changes: |
-    - kind: changed
-      description: "auto-bump to chart v1.8.3 triggered by release tag agent-v1.129.0"
+    - kind: added
+      description: "inter-service token mesh: the chart now generates the task-store, chat and internal API tokens so a fresh install needs zero hand-created Secrets"
+    - kind: added
+      description: "metrics auto-defaults METRICS_ADMIN_URL and METRICS_TASK_STORE_URL to the in-cluster Services, avoiding the SQLITE_CANTOPEN crash-on-boot"
+    - kind: added
+      description: "admin.apiKeys.extra merges caller-supplied entries into SHIPWRIGHT_ADMIN_API_KEYS (migration path off hand-wiring it via extraEnv)"
 dependencies:
   # Bitnami PostgreSQL subchart. Pinned to a chart version whose default image tag
   # is concrete (NOT `latest`), because Bitnami's 2025 catalog change moved newer
--- a/charts/shipwright/templates/metrics-deployment.yaml
+++ b/charts/shipwright/templates/metrics-deployment.yaml
@@ -87,9 +87,16 @@ spec:
                   name: {{ include "shipwright.metrics.secretName" . }}
                   key: METRICS_DATABASE_URL
             {{- end }}
-            {{- if .Values.metrics.provider.adminUrl }}
+            {{- /* Falls back to the in-cluster admin Service DNS name so a
+                   full-stack install needs no hand-written URLs. An explicit
+                   metrics.provider.adminUrl still wins (external admin). */}}
+            {{- $adminUrl := .Values.metrics.provider.adminUrl }}
+            {{- if and (not $adminUrl) .Values.admin.enabled }}
+            {{- $adminUrl = include "shipwright.admin.serviceUrl" . }}
+            {{- end }}
+            {{- if $adminUrl }}
             - name: METRICS_ADMIN_URL
-              value: {{ .Values.metrics.provider.adminUrl | quote }}
+              value: {{ $adminUrl | quote }}
             {{- end }}
             # Task-store mode: when taskStoreUrl is set alongside adminUrl, the
             # metrics service selects TASKSTORE mode (TaskStoreProvider over the
@@ -97,9 +104,18 @@ spec:
             # the local SQLite store, which has no writable path in this image and
             # crashes on boot (SQLITE_CANTOPEN). The token authenticates the
             # task-store HTTP client (METRICS_TASK_STORE_TOKEN).
-            {{- if .Values.metrics.provider.taskStoreUrl }}
+            {{- /* Falls back to the in-cluster task-store Service DNS name when
+                   that service is part of this release. This is load-bearing:
+                   without METRICS_TASK_STORE_URL the provider selector falls
+                   through to SQLite, which has no writable path in the published
+                   image and crashes on boot with SQLITE_CANTOPEN. */}}
+            {{- $taskStoreUrl := .Values.metrics.provider.taskStoreUrl }}
+            {{- if and (not $taskStoreUrl) .Values.taskStore.enabled }}
+            {{- $taskStoreUrl = include "shipwright.taskStore.serviceUrl" . }}
+            {{- end }}
+            {{- if $taskStoreUrl }}
             - name: METRICS_TASK_STORE_URL
-              value: {{ .Values.metrics.provider.taskStoreUrl | quote }}
+              value: {{ $taskStoreUrl | quote }}
             {{- end }}
             {{- if .Values.metrics.provider.taskStoreToken.existingSecret }}
             - name: METRICS_TASK_STORE_TOKEN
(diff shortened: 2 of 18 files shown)


charts/shipwright/Chart.yaml
```diff
@@ -30,11 +30,11 @@ annotations:
   # entry in CHANGELOG.md (keep the two in sync on every bump).
   artifacthub.io/changes: |
```
dodizzle, line 31: **Important (confidence 80)**: This annotation is out of sync with the CHANGELOG.md `[1.9.0]` entry — CONTRIBUTING.md requires "Keep the annotation and the CHANGELOG in sync." The CHANGELOG has 5 Added bullets plus full Changed/Compatibility sections (including the chat auto-wiring behavior change), but this only lists 3 `kind: added` entries — missing the `internal.enabled`/`internal.existingSecret` addition, and with no `kind: changed` entry for the breaking behavior change at all. Artifact Hub surfaces only this annotation to consumers browsing releases, so the more consumer-visible surface is silent about the change that most needs a pre-upgrade heads-up.

dodizzle (commented):
Verdict: COMMENT — Solid, well-tested Helm templating (helm lint clean, 332/332 unittests pass locally, base64/lookup round-tripping verified against admin/src/api-auth.ts). Two things worth fixing before merge: the artifacthub.io/changes annotation isn't in sync with the CHANGELOG (CONTRIBUTING.md requires it), and a pre-existing values.yaml comment on chat.adminToken now actively misdescribes the new default wiring behavior this PR introduces. Full write-up in the review file.

charts/shipwright/templates/metrics-deployment.yaml
```diff
@@ -87,26 +87,49 @@ spec:
                   name: {{ include "shipwright.metrics.secretName" . }}
                   key: METRICS_DATABASE_URL
             {{- end }}
-            {{- if .Values.metrics.provider.adminUrl }}
+            {{- /* Falls back to the in-cluster admin Service DNS name so a
+                   full-stack install needs no hand-written URLs. An explicit
+                   metrics.provider.adminUrl still wins (external admin). */}}
+            {{- $adminUrl := .Values.metrics.provider.adminUrl }}
+            {{- if and (not $adminUrl) .Values.admin.enabled }}
```
dodizzle: **Important (confidence 82)**: This auto-default (and the matching `taskStoreUrl` block at line 113) isn't gated on `internal.enabled`, but the corresponding token (`METRICS_INTERNAL_API_KEY`/`METRICS_TASK_STORE_TOKEN`) is — gated on `shipwright.internal.wired` or an explicit `existingSecret`. With `internal.enabled=false` (the chart's documented escape hatch, which CHANGELOG.md claims "restores pre-1.9 behaviour exactly") plus `taskStore.enabled=true`, metrics gets `METRICS_TASK_STORE_URL` auto-set with no token — flipping the provider selector into TASKSTORE mode with no credential, instead of the pre-1.9 no-op. Suggest gating this the same way the token is gated: `{{- if and (not $taskStoreUrl) .Values.taskStore.enabled (include "shipwright.internal.wired" .) }}` (and the admin-URL equivalent at line 94), or revising the compatibility claim if the URL auto-default is intentionally unconditional.

dodizzle (commented):
Verdict: COMMENT — Solid re-review: both prior findings (annotation/CHANGELOG sync, stale chat.adminToken comment) are cleanly fixed. Found one new important issue: the metrics URL auto-default isn't gated on `internal.enabled` the way the corresponding token is, so `internal.enabled=false` doesn't fully "restore pre-1.9 behaviour exactly" as CHANGELOG.md claims. Full write-up in the review file. (Posted as COMMENT — GitHub disallows self-approval via the API.)

charts/shipwright/templates/metrics-deployment.yaml
```diff
@@ -87,26 +87,49 @@ spec:
                   name: {{ include "shipwright.metrics.secretName" . }}
                   key: METRICS_DATABASE_URL
             {{- end }}
-            {{- if .Values.metrics.provider.adminUrl }}
+            {{- /* Falls back to the in-cluster admin Service DNS name so a
+                   full-stack install needs no hand-written URLs. An explicit
+                   metrics.provider.adminUrl still wins (external admin). */}}
+            {{- $adminUrl := .Values.metrics.provider.adminUrl }}
```
dodizzle, line 93: **Important (confidence 90)**: This auto-default (and the matching `taskStoreUrl` block at line 112) isn't gated on `internal.enabled`, but the corresponding tokens (`METRICS_TASK_STORE_TOKEN` at line 126, `METRICS_INTERNAL_API_KEY` at line 152) are — both gated on `shipwright.internal.wired`. With `internal.enabled=false` + `taskStore.enabled=true`, metrics gets `METRICS_TASK_STORE_URL` auto-set with no token — flipping the provider selector into TASKSTORE mode with no credential, instead of the pre-1.9 no-op. CHANGELOG.md's Compatibility section claims `internal.enabled=false` "restores pre-1.9 behaviour exactly," but pre-1.9 metrics had no URL auto-default at all — this is a newly introduced asymmetry, not a pre-existing issue. Suggest gating this block (and the admin-URL equivalent) the same way the token is gated: `{{- if and (not $taskStoreUrl) .Values.taskStore.enabled (include "shipwright.internal.wired" .) }}`, matching `admin-deployment.yaml`'s pattern of gating the URL and token together.

charts/shipwright/Chart.yaml
```diff
@@ -29,8 +29,12 @@ annotations:
   # Artifact Hub change log for this release. Must mirror the matching version
   # entry in CHANGELOG.md (keep the two in sync on every bump).
   artifacthub.io/changes: |
```
dodizzle, line 31: **Important (confidence 85)**: This annotation is out of sync with the CHANGELOG's `[1.9.0]` entry — 3 `kind: added` bullets here vs. 5 Added + a full Changed section (the chat auto-wiring behavior change) in CHANGELOG.md, with no `kind: changed` entry at all. CONTRIBUTING.md requires keeping the two in sync. Worth noting: this was actually fixed at an intermediate commit on this branch (`eef0c3a3`, which added the missing `internal.enabled`/`existingSecret` bullet and a `kind: changed` entry) but the fix is gone at the current head — looks like it got dropped during the rebase onto `main`. Same story for a stale `chat.adminToken` comment in values.yaml:727 (not in this diff, so flagged in the review body instead of inline).

dodizzle (commented):
Verdict: COMMENT — Solid, well-tested token-mesh implementation, but three things need attention: (1) `metrics-deployment.yaml`'s URL auto-default isn't gated on `internal.enabled` the way its paired token is, so `internal.enabled=false` silently breaks the CHANGELOG's claim of restoring pre-1.9 behavior exactly for metrics; (2) the Chart.yaml annotation/CHANGELOG sync and (3) the stale `chat.adminToken` values.yaml comment were both fixed at an intermediate commit on this branch but regressed back to their original state after a rebase/force-push — worth a quick sanity check that nothing else got dropped in that rebase. Full write-up (including the rebase-regression evidence) in the review file. (Posted as COMMENT — GitHub disallows self-approval via the API; there's also a still-open inline thread from a prior review cycle at this head.)

charts/shipwright/templates/metrics-deployment.yaml
dodizzle: **Important (confidence 88)**: `METRICS_ADMIN_URL` auto-defaults whenever `admin.enabled` is true, with no gate on `internal.enabled`/`shipwright.internal.wired` — but the paired token (`METRICS_INTERNAL_API_KEY`, line ~152) *is* gated on it. `admin-deployment.yaml:117-138` gates its URL+token pair together; this template doesn't follow that pattern. With `internal.enabled=false` (the documented escape hatch that CHANGELOG.md claims "restores pre-1.9 behaviour exactly") plus `admin.enabled=true`, metrics still gets the URL auto-set, flipping its provider selector into a credential-less mode instead of a true no-op. Same issue applies to the `taskStoreUrl` block at line 113 vs. `METRICS_TASK_STORE_TOKEN`. Still unaddressed after the last two review rounds — this remains the one concrete pre-merge fix needed.

Suggestion: gate both URL auto-defaults on `(include "shipwright.internal.wired" .)` as well, mirroring admin-deployment.yaml's combined gate.

dodizzle (commented):
Verdict: COMMENT — Re-review after a rebase (branch replayed onto an advanced main, plus #2485 merging into this branch and bumping the chart to 1.10.0). The one issue that still needs fixing before merge: metrics-deployment.yaml auto-defaults METRICS_ADMIN_URL/METRICS_TASK_STORE_URL without gating on internal.enabled, while the paired tokens are gated — breaking the documented internal.enabled=false compatibility escape hatch. The earlier Chart.yaml/CHANGELOG annotation-sync finding is now moot (superseded by #2485's version bump to 1.10.0, which is itself correctly synced). Full write-up in the review file. (Posted as COMMENT — GitHub disallows self-approval via the API; two related inline threads from prior review cycles remain open at this head.)

dodizzle:
Re: the Chart.yaml `artifacthub.io/changes` / CHANGELOG.md sync finding (thread on `charts/shipwright/Chart.yaml:31`) — confirming this is moot, per the most recent review round.

I re-verified directly against the current state of the branch:

- `charts/shipwright/Chart.yaml` is at `version: 1.10.0`, with an `artifacthub.io/changes` block listing two entries:
  - `kind: added` — "agent.credentials.* supplies deployment-wide Claude credentials to every provisioned agent, merged under each agent's own env rows"
  - `kind: fixed` — "agent.provisioning.pvc.size is now injected as SHIPWRIGHT_AGENT_PVC_STORAGE_GI; it was never rendered, so every provisioned agent silently requested the provisioner's 40Gi default"
- `CHANGELOG.md`'s `## [1.10.0] - 2026-08-07` entry has a matching `### Added` section for `agent.credentials.*` (same feature description) and a matching `### Fixed` section for the PVC size bug (same root cause and fix description).

The two are in sync at the current chart version, as PR #2485 (which bumped the chart to 1.10.0 and merged into this branch) already correctly synced them. No code change needed here — resolving the thread as REJECT/already-resolved.

dmcaulay (approved):
Verdict: APPROVE — Clean fix commit correctly gates the metrics URL auto-defaults on internal.enabled (mirroring the token gate), resolving the last finding from 4 rounds of prior review. Independently re-verified the token-mesh wiring, internal-secret.yaml's lookup/base64 handling, resolveDefaultAgentEnv's security boundary, and test coverage — no new issues.

PR #2485, opened 2026-08-07
feat(chart): supply deployment-wide Claude credentials to provisioned agents

dodizzle:
Third of four PRs toward a full `helm install` of the whole Shipwright stack on Minikube.

**Stacked on #2484** (which is stacked on #2483). Review those first.

## Why

This chart renders no agent Deployment by design — agents are provisioned at runtime by the admin service, and they fetch credentials from `GET /agents/:id/config` on each config sync. The only way to give an agent a Claude token was to paste one into every agent through the admin UI, so `helm install` could never produce a *working* agent on its own.

## How

The cheapest correct hook is a defaults merge in the config-bundle handler, not the manifest builder:

- `admin/src/api.ts` — `env: { ...defaultAgentEnv, ...bundle.env }`, so **per-agent rows always win**. This is a floor, not an override.
- `admin/src/main.ts` — new pure `resolveDefaultAgentEnv(env)`.
- Chart — `agent.credentials.*` → `agent-credentials-secret.yaml` → `secretKeyRef` into the admin Deployment.

~15 lines of app code. Because it flows through the config bundle, it also reaches **already-running** agents on their next sync — no pod restart, no re-provision. (Injecting into the pod manifest instead would only apply on re-provision, and `syncConfig`'s `Object.assign` would override it anyway.)

### Security boundary

Credentials are read from `SHIPWRIGHT_AGENT_DEFAULT_`-prefixed vars, never bare `ANTHROPIC_API_KEY`. Admin must not hand out a credential it happens to hold for its own use, so opting in is explicit. There's a test asserting a bare `ANTHROPIC_API_KEY` in admin's env is *not* picked up, and another asserting no bare var is injected into the container.

Blank values are treated as unset — an empty Helm value (the default) must not inject an empty credential that would mask a working per-agent one.

Inline values land in a chart-managed Secret and therefore sit in your values file and Helm release history; `existingSecret` is documented as the choice for anything beyond local dev.

## Bug fix: provisioned agents silently requested 40Gi PVCs

`values.yaml` claimed `agent.provisioning.pvc.size` was "not injected", and nothing ever rendered `SHIPWRIGHT_AGENT_PVC_STORAGE_GI` — so `agent-provisioner.ts:517` fell back to its **40Gi** default for every agent regardless of the configured value. Now injected (with `Gi` stripped); stale comment corrected.

*Upgrade note:* existing agents keep their current PVCs; the new size applies to agents provisioned from here on.

## Tests

- **Chart:** 345 passing, up from 332. New `agent_credentials_test` covers the Secret gating (inline / existingSecret / disabled / neither), admin wiring, custom key names, the no-bare-var boundary, and the PVC-size regression.
- **App:** 4 new smoke tests pin the merge precedence both ways (default inherited, per-agent wins, keys unioned, no-defaults passthrough); 6 new unit tests on `resolveDefaultAgentEnv` including blank/whitespace handling and the bare-var boundary.
- Full admin suite green: 1551 pass / 0 fail. `helm lint` + kubeconform 23/23. `task lint` + admin typecheck clean.

⚠️ Not yet validated against a real cluster — that lands with the Minikube profile PR.

### DIFF
--- a/admin/src/api.smoke.test.ts
+++ b/admin/src/api.smoke.test.ts
@@ -824,3 +824,101 @@ describe("combined server — regression guard: runtime middleware must not bloc
     expect(body.repos).toEqual([]);
   });
 });
+
+// ─── Deployment-wide default agent env ────────────────────────────────────────
+//
+// A Helm install supplies one set of Claude credentials for the whole deployment
+// (agent.credentials.*) instead of an operator pasting a token into every agent.
+// Admin merges those defaults UNDERNEATH each agent's own env rows when serving
+// the config bundle, so a per-agent value always wins. These tests pin that
+// precedence — getting it backwards would let a deployment-wide credential
+// silently override a deliberately-scoped per-agent one.
+
+describe("defaultAgentEnv merging in GET /:id/config", () => {
+  const AGENT_ID = "11111111-1111-1111-1111-111111111111";
+  const ADMIN_KEY = "sk_default_env_admin";
+
+  function buildApp(opts: {
+    bundleEnv?: Record<string, string>;
+    defaultAgentEnv?: Record<string, string>;
+  }) {
+    return createAgentRuntimeApp({
+      agentEnvService: {
+        async getConfigBundle() {
+          return {
+            agentId: AGENT_ID,
+            env: opts.bundleEnv ?? {},
+            allowedTools: [],
+          };
+        },
+      },
+      agentCronJobService: {
+        async list() {
+          return [];
+        },
+        async listWithRunSummary() {
+          return [];
+        },
+      },
+      agentService: {
+        async getById() {
+          return {
+            id: AGENT_ID,
+            name: "Test Agent",
+            repos: [],
+            authorAllowlist: [],
+          };
+        },
+      },
+      agentPluginService: {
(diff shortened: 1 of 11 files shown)


dmcaulay (approved):
Verdict: APPROVE — Clean, well-scoped, thoroughly tested change with a correctly-implemented security boundary (SHIPWRIGHT_AGENT_DEFAULT_ prefix, merge precedence, PVC-size fix all verified against the diff). One non-blocking suggestion: `shipwright.agent.effectiveCredentialsSecretName` in _helpers.tpl is defined but never referenced anywhere in the chart.

dmcaulay (approved):
Verdict: APPROVE — Re-reviewed after the branch rebased onto an advanced base (head moved 498c4ce7 → 7cb2e0c0, base picked up unrelated auto-bump/feature commits). Diffed every PR-authored file against the current base: content is byte-for-byte identical to the previously-approved commit, only the chart version bump (1.9.0 → 1.10.0) changed. No new issues. Prior non-blocking suggestion stands: `shipwright.agent.effectiveCredentialsSecretName` in `_helpers.tpl` is defined but never referenced anywhere in the chart.

PR #2488, opened 2026-08-07
feat(chart): add a full-stack Minikube profile and task minikube:up

dodizzle:
Last of four PRs toward a full `helm install` of the whole Shipwright stack on Minikube.

**Stacked on #2485 → #2484 → #2483.** Review those first.

## What lands here

- **`charts/shipwright/examples/values-minikube.yaml`** — admin + metrics + task-store + chat + bundled PostgreSQL + agent provisioning, no hand-created Secrets. Renders 31 resources, all kubeconform-valid.
- **`scripts/minikube.ts` + `task minikube:up` / `task minikube:down`** — pure `buildMinikubeCommands()` builder plus an injected-exec driver, mirroring `scripts/dev-tmux.ts`.
- **`helm test` now checks task-store and chat `/health`.** This is the part that makes the whole chain meaningful: both services run `prisma migrate deploy` as a boot preflight and refuse to serve if their database is missing or unmigrated. A green `helm test` therefore *proves* the chart-managed database wiring works.
- Docs: rewritten Minikube section, sizing table, a "what the chart generates for you" table, and the new env vars in `docs/configuration.md` + the site data.

## Why a script rather than documented commands

Four ordering constraints that fail confusingly rather than loudly:

1. VM sizing is only settable at `minikube start` — the agent pod alone requests 500m CPU / 2Gi memory, and an undersized VM schedules it then thrashes.
2. `minikube addons enable ingress` must precede the install, or the Ingress has no controller and nothing answers.
3. `helm dependency build` must precede the install — `Chart.lock` pins the PostgreSQL subchart via OCI.
4. `/etc/hosts` can only be written once the VM has an IP.

The 22 unit tests pin exactly that ordering, with no VM, Docker, or network.

## Stale docs corrected

`NOTES.txt` and the chart README both claimed *"the agent workload is added in a later task and is NOT yet rendered"*, and NOTES advertised an agent service port nothing listens on. Agents are provisioned at runtime by admin. NOTES now says so, points at `/admin/agents/new`, and warns when no deployment-wide Claude credential is configured. Also dropped the claim that an unset `METRICS_DATABASE_URL` falls back to a working SQLite store — it crashes with `SQLITE_CANTOPEN`.

## Sizing guidance

| VM size | Good for |
|---|---|
| `--cpus=4 --memory=8192` | Floor — platform plus **one** agent |
| `--cpus=6 --memory=12288` | An agent doing real work |
| Below 4 CPU / 6Gi | Agent schedules, then thrashes |

## Tests

353 chart tests (up from 345) + 22 new script unit tests. `helm lint`, `task lint`, `task check-config-docs`, full typecheck, and kubeconform (defaults 19/19, minikube profile 31/31) all green.

## ⚠️ Still unvalidated against a real cluster

Everything here passes static gates and CI's `ct install` on kind, but **no one has yet run `task minikube:up` on an actual Minikube.** Several failure modes in this chain are invisible to every automated gate — helm-unittest never executes `lookup`, and `ct install` sets an explicit Postgres password and never provisions an agent. The remaining validation (fresh install, `helm upgrade` twice to prove token stability, provisioning a real agent, and a backward-compat render diff against production values) is tracked and not yet done.


docs/deploy-kubernetes.md
```diff
@@ -147,79 +147,132 @@ for the full list of chat service env vars and their defaults.
...
-`networking.type=ClusterIP` too):
+Or port-forward instead (works with the default `networking.type=ClusterIP`):
 
 ```bash
-kubectl port-forward svc/shipwright-admin   3001:3001 -n shipwright   # → http://localhost:3001
-kubectl port-forward svc/shipwright-metrics 3460:3460 -n shipwright   # → http://localhost:3460/dashboard
+kubectl port-forward svc/shipwright-admin      3001:3001 -n shipwright
+kubectl port-forward svc/shipwright-metrics    3460:3460 -n shipwright
+kubectl port-forward svc/shipwright-task-store 3000:3000 -n shipwright
+kubectl port-forward svc/shipwright-chat       3001:3000 -n shipwright
```
dmcaulay: `admin.service.port` is 3001 and this line forwards chat's remote 3000 to local **3001** too — that collides with the admin line above it (`3001:3001`) if both are run together, which is the evident intent of listing them as a block for accessing the whole stack locally. Suggest a distinct local port, e.g. `3002:3000`.

dmcaulay (commented):
Verdict: COMMENT — the port-forward quick-reference in docs/deploy-kubernetes.md has a local-port collision between admin (3001) and chat (forwarded to local 3001 instead of a distinct port). Chart/script logic is otherwise solid and well-tested — see inline comment.

dmcaulay (commented):
Verdict: COMMENT — the port-forward quick-reference in docs/deploy-kubernetes.md has a local-port collision between admin (3001) and chat (forwarded to local 3001 instead of a distinct port). Chart/script logic is otherwise solid and well-tested — see inline comment.

docs/deploy-kubernetes.md
dmcaulay: `admin.service.port` is 3001 and this line forwards chat's remote 3000 to local **3001** too — that collides with the admin line above it (`3001:3001`) if both are run together, which is the evident intent of listing them as a block for accessing the whole stack locally. Suggest a distinct local port, e.g. `3002:3000`.

dmcaulay (commented):
Verdict: COMMENT — the previously flagged port-forward collision (docs/deploy-kubernetes.md:236, chat vs admin both on local 3001) is fixed here — chat now forwards to local 3002. A fresh full pass found no new issues: the ordering-sensitive minikube.ts script, injected-exec pattern, Helm helper references, new env var namespacing, and the local-dev-only Postgres password all check out. Posting as COMMENT rather than APPROVE only because the earlier review threads on this PR aren't yet formally closed out — no action needed from you beyond what's already landed.

dodizzle:
Reviewed the finding(s) above and did not implement changes — premise did not hold:

- Port-forward collision (docs/deploy-kubernetes.md, reviews at 16:20 and 17:36 UTC): already fixed on this branch — chat now forwards to local port 3002 (line 236), admin stays on 3001 (line 233). No collision remains.
- Confirmation review (18:10 UTC): review itself states "no action needed from you beyond what's already landed" and withheld a formal APPROVE only because it considered earlier review threads not yet formally closed — both inline threads are now resolved on GitHub, so nothing further to address here either.

dmcaulay (approved):
Verdict: APPROVE — the previously flagged port-forward collision (chat vs admin, both on local 3001) is fixed here and both inline threads are resolved. A fresh full pass found no new issues: scripts/minikube.ts's pure builder/injected-exec split, the Helm NOTES/health-check changes, and values-minikube.yaml all check out.

PR #2450, opened 2026-08-04
feat: RVG-1.1 re-trigger review candidacy on fresh author reply at unchanged HEAD

dmcaulay:
## Summary
- `getReviewCandidates()` (agent/src/check-review.ts) now re-adds a PR as a candidate when its record is already terminal at the current commit, but the PR author has posted a fresh comment with `createdAt` after `record.reviewedAt` — so a follow-up review pass can consume that reply and apply review.md's CPF-2.3 unaddressed-findings exclusion.
- Reuses the `reviewData` already fetched earlier in the loop for the RVD-1.1 live-review dedup check (hoisted, no second GraphQL call) and only counts comments from the PR's own author (`pr.author.login`), not the reviewing agent or third parties.
- Adds `reviewedAt?: string | null` to the local `PrRecord` interface (already present on the task-store `/prs` record, just not yet declared here).

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | A PR with reviewState posted/approved, unchanged commitSha, and no new PR-author comment since reviewedAt is still skipped (no regression) | MET |
| 2 | The same PR becomes an eligible candidate again once the PR author posts a comment with createdAt after the record's reviewedAt | MET |
| 3 | A comment from a non-author (another reviewer or bot) does not trigger re-candidacy | MET |
| 4 | Once a fresh review advances reviewedAt, the same original comment stops re-triggering candidacy | MET |
| 5 | Unit tests added to agent/src/check-review.unit.test.ts covering all four scenarios with injected fixtures | MET |

## Test Plan
- 6 new unit tests in `agent/src/check-review.unit.test.ts`: regression skip, fresh-reply retrigger, non-author-comment rejection, reviewedAt-watermark advancement, missing/null reviewedAt, and fetchPrReviews-failure fallback — plus a guard test confirming the unrelated staged-review skip stays unaffected.
- `bun test agent/src/check-review.unit.test.ts` — 69/69 pass.
- `task typecheck` — clean across all workspaces.
- `task ci` — full suite passes except one full-suite-only flake in `task-store/src/routes/prs.openapi.smoke.test.ts` (15/15 pass in isolation; unrelated to this diff, matches this repo's known shared-Bun-process full-suite flake pattern).
- [x] Pre-commit checks passing

Generated with [Claude Code](https://claude.com/claude-code)

PR #2382, opened 2026-07-31
feat(task-store): split hitl/blocked signals + data migration (HSR-1.1)

dodizzle:
## Summary
- Split the single `hitl` signal into two distinct signals: `Task.requiresHumanApproval` (Type B merge-approval-gate classification — does not gate dispatch candidacy) and `PullRequest.blocked` (a PR blocked in the pipeline).
- Removed the dead `Task.hitlNotifiedAt` field (zero consumers ever wrote to it) and the superseded `PullRequest.hitl`/`PullRequest.hitlNotifiedAt` fields.
- Hand-authored migration `20260731000000_split_hitl_blocked_signals` with data-preservation UPDATE statements between the ADD COLUMN and DROP COLUMN statements.
- Updated `task-store/src/openapi-schemas.ts`'s Zod schemas to match the new field set.

## ⚠️ Bundled PR — HSR-1.1 through HSR-1.8 land here together

This is task **HSR-1.1** of an 8-task bundle (HSR-1.1–HSR-1.8), all landing on this
one branch/PR per the task-store plan. HSR-1.1 only touches the schema, migration,
and `openapi-schemas.ts` — it intentionally does **not** update the consumer files
(`ready.ts`, `blocked-by.ts`, `task-service.ts`, `pull-request-service.ts`, `routes/prs.ts`)
that still reference the old `Task.hitlNotifiedAt` / `PullRequest.hitl` /
`PullRequest.hitlNotifiedAt` fields — that's the scope of HSR-1.2 through HSR-1.8,
which are already queued (`dependencies: ["HSR-1.1"]`) and will push further commits
onto this same branch/PR.

**CI is expected to be red on this commit** (typecheck fails in `pull-request-service.ts`,
`routes/prs.ts`, and several test files that still reference the removed fields) until
the remaining bundle tasks land. This is a known, self-declared interim state — see
acceptance criterion 5 below. Do not merge until all HSR-1.x tasks in this bundle are in.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `Task.requiresHumanApproval` exists; `Task.hitlNotifiedAt` removed; `PullRequest.blocked` exists; `PullRequest.hitl`/`hitlNotifiedAt` removed | MET |
[...]

PR #2498, opened 2026-08-08
fix(task-store): require active claim for legacy review-conflict guard

dmcaulay:
## Summary
- `PullRequestService.claim()`'s "legacy review conflict" guard 409'd purely on `commitSha` match + `reviewState !== 'pending'`, without checking `claimedBy` — so it blocked re-claiming an idle (unclaimed) PR whenever its review had already advanced past `pending` at the same commit.
- This defeated the fresh-author-reply re-candidacy exception in `agent/src/check-review.ts`'s `getReviewCandidates()` (#2450, #2461): a PR correctly re-queued for review after its author replied at an unchanged commit would still get 409'd on every claim attempt, permanently. Reproduced live on shipwright#2382 — stuck 45+ minutes across dozens of loop ticks, and (as later discovered) fully starving the whole dispatch loop, since #2382's age made it win FIFO selection every tick while never being excluded after its silent 409.
- Investigated whether the guard is dead code safe to delete outright — it isn't: `phase` is nullable and normally `null` while a PR is idle (reset on every `release()`/`complete()`), so the *primary* guard (which requires `phase` equality) can never catch a genuine concurrent double-claim during that idle window. The legacy guard is the only protection for that case. Fix: add `existing.claimedBy !== null` to it, matching the primary guard's condition — preserves real concurrent-claim protection, stops blocking re-claims when nobody actually holds the claim.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Legacy guard only throws `ConflictError` when `existing.claimedBy !== null` | MET |
| 2 | New integration test: claim() at unchanged commitSha with `reviewState !== 'pending'` and `claimedBy: null` succeeds (200), not 409 | MET |
| 3 | Existing "already claimed" conflict tests pass unmodified | MET |
| 4 | No unit test changes needed (integration-layer logic per repo convention) | MET |

## Test Plan
- New integration test added to `task-store/src/pull-request.integration.test.ts` (idle re-claim / fresh-author-reply case), verified RED (fails against pre-fix code) then GREEN.
- Full `pull-request.integration.test.ts` (69 tests) passes against a real local Postgres instance, including all existing concurrent-claim conflict tests unmodified.
- Full `task-store` package test suite: 607 pass, 0 fail.
- `task typecheck` and `task lint` clean across all workspaces.

### DIFF
--- a/docs/task-store.md
+++ b/docs/task-store.md
@@ -299,8 +299,8 @@ Body:
 Claim semantics:
 - No existing record → creates and returns `201`
 - Same `commitSha`, same `phase`, and already claimed by another agent → returns `409` (phase already locked)
-- Same `commitSha` and `reviewState !== pending` (review phase only) → returns `409` (already reviewed at this commit)
-- Different `commitSha` or `reviewState === pending` → updates and returns `200` (new cycle)
+- Already claimed (claimedBy !== null) AND same `commitSha` AND `reviewState !== pending` (review phase only) → returns `409` (already reviewed at this commit)
+- Not claimed, OR different `commitSha`, OR `reviewState === pending` → updates and returns `200` (new cycle)
 
 The `taskId` field is optional and does not trigger any side effects on the Task table — it is stored as metadata on the PR record only for reference.
 
--- a/task-store/src/pull-request-service.ts
+++ b/task-store/src/pull-request-service.ts
@@ -5,8 +5,9 @@
  *
  * claim() is atomic via a Prisma $transaction:
  *   1. Find existing record by @@unique([repo, prNumber])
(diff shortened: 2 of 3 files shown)


dmcaulay (commented):
Verdict: APPROVE — Narrow, correctly reasoned fix. The added `existing.claimedBy !== null` condition on the legacy review-conflict guard closes the idle-reclaim deadlock (blocked fresh-author-reply re-candidacy on shipwright#2382) without reopening a real concurrent-claim gap — the primary guard still fires once `claimedBy`+`phase` are set. Covered by a targeted integration test at the correct layer; docs and comments match the code.

PR #2500, opened 2026-08-08
fix: exclude 409'd items from same-tick re-selection in drain loop (RCG-1.2)

dmcaulay:
## Summary
- Excludes clean-409'd items (`claimTask`/`claimPr` resolving false/null, not throwing) from same-tick re-selection in `agent/src/loop-orchestrator.ts`'s drain loop, mirroring the existing thrown-error exclusion (CBD-2.1)
- Without this, an item that always loses the claim race (e.g. legitimately belongs to a sibling replica for the whole tick) gets re-collected and re-selected every iteration, starving every other candidate in every phase and repo for the tick's duration — live-reproduced by RCG-1.1's claim-guard bug fully starving the shipwright-loop cron for 3.5+ hours
- Adds two new unit tests mirroring the existing throw-path twins, asserting claim attempt count stays at 1 after a clean 409

## Acceptance Criteria
| 1 | Task-path 409 adds itemId to failedPreClaimTaskIds before continuing | MET | `failedPreClaimTaskIds.add(itemId);` before the task-path `continue` |
| 2 | PR-path 409 adds itemId to failedPreClaimPrIds before continuing | MET | `failedPreClaimPrIds.add(itemId);` before the PR-path `continue` |
| 3 | New unit test mirroring :2853 (task-path, claimTask resolves false) | MET | asserts `claimAttempts === 1`, `getCandidatesCalls === 2` |
| 4 | New unit test mirroring :2928 (PR-path, claimPr resolves null) | MET | asserts `claimAttempts === 1`, `getCandidatesCalls === 2` |
| 5 | No existing tests changed or retired | MET | diff only adds test hunks, zero deletions |

## Test Plan
- `bun test agent/src/loop-orchestrator.unit.test.ts` — 99/99 pass (97 pre-existing + 2 new)
- `task typecheck`, `task lint`, `task check-strings` — clean
- `task ci` full suite — 1 pre-existing failure (`extraAllowedTools` flake in `agent/src/claude.unit.test.ts`), confirmed reproducing on clean `main`, unrelated to this change


dmcaulay (commented):
Verdict: APPROVE — Correct, symmetric fix mirroring the proven CBD-2.1 throw-path exclusion pattern for the clean-409 case; new tests genuinely bound claim-attempt count, scope is tight, no CLAUDE.md or breaking-API concerns.

dodizzle (approved):
Verdict: APPROVE — Correct, tightly-scoped fix: verified failedPreClaimTaskIds/failedPreClaimPrIds are genuinely read as filters on the next candidate-collection pass, no regression to the shared-Set CBD-2.1 throw path, and the two new tests properly bound claim-attempt count to 1 after a clean 409 on both the task and PR paths.

PR #2526, opened 2026-08-10
fix: close PR claim()'s lost-update race

dodizzle:
## Summary
- `PullRequestService.claim()` re-validates its conflict guards (modern phase guard + legacy commitSha/reviewState guard) directly in the UPDATE's SQL `WHERE` clause instead of only in a preceding JS `if`, closing a lost-update race where two concurrent claimants could both pass a stale `findUnique()` snapshot under READ COMMITTED and the second writer would silently clobber the first's claim.
- Mirrors the existing single-conditional-UPDATE pattern already used by `TaskService.claim()` and the `claimedBy: null` optimistic-lock filter in `claimNext()`.
- When the WHERE clause matches 0 rows (Prisma P2025), a follow-up `findUnique` disambiguates: row still present → `ConflictError(409)` (unchanged message); row deleted → `NotFoundError`.
- Refreshed `docs/task-store.md`'s claim semantics section to describe the atomic WHERE-clause re-check.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Conflict conditions re-validated in the UPDATE's own SQL WHERE clause, not only a preceding JS `if`; a write that no longer matches affects 0 rows and surfaces as the existing `ConflictError(409)` | MET |
| 2 | Conflict-detection semantics unchanged from the caller's perspective — internal atomicity fix only | MET |
| 3 | New integration test in `pull-request-service.integration.test.ts` races two concurrent `claim()` calls against the same PR/commitSha on real Postgres, asserting exactly one succeeds and the other throws `ConflictError`; existing unit tests kept as-is | MET |

## Test Plan
- New `task-store/src/pull-request-service.integration.test.ts` (3 tests, gated on `DATABASE_URL_SHIPWRIGHT_TASK_STORE_TEST`, runs against real Postgres in CI):
  - Concurrent claims against an existing record: exactly one wins, the other gets `ConflictError`
  - Sequential claim guard preserved
  - Legacy review guard still surfaces `ConflictError` via the WHERE clause
- `task typecheck` and `task lint` pass
- `task test` — full suite passes except one pre-existing, unrelated flaky test (`agent/src/claude.unit.test.ts` extraAllowedTools) confirmed failing identically on a clean `origin/main` checkout; GitHub Actions CI on that same commit (`a7364d4c`) is green, so this is a local-sandbox-only flake, not a regression from this change.

### DIFF
--- a/docs/task-store.md
+++ b/docs/task-store.md
@@ -297,11 +297,12 @@ Body:
 | `phase` | no | Pipeline phase (`review`, `patch`, or `deploy`; default: `review`). When set, the phase is updated and reviewState is preserved. Phase-specific behavior on record creation: `review` sets `readyForReviewAt=now`; `deploy` sets `readyForDeployAt=now`; `patch` does not set a ready timestamp. |
 | `prCreatedAt` | no | ISO timestamp of the GitHub PR's actual creation time. Only applied when the claim creates a new record (`201`); ignored on subsequent claims (`200`) of an existing record since the field is immutable once set. |
 
-Claim semantics:
-- No existing record → creates and returns `201`
-- Same `commitSha`, same `phase`, and already claimed by another agent → returns `409` (phase already locked)
-- Already claimed (claimedBy !== null) AND same `commitSha` AND `reviewState !== pending` (review phase only) → returns `409` (already reviewed at this commit)
-- Not claimed, OR different `commitSha`, OR `reviewState === pending` → updates and returns `200` (new cycle)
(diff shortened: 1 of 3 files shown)


dmcaulay (approved):
Verdict: APPROVE — Correct fix for a genuine lost-update race in claim(); the WHERE-clause conflict re-check is an exact structural mirror of the original JS guards, the P2025 disambiguation runs same-transaction (no TOCTOU), no breaking changes, and the new integration tests are well-targeted.

PR #2528, opened 2026-08-10
fix(task-store): defense-in-depth on Task claim()

dodizzle:
## Summary
- Tighten `TaskService.claim()`'s atomic UPDATE `WHERE` clause to also require `"claimedBy" IS NULL`, so the "pending implies claimedBy null" invariant is enforced inside `claim()` itself instead of relying on every other write path (`release()`, `StaleClaimReaper`, the TPL-1.1 route-level PATCH denylist) to uphold it.
- A task with `status='pending'` and a non-null `claimedBy` now fails to claim (`ConflictError`, 409) instead of silently overwriting the existing claim.
- Added unit test coverage for the guard, the unchanged happy path, and the existing 404/409 disambiguation — verified the new guard test actually regresses (fails) when the WHERE clause fix is reverted, confirming it's not a tautological test.
- Refreshed `docs/task-store.md`'s `POST /tasks/:id/claim` description to reflect the tightened WHERE clause.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `claim()`'s raw SQL adds `AND "claimedBy" IS NULL` to WHERE clause | MET |
| 2 | pending+non-null-claimedBy no longer claimable, throws ConflictError | MET |
| 3 | Normal claim (pending+null claimedBy → success) unchanged; existing tests pass unmodified | MET |
| 4 | One new unit test added for the guard case; no existing tests retired | MET |

## Test Plan
- `bun test task-store/src/task-service.unit.test.ts` — 43/43 pass (4 new tests: happy path, claimedBy guard, not-found, non-pending status)
- Manually verified regression detection: reverted the WHERE clause guard locally, confirmed the new ConflictError test fails; restored the fix, confirmed it passes again
- `task typecheck` and `task lint` — clean
- `task ci` (full monorepo `test:coverage` gate) — 1 unrelated pre-existing failure in `task-store/src/routes/prs.openapi.smoke.test.ts` (a `validateRepo`/`repos.includes` cross-suite test-isolation flake, unrelated to this change — doesn't touch `prs.ts`, doesn't reproduce when that test file runs in isolation, and reproduces identically on `main` under the same full-suite conditions)

### DIFF
--- a/docs/task-store.md
+++ b/docs/task-store.md
@@ -169,7 +169,7 @@ Returns `204`. Agent tokens can only delete their own tasks.
 POST /tasks/:id/claim
 ```
 
-Atomically claims a pending task — a single conditional `UPDATE ... WHERE status='pending'`. Sets `status=in_progress`, `claimedBy`, `claimedAt`, `heartbeatAt`, and `startedAt` (or keeps existing if already set) in one round-trip. No request body is sent by agent tokens — the service pins `claimedBy` to the calling agent's ID server-side. Admin tokens must supply `{ claimedBy: string }` in the body. Returns `200` with the updated task on success, or `409` if already claimed or not in pending status.
+Atomically claims a pending task — a single conditional `UPDATE ... WHERE status='pending' AND "claimedBy" IS NULL`. Sets `status=in_progress`, `claimedBy`, `claimedAt`, `heartbeatAt`, and `startedAt` (or keeps existing if already set) in one round-trip. No request body is sent by agent tokens — the service pins `claimedBy` to the calling agent's ID server-side. Admin tokens must supply `{ claimedBy: string }` in the body. Returns `200` with the updated task on success, or `409` if already claimed or not in pending status.
 
 #### Heartbeat
 
--- a/task-store/src/routes/tasks.ts
+++ b/task-store/src/routes/tasks.ts
@@ -75,8 +75,10 @@ async function readJson(c: {
 // assignee/claimant match), so without this guard a repo-scoped agent token
 // could overwrite an actively claimed task's status back to "pending",
 // making it immediately re-claimable by a different agent while the
-// original session still holds it (TaskService.claim()'s atomic UPDATE
-// gates solely on status = 'pending', never on claimedBy IS NULL).
+// original session still holds it (TaskService.claim()'s atomic UPDATE now
+// also gates on claimedBy IS NULL, but a PATCH bypassing /claim or /release
+// could still directly clear claimedBy/status, defeating that protection
+// outside the atomic claim path).
 const LIFECYCLE_GUARD_KEYS = ["claimedBy", "claimedAt", "heartbeatAt"] as const;
 
 // admin tokens (agentId === null) are unrestricted, mirroring the existing
--- a/task-store/src/task-service.ts
+++ b/task-store/src/task-service.ts
@@ -452,7 +452,7 @@ export class TaskService implements TaskServiceLike {
           "heartbeatAt" = ${now},
           "startedAt" = COALESCE("startedAt", ${now}),
           "updatedAt" = now()
-      WHERE id = ${id} AND status = 'pending'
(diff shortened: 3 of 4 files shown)


dodizzle (commented):
Verdict: COMMENT — Core logic and tests are solid (WHERE clause change verified correct, guard test is genuinely non-tautological), but two things should be fixed before merge: (1) `task-store/src/routes/tasks.ts:78-79`'s comment justifying the PATCH lifecycle-field denylist now falsely says `claim()` "gates solely on status = 'pending', never on claimedBy IS NULL" — that's no longer true after this PR's WHERE-clause change, and it's the documented rationale for a still-necessary but now-mischaracterized guard. (2) The PR title fails the required `pr-title-lint` Conventional Commits check (CI is red) — rename to e.g. `fix(task-store): defense-in-depth on Task claim()`.

dmcaulay (commented):
Verdict: COMMENT — No new issues found; the claim() WHERE-clause fix is correct, well-tested, and non-breaking. Both findings from the prior review round at 95952ee (stale tasks.ts comment, PR title lint) are verifiably fixed in this commit. Posted as COMMENT rather than APPROVE only because the prior round's review has no recorded PR-author reply confirming resolution.

dodizzle:
Reviewed the finding(s) above and did not implement changes — both were already fixed in the current HEAD (commit db5307fc):

- Stale tasks.ts:78-79 comment (dodizzle, 05:41:41Z): already corrected — the comment now accurately describes claim()'s atomic UPDATE gating on claimedBy IS NULL.
- PR title fails pr-title-lint (dodizzle, 05:41:41Z): already renamed to "fix(task-store): defense-in-depth on Task claim()", matching Conventional Commits format. CI's pr-title-lint check is green.

Both fixes were independently confirmed by @dmcaulay's follow-up review at 06:21:11Z. Posting this reply so the review is recognized as addressed.

dmcaulay (commented):
Verdict: COMMENT — Prior findings (stale tasks.ts comment, PR title lint) are confirmed fixed at this commit. One new issue: the claim() docstring in task-store/src/task-service.ts (line 441, just above the modified WHERE clause) still says `WHERE id = $1 AND status = 'pending'`, omitting the new `AND "claimedBy" IS NULL` clause added two lines below — same class of staleness already caught and fixed once this cycle in tasks.ts. Not inline-commentable since that line isn't part of the diff hunk. Fix and this is a clean approve.

dodizzle (commented):
Verdict: COMMENT — All prior findings verified fixed at this commit: the `routes/tasks.ts` comment and `claim()` JSDoc now correctly describe the `claimedBy IS NULL` guard, and the PR title passes `pr-title-lint`. The WHERE clause change, the `ConflictError` guard path, and the new unit tests (double parses the real `$executeRaw` SQL text, so the guard test is non-tautological) all check out against the current source — no new issues found. Posted as COMMENT rather than APPROVE: this is a self-review (GitHub blocks self-APPROVE via the API), and @dmcaulay's round-3 review has no PR-author reply confirming the fix yet.

dodizzle:
Reviewed the finding above and did not implement additional changes — it was already
fixed in the current HEAD (commit d71c32fe, "fix: address review findings on #2528 —
correct stale claim() docstring"):

- Stale claim() docstring in task-service.ts:441 (dmcaulay, 07:14:27Z): already
  corrected — the docstring now reads `WHERE id = $1 AND status = 'pending' AND
  "claimedBy" IS NULL`, matching the actual WHERE clause two lines below.

Posting this reply so the review is recognized as addressed.

dmcaulay (approved):
Verdict: APPROVE — All findings from three prior review rounds (stale tasks.ts comment, PR title lint, stale claim() docstring) are verified fixed at this commit. The WHERE-clause fix is correct and atomic, the new ConflictError guard test genuinely regresses when reverted (independently re-verified), and the full suite passes (43/43). No new issues found.

PR #2565, opened 2026-08-11
fix: admin blocked-task badge/detail render crash + BlockedByEntry parity test

dodizzle:
## Summary
- Add the missing `{ type: "blocked"; reason: string | null }` variant to admin's hand-mirrored `BlockedByEntry` type, matching `task-store/src/blocked-by.ts` exactly.
- Fix `renderBlockerBadges` and the task-detail blockers renderer to handle this variant (escaped `reason`, "Blocked" fallback when null) instead of falling through to `taskLink(b.id)`/`escapeHtml(b.status)`, which crashed with `undefined` for this variant (Sentry issue 7665355727).
- Add a compile-time-only type-parity check (`admin/src/blocked-by-type-parity.check.ts` + dedicated `tsconfig.type-parity.json`) that fails the build if task-store's real union diverges from admin's mirror in the future — no runtime import, no new `admin/package.json` dependency.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `BlockedByEntry` gains `{ type: "blocked"; reason: string \| null }` matching task-store exactly | MET |
| 2 | `renderBlockerBadges` adds a `b.type==="blocked"` branch rendering escaped `reason` with fallback | MET |
| 3 | Task-detail blockers renderer gets the identical fix | MET |
| 4 | Compile-time-only type-parity test added, no runtime import, no new dependency | MET |
| 5 | Unit tests cover both renderers × {populated reason, null reason}, asserting no throw + correct escaped HTML | MET |

## Test Plan
- 6 new unit tests in `admin-ui-pages.unit.test.ts` (populated reason, null reason, XSS-escaping — 3 per renderer), all reproducing the exact Sentry crash before the fix and passing after.
- Full admin unit suite: 574 pass / 0 fail.
- `task lint`, `task check-strings`, `task check-version-sync`, `task check-config-docs`, `task typecheck` (all 7 packages, including the new type-parity project) all pass cleanly.
- Coverage: `admin/src/admin-ui-pages.ts` at 97.8%; aggregate repo-wide coverage 81.36% lines / 82.31% functions (threshold 80%/80%) — gate passes.
- `task test:coverage` has one pre-existing, unrelated failure (`extraAllowedTools > empty extraAllowedTools produces identical args...` in `agent/src/claude.unit.test.ts`) — confirmed to reproduce identically on `main` with none of this PR's changes applied.

### DIFF
--- a/admin/package.json
+++ b/admin/package.json
@@ -9,7 +9,7 @@
     "postinstall": "prisma generate --schema=prisma/schema.prisma",
     "test": "NODE_ENV=test bun test",
     "test:e2e": "bunx playwright test",
-    "typecheck": "tsc --noEmit",
+    "typecheck": "tsc --noEmit && tsc --noEmit -p tsconfig.type-parity.json",
     "db:generate": "prisma generate --schema=prisma/schema.prisma",
     "db:migrate": "prisma migrate deploy --schema=prisma/schema.prisma"
   },
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -83,7 +83,8 @@ function relativeTime(date: Date, now: Date = new Date()): string {
 // Inline type — mirrors task-store/src/blocked-by.ts without cross-package coupling.
 export type BlockedByEntry =
   | { type: "hitl"; notified?: true }
-  | { type: "dependency"; id: string; status: string };
+  | { type: "dependency"; id: string; status: string }
+  | { type: "blocked"; reason: string | null };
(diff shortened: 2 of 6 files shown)


dmcaulay (approved):
Verdict: APPROVE — Clean fix for the Sentry crash (missing `blocked` variant in admin's mirrored `BlockedByEntry`), verified against task-store's real type, with thorough test coverage and a useful compile-time parity guard against future drift.

PR #2580, opened 2026-08-11
fix: fall back to own-assignee match for repos-absent agent tokens in listReady/listBlocked

dodizzle:
## Summary
- Fix `listReady()` and `listBlocked()` in `task-service.ts` so a repos-absent agent token with a caller-supplied `assignee` filter falls back to the token's own tasks instead of an always-empty result on mismatch. Adds an `effectiveFilters()` helper that strips `assignee` from the filters object when `agentId` is present and `repos` is undefined/empty; `matchesTaskFilters()` itself is unchanged, so the admin-token and repo-scoped cases are untouched.
- Adds 8 new unit tests to `task-service.unit.test.ts` (4 per method) covering the regression case, the empty-array-repos variant, the unaffected-when-matching case, and a regression guard proving the repo-scoped AND-narrowing still works.
- Marks the "Known divergence from what actually shipped" section in `planning/task-store-ready-filters/PLAN.md` as resolved, referencing this task/PR.
- Auto-refreshed `docs/task-store.md` (assignee filter table row + agent-token-visibility section) for clarity on the fallback behavior — the original wording (`docs/task-store.md:55`) was already directionally correct and needed **no wording change** per this task's acceptance criteria; the refresh only expanded it for precision, it didn't correct an error.

## Acceptance Criteria
| 1 | listReady()/listBlocked() skip assignee filter when agentId present + repos undefined/empty; still AND-narrow when repos non-empty; still standalone-filter when agentId absent; matchesTaskFilters() unchanged | MET | New `effectiveFilters()` helper in `task-service.ts`, used by both methods before calling `matchesTaskFilters()`; `matchesTaskFilters()` body untouched |
| 2 | PLAN.md divergence section marked resolved, referencing ARF-1.1; docs/task-store.md:55 confirmed needing no wording change | MET | PLAN.md section rewritten and references ARF-1.1 explicitly |
| 3 | Unit tests added for listReady()/listBlocked() covering the fallback regression case, matching-assignee case, and repo-scoped regression guard | MET | 8 new tests added; all 83 tests in the file pass; pre-existing admin-token test untouched |

## Test Plan
- `cd task-store && bun test src/task-service.unit.test.ts` — 83 pass, 0 fail
- `task typecheck` — all workspaces pass
- `task lint` — clean
- `task ci` — full gate passes except one pre-existing, unrelated failure in `agent/src/claude.unit.test.ts` (`extraAllowedTools` test, env-var-order-dependent) — reproduced identically on a clean `main` checkout, confirmed not caused by this change

### DIFF
--- a/docs/task-store.md
+++ b/docs/task-store.md
@@ -52,7 +52,7 @@ Query params:
 | `session` | string | Filter by planning session slug |
 | `repo` | string, repeatable | Filter by repo (`org/repo` format). Repeat the param to match any repo in the list (e.g. `?repo=org/a&repo=org/b`). A single `?repo=` behaves identically to before (exact match). |
 | `org` | string, repeatable | Filter by org — matches any repo whose `org/repo` string starts with `<org>/`. Repeat the param to match any of several orgs (e.g. `?org=org-a&org=org-b`). Combines with `repo` as an AND filter (both narrow the same result set). |
-| `assignee` | string | Filter by assignee (admin tokens only; agent tokens without repo scope see only their own tasks). When used with repo-scoped agent tokens under `agentScope`, acts as an additional AND filter narrowing the visible set. |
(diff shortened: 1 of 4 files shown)


dmcaulay (approved):
Verdict: APPROVE — Small, correct, well-tested bug fix: listReady()/listBlocked() now fall back to the token's own tasks instead of an always-empty result when a repos-absent agent token supplies a mismatched assignee filter. No breaking API changes, no silent failures, accurate docs.

PR #2585, opened 2026-08-12
fix: key check-review.ts terminal-skip on reviewedCommitSha instead of commitSha

dodizzle:
## Summary
- Changes `getReviewCandidates()`'s terminal-skip check in `agent/src/check-review.ts` to key on `record.reviewedCommitSha` instead of `record.commitSha` when deciding whether a PR has already been reviewed at its current head
- `commitSha` is shared claim-lock bookkeeping, overwritten by any unrelated claim/patch/deploy action — a bump from such an action could silently mask a PR that was never actually re-reviewed at its current commit (the same trap class the staged-check immediately below already guards against via `reviewedCommitSha`)
- Refreshes `docs/agent.md`'s description of the terminal-skip guard to match

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Terminal-skip check compares `record.reviewedCommitSha` (not `record.commitSha`) against `pr.headRefOid` | MET |
| 2 | A record whose `commitSha` matches head but `reviewedCommitSha` does not is treated as eligible, not silently skipped | MET |
| 3 | A record whose `reviewedCommitSha` genuinely matches head still correctly skips (existing behavior preserved) | MET |
| 4 | Deploying this fix may surface newly-eligible review candidates repo-wide (intended effect, not a regression) | Documented in comment |
| 5 | Unit tests cover both the regression case and the preserved-skip case; no existing tests retired | MET — 2 new tests + 7 fixture updates, 73/73 passing |

## Test Plan
- `bun test agent/src/check-review.unit.test.ts` — 73/73 passing, including 2 new RCO-1.2 regression tests
- `task lint` — clean, no fixes needed


dmcaulay (approved):
Verdict: APPROVE — Small, correct, well-tested fix: keys check-review.ts's terminal-skip on reviewedCommitSha instead of the shared claim-lock commitSha field, matching an existing precedent (the staged-review guard a few lines below) and the project's documented dedup principle. Regression + preserved-behavior tests added, all 7 affected fixtures consistently updated, 73/73 passing (independently verified). No blocking findings.

PR #2588, opened 2026-08-12
feat: add per-pass summary logging to pr-state-reconciler (RCO-1.5)

dodizzle:
## Summary
- Add a per-pass end-of-run summary log line to `pr-state-reconciler.ts`'s `reconcilePrState()` and to each of `reconcileReviewState()`'s two scans (pending and posted/CHU-2.4), reporting `evaluated`/`patched`/`skippedClaimed`/`errored` counts.
- Previously these passes only logged on individual record/list failures — a clean, fully successful pass produced zero log output, making it impossible to distinguish "ran and succeeded" from "never ran" from logs alone. This gap hampered a recent incident investigation into whether the reconciler's 45-minute passes were running on schedule.
- Purely additive logging: `reconcileRecord()`, `reconcileReviewStateRecord()`, and `reconcilePostedReviewStateRecord()` now return an outcome string (`"patched" | "no-op" | "skipped-claimed"`) instead of `void` so the calling loops can tally counts — no reconciliation decision or write (PATCH) behavior changed.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `reconcilePrState()` emits one summary log line per invocation with evaluated/patched/skippedClaimed/errored counts | MET |
| 2 | `reconcileReviewState()`'s pending scan and posted scan (CHU-2.4) each emit their own summary log line | MET |
| 3 | Summary emitted even with zero records, and even when every repo's list call fails (failure counted into errored) | MET |
| 4 | No reconciliation decision or write behavior changes — existing tests pass unchanged | MET |
| 5 | Unit tests assert summary counts for representative scenarios | MET |

## Test Plan
- 7 new unit tests in `agent/src/pr-state-reconciler.unit.test.ts` covering: mixed outcomes, zero-record passes, and all-repos-list-failure, for both `reconcilePrState` and each of `reconcileReviewState`'s two scans.
- Full `agent/` test suite passes (93/93 in the modified file; 2 pre-existing unrelated failures in `claude.unit.test.ts` confirmed present on `main` independent of this change, caused by this environment having `ANTHROPIC_FALLBACK_MODEL` set).
- `bun run --filter='*' --sequential typecheck` — all 7 workspaces pass with zero errors.

### DIFF
--- a/agent/src/pr-state-reconciler.ts
+++ b/agent/src/pr-state-reconciler.ts
@@ -79,9 +79,9 @@
  * on GitHub's MERGED confirmation, as before.
  */
 
-import { DEFAULT_CLAIM_TTL_MS } from "@shipwright/lib/claim-ttl";
 import { existsSync, statSync } from "node:fs";
 import { join } from "node:path";
+import { DEFAULT_CLAIM_TTL_MS } from "@shipwright/lib/claim-ttl";
 import {
   classifyReviewState,
   createPrRecordQuery,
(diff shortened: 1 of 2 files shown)


dodizzle (commented):
Verdict: APPROVE — Purely additive per-pass summary logging (evaluated/patched/skippedClaimed/errored), correctly threaded through every early-return branch including the worktree-cleanup try/catch fall-through, with matching new unit tests. No reconciliation decision or write behavior changes. Posted as COMMENT because this is a self-authored PR (GitHub blocks self-APPROVE via the API).

Issue #66, opened 2026-06-07
[plan] shipwright-harness-extraction phase 2


dodizzle:
## Plan: shipwright-harness-extraction (Phase 2)

Phase 2 of the harness extraction — migration tooling and per-agent cutover tooling. Follows Phase 1 (SHE-1.1 through SHE-5.2).

### Technical Design

**Migration via API (SHE-6.1)**
Script at `agent/scripts/migrate.ts`. Calls vitals-os `GET /accounts/agents?role=AGENT` to enumerate all agents, then for each: `GET /accounts/agents/{id}/config` (decrypted env vars + tools) and `GET /accounts/agents/{id}/crons`. Writes to shipwright admin API (`POST /admin/api/agents/{id}/envs`, crons, tools). No direct DB access — both sides handle their own encryption. Idempotent via upsert.

External dependency: vitals-os list endpoint + shipwright infra deployment (tracked in vitals-os session `shipwright-harness-extraction-infra`).

**Cutover tooling (SHE-7.1)**
Two scripts at `agent/scripts/`:
- `cutover-values.ts` — given agent ID + shipwright image tag, prints Helm values patch for client-agent chart (new env vars in, vitals-os env vars out)
- `cutover-validate.ts` — calls `GET /agents/{id}/config` + `GET /agents/{id}/crons` on shipwright, asserts key credentials and crons are present; exits 0 on pass, non-zero with per-check report on fail

### Task Table

| Task | Title | Layer | Hours | Deps |
|------|-------|-------|-------|------|
| SHE-6.1 | Data migration script | CLI | 3 | SHE-3.1 |
| SHE-7.1 | Cutover tooling | CLI | 3 | SHE-6.1, SHE-2.1 |

### Dependency Graph

```
[SHE-1.2] → [SHE-3.1] → [SHE-6.1] → [SHE-7.1]
[SHE-2.1] ──────────────────────────────────────┘
```

PR #2601, opened 2026-08-12
feat: wire same-pass resolution exclusion into the unaddressed-findings gate

dodizzle:
## Summary
- Adds a fourth exclusion to the shared unaddressed-findings module (`isResolvedByPriorFindingsStatus`): a prior qualifying CURRENT_USER review is excluded when the current pass's `priorFindingsStatus[]` (PVD-1.2) marks it `resolved: true` with non-empty evidence, judged independently per finding (Option B — no coupling to `currentPassHasBlockingFindings` or other findings in the same pass)
- Closes the structural deadlock confirmed live on ok-wow/ok-wow-agency#66, where a third-party PR could never reach an automated APPROVE once a single self-authored COMMENT finding had been posted and fixed, because each pass posts a new review object rather than rewriting a prior one's body
- Updates `review.md` Step 9.5 documentation to describe all four exclusions (clean self-APPROVE, CPF-2.3 author-reply, DRO-1.2 later-clean-self-review, and the new PVD-1.3 exclusion) plus the Option B design rationale
- Refreshes `docs/agent.md` to reflect the fourth exclusion

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Fourth exclusion added to shared module, evaluated independently per finding | MET |
| 2 | review.md Step 9.5 docs updated with all four exclusions + Option B rationale | MET |
| 3 | Regression unit test reproduces PR #66's shape (resolved:true + evidence → APPROVE) | MET |
| 4 | Second unit test confirms Option B independence (old finding excluded despite fresh unrelated blocking finding) | MET |
| 5 | Additive unit tests (incl. negative case: resolved:false or empty evidence does not exclude) + review.content.test.ts updated; no existing tests retired | MET |

## Test Plan
- `NODE_ENV=test bun test plugins/shipwright/scripts/compute-unaddressed-findings.unit.test.ts plugins/shipwright/commands/review.content.test.ts` — 158 pass
- `bunx biome lint .` — clean
- Full `task ci` run — 6044 pass, 3 pre-existing unrelated failures (env-dependent, confirmed failing on `main` too)


plugins/shipwright/commands/review.md
```diff
@@ -614,13 +614,20 @@ bun run "${CLAUDE_PLUGIN_ROOT}/scripts/compute-unaddressed-findings.ts" \
...
+    '{currentUser: $currentUser, headRefOid: $headRefOid, reviews: $reviews, reviewThreads: $reviewThreads, comments: $comments, priorFindingsStatus: $priorFindingsStatus}')"
 # -> {"unaddressedFindings":true|false}
 ```
 
 `HEAD_REF_OID`, `REVIEWS_JSON`, `REVIEW_THREADS_JSON`, and `COMMENTS_JSON` are the
 `headRefOid`, `reviews`, `reviewThreads`, and `comments` values from Step 5.5's GraphQL
-response — pass them through unchanged, do not re-fetch or re-shape them. Assign the script's
+response — pass them through unchanged, do not re-fetch or re-shape them. `PRIOR_FINDINGS_STATUS_JSON`
+is the subagent's `priorFindingsStatus[]` array parsed in Step 7 (PVD-1.2's `{ ref, resolved, evidence }`
+attestations — see Step 5.5's `priorQualifyingReviews` and Step 7). Each entry's `ref` is derived
```
dodizzle, line 629: Step 5.5 (lines 289-291, not touched by this PR) still defines `priorQualifyingReviews`'s `ref` as "a short identifying label for each review (e.g. the commit short SHA plus `submittedAt`...)". Step 7 tells the subagent to echo back `ref` "matching the `ref` passed in above" — i.e. whatever format Step 5.5 built. But `reviewRef()` (new in this PR) is the FULL untruncated commit oid + `submittedAt`, and `isResolvedByPriorFindingsStatus` does an exact string match against it. Unless Step 5.5 is also updated to build `ref` via `reviewRef()`'s exact format, the `ref` values flowing Step 5.5 → subagent → `priorFindingsStatus[]` → here will never equal `reviewRef()`'s output, so this exclusion — the whole point of the PR — will likely never fire in practice. The unit tests bypass this because they construct `ref` directly via the exported `reviewRef()` helper rather than going through Step 5.5's prose. Please update Step 5.5 to explicitly build `ref` using `reviewRef()`'s exact full-oid+submittedAt format.

dodizzle (commented):
Verdict: COMMENT — the new isResolvedByPriorFindingsStatus/reviewRef logic is solid and well-tested in isolation, but Step 5.5's `ref` construction (unchanged by this PR, still "short SHA + submittedAt") is inconsistent with the new full-oid `reviewRef()` contract this PR relies on, which would likely prevent the PVD-1.3 exclusion from ever matching end-to-end. See inline comment.

dmcaulay (approved):
Verdict: APPROVE — the fix commit fully resolves the prior review round's ref-format concern (Step 5.5's `priorQualifyingReviews.ref` now built via `reviewRef()`'s exact full-oid format), and the new PVD-1.3 exclusion (`isResolvedByPriorFindingsStatus`) is correct, well-guardrailed, and thoroughly tested across unit/content/CLI layers. No blocking concerns.

PR #2604, opened 2026-08-12
fix(chart): make task minikube:up reliable on Apple Silicon and behind the shared ingress rewrite-target

dodizzle:
## Summary
- Ordering race: `helm install` could run before the ingress admission webhook was ready — added a `kubectl wait` step in `scripts/minikube.ts`
- `metrics`, `taskStore`, and `chat` had no memory overrides in `values-minikube.yaml` (only `admin` did), too small for `prisma migrate deploy` at startup — brought all four in line at 512Mi
- Metrics liveness/readiness probes and the `helm test` connectivity check were hardcoded to un-prefixed paths, but the app mounts its whole router under `provider.basePath` when set — fixed both, with new `helm-unittest` coverage
- `nginx.ingress.kubernetes.io/rewrite-target: /$2` is an ingress-wide annotation, but only the task-store path used the two-capture-group regex it needs — it silently mangled the metrics and admin paths once task-store's exposure turned it on. Fixed with a no-op-rewrite capture-group pattern for both, conditional on whether task-store exposure is active
- `task minikube:up` now automatically starts a background `kubectl port-forward` to the ingress controller and prints working URLs, instead of the docker-driver's unreachable `minikube ip` on port 80 (which is also not the ingress controller's actual NodePort)

## Test plan
- [x] `bun test scripts/minikube.unit.test.ts` — 26 pass
- [x] `helm unittest charts/shipwright` — 303 pass across 22 suites
- [x] `bunx biome check` on all changed files
- [x] Full `task minikube:up` run against a live Minikube cluster — all 5 pods `Running`, `helm test` `Succeeded`
- [x] Verified ingress routing end-to-end via `kubectl port-forward` + `curl` with `Host: shipwright.local`: admin console, metrics dashboard, and task-store health all resolve correctly post-fix (previously 404/mangled)

### DIFF
--- a/charts/shipwright/Chart.yaml
+++ b/charts/shipwright/Chart.yaml
@@ -9,7 +9,7 @@ type: application
 # Chart version (SemVer). Bump on every chart change, independent of appVersion.
 # See CHANGELOG.md and the README "Versioning" section — CI enforces this bump
 # via `ct lint --check-version-increment`.
-version: 1.11.31
+version: 1.11.32
 # Version of the Shipwright application this chart deploys by default.
 appVersion: "0.1.0"
 home: https://shipwrightharness.com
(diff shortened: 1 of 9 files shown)


dodizzle (commented):
Verdict: APPROVE — Clean fix for the Minikube dev-stack reliability issues (webhook race, undersized memory, basePath probe mismatch, shared rewrite-target path mangling). Verified the core claims against source (metrics basePath mounting, existing task-store capture-group precedent) and the new bun/helm-unittest coverage lands at the correct layers. No blocking concerns.

dodizzle (commented):
Verdict: APPROVE — Re-review after a trivial chart version bump (1.11.31 → 1.11.32) to satisfy CI's version-increment lint. Re-verified the nginx rewrite-target no-op capture-group regex, metrics basePath mounting, and resource sizing against source — no new issues. Prior clean approval stands.

PR #2612, opened 2026-08-13
feat: write back terminal reviewState from review.md's Step 3 dedup

dmcaulay:
## Summary
- `review.md`'s Step 3 ("Check if the PR was already reviewed at the current commit", the task-store-based defense-in-depth dedup) now PATCHes the already-fetched PR record to a terminal `reviewState` (`posted`, `reviewedCommitSha` at current head) before the existing `[silent]` stop, whenever the record doesn't already reflect that terminal state.
- Mirrors RVD-2.2's write-back on RVD-1.2's live-GitHub check, and the existing "unresolved human feedback" write-back at Step 5 — reuses the record already fetched in this step (`PR_RECORD_ID`), no extra query.
- Added a content-test case to `review.content.test.ts` asserting the PATCH exists, is gated on the record not already reflecting a terminal state, precedes the `[silent]` stop, and documents the no-extra-query reuse.

## No double-PATCH with RVD-2.2
RVD-2.2's write-back lives on RVD-1.2's Live-Review Pre-Check (the live-GitHub `$terminal` check), which runs *before* Step 3 in `review.md`'s flow and stops with `[silent]` immediately once `$terminal == true`. Step 3 is only ever reached when RVD-1.2's check did *not* already stop the run — so RVD-1.2/RVD-2.2's write-back and Step 3/RVD-2.3's write-back can't both fire in the same `review.md` execution. They're independent, non-overlapping defense-in-depth paths guarding two different data sources (live GitHub vs. task-store record) that can each be reached without the other having run, but never both in a single run.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Step 3 PATCHes record to terminal reviewState before `[silent]`, reusing already-fetched record, no extra query | MET |
| 2 | Content test added mirroring RVD-2.2/unresolved-feedback pattern | MET |
| 3 | PR description confirms no double-PATCH/conflict with RVD-2.2 | MET (see above) |
| 4 | Safe to deploy standalone (no dependency on RVD-2.1/RVD-2.2) | MET |

## Test Plan
- [x] `bun test plugins/shipwright/commands/review.content.test.ts` — 113/113 pass
- [x] `bun test plugins/shipwright/` — 1530/1530 pass
- [x] `task typecheck` — passes
- [x] `task lint` — clean
- [x] `task check-strings` — no banned strings

### DIFF
--- a/plugins/shipwright/commands/review.md
+++ b/plugins/shipwright/commands/review.md
@@ -1192,12 +1192,12 @@ gh pr view {pr} --repo {org}/{repo} --json headRefOid --jq '.headRefOid'
      loop orchestrator's generic `[silent]` handling would count it toward
      `SKIP_BLOCK_THRESHOLD`, risking a false HITL auto-block (see
      `agent/src/loop-orchestrator.ts`).
-   - If `record.reviewState` is not already `posted` or `approved` (defense-in-depth
-     write-back — this step's own entry condition already requires that today, but this
-     guards against a future change to that condition reaching this branch with a record
-     state not yet exercised here, e.g. RVD-2.1 widening what upstream selection considers
-     eligible), PATCH the already-fetched record (`PR_RECORD_ID`) to a terminal state
-     before stopping — no extra query, reuses the record fetched in step 2 above:
+   - PATCH the already-fetched record (`PR_RECORD_ID`) to write back a terminal state
+     before stopping — no extra query, reuses the record fetched in step 2 above. This is
+     unconditional (mirroring the Step 5 precedent), not guarded on `record.reviewState`:
+     the entry condition above already requires `record.reviewState` to be `posted` or
+     `approved` to reach this point, so a guard re-checking the same unmutated field would
+     be dead code:
      ```bash
      curl -sf -X PATCH \
        -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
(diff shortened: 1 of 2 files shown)


plugins/shipwright/commands/review.md
```diff
@@ -1192,6 +1192,19 @@ gh pr view {pr} --repo {org}/{repo} --json headRefOid --jq '.headRefOid'
      loop orchestrator's generic `[silent]` handling would count it toward
      `SKIP_BLOCK_THRESHOLD`, risking a false HITL auto-block (see
      `agent/src/loop-orchestrator.ts`).
+   - If `record.reviewState` is not already `posted` or `approved` (defense-in-depth
```
dmcaulay: This guard (`record.reviewState` is not already `posted` or `approved`) is the exact negation of the condition that already gates entry to this whole block (lines 1176-1178: `record.reviewState is posted or approved`). Nothing between those two checks re-fetches or mutates `record.reviewState` — so this branch can never evaluate true and the PATCH can never fire. This differs from the RVD-2.2 precedent it's meant to mirror, where the guard (`$terminal`) comes from an independent live-GitHub query rather than the same already-filtered field. Even in the counterfactual where it fired, `record.reviewedCommitSha` already equals `headRefOid` here by construction, so the PATCH would be a no-op. Suggest dropping the guard (PATCH unconditionally, like the Step 5 precedent) or moving any future-proofing guard onto the entry condition itself.

dmcaulay (commented):
Verdict: COMMENT — the new write-back branch (review.md:1195) is unreachable dead code: its guard is the exact negation of the enclosing block's own entry condition on the same unmutated `record.reviewState` field, so the PATCH can never fire and Acceptance Criterion #1 isn't actually satisfied by this diff.

dodizzle (approved):
Verdict: APPROVE — dead-code guard bug from the prior review round is fully fixed and confirmed reachable; the write-back is unconditional, correctly scoped, and non-overlapping with the RVD-2.2/Step 5 write-back paths. New content-test assertions genuinely validate the new prose. No blocking concerns.

PR #2614, opened 2026-08-13
fix: write back terminal reviewState from Live-Review Pre-Check (RVD-2.2)

dodizzle:
## Summary
- `review.md`'s Live-Review Pre-Check (RVD-1.2) now writes back a terminal `reviewState` to the task-store PR record when it detects a terminal review already exists at head on GitHub, instead of only releasing the pre-claim and stopping silently.
- Determines `PR_RECORD_ID` by reusing `PRECLAIM_RECORD_ID` (if a pre-claim marker was present) or looking the record up by `repo`+`prNumber`, then PATCHes `{"reviewState": "posted", "reviewedCommitSha": "{headRefOid}"}` — but only when the record doesn't already reflect this outcome at the current head, avoiding a double-write.
- Fixes the recurring mis-selection this task-store instance would otherwise see every tick: without this write-back, a stale `pending` (or `posted`-at-an-older-commit) record keeps this PR re-selectable indefinitely, even after RVD-2.1 lands (RVD-2.1 only fixes `check-review.ts`'s selection path — this is `review.md`'s own separate GraphQL-based detection with no corresponding selection-side fix).

## Reconciler safety (acceptance criterion 3)
This write-back cannot be undone by `pr-state-reconciler.ts`'s CHU-2.4 `reconcileReviewState()` sub-pass. That pass's `reconcilePostedReviewStateRecord()` reverts a `posted` record back to `pending` only when `hasAnyReviewAtHead()` returns `false` (`agent/src/pr-state-reconciler.ts:1002`: `if (hasAnyReviewAtHead(reviewData)) return "no-op";`). This write-back only ever runs when `$terminal == true`, which by construction means a real, terminal-labeled review object was found at `headRefOid` by the GraphQL query — so `hasAnyReviewAtHead()` is guaranteed `true` for this record on the next reconcile pass, and the reconciler takes its no-op branch, leaving the record untouched.

## Acceptance Criteria

| 1 | PATCHes task-store PR record to reviewState:posted + reviewedCommitSha before stopping, no double-write when already correct | MET | New PATCH block in review.md's `$terminal==true` branch, guarded by a condition checking `recordReviewState` ∈ {posted,approved} AND `recordReviewedCommitSha == headRefOid` before skipping the write |
| 2 | Content test added to review.content.test.ts asserting PATCH + reviewState/reviewedCommitSha | MET | New `describe` block with 6 `it` cases asserting PATCH call, reviewedCommitSha/headRefOid, PR_RECORD_ID resolution, double-write avoidance, ordering, and reconciler-safety citations |
| 3 | Explicit verification that reconciler cannot undo this, citing hasAnyReviewAtHead() | MET | New "Caveat" paragraph citing `agent/src/pr-state-reconciler.ts`'s `reconcileReviewState()` → `reconcilePostedReviewStateRecord()` and its `hasAnyReviewAtHead()` guard, verified against actual source |
| 4 | Safe to deploy standalone, no dependency on RVD-2.1/RVD-2.3 | MET | Change is confined to review.md prose + its content test; no coupling to check-review.ts or other RVD tasks |

## Test Plan
- `NODE_ENV=test bun test plugins/shipwright/commands/review.content.test.ts` — 116/116 pass (6 new test cases)
- `task lint` — clean
- Full `task ci` / `test:coverage`: 3 pre-existing failures in unrelated files (`task-store/src/routes/prs.openapi.smoke.test.ts`, `metrics/src/lib/errors.unit.test.ts`, `agent/src/claude.unit.test.ts`) — confirmed unrelated since this diff only touches `review.md` and `review.content.test.ts`

### DIFF
--- a/plugins/shipwright/commands/review.content.test.ts
+++ b/plugins/shipwright/commands/review.content.test.ts
@@ -584,6 +584,81 @@ describe("review.md — RVD-1.2 terminal-review skip emits [silent] + skip-reaso
   });
 });
 
+describe("review.md — RVD-1.2 terminal-review skip writes back task-store reviewState (RVD-2.2)", () => {
+  it("the terminal-review branch contains a PATCH call to /prs/{PR_RECORD_ID} with reviewState:posted", () => {
+    const preCheckIdx = content.indexOf("### Live-Review Pre-Check (RVD-1.2)");
+    const fastPathIdx = content.indexOf("### Pre-Claim Fast Path (CBD-1.4)");
+    expect(preCheckIdx).toBeGreaterThan(-1);
+    expect(fastPathIdx).toBeGreaterThan(preCheckIdx);
+    const section = content.slice(preCheckIdx, fastPathIdx);
+
+    expect(section).toContain("PATCH");
+    expect(section).toContain("$SHIPWRIGHT_TASK_STORE_URL/prs/");
+    expect(section).toContain("PR_RECORD_ID");
+    expect(section).toContain("reviewState");
+    expect(section).toContain("posted");
+  });
+
+  it("the PATCH body sets reviewedCommitSha to headRefOid", () => {
+    const preCheckIdx = content.indexOf("### Live-Review Pre-Check (RVD-1.2)");
+    const fastPathIdx = content.indexOf("### Pre-Claim Fast Path (CBD-1.4)");
+    const section = content.slice(preCheckIdx, fastPathIdx);
+
+    const patchIdx = section.indexOf("PATCH");
+    expect(patchIdx).toBeGreaterThan(-1);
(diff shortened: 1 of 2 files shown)


dodizzle (commented):
Verdict: APPROVE — Clean, narrowly-scoped write-back fix; verified the reconciler-safety claim against pr-state-reconciler.ts and confirmed the PATCH correctly only touches reviewedCommitSha (not the claim-lock commitSha field, per Rule 2 of the design constitution). Tests pass (116/116), CI green. One narrow non-blocking gap noted in the review file: the write-back no-ops when no PR record exists at all yet (vs. existing-but-stale) — not worth blocking on.

PR #2615, opened 2026-08-13
feat: tag patch.md's Step 3d no-op-at-dispatch bail with a structured skip-reason

dmcaulay:
## Summary
- Tags `patch.md`'s Step 3d "no PRs need attention" bail with a structured `[skip-reason:patch:deferred:no-op-at-dispatch:{pr}]` marker, emitted alongside the existing `[silent]`, matching the tagging convention already used by other legitimate-wait defers (e.g. `bundle-incomplete`).
- This lets the loop orchestrator's `SKIP_BLOCK_THRESHOLD` handling record exactly which PR found no work in `AgentCronRun.skipReason`, instead of falling back to the generic `command:no-work` reason.
- Refreshes `docs/agent.md`'s `AgentCronRun.skipReason` field description to include this new marker example alongside the existing `deploy` one.

## Why no task-store write-back (intentional, not an oversight)
Unlike `review.md`'s RVD-2.2/2.3 gaps, this is **not** a persisted-state drift bug. `check-patch.ts`'s `getPatchCandidates()` re-derives DIRTY/CI/findings status fresh from live GitHub on every tick — there's no cached "needs patch" field on the task-store record that can go stale, unlike `review.md`'s `classifyReviewState()`-based selection. Reaching Step 3d most likely reflects a genuine race (CI went green, or a human fixed the issue directly, between candidate selection and dispatch), not a drift that will recur on the next tick. A design note was added directly in `patch.md`'s Step 3d explaining this, so a future reader doesn't assume the missing write-back was an oversight. If telemetry later shows this recurring for the same PR repeatedly, that's the signal to open a real write-back follow-up task — not something to speculatively build now.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Step 3d emits `[skip-reason:patch:deferred:no-op-at-dispatch:{pr}]` immediately before `[silent]`, interpolating the PR number | MET |
| 2 | New test case in `patch.content.test.ts` asserting the tag, following the `bundle-incomplete` pattern | MET |
| 3 | No task-store write-back added; rationale documented in-file so it doesn't read as an oversight | MET |
| 4 | Safe to deploy standalone, independent of RVD-2.1/2.2/2.3 | MET |

## Test Plan
- [x] Added `describe("patch.md — no-op at dispatch skip-reason tag (RVD-2.4)")` in `plugins/shipwright/commands/patch.content.test.ts` (3 new test cases; RED confirmed before the `patch.md` edit, GREEN after).
- [x] `bun test plugins/shipwright/commands/patch.content.test.ts` — 146/146 pass.
- [x] `task typecheck` — passes across all workspaces.
- [x] `task ci` — 1 pre-existing failure (`agent/src/claude.unit.test.ts`'s `extraAllowedTools` test), unrelated to this diff — known in-sandbox flake, reproduces on clean `main`.


dodizzle (approved):
Verdict: APPROVE — Small, well-verified prose-only addition of a structured skip-reason marker to patch.md's Step 3d no-op bail, following deploy.md's existing bundle-incomplete pattern, with a matching content test. No blocking issues.

PR #2633, opened 2026-08-13
fix: dont re-gate an already-reviewed PR on the author allowlist

dmcaulay:
## Summary
- `getReviewCandidates()` in `agent/src/check-review.ts` now only applies the author-allowlist exclusion when no `PrRecord` exists yet for the PR — once a PR has already cleared the allowlist gate and entered the review pipeline (a `PrRecord` exists), subsequent commits from the same author stay eligible for re-review instead of being permanently re-excluded.
- Added a `authorAllowlistRef: string[]` field to the `not-allowlisted` `ReviewCandidacyTrace` variant, populated from the live `agentAuthorAllowlistRef` singleton at decision time, so a future repeat of the ok-wow/ok-wow-ai#1919 mystery is diagnosable from Sentry logs alone.
- Refreshed `docs/agent.md` to reflect the new allowlist-gating timing.

## Acceptance Criteria
| 1 | New unit test: non-allowlisted author + existing PrRecord NOT excluded, proceeds to normal dedup checks | MET | `check-review.unit.test.ts` — "isAuthorAllowed does not re-exclude a non-allowlisted author once a PrRecord already exists (AAL-3.2)" |
| 2 | Existing coverage: non-allowlisted author, no PrRecord, still excluded (initial-gate behavior unchanged) | MET | `check-review.unit.test.ts` — "isAuthorAllowed still excludes a non-allowlisted author when no PrRecord exists yet (AAL-3.2: initial-gate behavior unchanged)" |
| 3 | Bonus: debug log with live authorAllowlistRef at exclusion point | MET | `ReviewCandidacyTrace`'s `not-allowlisted` variant now carries `authorAllowlistRef: agentAuthorAllowlistRef.get()` |

## Test Plan
- `bun test agent/src/check-review.unit.test.ts` — 93/93 pass (new + existing tests green)
- `task lint`, `task check-strings`, `task typecheck` (agent package) — clean
- `task ci` full suite — 6088 pass, 1 pre-existing unrelated flake (`agent/src/claude.unit.test.ts > extraAllowedTools`, reproduces identically on clean `main`)


dmcaulay (commented):
Verdict: COMMENT — Solid fix for the ok-wow-ai#1919 re-gating bug, but the queryPrRecord fetch reorder introduces an untested edge case: a transient fetch failure now collapses into "no record" and can re-trigger the exact permanent-exclusion bug this PR fixes.

agent/src/check-review.ts
```diff
@@ -403,7 +416,35 @@ export async function getReviewCandidates(
...
+    // self-request-defeat concern the isRequestedReviewer bypass above
+    // exists to avoid: a non-allowlisted author still cannot manufacture an
+    // initial PrRecord solo — reaching this record-exists branch still
+    // requires having passed the allowlist gate (or the requested-reviewer
+    // bypass) at least once, exactly as today. The live authorAllowlistRef
+    // value is logged alongside the decision so a repeat of the ok-wow-ai#1919
+    // mystery (how the first two cycles got through despite the allowlist)
+    // is diagnosable from Sentry logs alone, without needing to reproduce.
     if (
+      record == null &&
```
dmcaulay: The allowlist re-gate exception is keyed on `record == null`, but `record` is also `null` when `queryPrRecord` throws (the catch block above swallows the error with no distinguishing state). Before this diff, a `queryPrRecord` failure never touched the allowlist decision since the fetch happened after that check. Now, if the fetch transiently fails for a PR that already has a legitimate `PrRecord` (already cleared the allowlist once, non-allowlisted author), the code can't tell "fetch failed" apart from "record never existed" — the allowlist re-fires and wrongly excludes an actively-reviewed PR. That's the exact ok-wow-ai#1919 failure mode this PR exists to fix, reintroduced for the duration of any transient failure. The existing "queryPrRecord throws" test doesn't set `isAuthorAllowed`, so this interaction is untested. Consider distinguishing "fetch failed" from "no record" explicitly and skipping the re-gate only on a genuine fetch failure.

dodizzle (approved):
Verdict: APPROVE — Clean fix for the AAL-3.2 allowlist re-gating bug; the queryPrRecordFailed guard correctly avoids conflating a fetch failure with "no record yet," and all three acceptance criteria are met with tests that verify what they claim.

dodizzle (approved):
Verdict: APPROVE — Re-reviewed after the head moved via a merge from main (no change to the PR's own diff). Still a clean, well-scoped fix for the AAL-3.2 allowlist re-gating bug with correct test coverage. Note: on closer trace, `queryPrRecordFailed` is logically redundant with `record == null` (a thrown `queryPrRecord()` call can never leave `record` non-null), so the comment claiming it "distinguishes a transient fetch failure from a genuine no-record state" overstates what the code does — worth a follow-up comment cleanup, but not blocking.

PR #2634, opened 2026-08-13
feat: treat review-thread replies as a fresh-author-reply signal in check-review dedup

dodizzle:
## Summary
- `hasFreshAuthorReply` in `check-review.ts` now also treats a PR author's inline reply on a review thread as a fresh-author-reply signal, not just top-level PR comments.
- `fetchPrReviews`'s GraphQL query widens `reviewThreads[].comments` from `first: 1` (no `createdAt`) to `last: 20` with `createdAt`, so replies past the thread-opening comment are visible.
- `ReviewThread.comments.nodes` (shared type in `check-patch.ts`) gains an optional `createdAt` field — additive only, no changes needed in `check-patch.ts` or `pr-state-reconciler.ts`, which don't read thread-comment authorship/freshness.

## Why
GitHub represents an inline reply on a review thread as its own zero-body `COMMENTED` review stamped to head — indistinguishable from a real review to `hasAnyReviewAtHead()`. Previously, `hasFreshAuthorReply` only scanned top-level comments, so an author who replied inline (rather than posting a new top-level comment) never tripped the "fresh reply" exception, and the PR was incorrectly skipped as `already-reviewed-live`. Confirmed live on verygood-ops/skills#80.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `ReviewThread.comments.nodes` gains optional `createdAt`; query widened to `comments(last: 20)` with `createdAt` | MET |
| 2 | `hasFreshAuthorReply` returns true for a `reviewThreads` reply authored by `pr.author.login` after `reviewedAtMs`, OR'd with the existing top-level check | MET |
| 3 | New unit test(s) covering an inline review-thread reply flipping `already-reviewed-live` → `eligible`, sourced from `reviewThreads` instead of `comments` | MET |
| 4 | `check-patch.ts` / `pr-state-reconciler.ts` typecheck unchanged against the widened (additive, optional) type | MET |

## Test Plan
- `bun test agent/src/check-review.unit.test.ts` — new cases in both `describe("getReviewCandidates", ...)` (real production computation, exercises the actual `hasFreshThreadReply` code path) and `describe("traceReviewCandidacyDecision", ...)` (pure decision-logic pass-through), plus a boundary case (non-author or stale thread reply does not flip the outcome).
- `bun run --filter='*' --sequential typecheck` — clean across all workspaces, confirming `check-patch.ts`/`pr-state-reconciler.ts` compile unchanged against the widened shared type.


dodizzle (commented):
Verdict: APPROVE — Clean, well-scoped fix: widens hasFreshAuthorReply to also catch inline review-thread replies (not just top-level comments), with an additive-only type change and tests that exercise the real production code path. No blocking concerns.

PR #2646, opened 2026-08-14
docs(planning): add patch-comment-findings plan

dmcaulay:
## Summary
- Adds `planning/patch-comment-findings/PLAN.md`, the plan for RCT-1.1/RCT-1.2, queued to the task store.
- Root cause (confirmed against source, discussed with Dan in Slack before planning): a reviewer's blocking feedback left as a plain PR comment (not a formal GitHub review) is invisible to the automated patch loop forever — `hasUnaddressedFindings()` only treats formal review objects as findings.
- Fix lands at the review layer, not patch: broaden `check-review.ts`'s fresh-reply override (and `review.md`'s RVD-1.2 dispatch-time mirror) from "PR author replied" to "anyone but the reviewing agent commented" — so a third-party comment triggers a re-review, and review's own judgment either produces a real formal finding (which patch already knows how to act on) or a no-op. No changes needed to `patch.md`/`check-patch.ts`/`compute-unaddressed-findings.ts`.

## Test plan
- [ ] No code changes in this PR — plan doc only, tasks queued to the task store for `dev-task` to execute.

### DIFF
--- a/planning/patch-comment-findings/PLAN.md
+++ b/planning/patch-comment-findings/PLAN.md
@@ -1,0 +1,149 @@
+# Plan Session: patch-comment-findings
+
+**Repo:** app-vitals/shipwright
+**No PRODUCT-SPEC.md** — bug-fix session, scoped in Slack with Dan McAulay before planning.
+
+## Problem
+
+Another Shipwright-deployed agent (not App Vitals') reported that a reviewer's blocking
+feedback on a PR, posted as a plain GitHub PR comment (not a formal "Review changes"
+submission), was never picked up by the automated patch loop no matter how many cron
+cycles ran.
+
+Root cause, confirmed directly against `app-vitals/shipwright` source:
+`hasUnaddressedFindings()` in `plugins/shipwright/scripts/compute-unaddressed-findings.ts`
+(the single source of truth `agent/src/check-patch.ts`'s `getPatchCandidates` imports)
+only treats formal GitHub review objects (`reviews.nodes` with
+`state === "COMMENTED" || "CHANGES_REQUESTED"`) as a source of qualifying findings. It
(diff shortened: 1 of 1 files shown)


planning/patch-comment-findings/PLAN.md
```diff
@@ -0,0 +1,149 @@
...
+  pass, not just the one that triggered dispatch.
+- No mechanical bot filtering needed — review's LLM pass already judges content; a bot
+  status comment just produces a no-op pass, same as any other non-finding today.
+
+## Design (approved)
+
+**Layer:** Shared/CLI logic only — no DB, API, or frontend changes.
+
+1. **`agent/src/check-review.ts`** — extract the inline fresh-reply computation
+   (currently in `buildProductionDeps`, ~lines 528-546) into a named, exported pure
```
dmcaulay: This says the fresh-reply computation (lines 528-546) is "currently in `buildProductionDeps`" — but `buildProductionDeps` starts at line 624 in `agent/src/check-review.ts`, well past that range. Lines 528-546 are actually inline inside `getReviewCandidates` (starts at line 356). Worth fixing so whoever executes RCT-1.1 doesn't go looking in the wrong function.

dmcaulay (commented):
Verdict: COMMENT — one factual location error in the plan (misattributes the fresh-reply computation to `buildProductionDeps` when it's actually inline inside `getReviewCandidates`) should be fixed before this plan is executed; otherwise the root-cause analysis and design are accurate and well-reasoned.

dmcaulay (commented):
Verdict: APPROVE — the prior finding (function misattribution: buildProductionDeps → getReviewCandidates) is genuinely fixed in this commit; no new issues found. Posted as COMMENT since GitHub blocks self-APPROVE.

PR #2649, opened 2026-08-14
feat: broaden fresh-reply detection to any non-agent commenter in check-review.ts

dodizzle:
## Summary
- Extracts the inline "fresh reply" OR computation in `getReviewCandidates()` (check-review.ts) into a new exported pure function, `hasFreshNonAgentComment`, used at both consuming call sites inside `traceReviewCandidacyDecision` (the `hasAnyReviewAtHead` exclusion and the RCO-1.2 terminal-skip exclusion)
- Broadens the author filter from equality against `pr.author.login` to inequality against `currentUser`, so a comment from the PR author, a third-party reviewer, or anyone else — except the reviewing agent's own identity — now counts as a fresh reply that reopens review candidacy
- Replaces the test file's local `computeHasFreshAuthorReply` mirror with an import of the real production function, inverts the pre-existing "third-party reply does not flip" tests to assert they now DO flip candidacy, and adds new tests confirming a comment from the agent's own identity (`currentUser`) never re-triggers review on its own PR
- Refreshes `docs/agent.md` to reflect the renamed function/broadened filter

## Acceptance Criteria

| 1 | New exported `hasFreshNonAgentComment` replaces inline OR computation at both call sites | MET | `check-review.ts` defines and exports the function; `getReviewCandidates` calls it once, feeding the shared value used by both the `hasAnyReviewAtHead` exclusion and RCO-1.2 terminal-skip inside `traceReviewCandidacyDecision` |
| 2 | Filter broadened from `=== pr.author.login` to `!== currentUser` | MET | New function uses `c.author.login !== currentUser`; call site passes `currentUser` (already in scope) instead of `pr.author.login` |
| 3 | Test decision: inverted test, new self-comment guard test, mirror function replaced with import | MET | Old tests inverted, new self-guard tests added, local `computeHasFreshAuthorReply` removed and all call sites now use the imported production function; existing RFR-1.1/RVG-2.1 fixtures still pass under new semantics |
| 4 | Safe-to-deploy-standalone note: RCT-1.2 not implemented; `review.md`'s Step 14 Live-Review Pre-Check mirrors the same broadened non-agent-comment filter for consistency between the TS candidate-selection logic and the markdown command's own pre-check | MET | Diff touches `check-review.ts`, its test file, `docs/agent.md`, `review.md`'s Step 14 (`.author.login != $currentUser` replacing the old author-equality check, plus a `CURRENT_USER={the login resolved in Step 3}` assignment added ahead of the Step 14 graphql call so `$currentUser` is populated at the point Step 14 actually runs, since it executes before Step 9.5's own assignment), and `review.content.test.ts`; no changes to RVD-1.2's dispatch-time claim/checkout logic |

## Test Plan
- `bun test agent/src/check-review.unit.test.ts` — 104/104 passing
- `bunx biome lint .` — clean
- `bun run --filter='*' --sequential typecheck` — all packages pass
- `task ci` — passes except one pre-existing, unrelated flaky failure (`metrics/src/lib/errors.unit.test.ts`), confirmed failing identically on `main` prior to this change


plugins/shipwright/commands/review.md
```diff
@@ -1044,18 +1044,20 @@ precheck=$(gh api graphql -f query='
       }
     }
   }
-}' --jq '.data.repository.pullRequest as $pr | ([$pr.reviews.nodes[] | select(.commit.oid == $pr.headRefOid and (.body | test("verdict\\**\\s*:\\s*\\**(approve|comment)\\b"; "i"))) | .submittedAt] | max) as $maxTerminalSubmittedAt | {headRefOid: $pr.headRefOid, terminal: (if $maxTerminalSubmittedAt != null then ([$pr.comments.nodes[] | select(.author.login == $pr.author.login and (.createdAt > $maxTerminalSubmittedAt)) | .createdAt] | length == 0) else false end)}')
+}' --arg currentUser "$CURRENT_USER" --jq '.data.repository.pullRequest as $pr | ([$pr.reviews.nodes[] | select(.commit.oid == $pr.headRefOid and (.body | test("verdict\\**\\s*:\\s*\\**(approve|comment)\\b"; "i"))) | .submittedAt] | max) as $maxTerminalSubmittedAt | {headRefOid: $pr.headRefOid, terminal: (if $maxTerminalSubmittedAt != null then ([$pr.comments.nodes[] | select(.author.login != $currentUser and (.createdAt > $maxTerminalSubmittedAt)) | .createdAt] | length == 0) else false end)}')
```
dodizzle, line 1054: `$CURRENT_USER` is referenced here but never assigned before this point in the file's execution flow — Step 14's Live-Review Pre-Check runs before Step 4/5/9.5, and the only literal `CURRENT_USER=<value>` assignment anywhere in the file is at Step 9.5 (line 610), which runs much later. Step 3 only says to resolve the login and "substitute it directly" — it never assigns a shell variable. If `$CURRENT_USER` is taken literally rather than manually substituted, it evaluates to an empty string, and since no GitHub login is empty, `.author.login != $currentUser` becomes true for every comment — including the reviewing agent's own — silently defeating the self-exclusion that's this PR's stated purpose. Suggest adding an explicit `CURRENT_USER={the login resolved in Step 3}` assignment immediately before this graphql call, mirroring Step 9.5's pattern.

dodizzle (commented):
Verdict: COMMENT — review.md's Step 14 precheck likely reads an unassigned $CURRENT_USER (evaluates to empty string, silently defeating the self-exclusion this PR adds), and the PR's acceptance criteria incorrectly claim review.md was untouched.

agent/src/check-review.ts
```diff
@@ -105,6 +105,52 @@ export interface PrRecord {
...
+// throwing.
+export function hasFreshNonAgentComment(
+  reviewData: PrReviewData | undefined,
+  currentUser: string,
+  reviewedAtMs: number,
+): boolean {
+  const hasFreshTopLevelReply =
+    reviewData?.comments.nodes.some(
+      (c) =>
+        c.author.login !== currentUser &&
```
dmcaulay, line 139: The broadened filter now treats any non-`currentUser` commenter, including bot/CI accounts, as a fresh-reply trigger (previously only `pr.author.login` counted). The diff's own test rename ("does not trigger re-candidacy" → "now DOES trigger re-candidacy") shows this was a known consequence, but no bot-account exclusion or documented rationale was added — contrast with the nearby `isRequestedReviewer` bypass, which notes it was "confirmed with the team." No actively-commenting bot exists in this repo's current CI workflows today, so impact looks limited, but the gap is structural. Optional: exclude `[bot]`-suffixed logins, or document why bot comments are an acceptable re-trigger source.

dmcaulay (approved):
Verdict: APPROVE — Prior-review-flagged CURRENT_USER bug is verifiably fixed with matching test coverage; one non-blocking suggestion on bot-comment re-trigger risk.

PR #2689, opened 2026-08-17
fix(minikube): unblock dev login and enable the admin Chat tab out of the box

dodizzle:
## Summary

Three fixes that together make `task minikube:up` produce a stack you can actually log into and chat with.

**1. Dev auth was hard-blocked by the image.** `auth.mode=open` sets `ADMIN_DEV_AUTH=true` but relied on `NODE_ENV` being *unset* to keep the dev-auth path alive. The admin image bakes `ENV NODE_ENV=production` (`admin/Dockerfile`), so the guard in `dev-auth-guard.ts` disabled dev auth and `/admin/dev-login` returned 404. The chart now sets `NODE_ENV=development` explicitly in the open/none branch.

**2. The post-install output pointed at a dead end.** It advertised the admin console root, which redirects to `/admin/login` — a page offering only a Google sign-in button that this profile never configures. `buildAccessUrls` now exposes `devLogin` and `printNextSteps` leads with it.

**3. The admin console's Chat tab was never wired.** Both `SHIPWRIGHT_CHAT_SERVICE_URL`/`SHIPWRIGHT_CHAT_SERVICE_ADMIN_TOKEN` (admin) and `CHAT_SEED_ADMIN_TOKEN` (chat) were gated on a hand-created `chat.adminToken.existingSecret`, which the minikube profile deliberately avoids — so chat ran but nothing could talk to it. The chart now generates the token into its own chat Secret using the same lookup-persistence idiom as `admin-secret.yaml`, and both Deployments resolve through one helper so they always read the same value.

## Notes

- The chat Secret's two keys are now independently gated: the DSN on the bundled-PostgreSQL path, the token whenever chat is enabled without a caller-supplied Secret. It renders when either applies.
- `chat.enabled: false` still suppresses all chat wiring, even with `adminToken.existingSecret` set.
- Supplying your own `chat.adminToken.existingSecret` is unchanged.

## Testing

- `task helm:check` — lint, 329 unittest specs, kubeconform: all pass
- `bun test scripts/minikube.unit.test.ts` — 27 pass
- Dev login verified live on Minikube before the cluster was torn down: `/admin/dev-login` → 302 → `/admin/agents`, session cookie works (200). The chat wiring is verified by chart tests and `helm template`; it has not yet been exercised against a live cluster.

https://claude.ai/code/session_015a3jURSc5pXi9Kdoa5ixtb

### DIFF
--- a/charts/shipwright/CHANGELOG.md
+++ b/charts/shipwright/CHANGELOG.md
@@ -10,6 +10,12 @@ independent of `appVersion`. CI enforces this with
 `ct lint --check-version-increment`. Each release here must mirror the
 `artifacthub.io/changes` annotation in `Chart.yaml`.
 
+## [1.11.55] - 2026-08-17
+
+### Fixed
+
+- `auth.mode=open`/`none` deployments no longer have dev auth blocked by the admin image's baked-in `NODE_ENV=production` default — set explicitly in `admin-deployment.yaml`
+
 ## [1.11.54] - 2026-08-15
 
 ### Changed
--- a/charts/shipwright/Chart.yaml
+++ b/charts/shipwright/Chart.yaml
@@ -9,7 +9,7 @@ type: application
 # Chart version (SemVer). Bump on every chart change, independent of appVersion.
 # See CHANGELOG.md and the README "Versioning" section — CI enforces this bump
 # via `ct lint --check-version-increment`.
-version: 1.11.54
+version: 1.11.55
 # Version of the Shipwright application this chart deploys by default.
 appVersion: "0.1.0"
 home: https://shipwrightharness.com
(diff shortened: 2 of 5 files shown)


dodizzle (commented):
Verdict: APPROVE — Correctly root-caused (verified against dev-auth-guard.ts's hard block on NODE_ENV=production and the admin Dockerfile's baked-in default) and precisely scoped: only touches the open/none auth branch, leaves the google branch untouched, with matching test and doc updates. No blocking concerns.

dmcaulay (approved):
Verdict: APPROVE — Correctly root-caused (verified against dev-auth-guard.ts's hard block on NODE_ENV=production and the admin Dockerfile's baked-in default) and precisely scoped to the open/none branch, with matching test and doc updates. No blocking concerns.

PR #2694, opened 2026-08-18
feat(minikube): enable the admin Chat tab and open the dev-login URL automatically

dodizzle:
## Summary

Follow-up to #2689 (which fixed only the `NODE_ENV` dev-auth block). Two remaining problems with `task minikube:up`:

**1. The admin console's Chat tab was never wired.** The chat service ran, but nothing could talk to it: `SHIPWRIGHT_CHAT_SERVICE_URL` / `SHIPWRIGHT_CHAT_SERVICE_ADMIN_TOKEN` (admin) and `CHAT_SEED_ADMIN_TOKEN` (chat) were all gated behind a hand-created `chat.adminToken.existingSecret`, which the minikube profile deliberately avoids — so the admin pod had zero chat env vars.

The chart now generates the token into its own chat Secret using the same lookup-persistence idiom as `admin-secret.yaml` (generate on first install, reuse the cluster's value on upgrade). Both Deployments resolve name and key through one shared helper, so they cannot drift and `helm upgrade` cannot desynchronize them.

**2. The post-install output buried the only useful link.** It listed the admin console, metrics, task store, and agent-creation URLs — but the console root redirects to a Google sign-in page this profile never configures. Output is now exactly one URL, `/admin/dev-login`, and the script opens it for you.

## Notes

- The chat Secret's two keys are independently gated: the DSN on the bundled-PostgreSQL path, the token whenever chat is enabled without a caller-supplied Secret. The Secret renders when either applies.
- `chat.enabled: false` still suppresses all chat wiring, even with `adminToken.existingSecret` set.
- Supplying your own `chat.adminToken.existingSecret` is unchanged.
- Auto-open is best-effort: `open` on macOS, `xdg-open` on Linux, skipped elsewhere, and a launch failure never fails the bring-up. It is suppressed entirely when `/etc/hosts` lacks the `shipwright.local` mapping (the check ignores commented-out lines), so it cannot open a link that would not resolve.

## Testing

- `task helm:check` — lint, 329 unittest specs, kubeconform: all pass
- `bun test scripts/minikube.unit.test.ts` — 33 pass
- `task lint` — clean
- Output and auto-open decision verified against a real `/etc/hosts`
- Not yet exercised against a live cluster — the Minikube profile was deleted mid-session

https://claude.ai/code/session_015a3jURSc5pXi9Kdoa5ixtb

### DIFF
--- a/charts/shipwright/Chart.yaml
+++ b/charts/shipwright/Chart.yaml
@@ -9,7 +9,7 @@ type: application
 # Chart version (SemVer). Bump on every chart change, independent of appVersion.
 # See CHANGELOG.md and the README "Versioning" section — CI enforces this bump
 # via `ct lint --check-version-increment`.
-version: 1.11.57
+version: 1.11.58
 # Version of the Shipwright application this chart deploys by default.
 appVersion: "0.1.0"
 home: https://shipwrightharness.com
--- a/charts/shipwright/examples/values-minikube.yaml
+++ b/charts/shipwright/examples/values-minikube.yaml
@@ -121,6 +121,14 @@ chat:
   enabled: true
   database:
     existingSecret: ""
+  adminToken:
+    # Left empty on purpose: the chart then generates the chat admin token into
+    # its own chat Secret and injects that one value into BOTH the chat service
+    # (CHAT_SEED_ADMIN_TOKEN) and the admin console
+    # (SHIPWRIGHT_CHAT_SERVICE_ADMIN_TOKEN), so the admin console's Chat tab
+    # (/admin/chat) works with no hand-created Secret. Set this only to supply
+    # your own token via a pre-existing Secret.
+    existingSecret: ""
   resources:
     requests:
       cpu: 100m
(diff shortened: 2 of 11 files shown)


dodizzle (commented):
Verdict: APPROVE — Chat admin-token generation correctly reuses admin-secret.yaml's lookup-persistence idiom (no re-generation on helm upgrade), the `with`/`$` scoping fix in the two Deployments is necessary and correct, and the new minikube.ts helpers (hostsEntryPresent, buildOpenCommand, openInBrowser) are pure, well-tested, and fail open. No blocking issues found.

dodizzle (commented):
Verdict: APPROVE — Only change since the prior review is a one-line Chart.yaml version bump (1.11.57 → 1.11.58) to satisfy check-version-increment. No functional change; re-verified the Secret lookup-persistence idiom, `.`/`$` template scoping, and new minikube.ts helpers are all still correct. No blocking issues found.

dodizzle (commented):
Verdict: APPROVE — Re-reviewed after the CI-fix commit (eb19b16b), which adds unit test coverage for ingressHostResolves and openInBrowser plus incidental Biome reformatting. New tests use injected fakes exclusively (no mock.module()/global overrides), correctly cover the comment-stripping edge cases, and no functional code changed. No blocking issues found.

PR #2695, opened 2026-08-18
fix(agent): pin Go/npm/bun caches to the PVC and raise ephemeral storage to 4Gi

dodizzle:
Provisioned agents were being evicted mid-run with `Pod ephemeral local storage usage exceeds the total limit of containers 1Gi`, losing the in-flight task. Two independent causes, both fixed here.

## Cache pinning (`agent/src/setup.ts`)

`runMiseStartup` already pins `MISE_*` and `XDG_CACHE_HOME` to the PVC, but three package managers ignore XDG and write under `$HOME` — the container's ephemeral overlay:

- `GOMODCACHE` defaults to `$GOPATH/pkg/mod`, **not** `$XDG_CACHE_HOME`. A single `go mod download` on a Terratest module pulls ~1.7GiB — over the whole pod budget on its own. (`GOCACHE`, the build cache, already follows XDG and was never the problem.)
- `npm` writes to `$HOME/.npm`, `bun` to `$HOME/.bun/install/cache`.

All four are now pinned under `AGENT_HOME`. Side benefit: the caches survive a pod restart instead of re-downloading on every boot.

`CARGO_HOME` is deliberately left alone — mise owns the Rust toolchain install and sets it itself.

## Ephemeral budget (`admin/src/agent-manifest.ts`)

`1Gi` → `4Gi`, request and limit. 1Gi was just the Autopilot default made explicit and left no headroom for legitimately transient writes — e.g. the security-scan skill staging five tool binaries in a `mktemp -d`. Pinning the caches fixes the confirmed cause; this stops the next unpinned tool from repeating it.

## Verification

- Evictions were confirmed against a live cluster: every eviction correlated with a cron run that started 2–5 min earlier and never completed.
- 3 new tests (cache pinning, dir creation, `CARGO_HOME` left untouched, 4Gi budget). `bun test`: 6471 pass.
- Coverage-neutral: +7 lines, +7 covered — aggregate unchanged at 83.48%.

Takes effect on the next agent pod restart for the cache pins; the resource bump needs a reconcile per agent.

### DIFF
--- a/admin/src/agent-manifest.ts
+++ b/admin/src/agent-manifest.ts
@@ -43,10 +43,18 @@ export const AGENT_WORKTREE_DIR = `${AGENT_HOME_MOUNT_PATH}/workspace/worktrees`
  * grow until the kubelet evicts neighbouring pods under node memory pressure
  * (observed: ~11.6Gi used against the 2Gi request). No CPU limit — CPU
  * contention throttles instead of evicting.
+ *
+ * ephemeral-storage is 4Gi, not the 1Gi Autopilot default: 1Gi left no room for
+ * a single build step's scratch space, and agents were evicted mid-run ("Pod
+ * ephemeral local storage usage exceeds the total limit of containers 1Gi"),
+ * losing the in-flight task. Tool caches are pinned to the PVC in the agent's
+ * runMiseStartup, so this budget only has to cover genuinely transient writes —
+ * e.g. a scan skill staging tool binaries in a mktemp dir. Requests and limits
+ * are kept equal because Autopilot bills the request either way.
  */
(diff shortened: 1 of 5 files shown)


dodizzle (commented):
Verdict: APPROVE — Clean, well-tested infra fix: pins Go/npm/bun caches (which ignore XDG_CACHE_HOME) to the PVC and raises ephemeral-storage headroom from 1Gi to 4Gi, addressing confirmed mid-run pod evictions. Verified independently: BUN_INSTALL_CACHE_DIR and npm_config_cache both actually redirect their tools' caches, check-config-docs passes, lint clean, new tests pass locally. No blocking concerns. Posting as COMMENT since GitHub blocks self-APPROVE on own PRs.

PR #2703, opened 2026-08-18
feat: check-review.ts ledger-timestamp candidacy trigger (PFL-3.1)

dodizzle:
## Summary
- Add a second eligibility trigger to `traceReviewCandidacyDecision` in `agent/src/check-review.ts`: a PR is now eligible for review when any `PrFinding.at` recorded against it is newer than `record.reviewedAt`, even when `headRefOid` is unchanged
- New trigger is additive to the existing `already-reviewed-terminal` (RCO-1.2) terminal-skip check, alongside the existing `hasFreshAuthorReply` bypass — the `staged` and `already-reviewed-live` checks are untouched
- Adds a local `PrFinding` type and `findings?: PrFinding[]` field to `PrRecord`, mirroring the task-store shape (PFL-1.2) so it survives `createPrRecordQuery`'s generic JSON parse
- Refreshes `docs/agent.md` to document the new bypass

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | getReviewCandidates/traceReviewCandidacyDecision treats a PR as eligible when a PrFinding.at is newer than record.reviewedAt, even when headRefOid is unchanged | MET |
| 2 | Existing terminal-skip behavior (no new commit, no new ledger entry -> excluded) is unchanged | MET |
| 3 | Unit tests in check-review.unit.test.ts covering (a) fresh ledger entry makes PR eligible, (b) no new entry preserves exclusion, (c) existing RCT-1.1/RVG-1.1 fresh-comment tests pass unmodified | MET |

## Test Plan
- 7 new unit tests added to `check-review.unit.test.ts` (fresh finding bypass, stale/no findings preserve exclusion, empty/undefined findings, exact-timestamp non-fresh, epoch-0 `reviewedAt` fallback, multiple findings with only one fresh)
- `bun test agent/src/check-review.unit.test.ts` → 110/110 pass, including all pre-existing RCT-1.1/RVG-1.1 tests unmodified
- `task ci` → green (one pre-existing, unrelated failure in `metrics/src/lib/errors.unit.test.ts`, confirmed failing on `main` too)


agent/src/check-review.ts
```diff
@@ -287,12 +315,29 @@ export function traceReviewCandidacyDecision(args: {
...
+  // PFL-2.2's rebuttal writes) newer than the last review pass is a second,
+  // independent bypass of the terminal-skip exclusion below, alongside
+  // hasFreshAuthorReply — new ledger activity since the last review means
+  // there's something worth re-reviewing even though the head commit hasn't
+  // moved. Mirrors the `new Date(record?.reviewedAt ?? 0).getTime()` epoch-0
+  // fallback pattern used for reviewedAtMs elsewhere in this file, and uses
+  // the same strict `>` inequality as hasFreshAuthorReply's own freshness
+  // check (a finding stamped exactly at reviewedAt does not count).
+  const reviewedAtMs = new Date(record.reviewedAt ?? 0).getTime();
+  const hasFreshLedgerFinding =
```
dmcaulay: This `hasFreshLedgerFinding` bypass is unreachable in the scenario this PR targets.

`getReviewCandidates()` calls `traceReviewCandidacyDecision()` twice per PR with the same `reviewData` (agent/src/check-review.ts:626 and :668). The first call at line 626 exists solely to check the live-GitHub dedup:
```ts
if (hasAnyReviewAtHead(reviewData) && !hasFreshAuthorReply) {
  ...continue...
}
```
`hasAnyReviewAtHead` (check-helpers.ts) returns true for *any* review at the current head commit regardless of content, and `hasFreshAuthorReply` only considers PR comments/thread replies — never ledger findings. This check is untouched by this PR and runs *before* the code you're adding here, which is only reached by the *second* call (line 668).

In the exact scenario PFL-3.1 targets — a review already posted and still live at the PR's current head commit, plus a fresh ledger/rebuttal finding recorded afterward, no new commit — `hasAnyReviewAtHead(reviewData)` is `true` and `hasFreshAuthorReply` is `false`, so the loop `continue`s at line 626 before this bypass is ever evaluated. It can only fire when `reviewData` is empty/absent (fetch failure, or genuinely no review at head at all) — not the case the acceptance criteria describe.

None of the 7 new unit tests catch this since they all call `traceReviewCandidacyDecision()` directly via `baseArgs()`, which defaults `reviewData: undefined` — the earlier `if (reviewData) {...}` branch is never exercised together with a fresh finding.

Suggestion: have the line-626 call site's live-review check also consult `hasFreshLedgerFinding` (or compute it once above both call sites and thread it through as an additional bypass alongside `hasFreshAuthorReply`), and add a `getReviewCandidates()`-level test combining a populated `reviewData` + an existing terminal review at head + a fresh ledger finding.

dmcaulay (commented):
Verdict: COMMENT — the new ledger-finding bypass in traceReviewCandidacyDecision is correctly implemented in isolation but is shadowed by an earlier, unmodified `already-reviewed-live` check in getReviewCandidates()'s actual call path (agent/src/check-review.ts:626), making it unreachable for the exact scenario (existing review at head + fresh ledger finding, no new commit) this PR was built to handle. See inline comment for details.

dmcaulay (approved):
Verdict: APPROVE — the shadowing bug from the prior review (ledger-finding bypass unreachable through getReviewCandidates()'s early live-review-dedup call site) is fixed: `hasFreshLedgerFinding` is now computed at the top of `traceReviewCandidacyDecision` and guards both the live-review-dedup and terminal-skip checks, with a dedicated new test proving the target scenario now resolves to eligible. `bun test agent/src/check-review.unit.test.ts` passes 112/112 and CI is green. No new issues found.

PR #2712, opened 2026-08-18
feat(admin): let /admin/agents/new provision an in-cluster agent without Slack

dodizzle:
## Summary
- Adds a runtime choice (self-hosted vs provisioned in-cluster) to `/admin/agents/new`, so minikube/local users can get a real K8s workload without a Slack App Configuration Token — talk to the agent via `/admin/chat` instead.
- `AgentProvisioner` gains `canProvision` so the in-cluster option is only offered when `SHIPWRIGHT_K8S_PROVISIONING` is actually enabled.
- System crons are seeded and an optional `CLAUDE_CODE_OAUTH_TOKEN` can be set right on the form.
- Agent runtime: `cutover-validate`'s `SLACK_BOT_TOKEN` check is now informational (a chat-only agent no longer fails cutover), and cron delivery skips gracefully instead of throwing `invalid_auth` when no Slack client is configured.
- `/admin/provision` (the Slack app wizard) is untouched.

## Test plan
- [x] `bun run --filter='*' typecheck`
- [x] `task lint`
- [x] `bun test` (6562 pass; 1 pre-existing unrelated failure on `main` in `check-deploy.unit.test.ts`)
- [x] New/updated unit, smoke, and integration tests for the runtime picker, provisioning rollback, non-fatal cron reconcile, and Slack-absent cron delivery
- [ ] Manual verification on minikube (`task minikube:up` → create agent with in-cluster runtime → confirm pod + chat round-trip)

### DIFF
--- a/admin/e2e/test-server.ts
+++ b/admin/e2e/test-server.ts
@@ -233,6 +233,21 @@ function buildMockDeps(): AdminUIDeps {
         missingRequiredEnv: [],
       }),
     },
+    provisioner: {
+      canProvision: false,
+      provision: async () => ({
+        resourceName: "r",
+        secretName: "s",
+        deploymentName: "d",
+      }),
+      deprovision: async () => {},
+      reconcile: async () => ({
+        recreated: [],
+        updated: [],
+        orphans: [],
+        failed: [],
+      }),
+    },
     sessionSecret: ADMIN_E2E_SESSION_SECRET,
     googleClientId: "e2e-google-client-id",
     googleClientSecret: "e2e-google-client-secret",
--- a/admin/src/admin-dev-login.smoke.test.ts
+++ b/admin/src/admin-dev-login.smoke.test.ts
@@ -201,6 +201,7 @@ function makeMockDeps(overrides?: Partial<AdminUIDeps>): AdminUIDeps {
       exchangeOAuthCode: async () => ({ botToken: "xoxb-mock-bot-token" }),
     },
     provisioner: {
+      canProvision: false,
       provision: async () => ({
         resourceName: "r",
         secretName: "s",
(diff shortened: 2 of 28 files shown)


dmcaulay (approved):
Verdict: APPROVE — Well-scoped Slack-optional in-cluster provisioning path. Every AgentProvisioner implementer/test double was updated for the new canProvision field, all post-optional-Slack code paths in cron-handler.ts/index.ts are properly null-guarded, provisioning rollback correctly cascade-deletes the patched CLAUDE_CODE_OAUTH_TOKEN env row, and the new form fields introduce no XSS surface. No blocking findings.

dodizzle (commented):
Verdict: APPROVE — Slack-optional in-cluster provisioning is well-scoped and cleanly implemented: rollback ordering is safe (AgentEnv cascades on Agent delete, so a provisioning failure cleans up the stored Claude token too), all four repo-wide AgentProvisioner implementers picked up the new canProvision flag, and every slack/deps.slack call site is properly null-guarded after the WebClient | undefined widening. New test coverage (rollback, rollback-failure-swallowed, provisioning-disabled short-circuit, non-fatal cron reconcile, chat-only cron delivery, cutover-validate's informational SLACK_BOT_TOKEN check) directly backs the changed logic. No blocking findings.

dodizzle (commented):
Verdict: APPROVE — Re-reviewed after the new commit (dd157497), which only fixes CI: the e2e test-server's mock AdminUIDeps was missing the now-required provisioner field, and a regression smoke test was added to catch this class of bug earlier. No new concerns; the implementation remains clean, well-guarded (provisioning rollback, secret handling, Slack-optional wiring), and thoroughly tested.

PR #2724, opened 2026-08-20
feat: add Okta OIDC auth client

dodizzle:
## Summary
- Add `admin/src/okta-auth-client.ts` — `OktaAuthClient` interface + `HttpOktaAuthClient` implementation, mirroring `HttpGoogleAuthClient`'s fetchFn-DI pattern
- Endpoints (`/v1/authorize`, `/v1/token`, `/v1/userinfo`) are derived from a caller-supplied static `issuer` config — no live `Issuer.discover()` `.well-known` call, so construction is synchronous and side-effect free
- Add `openid-client` to `admin/package.json` dependencies

## Acceptance Criteria
| 1 | OktaAuthClient exposes getAuthorizationUrl, exchangeCode, getUserInfo, matching GoogleAuthClient shape | MET | Interface + `HttpOktaAuthClient` implement all three methods |
| 2 | No live network call during construction (static config, not Issuer.discover()) | MET | Constructor only normalizes the issuer string synchronously |
| 3 | Integration test with injected fetchFn + recorded fixtures, mirroring google-auth-client pattern, no existing tests retired | MET | `okta-auth-client.integration.test.ts` + `fixtures/okta-auth-cassette.json` added; `google-auth-client.integration.test.ts` untouched |

## Test Plan
- `NODE_ENV=test bun test admin/src/okta-auth-client.integration.test.ts` — 10/10 pass
- `bunx biome lint` on new files — clean
- `bun run typecheck` — clean

### DIFF
--- a/admin/package.json
+++ b/admin/package.json
@@ -20,7 +20,6 @@
     "@sentry/hono": "^10.63.0",
     "@shipwright/lib": "*",
     "hono": "^4.13.1",
-    "openid-client": "^6.8.1",
     "yaml": "^2.9.0",
     "zod": "^3",
     "zod-to-json-schema": "^3.25.2"
--- a/admin/package.json
+++ b/admin/package.json
@@ -20,6 +20,7 @@
     "@sentry/hono": "^10.63.0",
     "@shipwright/lib": "*",
     "hono": "^4.13.1",
+    "openid-client": "^6.8.1",
     "yaml": "^2.9.0",
     "zod": "^3",
     "zod-to-json-schema": "^3.25.2"
--- a/admin/src/fixtures/okta-auth-cassette.json
+++ b/admin/src/fixtures/okta-auth-cassette.json
@@ -1,0 +1,36 @@
+{
+  "exchangeCode_success": {
+    "status": 200,
+    "body": {
+      "access_token": "okta.test-access-token",
+      "refresh_token": "okta.test-refresh-token",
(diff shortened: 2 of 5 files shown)


admin/package.json
```diff
@@ -20,6 +20,7 @@
     "@sentry/hono": "^10.63.0",
     "@shipwright/lib": "*",
     "hono": "^4.13.1",
+    "openid-client": "^6.8.1",
```
dodizzle: This adds `openid-client` (plus a new transitive `oauth4webapi` dep and a `jose` 6.2.3→6.2.9 bump) as a direct dependency, but `admin/src/okta-auth-client.ts` implements the OIDC flow entirely by hand with raw `fetch` calls — mirroring `google-auth-client.ts`. A repo-wide grep finds zero imports of `openid-client` anywhere. Suggest dropping this line (and the corresponding `bun.lock` entries) unless there's a planned follow-up that will actually use the library — if so, worth calling out in the PR description.

dodizzle (commented):
Verdict: COMMENT — Clean, well-tested client implementation mirroring google-auth-client.ts, but the openid-client dependency added to admin/package.json is unused (the client is hand-rolled with raw fetch) — see inline comment.

dmcaulay (approved):
Verdict: APPROVE — Clean re-review: prior unused-openid-client-dependency finding is fully resolved (package.json/bun.lock no longer reference it), and no new issues found.

PR #2728, opened 2026-08-20
feat: tee hitl's in-process console output to a log file

dmcaulay:
## Summary
- Add `logFilePath` to `HitlConfig`/`buildHitlConfig()` (default `~/.shipwright/hitl.log`, overridable via `SHIPWRIGHT_HITL_LOG_FILE`)
- Add a pure `createTeeWriter(original, appendLine)` helper and `installLogFileTee(path)` real-I/O glue that monkey-patches `process.stdout.write`/`process.stderr.write` *and* `console.log`/`warn`/`error` (Bun's console bypasses `process.stdout.write`, confirmed empirically) so in-process output from any module (e.g. `check-review.ts`'s candidacy-skip traces) is captured, each line ISO-timestamped in the file while terminal output is unchanged
- Wire `installLogFileTee(CONFIG.logFilePath)` into the top of the entrypoint before `runPreflight()`, and route the top-level fatal-error path through `log()` instead of `console.error()` so crashes are captured too
- Refresh `docs/configuration.md` with the new `SHIPWRIGHT_HITL_LOG_FILE` env var

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `buildHitlConfig()` returns `logFilePath` defaulting to `join(hitlHome, "hitl.log")`, overridden by `SHIPWRIGHT_HITL_LOG_FILE` | MET |
| 2 | Running `task hitl` writes both `[hitl]` and in-process `[check-review]` lines to `~/.shipwright/hitl.log`, ISO-timestamped, terminal unchanged | MET |
| 3 | Top-level fatal-error path captured in log file | MET |
| 4 | Unit tests for `createTeeWriter` and `buildHitlConfig`'s `logFilePath` (default + env override) | MET |

## Test Plan
- `bun test scripts/hitl.unit.test.ts scripts/hitl.bootstrap.unit.test.ts scripts/hitl.integration.test.ts` — 106 pass, 0 fail
- `biome check scripts/hitl.ts scripts/hitl.unit.test.ts` — clean
- Manual smoke: `installLogFileTee()` verified live to tee `console.log`/`warn`/`error` output to a file with ISO timestamp prefixes, terminal unchanged

### DIFF
--- a/scripts/hitl.ts
+++ b/scripts/hitl.ts
@@ -65,6 +65,12 @@ import {
   type WorkTaskCandidate,
   selectNextWorkItem,
 } from "../agent/src/work-selector.ts";
+import {
+  type ConsoleMethod,
+  type LogFileTeeTarget,
+  type WriteFn,
+  realLogFileTeeTarget,
+} from "./hitl-log-file-tee-target.ts";
 
 // ---------------------------------------------------------------------------
 // Allowed tools — FLOOR_TOOLS + web access, minus Bash/Agent (both can
@@ -238,13 +244,6 @@ function sleep(ms: number): Promise<void> {
 // ---------------------------------------------------------------------------
 
 /**
- * A stdout/stderr-shaped write function: takes a chunk (plus whatever
- * encoding/callback args Node's Writable#write overloads pass) and returns a
- * boolean (backpressure signal).
- */
-type WriteFn = (chunk: string | Uint8Array, ...args: unknown[]) => boolean;
-
-/**
  * Builds a replacement `write` function that calls `original` first
  * (preserving the real terminal output, side effects, and return value —
  * i.e. backpressure signaling isn't broken) and then forwards the chunk's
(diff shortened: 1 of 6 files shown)


scripts/hitl.ts
```diff
@@ -225,6 +230,82 @@ function sleep(ms: number): Promise<void> {
...
+ * than calling process.stdout.write()/process.stderr.write() (confirmed
+ * empirically — patching only the stream `write` methods does not intercept
+ * console.log in Bun, unlike Node). Since this file's log() helper and
+ * check-review.ts's logSkippedCandidacy() (called directly inside runLoop())
+ * both go through console.log, console.log/warn/error are additionally
+ * wrapped here — calling the original (unchanged terminal output) and then
+ * appendLine with the formatted message — so in-process console output from
+ * any module is actually captured, not just direct stream writes.
+ */
+export function installLogFileTee(path: string): void {
```
dmcaulay: `installLogFileTee` (real-I/O glue: `mkdirSync`, `createWriteStream`, monkey-patches `process.stdout.write`/`process.stderr.write`/`console.log/warn/error`) has no test coverage — only the pure `createTeeWriter` helper and `buildHitlConfig`'s `logFilePath` field are tested. This is the confirmed cause of the CI "coverage must not decrease" failure and conflicts with CLAUDE.md's "tests land with the code, same PR" rule. A subprocess-based `*.integration.test.ts` (spawn `bun run` against a small harness script, assert on the resulting log file) would satisfy both the coverage gate and the repo's test-isolation hard rule against `global.*` overrides in the shared test process.

scripts/hitl.ts
```diff
@@ -225,6 +230,82 @@ function sleep(ms: number): Promise<void> {
...
+  mkdirSync(dirname(path), { recursive: true });
+  const sink = createWriteStream(path, { flags: "a" });
+
+  const appendLine = (line: string) => {
+    sink.write(`${new Date().toISOString()} ${line}`);
+  };
+
+  process.stdout.write = createTeeWriter(
```
dmcaulay: No guard against double-invocation — a second `installLogFileTee()` call would double-wrap `process.stdout.write`/`console.log`, causing duplicate log lines. Only called once today; consider a module-level `installed` flag to make this safe against future regressions.

scripts/hitl.ts
```diff
@@ -1119,7 +1203,7 @@ if (import.meta.main) {
 
     await runLoop();
   } catch (err) {
-    console.error(`[hitl] fatal: ${err}`);
+    log(`fatal: ${err}`);
```
dmcaulay: This changed from `console.error(...)` to `log(...)`, which writes via `console.log`, not `console.error` — the fatal-error path now goes to stdout instead of stderr. Low-impact for a local dev tool, but worth confirming intentional; `console.error` is teed too, so keeping it would preserve stderr semantics without losing file capture.

dmcaulay (commented):
Verdict: COMMENT — `installLogFileTee()` is untested and is the confirmed cause of the CI "coverage must not decrease" failure (83.82% < 83.88%); one suggestion each on double-invocation safety and the fatal-path stderr→stdout shift.

scripts/hitl.unit.test.ts
```diff
@@ -296,6 +302,111 @@ describe("parseHitlRepos", () => {
...
+    const tee = createTeeWriter(original, appendLine);
+    const result = tee("plain string chunk\n");
+
+    expect(result).toBe(false);
+    expect(originalCalls).toHaveLength(1);
+    expect(appendLineCalls).toEqual(["plain string chunk\n"]);
+  });
+});
+
+describe("installLogFileTee", () => {
```
dodizzle: This describe block does real file I/O (tmpdir(), createWriteStream, polling readFileSync) and monkey-patches process.stdout.write/process.stderr.write/console.log/warn/error directly (restored in finally) — but it's in a `*.unit.test.ts` file. Per the root CLAUDE.md's Test conventions table, `*.unit.test.ts` is "pure logic, no I/O"; `*.integration.test.ts` is for "real dependency behavior via recorded fixtures / injected doubles". The code comment cites `withCapturedLogs()` in `agent/src/loop-orchestrator.unit.test.ts` as precedent, but that helper only does in-memory console capture — no real filesystem, no real stream patching — so it doesn't actually establish precedent for this. Consider moving this block into a new `scripts/hitl.integration.test.ts`.

dodizzle (commented):
Verdict: COMMENT — Core mechanism looks solid and CI is green, but the new installLogFileTee test does real file I/O and global console/stream monkey-patching in a *.unit.test.ts file, against the repo's own explicit unit-vs-integration layer convention.

scripts/hitl.ts
```diff
@@ -225,6 +230,105 @@ function sleep(ms: number): Promise<void> {
...
+ *
+ * Guarded against double-invocation: a second call would double-wrap
+ * process.stdout.write/console.log (each wrapping the already-wrapped
+ * version from the first call), causing duplicate log lines. Only called
+ * once today (at top-level entrypoint startup), but the guard makes that
+ * safe against future regressions.
+ */
+let logFileTeeInstalled = false;
+
+export function installLogFileTee(path: string): void {
```
dodizzle: `installLogFileTee()` (lines 308-330) is exercised only inside `hitl.integration.test.ts`'s subprocess-spawned test (`Bun.spawn(["bun", "run", TEE_FIXTURE_PATH, ...])`), required by this repo's isolation contract since it monkey-patches `console`/`process.stdout`/`process.stderr` globally. Bun's `--coverage` only instruments the invoking process, so these lines are invisible to lcov despite being functionally tested. Verified locally: `scripts/hitl.ts` line coverage is 85.11% on `main` vs 84.62% on this head, with 308-328 newly uncovered — this is the confirmed cause of the red "coverage must not decrease" CI job. Consider excluding these lines from coverage accounting (if Bun's lcov reporter supports ignore comments) or shrinking this function to a thinner shim around a smaller in-process-testable core.

scripts/hitl.ts
```diff
@@ -225,6 +230,105 @@ function sleep(ms: number): Promise<void> {
...
+ * to it prefixed with an ISO timestamp. Split out from installLogFileTee()
+ * so the actual filesystem behavior (dir creation, append semantics,
+ * timestamp formatting) can be exercised in-process against a real tmpdir —
+ * no global patching involved, so it's safe to unit-test directly (unlike
+ * the console/stream monkey-patching in installLogFileTee(), which needs a
+ * subprocess to test safely — see hitl.integration.test.ts).
+ */
+export function createAppendLineSink(path: string): (line: string) => void {
+  mkdirSync(dirname(path), { recursive: true });
+  const sink = createWriteStream(path, { flags: "a" });
```
dodizzle, line 291: `createWriteStream(path, { flags: "a" })` has no `.on('error', ...)` listener, and `hitl.ts` has no process-level `uncaughtException` handler either. An unhandled `'error'` event on a Node/Bun stream throws and crashes the process. Since `installLogFileTee` runs at the top of the entrypoint before the service loop starts, a transient log-write failure (unwritable dir, full disk, remounted read-only volume) would take down the entire long-running HITL runner over a best-effort logging feature. Suggest attaching an `'error'` handler here that logs and continues rather than propagating.

dodizzle (commented):
Verdict: COMMENT — CI is confirmed red from a genuine coverage-gate regression caused by this PR (installLogFileTee's lines are only exercised via a spawned subprocess, invisible to Bun's --coverage instrumentation), and the new log-file write stream has no 'error' handler, so an I/O failure would crash the whole long-running HITL runner. Both prior review findings on this PR (untested installLogFileTee / unit-vs-integration layer split) are confirmed resolved.

dodizzle (approved):
Verdict: APPROVE — Both prior findings (coverage-gate regression from installLogFileTee's subprocess-only exercise, and the missing write-stream error handler) are genuinely resolved. The coverage regression is fixed correctly: the untestable global-patching glue is isolated into its own file (scripts/hitl-log-file-tee-target.ts) and excluded from the coverage gate per the repo's existing 'process entrypoints' precedent, while all pure logic stays in hitl.ts, unit-tested via an injected fake LogFileTeeTarget. createAppendLineSink now has a sink.on('error', ...) handler, verified by a test that forces a real EISDIR failure. CI is green. No new issues found.

Issue #89, opened 2026-06-08
SWW-2.2: Body sections — problem, how-it-works, differentiators, demo, social proof


dodizzle:
```shipwright
{
  "id": "SWW-2.2",
  "title": "Body sections — problem, how-it-works, differentiators, demo, social proof",
  "status": "merged",
  "session": "shipwright-website",
  "repo": "shipwright",
  "layer": "Frontend",
  "branch": "feat/sww-2-2-body-sections",
  "dependencies": [
    "SWW-1.1"
  ],
  "pr": 99,
  "hours": 5,
  "source": "gh:app-vitals/shipwright#86",
  "addedAt": "2026-06-08T00:45:58Z",
  "startedAt": "2026-06-08T20:00:00Z",
  "prCreatedAt": "2026-06-07T20:28:00Z",
  "mergedAt": "2026-06-08T03:25:40Z",
  "deployedAt": "2026-06-08T03:26:46Z"
}
```

PR #2730, opened 2026-08-20
fix: remove release-before-patch race in review.md terminal-skip path

dmcaulay:
## Summary
- Removes the `POST /prs/${PRECLAIM_RECORD_ID}/release` call in `review.md`'s Live-Review Pre-Check (RVD-1.2) `$terminal == true` branch. That branch previously released the inherited pre-claim (setting `reviewState:"pending"`) as a separate step *before* the later PATCH set `reviewState:"posted"` — if execution stopped between the two calls, the PR was left unclaimed and pending, strictly worse than before the run started and guaranteed to be re-selected next tick. Confirmed live on PR #89 in verygood-ops/skills.
- The branch now relies solely on the existing atomic PATCH (`reviewState: "posted"`, `reviewedCommitSha: headRefOid`) — the task-store already auto-clears `claimedBy`/`claimedAt`/`heartbeatAt`/`phase` server-side whenever `reviewState` transitions to `posted`, exactly as Step 5's and Step 14.3's skip paths already rely on without ever calling `/release` themselves.
- Updated the surrounding prose to describe the single-PATCH atomic behavior (removed the now-inaccurate "release first so it re-enters the pool" rationale) and added a regression-guard test.

## Acceptance Criteria

| 1 | No `/release` call remains in `$terminal == true` branch | MET | 6-line `/release` curl block fully deleted; only the existing PATCH remains |
| 2 | Prose updated to single-PATCH atomic framing, matching Step 5/14.3 | MET | New prose cites Step 5/14.3 and explains claim auto-clear on `reviewState: posted` |
| 3 | No functional change to `$terminal == false` branch or Pre-Claim Fast Path | MET | Diff touches only the terminal==true prose block plus the test file |
| 4 | RVD-2.2 regression guard added; no tests retired | MET | New assertion added; 3 pre-existing tests rewritten (not deleted) since they encoded the old buggy two-step behavior — all 133 tests in the file pass |
| 5 | Safe to deploy standalone | MET | Markdown spec + content-test only, no schema/interface change |

## Test Plan
- `bun test plugins/shipwright/commands/review.content.test.ts` → 133/133 pass, 100% coverage
- `bun plugins/shipwright/scripts/check-banned-strings.ts` → clean
- `bun scripts/check-config-docs.ts` → clean
- `bun run scripts/sync-version.ts --check` → clean
- `bunx biome lint` on changed files → clean
- [x] Pre-commit checks passing (typecheck/full `task ci` gated by CI — see note below)

Note: `task typecheck` and `task test:coverage`'s full workspace run could not be executed locally in this sandboxed worktree (isolated `node_modules` install is blocked by the sandbox). The change is markdown + test-assertion only with no new TS logic, and CI will run the full `task ci` chain on this PR.

### DIFF
--- a/docs/task-store.md
+++ b/docs/task-store.md
@@ -379,7 +379,7 @@ Returns `404` if the PR doesn't exist.
 PATCH /prs/:id
 ```
 
-Writable fields: `staged`, `commitSha`, `reviewedCommitSha`, `taskId`, `agentId`, `state`, `mergedAt`, `reviewState`, `phase`, `readyForReviewAt`, `readyForPatchAt`, `readyForDeployAt`, `blocked`, `blockedReason`. All other fields are managed by lifecycle endpoints. Returns `400` if no writable fields are provided.
+Writable fields: `staged`, `commitSha`, `reviewedCommitSha`, `taskId`, `agentId`, `state`, `mergedAt`, `reviewState`, `reviewedAt`, `phase`, `readyForReviewAt`, `readyForPatchAt`, `readyForDeployAt`, `blocked`, `blockedReason`. All other fields are managed by lifecycle endpoints. Returns `400` if no writable fields are provided.
 
 **Note:** PATCH does not record an audit event; only the lifecycle endpoints (`POST /prs/:id/claim`, `POST /prs/:id/complete`, `POST /prs/:id/patch`, `POST /prs/:id/release`, `POST /prs/:id/skip`, `POST /prs/:id/skip/reset`) record field-level transitions as `PullRequestEvent` rows. PATCH is designed for late-stage corrections (e.g., force-setting `state=merged` after GitHub confirms it) where the transactional guarantees and field-diff auditing of a lifecycle endpoint are not needed.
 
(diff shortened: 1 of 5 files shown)


dodizzle (approved):
Verdict: APPROVE — Closes a verified crash-window race (release-then-patch two-step) in review.md's terminal-skip path with an atomic single-PATCH write-back; no blocking findings.

dodizzle (approved):
Verdict: APPROVE — RWA-1.2 folds in a well-tested, mechanically-related second fix (advances `reviewedAt` at all three terminal-skip write-back sites, closing a `hasFreshNonAgentComment` perpetual-retrigger gap) alongside the original release-before-patch race fix. No blocking findings.

dmcaulay (commented):
Verdict: APPROVE — Removes a genuine crash-window race (release-then-patch two-step) in review.md's terminal-skip path in favor of the existing atomic PATCH, which already clears claim fields server-side on reviewState:posted/approved. Also makes reviewedAt PATCH-writable and advances it at all three terminal-skip write-back sites, closing a real perpetual-retrigger watermark gap in check-review.ts's hasFreshNonAgentComment. Verified both changes directly against pull-request-service.ts, check-review.ts, and prs.ts; ran the affected test suites locally (136/136 content tests, 21/21 smoke tests) — all pass. No blocking findings. One suggestion-level note (not gating): the PR description only covers the /release removal, not the bundled reviewedAt write-back change.

PR #2733, opened 2026-08-20
test: Add smoke tests for Okta-authenticated access control

dodizzle:
## Summary
- Extends `admin/src/admin-ui.smoke.test.ts` with 4 new smoke test cases exercising the existing `isAdmin`/`AgentMember`/`assertAgentAccess` access-control behavior under Okta-authenticated sessions
- Verification-only — no production/access-control code was modified. Duplicates the existing Google-authenticated access-control cases with Okta-flavored session cookies to confirm the shared `completeLogin()` session shape makes access checks provider-agnostic
- No existing tests were modified or retired

## Acceptance Criteria

| 1 | Okta admin sees all agents and can create one | MET | GET /admin/agents shows 'Test Agent'; POST /admin/agents returns 302 to created agent |
| 2 | Okta non-admin sees only their AgentMember agent(s), 403s elsewhere | MET | List filtered to 'My Okta Agent' only; outsider gets 403 on agent detail; non-admin gets 403 on /admin/provision |
| 3 | Okta non-admin can configure their own agent's settings | MET | POST /admin/agents/{id}/envs returns 302, agentEnvService.upsert invoked |
| 4 | /admin/chat returns 403 for Okta non-admin | MET | GET /admin/chat with non-admin Okta session returns 403 |
| 5 | New test cases only, duplicated from Google cases, no tests retired | MET | Diff is purely additive (302 insertions, 0 deletions); mirrors existing 'member access control' describe block pattern |

## Test Plan
- `task lint`, `task typecheck` (all workspaces), `task check-config-docs`, `task check-version-sync` — all pass
- `bun test` — admin workspace: 1598 pass / 200 skip / 0 fail; repo-wide: 6630 pass / 432 skip / 1 fail (pre-existing failure in `agent/src/claude.unit.test.ts`, confirmed failing identically on a clean `origin/main` checkout — unrelated to this change, environment-dependent)
- Manually verified each new test case genuinely exercises its access-control path by temporarily breaking the corresponding `isAdmin`/`assertAgentAccess` gate and confirming the test fails, then restoring

### DIFF
--- a/admin/src/admin-ui.smoke.test.ts
+++ b/admin/src/admin-ui.smoke.test.ts
@@ -6212,10 +6212,67 @@ describe("admin UI — public task board", () => {
 // admin-ui.ts is shared across providers, so the session JWT shape (and thus
 // every downstream access check) is provider-agnostic. No new access-control
 // logic is introduced here.
+//
+// The first test below closes the gap flagged in review: rather than minting
+// the session JWT directly via makeSessionCookie() (which never touches Okta
+// code), it drives the real GET /admin/auth/okta/callback handler — token
+// exchange + userinfo normalization via the makeOktaClient mock — and then
+// reuses the cookie that handler actually sets on a subsequent authenticated
+// request, proving the callback-minted session drives access control exactly
+// like the directly-minted ones the remaining tests below use.
 
 describe("admin UI — Okta-authenticated access control", () => {
   const OKTA_MEMBER_EMAIL = "okta-member@example.com";
   const OKTA_OTHER_AGENT_ID = "agent-okta-other-456";
+
+  it("session minted by the real Okta callback route grants access like a directly-minted session", async () => {
+    // Drive the actual /admin/auth/okta/callback handler: CSRF state check,
+    // exchangeCode() + getUserInfo() via the makeOktaClient mock, allowlist
+    // check, then completeLogin() mints the session cookie.
+    const nonce = "test-okta-nonce-real-callback";
+    const oauthState = encodeURIComponent(JSON.stringify({ nonce }));
+    const params = new URLSearchParams({ state: nonce, code: "auth-code-123" });
+    const callbackApp = createAdminUIApp(
+      makeMockDeps({
+        oktaClient: makeOktaClient({
+          getUserInfo: () =>
+            Promise.resolve({
+              sub: "okta-sub-real-callback",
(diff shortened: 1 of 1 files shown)


admin/src/admin-ui.smoke.test.ts
```diff
@@ -6201,3 +6201,305 @@ describe("admin UI — public task board", () => {
...
   });
 });
+
+// ─── Okta-authenticated access control ────────────────────────────────────────
+//
+// ESR-2.1: verification-only coverage. These duplicate the Google-authenticated
+// access-control cases above (agents-list filtering, per-agent 403s,
+// admin-only-route 403s, /admin/chat 403) under an Okta-issued session, to
+// confirm isAdmin/AgentMember/assertAgentAccess behave identically regardless
+// of which OAuth provider minted the session cookie — completeLogin() in
```
dodizzle, line 6211: This rationale doesn't hold up under inspection. `completeLogin()` (`admin-ui.ts:567-609`) is genuinely shared across the Google and Okta callback routes and mints a `{userId, email, isAdmin}` session with no provider field — so it's true there's nothing to branch on downstream. But neither these new tests nor the pre-existing Google-flavored ones they duplicate (line 2852) ever call `completeLogin()` or hit an OAuth callback route at all: `makeSessionCookie()` (line 78) signs the session JWT directly. That means this diff can't actually confirm provider-agnostic behavior — it's a cosmetic duplicate of existing coverage (different string literals only), since no code here inspects provider identity in the first place. If the goal is to verify Okta-specific behavior, the one path worth testing is the `/admin/auth/okta/callback` handler + userinfo normalization (the file already has a `makeOktaClient` mock helper for this).

dodizzle (commented):
Verdict: COMMENT — Safe test-only diff, all stated ACs mechanically met, but the new tests never exercise Okta-specific code (they bypass OAuth entirely via direct JWT minting, same as the Google-flavored tests they duplicate), so they can't actually verify the PR body's provider-agnostic claim.

dodizzle (commented):
Verdict: APPROVE — Follow-up commit closes the gap from my prior review: the new first test now drives the real GET /admin/auth/okta/callback handler (token exchange + userinfo normalization via the makeOktaClient mock) and reuses the cookie it mints, rather than bypassing OAuth via direct session minting. All stated acceptance criteria remain met, no new issues.

PR #2763, opened 2026-08-21
feat: add restrictSlackToMembers toggle to Agent

dodizzle:
## Summary
- Adds a `restrictSlackToMembers: boolean` column to the `Agent` model (default `false`) via an additive Prisma migration
- Wires the field end-to-end through the admin service layer (`agents.ts`), the CRUD API (`agents-api.ts`, `openapi-schemas.ts`), and the admin UI (create form + a new "Slack access" settings card on the edit page), mirroring the existing `authorAllowlist` wiring pattern
- When the flag is set to `true` on an agent with zero `AgentMember` rows, the save still succeeds but a non-blocking warning banner ("this agent has no members — enabling this will block all Slack senders") is surfaced in both the JSON API response and the admin UI

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | New `restrictSlackToMembers` boolean column exists on Agent, default false, via an additive Prisma migration | MET |
| 2 | Create and update agent flows accept/persist `restrictSlackToMembers`, matching the existing authorAllowlist wiring pattern | MET |
| 3 | Admin UI create/edit forms expose a checkbox for this field | MET |
| 4 | Saving `restrictSlackToMembers: true` on an agent with zero AgentMember rows succeeds but renders a warning banner; saving with existing members renders no warning | MET |
| 5 | Integration test coverage for create/update round-tripping, plus a zero-members warning banner case; smoke tests extended for the warning route; no existing tests retired | MET |

## Test Plan
- Integration tests in `admin/src/agents-api.integration.test.ts`: create/update round-trip, field-omission-preserves-value, zero-members-warning present/absent (against a real Postgres DB)
- Smoke tests in `admin/src/agents-api.smoke.test.ts` and `admin/src/admin-ui.smoke.test.ts`: route contracts, warning banner rendering, no-warning-when-restrict-false, 403 for non-admin
- Unit tests in `admin/src/admin-ui-pages.unit.test.ts`: checkbox presence/checked-state, warning banner rendering
- [x] `bun run --filter='*' --sequential typecheck` — all 7 workspaces pass
- [x] `bun test admin/` — 1620 pass, 2 pre-existing environment-only failures (confirmed identical on `main`, unrelated to this change)
- [x] Coverage on all changed files ≥ 89.8%, at or above the repo's 90/89 floor


admin/src/admin-ui.ts
```diff
@@ -1356,6 +1397,34 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
     },
   );
 
+  // ─── Slack access settings (restrictSlackToMembers) ────────────────────────
+
+  app.post("/admin/agents/:id/settings", requireAuth, async (c) => {
+    const agentId = c.req.param("id");
+    if (!(await assertAgentAccess(agentId, c.var.userEmail, c.var.isAdmin))) {
```
dodizzle: **Critical (confidence 92)**: This route authorizes via `assertAgentAccess`, which returns `true` for admins **or any non-admin `AgentMember`**. That's inconsistent with the other two surfaces this PR adds for the same field: `PATCH /agents/:id` in `agents-api.ts:1103` strictly requires `isAdmin` (and `docs/agent-api.md:84` documents it as admin-only), and the "Slack access" card in `admin-ui-pages.ts:1224` only renders for admins. The sibling member-mutation routes (`POST /admin/agents/:id/members[/delete]`) also use a strict `!c.var.isAdmin` check. As written, a non-admin member can bypass the UI and the JSON API's admin-only rule by POSTing directly here to flip `restrictSlackToMembers` — no test in this PR covers a non-admin *member* (only a non-admin *outsider*) hitting this route. Suggest `if (!c.var.isAdmin) return new Response("Forbidden", { status: 403 });`, matching the sibling routes.

dodizzle (commented):
Verdict: COMMENT — POST /admin/agents/:id/settings uses assertAgentAccess (admin OR any AgentMember), letting a non-admin member toggle restrictSlackToMembers even though this PR gates the equivalent JSON PATCH route and the UI settings card to admins only.

dmcaulay (approved):
Verdict: APPROVE — The critical auth gap from the prior review (POST /admin/agents/:id/settings using assertAgentAccess instead of strict admin-only) is fixed correctly, verified in source against the sibling /members route and the JSON API's isAdmin check. New regression test covers the non-admin AgentMember case specifically, not just an outsider. Migration is additive, zero-members warning logic is consistent, CI is green.

PR #2723, opened 2026-08-20
feat: add Helm chart support for auth.mode=okta

dodizzle:
## Summary
- Add `auth.mode=okta` support to the shipwright Helm chart, mirroring the existing `auth.mode=google` implementation field-for-field
- Adds an `auth.okta` values block (issuer/clientId/clientSecret/allowedEmails + `existingSecret` support), a `values.schema.json` conditional requiring those fields when `existingSecret` is unset, and template wiring in `admin-deployment.yaml`/`admin-secret.yaml`/`NOTES.txt`
- `auth.mode` remains a single-value enum (`open`/`google`/`okta`) — one Helm deployment runs exactly one provider at a time; `google` behavior is untouched

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | helm template with auth.mode=okta renders OKTA_ISSUER/OKTA_CLIENT_ID/OKTA_CLIENT_SECRET/SHIPWRIGHT_ADMIN_ALLOWED_EMAILS correctly from both inline values and existingSecret | MET |
| 2 | values.schema.json rejects auth.mode=okta without required auth.okta.* fields | MET |
| 3 | extend tests/admin_auth_modes_test.yaml with okta-mode cases mirroring google-mode cases, no tests retired | MET |

## Test Plan
- `helm unittest charts/shipwright` — 344/344 passing (15 new okta cases added, zero retired)
- `helm lint charts/shipwright` — clean
- `bun scripts/check-config-docs.ts` — 39/39 env vars documented
- Manual `helm template` renders verified for both inline-values and `existingSecret` paths
- [x] Pre-commit checks passing

Generated with [Claude Code](https://claude.com/claude-code)

PR #2764, opened 2026-08-21
fix(chart): emit a 64-char-hex SHIPWRIGHT_ENCRYPTION_KEY so admin can encrypt agent envs

dodizzle:
## Problem
Creating/crediting an agent on any install that relies on chart-generated keys (e.g. minikube) returns **500 Internal Server Error**. Root cause: the chart generated `SHIPWRIGHT_ENCRYPTION_KEY` as `randAlphaNum 32 | b64enc` → a **32-char** value in the pod env, but `admin/src/crypto.ts` does `Buffer.from(key, "hex")` and requires a **64-char hex** string (32 bytes for AES-256-GCM). The mismatch throws `RangeError: Invalid key length` on the first encrypted-env write. `SESSION_SECRET` uses the same generator but is fine — it's never hex-decoded.

## Fix
- **Chart** (`admin-secret.yaml`): generate the key as `randAlphaNum 32 | sha256sum | b64enc` — `sha256sum` yields exactly 64 hex chars (32 bytes). Rendered key now decodes to `^[0-9a-f]{64}$`.
- **Admin fail-fast** (`token-crypto.ts` + `crypto.ts`): `makeTokenCrypto()` now validates the key at construction and throws an actionable error (`must be a 64-character hex string … openssl rand -hex 32`) at boot, instead of an opaque runtime 500 on first write. Added `isValidKeyHex()`; `makeTokenCrypto()` takes an optional key param for testability (no env mutation).

## Tests
- helm-unittest: asserts the rendered `SHIPWRIGHT_ENCRYPTION_KEY` base64-decodes to 64-char hex (all 344 pass).
- admin unit: `isValidKeyHex` cases + `makeTokenCrypto` rejects wrong-length/non-hex keys, accepts valid, still identity when unset (23 pass).

## Notes
- Existing installs preserve their key via the `lookup` idiom, so an already-broken cluster needs its key replaced (fresh `minikube:down && minikube:up`, or rotate the secret) — the template fix only affects newly generated keys.
- Chart version intentionally not bumped (handled by `auto-bump-chart` on merge).


dmcaulay (commented):
Verdict: COMMENT — Correct, well-tested fix for the chart-generated encryption key length mismatch, but the PR currently fails its own enforced `ct lint --check-version-increment` CI gate: `charts/shipwright/templates/admin-secret.yaml` changed without a `Chart.yaml` version bump. The PR description's note about `auto-bump-chart` handling this describes a different, post-merge mechanism (triggered by a service release tag) and doesn't address this pre-merge gate — precedent (#2723) required a manual bump for a chart-template-only change. Suggest bumping `charts/shipwright/Chart.yaml`'s version (+ a CHANGELOG.md entry) before merge.

dodizzle (commented):
Verdict: APPROVE — Well-scoped, well-tested fix for the chart-generated encryption key length mismatch (32-char randAlphaNum vs. required 64-char hex). The prior reviewer's chart-version-bump finding is verifiably resolved in this commit (Chart.yaml 1.12.11 → 1.12.12, CHANGELOG.md entry, artifacthub.io/changes annotation all consistent). No new issues found.

dodizzle:
Re: @dmcaulay's chart-version-bump finding — confirmed addressed. `charts/shipwright/Chart.yaml` is now at version `1.12.19` (bumped past main's concurrent auto-bump to `1.12.18` during merge-conflict resolution), satisfying `ct lint --check-version-increment`. The `artifacthub.io/changes` annotation and the `charts/shipwright/CHANGELOG.md` `## [1.12.19]` entry both correctly describe this PR's actual fix (`admin-secret.yaml` generating a 64-char-hex `SHIPWRIGHT_ENCRYPTION_KEY`), not the auto-bump change it superseded. No further changes needed.

dmcaulay (approved):
Verdict: APPROVE — Chart-version-bump finding from the prior review is resolved (Chart.yaml now at 1.12.19 with a matching CHANGELOG entry), the crypto/chart fix is correct and well-tested, and no new issues surfaced in this pass.

PR #2767, opened 2026-08-21
fix(admin): make /admin/agents/new the primary agent-create CTA

dodizzle:
## Problem
On a fresh stack (e.g. minikube, `auth.mode=open`), the Agents page funnels admins into `/admin/provision` — the **Slack-app bootstrap wizard**, whose "Slack App Configuration Token" (`xoxp`) field is unconditionally `required`. A dev/no-Slack setup has no such token, so agent creation dead-ends there. The general, Slack-optional path (`/admin/agents/new`) was only offered as a secondary "New local agent" button.

## Fix
Reframe the Agents-page CTAs so the path that never dead-ends is primary:
- Empty state links to `/admin/agents/new` ("Create one →") instead of `/admin/provision`.
- Primary button is **"+ New agent"** → `/admin/agents/new` (creates in-cluster or self-hosted, no Slack needed).
- `/admin/provision` is kept, demoted to a secondary **"Connect Slack app"** action for setups that use Slack.

The Slack wizard and its route are unchanged and still reachable — only the prominence/labeling of the CTA changes.

## Tests
- `admin-ui-pages.unit.test.ts`: empty state + primary CTA now target `/admin/agents/new`; Slack wizard asserted as secondary; non-admin sees neither.
- `admin-local-agent.smoke.test.ts`: admin sees "+ New agent" → `/admin/agents/new`; non-admin does not.
- All 614 tests across the two suites pass; typecheck + lint clean.

## Follow-up
Pairs with #2764 (the encryption-key format fix). Not addressed here: the toolbar "Provision" nav link still points at the Slack wizard — left as a separate consideration.

### DIFF
--- a/admin/src/admin-local-agent.smoke.test.ts
+++ b/admin/src/admin-local-agent.smoke.test.ts
@@ -1,6 +1,6 @@
 /**
  * admin/src/admin-local-agent.smoke.test.ts
- * Smoke tests for the "New local agent" create flow.
+ * Smoke tests for the "New agent" create flow (/admin/agents/new).
  *
  * Tests:
  * - GET /admin/agents/new — admin session returns 200 with form containing name input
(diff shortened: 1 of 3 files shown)


dodizzle (commented):
Verdict: APPROVE — Small, well-tested, purely presentational CTA reprioritization (no API/auth/behavior changes). Posted as COMMENT since this is a self-review; GitHub blocks self-APPROVE via the API.

PR #2769, opened 2026-08-21
fix(admin): relabel the toolbar 'Provision' nav link to 'Slack'

dodizzle:
## Problem
The admin toolbar's **"Provision"** nav item points at `/admin/provision` — the **Slack-app bootstrap wizard** (requires an `xoxp` token), not a generic agent-provisioning page. The generic label reads as "create an agent here," funneling dev/no-Slack users into the Slack-only flow.

## Fix
Relabel the nav item **"Provision" → "Slack"** (`lib/web/toolbar.ts`). Link target (`/admin/provision`) and active-state highlighting are unchanged — only the label. Agent creation lives under **Agents → + New agent** (`/admin/agents/new`).

## Tests
- `lib/web/toolbar.unit.test.ts`: asserts the nav link is labeled "Slack" → `/admin/provision` and no longer presents a generic "Provision" item (19 pass).
- Existing `admin-ui-pages.unit.test.ts` active-state assertions match on the href+class substring, so they're unaffected.

## Context
Third of three related fixes for the minikube agent-provisioning flow: #2764 (encryption-key format — the hard 500), #2767 (agents-page primary CTA), and this toolbar relabel.

### DIFF
--- a/lib/web/toolbar.ts
+++ b/lib/web/toolbar.ts
@@ -215,7 +215,7 @@ export function renderShipwrightToolbar(
     <a href="${adminBase}/admin/agents" class="vos-wordmark">Shipwright</a>
     <div id="vos-nav-content" class="vos-nav">
       <a href="${adminBase}/admin/agents" class="vos-nav-link${active("/admin/agents")}">Agents</a>
-      <a href="${adminBase}/admin/provision" class="vos-nav-link${active("/admin/provision")}">Provision</a>
+      <a href="${adminBase}/admin/provision" class="vos-nav-link${active("/admin/provision")}">Slack</a>
       <a href="${adminBase}/admin/tasks?state=ready" class="vos-nav-link${active("/admin/tasks")}">Tasks</a>
       <a href="${adminBase}/admin/prs" class="vos-nav-link${active("/admin/prs")}">PRs</a>
       <a href="${adminBase}/admin/chat" class="vos-nav-link${active("/admin/chat")}">Chat</a>
--- a/lib/web/toolbar.unit.test.ts
+++ b/lib/web/toolbar.unit.test.ts
@@ -30,6 +30,12 @@ describe("renderShipwrightToolbar", () => {
       expect(html).toContain("Metrics");
     });
 
+    test("Slack wizard link is labeled 'Slack' (not the generic 'Provision')", () => {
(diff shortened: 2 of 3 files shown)


lib/web/toolbar.ts
```diff
@@ -215,7 +215,7 @@ export function renderShipwrightToolbar(
     <a href="${adminBase}/admin/agents" class="vos-wordmark">Shipwright</a>
     <div id="vos-nav-content" class="vos-nav">
       <a href="${adminBase}/admin/agents" class="vos-nav-link${active("/admin/agents")}">Agents</a>
-      <a href="${adminBase}/admin/provision" class="vos-nav-link${active("/admin/provision")}">Provision</a>
+      <a href="${adminBase}/admin/provision" class="vos-nav-link${active("/admin/provision")}">Slack</a>
```
dmcaulay, line 218: This shared toolbar is also rendered by `metrics/src/dashboard/dashboard-page.ts`. Its golden snapshot fixture (`metrics/src/dashboard/__snapshots__/dashboard-page.unit.test.ts.snap`, 4 occurrences) still hardcodes the old "Provision" text, so this label change breaks 4 snapshot tests in the metrics workspace — confirmed reproducible locally, passing on main. Please regenerate that snapshot and include it in this PR.

dmcaulay (commented):
Verdict: COMMENT — This label change is correct, but it breaks 4 snapshot tests in `metrics/src/dashboard/dashboard-page.unit.test.ts` (metrics dashboard shares this same toolbar component) — the golden snapshot at `metrics/src/dashboard/__snapshots__/dashboard-page.unit.test.ts.snap` still hardcodes the old "Provision" label. Reproduced locally on this branch; confirmed passing on main. Please regenerate the snapshot (`cd metrics && bun test --update-snapshots src/dashboard/dashboard-page.unit.test.ts`) and include it in this PR.

dmcaulay (approved):
Verdict: APPROVE — The prior finding (stale metrics dashboard snapshot after the shared toolbar label change) is fully resolved: all 4 occurrences in `metrics/src/dashboard/__snapshots__/dashboard-page.unit.test.ts.snap` now correctly say "Slack". No remaining stale occurrences found anywhere else in the repo. Clean, well-scoped fix.

PR #2788, opened 2026-08-22
feat: enforce Slack membership check across message, mention, and reaction handlers

dmcaulay:
## Summary
- Adds a single shared `isAllowedSlackSender()` / `shouldRejectSlackSender()` gate in `agent/src/slack.ts`, called at the top of `app.message()`, `app.event('app_mention')`, and `app.event('reaction_added')`, before `runner()` is invoked in each
- Fails open (no behavior change) when the membership ref hasn't synced or `restrict` is false; when `restrict` is true, allows only senders whose resolved email case-insensitively matches a synced member email, otherwise returns early with no `runner()` call, no reply, and no session creation
- Logs a `[slack] rejected sender (not a member): user=...` line for rejected senders; nothing is surfaced to the rejected user
- Wires `resolveUserEmail` and `agentSlackMembershipRef` into `createSlackApp()` in `agent/src/index.ts`

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Single shared check used by all three handlers | MET |
| 2 | Unsynced ref or restrict=false: processed as today | MET |
| 3 | restrict=true + matching email (case-insensitive): processed as today | MET |
| 4 | restrict=true + non-match/undefined email: runner not called, no reply, no session | MET |
| 5 | Test coverage extending slack.unit.test.ts, no existing tests retired | MET |

## Test Plan
- `agent/src/slack.unit.test.ts`: 27/27 passing — direct `isAllowedSlackSender()` unit cases plus per-handler (message/app_mention/reaction_added) coverage of unsynced/restrict-off/member-match/non-member/undefined-email scenarios using an injected fake client, runner, and membership ref
- `task typecheck` — clean
- `task lint` — clean

### DIFF
--- a/agent/src/index.ts
+++ b/agent/src/index.ts
@@ -67,7 +67,7 @@ import { ensureAgentHome, installPlugins, runMiseStartup } from "./setup.ts";
 import { HttpShipwrightRuntimeClient } from "./shipwright-runtime-client.ts";
 import { createSlackApp, hasSlackCredentials } from "./slack.ts";
 import { sendBackOnlineDm } from "./startup-dm.ts";
-import { resolveDisplayName } from "./users.ts";
+import { resolveDisplayName, resolveUserEmail } from "./users.ts";
 import { synthesizeSpeech } from "./voice.ts";
 import {
   HttpWorkQueueReporter,
@@ -565,6 +565,8 @@ if (hasSlackCredentials(slackAppConfig)) {
     (key) => sessions.get(key),
     undefined, // blocksConverter — default
     chatTokenReporter,
+    (userId, client) => resolveUserEmail(userId, client),
+    agentSlackMembershipRef,
   );
 
   await app.start();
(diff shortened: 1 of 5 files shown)


dmcaulay (commented):
Verdict: COMMENT — Core gate logic and handler wiring are correct and well-tested (27/27 unit tests, all 5 acceptance criteria met), but `agent/src/slack.integration.test.ts:3021` still calls `_createSlackApp(...)` directly with 16 positional args, missing the two new trailing params and silently falling through to the real `agentSlackMembershipRef` singleton default — the exact leak this PR's own fix commits (d9ac0fa0, a16c583c) set out to close elsewhere. Not currently flaky (neither message fixture in that test sets `user`), but a future edit adding a `user` field would reintroduce the file-order-dependent CI failure. Full review in the attached comment.

dmcaulay (commented):
Verdict: APPROVE — Follow-up commit correctly fixes the one outstanding finding from the prior review: `agent/src/slack.integration.test.ts:3021`'s direct `_createSlackApp(...)` call now passes a fresh `createAgentSlackMembershipRef()` and `resolveUserEmailFn`, closing the last remaining singleton-leak gap. Verified independently: 195/195 Slack tests pass, typecheck clean, and the fix's arg order/count matches `createSlackApp`'s signature. No new issues.

PR #2804, opened 2026-08-25
feat: add explicit chaining instruction to error-patrol-maintenance cron prompt

dodizzle:
## Summary
- Insert an explicit continuation instruction into `error-patrol-maintenance`'s `prompt:` block in `agent-types/coding/manifest.yaml`, between the three `/shipwright:X` command lines and the "After the chain completes..." line
- The instruction tells the model to invoke each command immediately after the previous one finishes, in the same session, and to disregard a command's own "run /X next" output as a stop signal
- Adds a content-assertion test to `agent-types/coding/manifest.content.test.ts` guarding against a future edit silently dropping the instruction

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `error-patrol-maintenance` prompt block contains an explicit continuation instruction | MET |
| 2 | YAML still parses via `parseAgentTypeManifest`, no structural schema changes | MET |
| 3 | New content-assertion test added to `manifest.content.test.ts` | MET |

## Test Plan
- [x] New content-assertion test in `manifest.content.test.ts` fails before the fix, passes after
- [x] All 24 tests in `manifest.content.test.ts` pass (23 pre-existing + 1 new)
- [x] `task ci` run — 30 pre-existing failures confirmed unrelated (present on `main` too, in unrelated files: `metrics/src/lib/env.unit.test.ts`, `task-store-provider.integration.test.ts`, `prs.openapi.smoke.test.ts`); no regressions introduced by this change

### DIFF
--- a/admin/src/agent-cron-jobs.integration.test.ts
+++ b/admin/src/agent-cron-jobs.integration.test.ts
@@ -144,7 +144,8 @@ const GOLDEN_CODING_CRONS: GoldenCron[] = [
     name: "error-patrol-maintenance",
     schedule: "0 4 * * *",
     prompt:
-      '/shipwright:error-scan\n/shipwright:error-fix\n/shipwright:error-resolve\nAfter the chain completes, write state/error-patrol-ledger.json\'s lastRun field: "<ISO timestamp>". Use [silent] if no new or regressed issues are found.',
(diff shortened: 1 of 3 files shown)


dodizzle (commented):
Verdict: APPROVE — Small, well-scoped prompt change: verified the YAML block-scalar join and the golden-fixture test's JS string concatenation produce byte-identical output, ran the new content test locally (24 pass, 0 fail), confirmed error-scan's own output does contain the 'run /X next' stop-signal phrasing this instruction addresses, and lint/CI are clean. No blocking issues. Sibling crons (entropy-patrol-maintenance, security-patrol-maintenance) have the same latent gap but are out of scope for this PR — worth a follow-up.

PR #2820, opened 2026-08-25
feat: Ingress TLS + cert-manager ingress-shim (nginx)

dodizzle:
## Summary
- Wire `spec.tls` rendering + the `nginx.ingress.kubernetes.io/ssl-redirect` annotation into `templates/ingress.yaml`, gated on TLS being on (either `networking.ingress.tls.enabled` or `tls.certManager.enabled`) and, for ssl-redirect, on the nginx controller flavor
- Add the cert-manager ingress-shim annotation (`cert-manager.io/issuer` or `cert-manager.io/cluster-issuer`, keyed by the effective issuer name/kind) so cert-manager manages the Certificate itself — no chart-rendered Certificate CR on the ingress path, so TLS works on first install regardless of cert-manager readiness
- Broaden the CNH-2.1 `shipwright.ingress.tlsEnabled` helper and refresh stale `values.yaml`/docs comments now that the groundwork is actually wired in

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | All existing cases in tests/ingress_test.yaml pass unchanged | MET |
| 2 | New helm-unittest cases for TLS + cert-manager scenarios (ClusterIssuer, chart-managed Issuer, bring-your-own secretName, redirect=false) | MET |
| 3 | Schema cases for tls.certManager+ingress/ClusterIP, missing issuer, unknown tls key, bad controller enum | MET |
| 4 | `helm template` with default values remains byte-identical to the previous chart version | MET |

## Test Plan
- `helm unittest charts/shipwright` — 367/367 tests, 26/26 suites green (including `compat_clusterip_test.yaml` and `certificate_test.yaml` unchanged)
- `helm lint charts/shipwright` — clean
- `helm template charts/shipwright` — default values byte-identical vs. main (modulo intentionally-random secrets)

### DIFF
--- a/docs/deploy-kubernetes.md
+++ b/docs/deploy-kubernetes.md
@@ -45,9 +45,11 @@ served at `/` and the metrics dashboard at `/dashboard`, on a single host:
 | `gateway` | A Gateway API `Gateway` + `HTTPRoute`s | GKE (managed L7) |
 
 `ingress` and `gateway` are mutually exclusive — only one of the two is ever
-rendered. TLS via cert-manager (`tls.certManager.enabled`) currently applies to
-the **gateway** path only; on the ingress path TLS is configured through
-controller-specific annotations (see the EKS section).
+rendered. TLS via cert-manager (`tls.certManager.enabled`) applies to **both**
+the gateway and ingress paths (see the [GKE](#gke-gateway-api--cert-manager)
+and [EKS](#eks-alb-ingress--cert-manager) sections for details on how each
+implements it). On the ingress path, cert-manager is integrated via the
+ingress-shim annotation mechanism rather than a chart-rendered Certificate.
 
 ### Exposing the task-store externally (opt-in)
 
@@ -473,6 +475,8 @@ In practice prefer a values file for the annotations:
 
 ### Key values
 
+The example below uses **AWS Certificate Manager (ACM)** for TLS — the traditional EKS path that requires no cert-manager:
+
 ```yaml
 networking:
   type: ingress
@@ -487,7 +491,7 @@ networking:
       alb.ingress.kubernetes.io/certificate-arn: <your-acm-certificate-arn>
 tls:
   certManager:
-    enabled: false      # ingress-path TLS is via ALB/ACM annotations, not the Gateway Certificate
+    enabled: false      # ACM path: leave cert-manager off
 auth:
   mode: google
   google:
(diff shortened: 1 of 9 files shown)


docs/deploy-kubernetes.md
```diff
@@ -496,6 +500,39 @@ auth:
...
 
+Alternatively, use **cert-manager** to fully automate certificate provisioning and renewal:
+
+```yaml
+networking:
+  type: ingress
+  ingress:
+    className: alb
+    host: shipwright.example.com
+    controller: nginx              # required by the ingress-shim annotations
```
dodizzle: This is incorrect: the cert-manager ingress-shim annotation is controller-agnostic (see `$certManagerAnnotation := and .Values.tls.certManager.enabled $issuerName` in `templates/ingress.yaml` — not gated on `shipwright.bundled.nginx`). `controller` only gates the nginx-specific `ssl-redirect` annotation. Since this example also has `redirect: true` (the default, line 514) alongside `controller: nginx`, following it verbatim on this ALB (`className: alb`) deployment renders `nginx.ingress.kubernetes.io/ssl-redirect: "true"` — an annotation the AWS Load Balancer Controller doesn't understand, contradicting the comment two lines below that ALB handles the listener itself. Suggest dropping `controller: nginx` here (or setting `redirect: false` with a note that ALB handles its own redirect) and fixing this comment.

dodizzle (commented):
Verdict: COMMENT — Template logic and new test coverage are solid, but the new EKS cert-manager doc example sets `controller: nginx` on an ALB (`className: alb`) deployment with an incorrect comment claiming it's required by the ingress-shim annotation (it isn't — that annotation is controller-agnostic), and combined with the default `redirect: true` it would render a meaningless/misleading nginx `ssl-redirect` annotation on an ALB Ingress if followed verbatim.

dmcaulay (commented):
Verdict: COMMENT — Templating logic, tests, and version bump are solid, and the prior review's `controller: nginx` finding is genuinely fixed — but the fix introduces a new inaccuracy: `alb.ingress.kubernetes.io/listen-ports` alone does not perform HTTP->HTTPS redirect on ALB (that needs the separate `actions.ssl-redirect` annotation), so the new doc example's `redirect: false` comment is misleading.

docs/deploy-kubernetes.md
```diff
@@ -496,6 +500,38 @@ auth:
...
+
+```yaml
+networking:
+  type: ingress
+  ingress:
+    className: alb
+    host: shipwright.example.com
+    tls:
+      enabled: true                # render spec.tls on the Ingress
+      redirect: false              # ALB handles its own HTTP->HTTPS redirect via listen-ports/annotations; this flag only applies to the nginx controller
```
dmcaulay: "ALB handles its own HTTP->HTTPS redirect via listen-ports/annotations" is inaccurate — `alb.ingress.kubernetes.io/listen-ports` only opens the HTTP/HTTPS listeners, it doesn't redirect between them. The actual mechanism is the separate `alb.ingress.kubernetes.io/actions.ssl-redirect` annotation plus a matching rule, which isn't present in this example. As written, a user following this example verbatim gets an ALB serving plain HTTP with no redirect to HTTPS. Either add the `actions.ssl-redirect` annotation/rule, or soften the comment to note ALB does not auto-redirect.

dodizzle (commented):
Verdict: APPROVE — Both prior review findings (the misleading `controller: nginx` on the ALB doc example, and the inaccurate `listen-ports` redirect claim) are verified resolved in the current diff. Template logic, helm-unittest coverage, and docs are solid.

PR #2822, opened 2026-08-26
feat: chart-created cert-manager Issuer for external cert-manager (CNH-5.1)

dmcaulay:
## Summary
- Implements the body of `shipwright.certManager.issuerManifest` (groundwork from CNH-2.1) and adds a new guard-only `templates/cert-manager-issuer.yaml`, rendering a chart-managed cert-manager (Cluster)Issuer when `tls.certManager.issuer.create=true` and no explicit `tls.certManager.issuerRef.name` is set (an explicit `issuerRef.name` always suppresses chart-managed creation).
- Supports two issuer types: `letsencrypt` (ACME email/server/privateKeySecretRef/HTTP-01 solver via the modern `ingressClassName` field, `<issuer>-acme-account` secret name) and `selfsigned` (`spec.selfSigned: {}`). `Issuer` kind renders an explicit `metadata.namespace`; `ClusterIssuer` renders none. Default kind changed to `Issuer` (was `ClusterIssuer`) per spec — safe because default `issuer.create=false` means nothing renders by default either way.
- Wires the existing (CNH-2.1) required-email validation guard into the new template's render path so a missing email fails the render directly (previously only reachable via `NOTES.txt`).

## Acceptance Criteria
| Criterion | Status |
|---|---|
| New helm-unittest suite `cert_manager_issuer_test.yaml` — letsencrypt fields, missing-email failure, selfsigned, ClusterIssuer/no-namespace, issuerRef.name suppression, enabled-alone forward-compat guard (CNH-7.1) | ✅ MET (8 cases, all passing) |
| kubeconform passes against datreeio cert-manager.io/{issuer,clusterissuer}_v1.json | ✅ MET (verified both letsencrypt/Issuer and selfsigned/ClusterIssuer variants) |
| `helm template` with default values remains byte-identical | ✅ MET (default `issuer.create=false` → zero Issuer/ClusterIssuer documents render) |

## Test Plan
- [x] `helm unittest charts/shipwright` — 403/403 passing (8 new cases)
- [x] `helm lint charts/shipwright` — clean
- [x] `task helm:check` (lint → unittest → kubeconform validate, both nginx and traefik profiles) — passing
- [x] Manual kubeconform validation of the new Issuer/ClusterIssuer manifest shapes against the datreeio cert-manager.io CRD schemas
- [x] Confirmed default-values `helm template` output renders zero `kind: Issuer`/`kind: ClusterIssuer` documents (byte-identical guarantee)
- [x] Chart.yaml version bumped 1.14.0 → 1.15.0 with matching CHANGELOG.md entry

### DIFF
--- a/charts/shipwright/templates/_validation.tpl
+++ b/charts/shipwright/templates/_validation.tpl
@@ -46,15 +46,22 @@ Checks (each `fail`s independently — first failing check wins):
      className" rule as check 2 — this is intentionally the same underlying
      contradiction stated as a second, explicit acceptance-criteria-mandated
      check: "an explicit className contradicts a bundled class").
-  5. `tls.certManager.issuer.create=true` with `type=letsencrypt` (CNH-5.1's
-     chart-managed Issuer) while `networking.type=gateway`. The letsencrypt
-     branch of shipwright.certManager.issuerManifest always targets an
-     HTTP-01 solver via an Ingress (shipwright.ingress.className), but
-     templates/ingress.yaml only renders when networking.type=ingress — in
-     gateway mode there is no Ingress for the solver to ever satisfy, so the
-     chart-managed Issuer would render schema-valid but permanently unable to
-     complete its ACME challenge. selfsigned (no ACME/no solver) and
-     bring-your-own via issuerRef.name are both unaffected.
+  5. `tls.certManager.enabled=true` with `tls.certManager.issuer.create=true`
+     and `type=letsencrypt` (CNH-5.1's chart-managed Issuer) while
+     `networking.type=gateway`. The letsencrypt branch of shipwright.
+     certManager.issuerManifest always targets an HTTP-01 solver via an
+     Ingress (shipwright.ingress.className), but templates/ingress.yaml only
+     renders when networking.type=ingress — in gateway mode there is no
+     Ingress for the solver to ever satisfy, so the chart-managed Issuer
+     would render schema-valid but permanently unable to complete its ACME
+     challenge. selfsigned (no ACME/no solver) and bring-your-own via
+     issuerRef.name are both unaffected. The `enabled` guard mirrors
+     templates/cert-manager-issuer.yaml's render guard: with
+     tls.certManager.enabled=false (the default) the whole cert-manager
+     integration is off and nothing would render, so this check must not
+     hard-fail the render on that schema-valid-but-inert combination —
+     `values.schema.json`'s issuerRef/issuer.create `anyOf` constraint itself
+     only activates when `enabled=true`.
 */}}
 {{- define "shipwright.validate" -}}
 {{- $ingress := .Values.networking.ingress -}}
@@ -73,7 +80,7 @@ Checks (each `fail`s independently — first failing check wins):
 {{- if and .Values.tls.certManager.issuer.create (eq .Values.tls.certManager.issuer.type "letsencrypt") (not .Values.tls.certManager.issuer.email) -}}
 {{- fail "tls.certManager.issuer.email is required when tls.certManager.issuer.create=true and tls.certManager.issuer.type=letsencrypt (Let's Encrypt requires a contact email on every ACME account)." -}}
 {{- end -}}
-{{- if and .Values.tls.certManager.issuer.create (eq .Values.tls.certManager.issuer.type "letsencrypt") (eq .Values.networking.type "gateway") -}}
+{{- if and .Values.tls.certManager.enabled .Values.tls.certManager.issuer.create (eq .Values.tls.certManager.issuer.type "letsencrypt") (eq .Values.networking.type "gateway") -}}
 {{- fail "tls.certManager.issuer.create=true with tls.certManager.issuer.type=letsencrypt is not supported when networking.type=gateway — the chart-managed Issuer's HTTP-01 solver targets an Ingress, which never renders in gateway mode (templates/ingress.yaml only renders for networking.type=ingress), so the ACME challenge could never complete. Use tls.certManager.issuer.type=selfsigned, switch networking.type to ingress, or bring your own Issuer via tls.certManager.issuerRef.name instead." -}}
 {{- end -}}
 {{- $bundledClass := include "shipwright.bundledIngressClass" . -}}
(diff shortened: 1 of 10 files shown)


charts/shipwright/templates/cert-manager-issuer.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if and (include "shipwright.certManager.createIssuer" .) (not (include "shipwright.certManager.viaHook" .)) }}
```
dmcaulay: This guard checks `shipwright.certManager.createIssuer && !viaHook`, but never checks `tls.certManager.enabled`. `values.yaml` documents `tls.certManager.enabled: false` as the master off switch for the whole cert-manager integration ("Disabled by default"), and the sibling `templates/certificate.yaml` does gate on it. As written, setting `tls.certManager.issuer.create=true` while leaving `tls.certManager.enabled` at its default `false` still renders a live Issuer/ClusterIssuer — contradicting the documented default-disabled behavior. `values.schema.json`'s cross-field validation for issuerRef/issuer.create is also gated behind `enabled==true`, so nothing catches this combination at schema-validation time either.

Suggest: `{{- if and .Values.tls.certManager.enabled (include "shipwright.certManager.createIssuer" .) (not (include "shipwright.certManager.viaHook" .)) }}`, mirroring certificate.yaml.

dmcaulay (commented):
Verdict: COMMENT — The new templates/cert-manager-issuer.yaml render guard never checks tls.certManager.enabled (the documented master off-switch), so issuer.create=true alone still creates a live Issuer/ClusterIssuer even with enabled=false. Everything else — the helm-unittest coverage, kind/namespace handling, and secret-name rename — is correct and well tested.

charts/shipwright/templates/cert-manager-issuer.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if and .Values.tls.certManager.enabled (include "shipwright.certManager.createIssuer" .) (not (include "shipwright.certManager.viaHook" .)) }}
```
dodizzle, line 1: This guard omits the `networking.type` check that the sibling `templates/certificate.yaml` uses (`eq .Values.networking.type "gateway"`, documented as preventing render on a non-matching networking setup). The default `issuer.type=letsencrypt` branch always renders `spec.acme.solvers[0].http01.ingress.ingressClassName` via `shipwright.ingress.className`, which resolves unconditionally regardless of `networking.type` (falls back to `networking.ingress.controller`, default `"nginx"`). `templates/ingress.yaml` only renders when `networking.type == "ingress"`, and `values.schema.json`'s cert-manager `anyOf` block allows both `"ingress"` and `"gateway"` with no cross-field constraint. So `networking.type=gateway` + `tls.certManager.enabled=true` + `issuer.create=true` + default `type=letsencrypt` is schema-valid but renders a live Issuer whose HTTP-01 solver targets an Ingress class that will never exist in that deployment — cert-manager can never complete the challenge. The new test suite's only `networking.type=gateway` case (lines 123-131) uses `issuerRef.name`, never `issuer.create=true` with the default letsencrypt type, so this is untested.

Suggest gating the letsencrypt branch's solver on `networking.type=ingress` (mirroring `certificate.yaml`), or rejecting this combination in `shipwright.validate`.

dodizzle (commented):
Verdict: COMMENT — Solid PR; the prior review's `tls.certManager.enabled` guard fix is correctly resolved and tested. One important gap remains: the new render guard has no `networking.type` check, so a schema-valid gateway-mode + `issuer.create=true` + default `letsencrypt` combination renders an Issuer with a broken HTTP-01 solver (no Ingress exists in gateway mode to satisfy it). See inline comment.

charts/shipwright/templates/_validation.tpl
```diff
@@ -64,6 +73,9 @@ Checks (each `fail`s independently — first failing check wins):
 {{- if and .Values.tls.certManager.issuer.create (eq .Values.tls.certManager.issuer.type "letsencrypt") (not .Values.tls.certManager.issuer.email) -}}
 {{- fail "tls.certManager.issuer.email is required when tls.certManager.issuer.create=true and tls.certManager.issuer.type=letsencrypt (Let's Encrypt requires a contact email on every ACME account)." -}}
 {{- end -}}
+{{- if and .Values.tls.certManager.issuer.create (eq .Values.tls.certManager.issuer.type "letsencrypt") (eq .Values.networking.type "gateway") -}}
```
dodizzle: This new check tests `issuer.create` + `type=letsencrypt` + `networking.type=gateway` but never checks `tls.certManager.enabled`, unlike every other render path in this chart (e.g. the render guard in `templates/cert-manager-issuer.yaml:1` requires `enabled=true`). Since `values.schema.json`'s `anyOf` constraint on `issuerRef`/`issuer.create` only activates when `enabled=true`, the combination `enabled=false, issuer.create=true, type=letsencrypt (default), networking.type=gateway` is schema-valid and reachable — and this check now hard-fails the render even though the whole cert-manager path is off and nothing would actually render. Reproduced directly:

```
helm template test-release . \
  --set tls.certManager.enabled=false \
  --set tls.certManager.issuer.create=true \
  --set networking.type=gateway \
  --set tls.certManager.issuer.email=admin@example.com
# Error: ... not supported when networking.type=gateway ...
```

`shipwright.validate` runs unconditionally from `templates/NOTES.txt` on every install/upgrade/template, independent of `enabled`. The new test at `tests/validation_test.yaml:79-86` exercises this exact combination without setting `enabled`, so it asserts the current (buggy) behavior as correct.

Suggest adding `.Values.tls.certManager.enabled` to this `and` condition, mirroring the render guard.

dodizzle (commented):
Verdict: COMMENT — The gateway-mode fail-fast fix correctly resolves the prior review's finding, but the new validation check itself omits the `tls.certManager.enabled` guard, causing a reachable false-positive render failure when the feature is fully disabled. See inline comment.

dodizzle (approved):
Verdict: APPROVE — Both findings from prior review rounds (missing `networking.type=gateway` check, and the follow-up `tls.certManager.enabled` gap) are now correctly resolved in a single combined validation guard (`_validation.tpl:83`), with matching test coverage in `validation_test.yaml`. No new issues found.

PR #2824, opened 2026-08-26
feat: cert-manager bootstrap hook for the bundled cert-manager path

dodizzle:
## ⚠️ MERGE GATE — DO NOT AUTO-MERGE

This is the highest-complexity task in the cloud-native-helm-chart effort and is flagged for **human oversight** before it reaches the Vitals OS deployment of shipwright. Leave this PR at reviewed/approved + CI green and **stop there** — do not let the standard deploy loop auto-merge or auto-deploy it. The paired HITL task **CNH-7.3** performs the actual merge.

## Summary
- Adds a post-install/post-upgrade bootstrap Job (`<fullname>-cert-manager-bootstrap`) that applies the chart-managed Issuer/Certificate via `kubectl apply --server-side` once the bundled cert-manager subchart's CRDs and webhook are Available — needed because when cert-manager ships in the *same* Helm release, its CRDs are templates too and don't exist at first-install apply time for a plain inline `cert-manager.io/v1` manifest.
- Ships the Issuer/Certificate bodies inside a ConfigMap (reusing the existing `shipwright.certManager.issuerManifest`/`certificateManifest` helpers verbatim — one source of truth), adds RBAC (SA/Role/ClusterRole as regular non-hook resources so the token survives upgrades), and a pre-delete cleanup Job (gated on `tls.certManager.bootstrap.cleanupOnDelete`, default true) that removes the applied CRs while intentionally keeping the issued TLS Secret and ACME account Secret.
- Wires the previously-dead-code `shipwright.certManager.viaHook` helper (`tls.certManager.enabled && cert-manager.enabled && (issuer.create || networking.type=gateway)`) so the existing inline `certificate.yaml`/`cert-manager-issuer.yaml` guards — already wired for this in CNH-2.1/CNH-5.1 — now correctly suppress inline rendering in favor of the hook path.

## Acceptance Criteria
| 1 | New `tests/cert_manager_bootstrap_job_test.yaml` (helm-unittest) covering hook annotations/weight/delete-policy, init-container wait targets, apply args, ConfigMap contents per networking type, SA name, Role rules, ClusterRole gating, runAsNonRoot, overrides, cleanup Job gating, zero-docs when disabled | MET | 22 test cases, all passing |
| 2 | `certificate_test.yaml` asserts no inline Certificate when cert-manager.enabled=true + gateway + tls.certManager.enabled=true | MET | New case added, `hasDocuments: count: 0` |
| 3 | `helm template` with default values byte-identical to previous chart version | MET | Verified via diff against parent commit — only random secret bytes differ |
| 4 | PR left open at reviewed/approved + CI green — no auto-merge/auto-deploy; CNH-7.3 (HITL) performs the merge | Enforced by process, see banner above | N/A |

## Test Plan
- [x] `task helm:check` — helm lint + 443/443 helm-unittest tests pass + kubeconform strict validation (default + traefik profiles) clean
- [x] Default-values render confirmed byte-identical to previous chart version (modulo random secret bytes)
- [x] `task ci` lint/check-strings/check-version-sync/check-config-docs/typecheck all green; the 29 `bun test` failures observed locally in `metrics/`/`task-store/` are pre-existing test-process leakage present on `main` too (confirmed via a clean checkout — unrelated to this chart-only diff)
- [x] Docs refreshed (`docs/deploy-kubernetes.md`) to document the new bootstrap Job behavior

### DIFF
--- a/charts/shipwright/CHANGELOG.md
+++ b/charts/shipwright/CHANGELOG.md
@@ -10,6 +10,12 @@ independent of `appVersion`. CI enforces this with
 `ct lint --check-version-increment`. Each release here must mirror the
 `artifacthub.io/changes` annotation in `Chart.yaml`.
 
+## [1.17.0] - 2026-08-26
+
+### Added
+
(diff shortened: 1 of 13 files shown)


dodizzle (commented):
Verdict: APPROVE — Careful, well-documented, and well-tested cert-manager bootstrap hook. Verified the RBAC-non-hook claim (no helm.sh/hook annotations in cert-manager-bootstrap-rbac.yaml), the hook-weight ordering against postgres-db-bootstrap-job.yaml (-5), and the default-render no-op (both cert-manager.enabled/tls.certManager.enabled default false). No blocking findings. Note this PR's own banner: flagged for human oversight, paired HITL task CNH-7.3 performs the actual merge — this review doesn't override that gate.

PR #2825, opened 2026-08-26
feat: scope PR redispatch cooldown key by phase

dmcaulay:
## Summary
- Scope the PR redispatch cooldown key in `agent/src/loop-orchestrator.ts` to `${pr.id}:${pr.phase ?? "review"}` instead of bare `pr.id`, so each phase (review/patch/deploy) tracks its own redispatch history independently.
- Update the doc comment above `PR_REDISPATCH_COOLDOWN_MS` to describe the phase-scoped design and remove the disproven claim that a shared key was safe (a real review dispatch on an unchanged commit was confirmed live to suppress an unrelated patch need — e.g. resolving a live merge conflict — for the full 25-minute cooldown window).
- Add a regression unit test asserting a review dispatch does not suppress a same-PR/same-commit patch dispatch within the cooldown window; the three existing same-phase cooldown tests are retained unchanged.

## Acceptance Criteria

| 1 | `isPrDispatchSuppressed` lookup key and `lastPrDispatch.set()` write key both use `${pr.id}:${pr.phase ?? "review"}` | MET | New `prDispatchCooldownKey()` helper used at both the lookup and write sites |
| 2 | Doc comment above `PR_REDISPATCH_COOLDOWN_MS` updated, no longer claims shared key is safe | MET | Comment rewritten to describe phase-scoping and cites the disproven PH-1.3 assumption |
| 3 | New regression unit test added; 3 existing cooldown tests retained unchanged | MET | New test covers the cross-phase suppression bug; existing suppression/expiry/commit-change tests untouched |
| 4 | No DB/API/schema changes; only relevant files touched | MET | Diff touches `agent/src/loop-orchestrator.ts`, its unit test, and one doc-refresh line in `docs/agent.md` |

## Test Plan
- `bun test agent/src/loop-orchestrator.unit.test.ts` — 102/102 pass, including the new regression test (confirmed RED before the fix, GREEN after)
- `bunx tsc --noEmit` (agent package) — clean
- Coverage on `agent/src/loop-orchestrator.ts`: 96.55% lines / 92.34% branches


agent/src/loop-orchestrator.unit.test.ts
```diff
@@ -3521,6 +3521,40 @@ describe("createLoopOrchestrator", () => {
     expect(creates).toHaveLength(2);
   });
 
+  test("dispatching a review candidate for a PR does not suppress a patch candidate for the same PR id/commit within the cooldown window", async () => {
+    // Regression coverage for the ok-wow-infra#2373-shaped bug: the cooldown
```
dmcaulay: This reads like a real external org/repo/issue citation ("ok-wow-infra#2373") rather than a placeholder — it doesn't match this file's established `acme/x#N` fixture convention (used 15+ times elsewhere in this file, including twice in this exact test). Per CLAUDE.md's "Before you commit — this repository is going public" section, internal PR/issue/org citations should be scrubbed before commit since git history is permanent; `task check-strings`'s banned-patterns list doesn't cover this string, so it won't be caught by CI. Suggest replacing with a generic placeholder consistent with the file's convention, e.g. "the acme/x#5-shaped bug".

dmcaulay (commented):
Verdict: COMMENT — Clean implementation of the phase-scoped cooldown fix, but the new test comment's "ok-wow-infra#2373" reference looks like a real external org/repo/issue citation that should be scrubbed before it lands in permanent public git history.

dodizzle (approved):
Verdict: APPROVE — Correct, narrowly-scoped fix: cooldown key is now phase-scoped (`${pr.id}:${phase}`) applied consistently at both the read and write sites, with a well-constructed regression test and matching doc updates. No blocking issues.

PR #2829, opened 2026-08-26
fix: default allow_self_review to false on missing/unparseable policy

dodizzle:
## Summary
- Flip the code-level fallback for `allow_self_review` from default-`true` to default-`false` in `agent/src/check-helpers.ts` and its mirrored source `plugins/shipwright/scripts/check-helpers.ts` — a missing/corrupt policy file now defaults to requiring human review instead of skipping it.
- Correct four docs/site locations (`docs/configuration.md`, `site/src/data/config-vars.ts`, `site/src/content/docs/reference.mdx`) that documented the old fallback as "the default" instead of the real, template-seeded `false` value; `docs/configuration.md` also retains a short technical note that `true` is the internal fallback if the policy file is ever missing.
- Updated the two existing unit test assertions that expected the old `true` default, plus a downstream test in `check-deploy.unit.test.ts` that also asserted the old default.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `agent/src/check-helpers.ts`: `parseAllowSelfReview`/`readAllowSelfReview` default to `false` on missing/unparseable input | MET |
| 2 | `plugins/shipwright/scripts/check-helpers.ts` mirrors the identical change | MET |
| 3 | Docs state `false` default with reworded opt-in description + retained fallback note | MET |
| 4 | Existing unit tests flipped to assert `false`; plugin-side coverage gap intentionally not backfilled | MET |

## Test Plan
- `bun test agent/src/check-helpers.unit.test.ts agent/src/check-deploy.unit.test.ts` — 202 pass, 0 fail
- `task lint`, `task typecheck`, `task check-config-docs`, `task check-version-sync` — all pass
- `task ci`'s `test:coverage` surfaced 29 pre-existing failures unrelated to this change (metrics env-var test pollution, task-store smoke-test ordering); confirmed identical on `main`


dodizzle (commented):
Verdict: COMMENT — Core fix in check-helpers.ts is correct, but merge.md, deploy.md, and review.md still document/gate on the old `allow_self_review` default of `true`, leaving the same permissive-fallback gap live in the merge/deploy self-approval path.

dodizzle (commented):
Verdict: COMMENT — Core allow_self_review default flip is correct and fully mirrored, and this revision fixed the deploy.md/merge.md/review.md gap flagged in my prior review. One more doc location, site/src/content/docs/configuring-autonomy.mdx (lines 135 and 152), still states the stale `true` default and should be updated to match.

dodizzle (commented):
Verdict: APPROVE — Both findings from prior review rounds (stale `true` default in merge.md/deploy.md/review.md, and in configuring-autonomy.mdx) are confirmed resolved by direct file inspection. Core fallback fix is correct and symmetric across both mirrored `check-helpers.ts` copies, docs are now fully consistent, and tests pass (202/202). No new issues.

PR #2843, opened 2026-08-26
feat: downstream-compatibility render-diff script (CNH-8.3)

dmcaulay:
## Summary
- Add `scripts/check-downstream-compat.sh <wrapper-dir>`, which renders a downstream "wrapper" chart's `shipwright` dependency three ways — as pinned (A), swapped to a published `gh-pages` baseline version (B), and swapped to a chart packaged from this repo's working tree (C) — and diffs them: A-vs-B allows only image-tag-only differences, B-vs-C must be completely empty.
- Add the `helm:downstream-compat` Taskfile target (requires `WRAPPER` env var, following the existing `db:provision` precondition pattern).
- Document the target and a copy-pasteable self-test smoke run in CONTRIBUTING.md, using this repo's own `charts/shipwright` as a stand-in wrapper fixture.
- The script derives the published Helm repo URL entirely at runtime (flag, wrapper's own `Chart.yaml` dependency entry, or `git remote get-url origin` + GitHub Pages convention) — no hardcoded org/repo/hostname identifiers.

## Acceptance Criteria
| 1 | Test decision: no unit-test seam; documented self-test smoke run in CONTRIBUTING.md using charts/shipwright as stand-in wrapper, confirming empty diff comparing chart to itself | MET | CONTRIBUTING.md "Self-test smoke run" section; independently re-ran `./scripts/check-downstream-compat.sh charts/shipwright` and confirmed `A vs B: identical.` / `B vs C: identical.` / `PASSED`, exit 0 |
| 2 | Script has no hardcoded org/repo/hostname identifiers (`task check-strings` passes); `helm:downstream-compat` Taskfile target documented in CONTRIBUTING.md | MET | `task check-strings` passes; repo URL resolution is fully runtime-derived; Taskfile target + CONTRIBUTING.md docs both present |

## Test Plan
- [x] `task lint` — passes
- [x] `task check-strings` — passes ("No banned strings found")
- [x] `task typecheck` — passes across all workspaces
- [x] Self-test smoke run (`./scripts/check-downstream-compat.sh charts/shipwright`) — passes end-to-end, reports empty A-vs-B and B-vs-C diffs
- [x] Manually exercised the non-self-referential path with a synthetic real wrapper chart (real `shipwright` dependency, local HTTP Helm repo) — confirmed image-tag-only diffs are allowlisted and genuine non-image-tag diffs fail loudly on both A-vs-B and B-vs-C
- Note: `task ci`'s `test:coverage` step surfaces 29 pre-existing failures in `metrics/src/lib/errors.unit.test.ts` and `task-store/src/routes/prs.openapi.smoke.test.ts` — both files are untouched by this PR and pass 100% in isolation on a clean `main` checkout; this is known full-suite cross-test flakiness, unrelated to this change (no TS files are touched here).

### DIFF
--- a/scripts/check-downstream-compat.sh
+++ b/scripts/check-downstream-compat.sh
@@ -289,6 +289,13 @@ wrapper_dep_field() {
   yq -r ".dependencies[] | select(.name == \"${DEP_NAME}\") | .${field} // \"\"" "${WRAPPER_CHART_YAML}" 2>/dev/null | head -n1
 }
 
+# `helm template` a chart dir with the resolved VALUES_ARGS, writing to $2.
+render_variant() {
+  local chart_dir="$1"
+  local out="$2"
+  helm template compat-check "${chart_dir}" "${VALUES_ARGS[@]}" > "${out}"
+}
+
 # Normalize output that legitimately (and expectedly) differs between two
 # otherwise-identical renders, so it never shows up as a spurious diff line:
 #
@@ -421,7 +428,7 @@ if [[ -f "${RENDER_A_DIR}/Chart.lock" || -d "${WRAPPER_DIR}/charts" ]]; then
     cp "${WRAPPER_DIR}/Chart.lock" "${RENDER_A_DIR}/Chart.lock"
   fi
 fi
-helm template compat-check "${RENDER_A_DIR}" "${VALUES_ARGS[@]}" > "${RENDER_A}"
+render_variant "${RENDER_A_DIR}" "${RENDER_A}"
 
 # ---------------------------------------------------------------------------
 # Determine whether the wrapper has a matching dependency to swap
@@ -449,7 +456,7 @@ if [[ "${HAS_DEP}" == "true" ]]; then
   copy_wrapper "${RENDER_B_DIR}"
   yq -i "(.dependencies[] | select(.name == \"${DEP_NAME}\")).version = \"${EFFECTIVE_BASELINE}\" | (.dependencies[] | select(.name == \"${DEP_NAME}\")).repository = \"${REPO_URL_RESOLVED}\"" "${RENDER_B_DIR}/Chart.yaml"
   helm dependency update "${RENDER_B_DIR}" >/dev/null
-  helm template compat-check "${RENDER_B_DIR}" "${VALUES_ARGS[@]}" > "${RENDER_B}"
+  render_variant "${RENDER_B_DIR}" "${RENDER_B}"
 else
   cp "${RENDER_A}" "${RENDER_B}"
 fi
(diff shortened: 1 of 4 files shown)


dodizzle (commented):
Verdict: COMMENT — One real correctness bug in the Render A restore-guard (scripts/check-downstream-compat.sh:421): it checks `-f "${RENDER_A_DIR}/Chart.lock"`, a path `copy_wrapper()` just deleted, instead of `-f "${WRAPPER_DIR}/Chart.lock"`. This can silently drop a wrapper chart's Chart.lock from Render A whenever the wrapper has a lock file but no vendored `charts/` dir — a common layout — undetected by the self-test since `charts/shipwright` happens to have both. Otherwise a well-designed script: runtime-derived repo URL, sound image-tag-diff allowlisting, and thorough docs.

scripts/check-downstream-compat.sh
```diff
@@ -0,0 +1,626 @@
...
+fi
+
+# ---------------------------------------------------------------------------
+# Render A — the wrapper as-is
+# ---------------------------------------------------------------------------
+
+echo "[check-downstream-compat] rendering A (wrapper as-is)..."
+RENDER_A_DIR="${SCRATCH_DIR}/wrapper-as-is"
+copy_wrapper "${RENDER_A_DIR}"
+if [[ -f "${RENDER_A_DIR}/Chart.lock" || -d "${WRAPPER_DIR}/charts" ]]; then
```
dodizzle: This guard checks `-f "${RENDER_A_DIR}/Chart.lock"` — the destination path `copy_wrapper()` on the previous line just deleted, so this operand is always false — instead of `-f "${WRAPPER_DIR}/Chart.lock"` (the actual source, which the inner check three lines below correctly uses). Because the outer condition ORs with `-d "${WRAPPER_DIR}/charts"`, a wrapper that has a `Chart.lock` but no vendored `charts/` directory (lock committed, `.tgz` dependency archives gitignored — a common layout) skips this whole restore block, and Render A silently templates with no lock file at all, defeating "Render A truly reflects whatever the wrapper currently has pinned." The script's trailing `helm lint`/`helm template` checks won't catch this either, since those run against the untouched `WRAPPER_DIR`, not `RENDER_A_DIR`.

Suggested fix: `if [[ -f "${WRAPPER_DIR}/Chart.lock" || -d "${WRAPPER_DIR}/charts" ]]; then`

scripts/check-downstream-compat.sh
```diff
@@ -0,0 +1,626 @@
...
+import re
+import sys
+
+diff_path, out_path = sys.argv[1], sys.argv[2]
+with open(diff_path) as f:
+    lines = f.readlines()
+
+# Collect contiguous +/- blocks (hunks of removed/added lines) and pair them
+# positionally to check for image-tag-only differences.
+IMAGE_TAG_RE = re.compile(r'^(\s*(?:-\s*)?image:\s*\S+):[^:\s]+(\s*)$')
```
dodizzle: `IMAGE_TAG_RE = re.compile(r'^(\s*(?:-\s*)?image:\s*\S+):[^:\s]+(\s*)$')` strips everything after the last `:` on an `image:` line and treats it as an allowlisted "tag." Because of greedy backtracking this also matches digest-pinned images (`image: repo@sha256:<hex>`) — two genuinely different digests normalize to the identical string (verified: `norm("...@sha256:aaaa1111")` and `norm("...@sha256:bbbb2222")` produce the same output). `charts/shipwright/values.yaml:969` documents digest-pinning as the recommended pattern for production (`# Pin to a concrete digest for production`). Since Render A/B share the wrapper's `values.yaml` and differ only in the shipwright chart version supplying the dependency, a real content change behind an unchanged digest-pinned image reference between the pinned and baseline versions would be silently swallowed by this allowlist — exactly the class of change this diff exists to catch.

Suggestion: exclude `@sha256:` digest references from the image-tag allowlist match — treat a digest change as a real diff, not a normalized tag.

dodizzle (commented):
Verdict: COMMENT — Prior restore-guard bug confirmed fixed with a targeted regression test. One remaining gap: the image-tag allowlist regex also normalizes digest-pinned images, so a real digest change between the pinned and baseline shipwright versions would be silently allowlisted as a tag-only diff.

dodizzle (approved):
Verdict: APPROVE — Both prior findings (Render A restore-guard path, digest-pinned image allowlisting) are fixed with targeted regression tests that extract and execute the real script logic. No new issues found.

PR #2852, opened 2026-08-27
fix: parse plain-YAML agent-policy.md booleans in check-helpers

dmcaulay:
## Summary
- `parseAllowSelfReview`/`parseCleanupMergedWorktrees` in `agent/src/check-helpers.ts` (and `parseAllowSelfReview` in `plugins/shipwright/scripts/check-helpers.ts`) only matched the markdown-table (`` `field` | true ``) and bold-colon (`**field**: true`) formats.
- `state/agent-policy.md` is seeded and documented (`docs/configuration.md`) as plain YAML frontmatter (`field: true`, no backticks/bold) — that format silently failed to match, so `allow_self_review: true` parsed as `false` and self-authored PRs kept getting excluded from `check-review`'s candidacy with `check: "self-review"`.
- Widened both regexes to accept optional backtick/bold wrapping around the key with either `:` or `|` as the separator, covering table, bold, and plain-YAML styles.

## Test plan
- [x] Added unit tests for the plain-YAML case in `agent/src/check-helpers.unit.test.ts` (both `parseAllowSelfReview` and `parseCleanupMergedWorktrees`)
- [x] `bun test agent/src/check-helpers.unit.test.ts` — 153 pass
- [x] `bun test plugins/shipwright/scripts/check-helpers.unit.test.ts agent/src/check-review.unit.test.ts` — 140 pass
- [x] Typecheck clean in both `agent/` and `plugins/shipwright/` workspaces
- [x] Verified against the live `~/.shipwright/workspace/state/agent-policy.md` — `parseAllowSelfReview` now returns `true`

### DIFF
--- a/agent/src/check-helpers.ts
+++ b/agent/src/check-helpers.ts
@@ -99,7 +99,7 @@ export function parseCleanupMergedWorktrees(content: string): boolean {
 
 export function parseCleanupAfterDays(content: string): number {
   const match = content.match(
-    /(?:`cleanup_after_days`\s*\|\s*|\*\*cleanup_after_days\*\*:\s*)(\d+)/,
+    /`?\*{0,2}\bcleanup_after_days\b\*{0,2}`?\s*[:|]\s*\*{0,2}(\d+)\b/i,
   );
   if (!match) return 14; // default 14 if missing
   const parsed = Number.parseInt(match[1], 10);
--- a/agent/src/check-helpers.ts
+++ b/agent/src/check-helpers.ts
@@ -85,14 +85,14 @@ export interface Task {
 
 export function parseAllowSelfReview(content: string): boolean {
   const match = content.match(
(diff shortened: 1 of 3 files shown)


dodizzle (commented):
Verdict: COMMENT — The core fix is correct and well-tested, but `parseCleanupAfterDays` (agent/src/check-helpers.ts:100, same file) has the identical plain-YAML parsing gap this PR fixes for its two sibling functions — it still only matches table/bold-colon styles, so a plain-YAML `cleanup_after_days: N` in `state/agent-policy.md` silently falls back to the hardcoded default (14) in `worktree-reaper.ts`. It currently works only by coincidence since the default matches the fallback. Worth widening the same way and adding a plain-YAML test case, since that line falls outside this diff's hunks so it can't be left as an inline comment.

dodizzle (approved):
Verdict: APPROVE — Prior finding (parseCleanupAfterDays plain-YAML parsing gap) is fully resolved: the regex was widened to match the sibling functions' pattern and a matching plain-YAML unit test was added. No new issues.

PR #2866, opened 2026-08-27
docs: convert cron troubleshooting prose to symptom/check table

dodizzle:
## Summary
- Converts the "Two outcomes that are expected, not bugs" troubleshooting section in `cron-jobs.mdx` from nested prose into a Symptom / Check / Meaning table
- Matches how an operator actually uses this section during an incident (arrives with a symptom, not reading linearly) — no diagnostic content dropped
- Matches the existing table separator style used elsewhere in the same doc

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Section content is a table with at minimum Symptom, Check, and Meaning columns; same diagnostic content as current prose, no information dropped | MET |
| 2 | No test change needed -- same-page prose-to-table conversion with no heading/anchor rename, and no existing spec asserts on this section's specific text | MET |

## Test Plan
- Confirmed no existing spec (`site/tests/docs-cron-jobs.spec.ts`) asserts on the removed prose text
- `bunx astro build` succeeds; verified the rendered `dist/docs/cron-jobs/index.html` contains the new table
- Heading text/anchor unchanged; only the section body was restructured


site/src/content/docs/cron-jobs.mdx
```diff
@@ -170,29 +170,14 @@ conclusion.
...
-check whether `shipwright-loop` itself is enabled (see "Loop-driven vs standalone crons" above): a
-reconciled pipeline cron is a child of `shipwright-loop` and never gets its own independent
-schedule, so its `enabled` toggle only takes effect once `shipwright-loop` reads it and dispatches
-that phase directly — there's no standalone tick to show up as `silent` or `skipped` in the
-meantime.
+Use this table when a cron produces no output or shows an unexpected outcome. Arrive with a symptom (what you're observing), then follow the check and meaning columns to confirm whether it's a normal state or a real issue.
+
+| Symptom | Check | Meaning |
+|---|---|---|
+| A cron run shows outcome `silent` in Cron Logs | Look at the run details — this is normal. The run completed successfully and simply had no qualifying work to report. | Most cron ticks find no work and end silently by design. This is the system working correctly, not a bug. Before assuming something's wrong, confirm the run finished (check its duration). |
```
dmcaulay: This row adds "Before assuming something's wrong, confirm the run finished (check its duration)" — that instruction isn't in the removed prose, which only said a `silent` outcome "completed successfully and simply had nothing worth posting." The doc's documented Outcome values (`posted`/`dm`/`silent`/`skipped`/`error`, line 144) are all terminal states, so there's no in-progress state a `silent` outcome could be confused with — "check its duration" doesn't map to anything in the original content. Minor, but worth dropping or reverting to the original phrasing so this stays a faithful restatement rather than adding new guidance.

dmcaulay (commented):
Verdict: COMMENT — One added diagnostic instruction ("check its duration") isn't present in the original prose and isn't grounded in the doc's documented outcome states; otherwise a clean, faithful prose-to-table conversion.

dmcaulay (approved):
Verdict: APPROVE — Faithful prose-to-table conversion, no information loss, and the prior finding (ungrounded "check its duration" instruction) has been removed and confirmed resolved.

PR #2899, opened 2026-08-29
fix(chat): heartbeat long-running replies instead of a flat 90s UI timeout

dodizzle:
## Summary
- Chat replies that take multiple Claude turns (e.g. a multi-task status question) were completing successfully, but the admin chat UI showed "Request timed out. Please try again." because its poller gave up after a flat 90s regardless of whether the agent was still working.
- The agent's chat poller now sends a throttled heartbeat (`POST /threads/:id/messages/:id/heartbeat`, min 3s apart) via the Claude runner's per-turn progress callback while it works a claimed message.
- The admin UI's poll loop now extends its timeout whenever the pending message's `claimedAt`/`heartbeatAt` advances, instead of tripping the same flat cutoff — a message that's never claimed, or claimed but silent, still times out at ~90s (unchanged); a message getting heartbeats keeps extending up to a 10-minute hard cap.

## Test plan
- [x] `chat/` unit/smoke/integration tests (new `heartbeat()` route + service tests) — `bun test chat/`
- [x] `agent/` unit/integration tests (new heartbeat-during-run poller tests, throttling, failure isolation) — `bun test agent/`
- [x] `admin/` unit/smoke tests (new content test asserting the flat `MAX_POLLS` cutoff is gone and heartbeat-aware constants exist) — `bun test admin/`
- [x] `bun run --filter=@shipwright/chat --filter=@shipwright/agent --filter=@shipwright/admin typecheck`
- [x] `bunx biome check` / `task lint`
- [x] Verified the new Prisma migration (`heartbeatAt` on `Message`) applies cleanly against a fresh Postgres

https://claude.ai/code/session_01VzZcAciAPGjwcVxXoz7TvL


chat/src/message-service.ts
```diff
@@ -205,6 +214,18 @@ export class MessageService implements MessageServiceLike {
     }
   }
 
+  async heartbeat(id: string): Promise<Message | null> {
```
dmcaulay, line 218: `heartbeat()` only checks the message exists — it never verifies it's actually claimed (`claimed === true`) or unreplied (`repliedAt === null`) before bumping `heartbeatAt`. `reply()` three methods below guards `repliedAt !== null`, and `task-store/src/routes/tasks.ts`'s heartbeat route calls `requireOwnership` (tied to the actual claim) before touching `heartbeatAt` — this method diverges from both. The chat service's own smoke test (`messages.smoke.test.ts:316-334`) confirms a never-claimed message can be heartbeated and returns 200. Blast radius is currently limited (nothing chat-side reads `heartbeatAt` besides the admin UI's display/timeout logic), but worth guarding with a `claimed`/`repliedAt` check mirroring `reply()`'s pattern before this gets relied on more heavily.

dmcaulay (commented):
Verdict: COMMENT — one important finding: heartbeat() doesn't validate the message is claimed/unreplied before bumping heartbeatAt, unlike its sibling reply() and the task-store precedent.

dmcaulay (approved):
Verdict: APPROVE — the prior finding (heartbeat() missing a claimed/repliedAt guard) is fully resolved: it now uses `where: { id, claimed: true, repliedAt: null }` with P2025-as-not-found handling, exactly mirroring the sibling `claim()` method, and is backed by new integration/smoke tests covering not-found, unclaimed, and already-replied cases. No new issues found — the migration is a nullable additive column (rolling-deploy safe), the API surface is additive, and test coverage matches the layer conventions.

PR #2902, opened 2026-08-29
fix(chat): make the reply heartbeat interval-driven and owner-scoped

dodizzle:
Follow-up to #2899. The heartbeat landed there stops during exactly the silence it was built to cover.

## The defect

The heartbeat was driven by `onProgress`, which fires only when a new assistant `message.id` is seen (`claude.ts:362` dedupes on it). A long-running tool call — a multi-minute `Bash`, say — emits one `tool_use` line and then nothing until it returns. So the heartbeat goes quiet and the UI declares a stall on a perfectly healthy run.

Liveness is a property of the agent process, not of the model's output cadence, so it now gets its own timer.

## Changes

- **Interval-driven heartbeat** (`agent/src/chat-poller.ts`) — a 3s timer running for the whole reply, cleared in a `finally` so it stops on success, throw, or early return. Replaces the progress-driven throttle.
- **Owner-scoped `heartbeat()`** (`chat/src/message-service.ts`) — extends #2899's guard with `claimedBy` and `role: "user"`. Previously any caller authorized for the thread could keep *any* claimed message looking alive; now only the worker holding the claim can. The route derives `claimedBy` exactly as `claim` does.
- **Overlap guard on `start()`** — `setInterval` fired regardless of whether the previous `pollOnce` was still running, stacking concurrent iterations for as long as a slow poll ran. The existing `if (timer) return` only guarded double-`start()`.
- **Null-deref in the poll loop** (`admin/src/admin-ui-pages.ts`) — `lastUserMessageTime.getTime()` throws when it is null on a fresh page load, and the `> null` cutoff coerces to `> 0`, treating every assistant message as a new reply. Elapsed now reads the server's `createdAt`, so it also survives a reload.
- **Regenerated `chat/openapi.json`** — #2899 added the route and schema field but not the spec.
- **`ChatMessage.claimed` / `.claimedAt`** (`admin/src/http-chat-client.ts`) — the UI reads `pending.claimedAt`, which was never declared on the interface.

## Testing

- Interval heartbeat covered by a fake `setInterval`/`clearInterval` pair (injected, no wall-clock waits). The key regression test uses a runner that reports **no** progress at all and asserts the heartbeat still fires.
- Ownership rejection covered at both the service layer (integration, real Postgres) and the route layer (smoke).
- Deduped two of my integration tests in favour of #2899's stronger versions, which also assert `heartbeatAt` stays null.
- `chat/` 168 pass against real Postgres; `chat/ agent/ admin/` 3398 pass. The one failure, `check-deploy.unit.test.ts > repos reflects an empty scan`, is pre-existing and environmental — it fails identically on clean `main`.

https://claude.ai/code/session_01VzZcAciAPGjwcVxXoz7TvL

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -4441,10 +4441,13 @@ export function renderChatThreadPage(
       .then(function(r) { return r.json(); })
       .then(function(data) {
         var msgs = data.messages || [];
+        // Guard the null cutoff: comparing a Date against null coerces to a
+        // compare against 0, which would treat every assistant message in the
+        // thread as a new reply.
         var cutoff = lastUserMessageTime;
-        var replies = msgs.filter(function(m) {
+        var replies = cutoff ? msgs.filter(function(m) {
           return m.role === 'assistant' && new Date(m.createdAt) > cutoff;
-        });
+        }) : [];
         if (replies.length > 0) {
           stopPolling();
           removeThinkingIndicator();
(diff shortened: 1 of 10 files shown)


dmcaulay (approved):
Verdict: APPROVE — Correct, well-tested fix for #2899's heartbeat defect (progress-driven heartbeat went silent during long tool calls); ownership scoping and overlap guard are solid, no blocking issues.

PR #2908, opened 2026-08-29
refactor: move chat page inline styles into classes (CFB-1.3)

dodizzle:
## Summary
- Migrated ~10 inline-styled elements on the chat thread page (`renderChatThreadPage()`) into a new `chatThreadStyles` CSS constant, beside `threadPaneStyles`/`chatPageStyles`, so the existing mobile media queries can actually take effect (inline styles were outranking them)
- Exported `CHAT_BUBBLE_CLASS`/`CHAT_BUBBLE_INNER_CLASS`/`CHAT_BUBBLE_INNER_WIDE_CLASS` constants and interpolated them into both the server-rendered bubble markup and the inline JS `addBubble()` client builder, so the two can no longer drift apart
- Dropped the now-unneeded `!important` on `.chat-bubble-inner` in the mobile media query, since no inline `max-width` remains for it to fight

## Acceptance Criteria
| 1 | Strict 1:1 lift — byte-identical class bodies | MET | chatThreadStyles const bodies match removed inline style values exactly |
| 2 | Class-name constants shared by server + inline JS | MET | CHAT_BUBBLE_CLASS/CHAT_BUBBLE_INNER_CLASS exported and interpolated into both renderMessageBubble() and addBubble() |
| 3 | Unit test: 5 migrated elements emit no style= | MET | 13 new unit tests in admin-ui-pages.unit.test.ts cover page wrapper, header, messages container, bubble wrapper, bubble inner (+more) |
| 4 | Desktop e2e: .chat-bubble-inner computed width ratio 70% at 1280px | MET | New e2e spec chat-thread-page.e2e.ts asserts ratio in (0.65, 0.71] at 1280px viewport |

## Test Plan
- `admin/src/admin-ui-pages.unit.test.ts` — 635/635 pass (13 new tests for the class migration)
- `admin/e2e/chat-thread-page.e2e.ts` (new) — 3/3 pass, including the AC4 computed-width proof at 1280px
- Full `admin/` Playwright e2e suite — 28/28 pass (no regressions)
- `task lint`, `task typecheck` — clean
- Coverage on the changed file (`admin/src/admin-ui-pages.ts`): 98.52% lines / 98.05% branches
- `task ci`'s `test:coverage` step surfaces 31 pre-existing, unrelated failures (metrics/env var pollution, task-store openapi smoke fixtures, a check-test-readiness fixture) — verified these reproduce identically against a clean `origin/main` checkout and an unrelated worktree, so they predate this change

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -4485,7 +4485,13 @@ export function renderChatThreadPage(
 
   return renderAdminPage({
     title: `${rawTitle} - Shipwright Admin`,
-    extraStyles: `${threadPaneStyles}${chatPageStyles}${chatThreadStyles}
+    // NOTE: chatThreadStyles must be concatenated before chatPageStyles here —
+    // chatPageStyles's mobile @media (max-width:640px) override for
+    // .chat-bubble-inner has the same specificity as chatThreadStyles's
+    // unconditional base rule, so whichever is concatenated *last* wins the
+    // cascade. Putting chatPageStyles last ensures the mobile override
+    // actually takes effect at ≤640px instead of being silently shadowed.
+    extraStyles: `${threadPaneStyles}${chatThreadStyles}${chatPageStyles}
   `,
     body: `${renderAdminToolbar(userName, activePath)}
   <div class="vos-page chat-thread-page">
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -3874,9 +3874,45 @@ const chatPageStyles = `
       /* chat-thread-layout: flex wrapper for thread+message area; stacks to column on mobile */
       .chat-thread-layout { flex-direction:column }
       .chat-thread-sidebar { display:none }
-      .chat-bubble-inner { max-width:90% !important }
+      .chat-bubble-inner { max-width:90% }
       #message-input { font-size:16px }
     }`;
+
+// ─── Chat thread page — class-driven bubble/composer styling (CFB-1.3) ────────
+// These class names are shared verbatim between the server-rendered HTML
+// (renderMessageBubble()) and the inline JS bubble builder (addBubble()) so
+// the two renderers physically cannot drift apart — see the interpolation of
+// CHAT_BUBBLE_CLASS/CHAT_BUBBLE_INNER_CLASS into `inlineScript` below.
+export const CHAT_BUBBLE_CLASS = "chat-bubble";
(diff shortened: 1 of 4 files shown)


admin/src/admin-ui-pages.ts
```diff
@@ -3874,10 +3874,46 @@ const chatPageStyles = `
       /* chat-thread-layout: flex wrapper for thread+message area; stacks to column on mobile */
       .chat-thread-layout { flex-direction:column }
       .chat-thread-sidebar { display:none }
-      .chat-bubble-inner { max-width:90% !important }
+      .chat-bubble-inner { max-width:90% }
```
dodizzle, line 3877: This mobile-media-query rule no longer has `!important`. Meanwhile `chatThreadStyles` (line ~3904) adds a new **unconditional** `.chat-bubble-inner { max-width:70% }` rule, concatenated *after* `chatPageStyles` in `extraStyles` (line ~4488). Same specificity (0-1-0), no `!important` on either side — so per CSS cascade rules, the later unconditional rule wins even when this media query matches (≤640px). Net effect: mobile bubbles now get clamped to 70% instead of 90%, the opposite of this PR's stated goal ("so the existing mobile media queries can actually take effect"). `main` never had this conflict since the previous inline `style=` attribute always won regardless of media-query state, so this is a new regression. The added e2e test only checks a 1280px desktop viewport, so it doesn't catch this.

Suggest: reorder the concatenation so the mobile override lands after `chatThreadStyles`'s base rule (or move the override into `chatThreadStyles` itself, after the base rule), or restore `!important` here — plus a mobile-viewport assertion to lock in the fix.

dodizzle (commented):
Verdict: COMMENT — A critical CSS cascade-order regression: the new unconditional `.chat-bubble-inner { max-width:70% }` rule in `chatThreadStyles` is concatenated after `chatPageStyles`'s mobile `@media` override, and with `!important` now removed, the later unconditional rule silently wins at viewports ≤640px — clamping mobile bubbles to 70% instead of the intended 90%, the opposite of this PR's stated goal. Not caught by the new e2e test, which only asserts computed width at a 1280px desktop viewport. See inline comment.

dodizzle (commented):
Verdict: APPROVE — The previously flagged critical CSS cascade regression (mobile `.chat-bubble-inner` clamped to 70% instead of 90%) is fixed by reordering `extraStyles` concatenation so `chatPageStyles`'s mobile override lands last, and is now proven by a new 375px-viewport e2e test that would fail without the fix. No new issues found.

PR #2925, opened 2026-08-31
feat: live progress, elapsed timer and stall state in chat UI (CFB-2.3)

dmcaulay:
## Summary
- Layer 1: a 1s client-side ticker computing `now - message.createdAt` with zero network dependency — the "never go silent" guarantee, immune to server/network failure.
- Layer 2: milestone text refreshes off the existing `messages.json` poll (adaptive 2s while pending, 10s idle); hoists `renderMessageBubble` into a module-level `renderChatMessageBubble`, deletes `simpleMarkdown`, and returns `bubbleHtml` from `messages.json` so polled bubbles are byte-identical to server-rendered ones.
- Idle polling now renders every new message (not just the last reply) via an id-based `renderedIds` dedupe set, fixing agent-initiated messages, mid-run reloads, and multi-message replies in one pass; adds a skew-free stall indicator (`STALL_WARN_AFTER_MS = 120000`) that honors `prefers-reduced-motion`.
- Makes `admin/e2e/test-server.ts`'s `chatClient` genuinely injectable (`ADMIN_E2E_CHAT_CLIENT_ABSENT`) with a new `chat-degraded.e2e.ts` suite proving the degraded alert renders without it — guarding against the other chat e2e suites silently passing on a fallback page.

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Elapsed ticker updates every ~1s, zero-network guarantee | MET |
| 2 | Idle polling renders agent-initiated + all new assistant messages | MET |
| 3 | simpleMarkdown deleted; polled bubbles byte-identical | MET |
| 4 | Stall state at 120s unchanged progressSeq, prefers-reduced-motion honored | MET |
| 5 | New chat-progress.e2e.ts asserts elapsed increments + stall appearance | MET |
| 6 | test-server.ts gains injectable chatClient; degraded alert without it | MET |

## Test Plan
- `admin/e2e/chat-progress.e2e.ts` (new) — elapsed ticker increments across two DOM reads ~2s apart, ticker survives a blocked poll endpoint, stall indicator appears once `progressSeq` freezes past a (test-shortened) threshold.
- `admin/e2e/chat-degraded.e2e.ts` (new) — proves the degraded alert renders on `/admin/chat` and the thread route when `chatClient` is absent.
- `admin/src/admin-ui-pages.unit.test.ts`, `admin/src/chat.smoke.test.ts`, `admin/src/http-chat-client.unit.test.ts` — server-side bubble rendering, `?since=` filtering, stall CSS, inline JS constants.
- [x] `task typecheck` clean
- [x] `task lint` clean
- [x] Coverage on changed files: admin-ui-pages.ts 98.5%/98.3%, admin-ui.ts 92.3%/90.4%, http-chat-client.ts 96.6%/100% (all above 90/89 threshold)
- [x] `task ci` full-suite failures (31, in `metrics/`/`task-store/`) confirmed pre-existing on clean `main` — sandbox env-var leaks (`SHIPWRIGHT_ENCRYPTION_KEY`, `SHIPWRIGHT_METRICS_PUBLIC_REPO`), unrelated to this diff

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -15,6 +15,7 @@ import {
 } from "./admin-ui-styles.ts";
 import type { ManualStep } from "./agent-deletion-checklist.ts";
 import type { AgentTypeOption } from "./agent-type-manifest-loader.ts";
+import { PROGRESS_LABELS } from "@shipwright/lib/progress-phases";
 import { parseChatMarkers } from "./chat-markers.ts";
 import type {
   ChatMessage,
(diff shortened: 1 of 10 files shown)


admin/src/admin-ui-pages.ts
```diff
@@ -15,6 +15,7 @@ import {
 } from "./admin-ui-styles.ts";
 import type { ManualStep } from "./agent-deletion-checklist.ts";
 import type { AgentTypeOption } from "./agent-type-manifest-loader.ts";
+import { PROGRESS_LABELS } from "@shipwright/lib/progress-phases";
```
dodizzle, line 18: Nit: this import is sandwiched between two relative imports rather than sorted per Biome's `organizeImports` (verified with the repo's pinned `@biomejs/biome@1.9.4` — `biome check` flags it). Doesn't fail CI since `task lint` only runs `biome lint .`, not `biome check`, so this is non-blocking — just a follow-up `biome check --write` cleanup.

dodizzle (approved):
Verdict: APPROVE — Well-tested, behavior-preserving rewrite of the chat live-progress mechanism (elapsed ticker, skew-free stall detection, id-based dedupe fixing a real 'only last reply renders' bug). One cosmetic import-ordering nitpick noted inline, not CI-enforced.

PR #2930, opened 2026-08-31
feat: milestone-updated progress message in Slack threads (CFB-5.1)

dodizzle:
## Summary
- Widened the local `ClaudeRunner` type in `agent/src/slack.ts` to accept the optional third `onProgress` param — the actual runner returned by `createRunClaude()` always supported it, but the narrower local type silently discarded it, which was the sole reason progress never reached Slack.
- Added `agent/src/slack-progress.ts` (`SlackProgress`): a lazy first `chat.postMessage` (fires only once a run has been active ≥3s AND a milestone exists, so a fast reply adds no channel noise), subsequent milestones flushed via `chat.update` throttled to a 15s trailing-edge window (coalescing to the newest milestone), and a Retry-After-aware backoff on 429 that suppresses further updates until the real backoff window clears. `SlackProgress` also owns the `assistant.threads.setStatus` calls, de-duplicating the copy-pasted `setStatus` closures previously in the `message` and `app_mention` handlers. `finish()` deletes the progress message (the real reply is the artifact) — falling back to a collapsed `chat.update` if delete fails — and is a no-op beyond clearing status when no message was ever posted.
- Wired `SlackProgress` into all three event handlers (`message`, `app_mention`, `reaction_added`); every Slack API call in the new module is wrapped in `.catch(()=>{})` so a progress failure can never break the real reply.
- Fixed a race during simplify: if a run completed while the lazy `postMessage` was still in flight, `finish()` would see `messageTs` unset and skip cleanup, orphaning the message once the post landed after `finish()` had already returned — `finish()` now awaits the in-flight post promise first.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | ClaudeRunner type in slack.ts accepts onProgress; progress reaches all three handlers | MET |
| 2 | chat.update throttled to 15s trailing-edge, coalescing to the newest milestone | MET |
| 3 | 429 triggers a Retry-After-aware backoff for the remainder of the run | MET |
| 4 | A sub-3s reply posts no progress message at all | MET |
| 5 | Existing slack.integration.test.ts setStatus assertions updated to the milestone-driven sequence, not deleted | MET |

## Test Plan
- `agent/src/slack-progress.unit.test.ts` (new, 30 tests): lazy-post threshold, sub-3s no-op, trailing-edge coalescing, 429 backoff + resumption, `finish()` delete/fallback/no-op paths, in-flight-post race, and failure-swallowing — all via injected `Clock`/timer doubles, no real sleeps.
- `agent/src/slack.integration.test.ts` and `agent/src/slack-chat-conversation.integration.test.ts`: updated to assert the new `onProgress` third arg is passed to the runner in all three handlers, plus new AC #1/#4 assertions and a `reaction_added` setStatus-sequence test (that handler had none before).
- [x] `bun test` — all affected suites pass (159 + 30 + relevant chat-conversation tests)
- [x] `task typecheck` — clean
- [x] `task ci` — 31 pre-existing failures in `metrics`/`task-store` unrelated to this change (env-var/Sentry test quirks), zero new failures


agent/src/slack-progress.ts
```diff
@@ -0,0 +1,271 @@
...
+    try {
+      const result = await this.client.chat.postMessage({
+        channel: this.channel,
+        thread_ts: this.threadTs,
+        text: this.latestLabel,
+      });
+      const ts = (result as { ts?: string } | undefined)?.ts;
+      if (ts) {
+        this.messageTs = ts;
+        this.hasUnflushedUpdate = false;
```
dodizzle: A milestone that arrives while `postInitialMessage()` is still in flight (`postInFlight === true`, `messageTs` unset) only updates `latestLabel`/`hasUnflushedUpdate` and returns early in `onProgress()` — it isn't scheduled for any future flush. Since `chat.postMessage`'s `text` is captured from `this.latestLabel` synchronously at call time (before the `await` suspends), that intervening milestone never makes it into the post. Then here on success, `hasUnflushedUpdate` is unconditionally reset to `false` even though `latestLabel` may have changed again in the interim — so the internal state falsely reports nothing pending, and the posted message can display a stale phase for the rest of the run unless another `onProgress` call happens to arrive later.

Suggestion: capture the label actually sent to `chat.postMessage` before the `await`, and only clear `hasUnflushedUpdate` here if `this.latestLabel` still matches it — otherwise leave it `true` and call `scheduleThrottledUpdate()` so the newer milestone flushes promptly.

dodizzle (commented):
Verdict: COMMENT — a milestone that arrives while the lazy `chat.postMessage` is still in flight can be silently dropped forever (see inline comment on `slack-progress.ts`); everything else in this PR — the `ClaudeRunner` type widening, handler wiring, 429 backoff, and test isolation — checks out.

dodizzle (commented):
Verdict: APPROVE — the lazy-post race flagged in the prior round (a milestone arriving while chat.postMessage was still in flight could be silently dropped) is fixed correctly in agent/src/slack-progress.ts's postInitialMessage(): sentLabel is now captured before the await and hasUnflushedUpdate only clears when latestLabel still matches it, pinned down by a dedicated unit test. Nothing else surfaced — handler wiring, 429 backoff, and test isolation all check out.

PR #2936, opened 2026-08-31
feat: cancel an in-flight reply via the heartbeat tick (CFB-2.4)

dmcaulay:
## Summary
- Makes the existing 3s agent↔server heartbeat tick bidirectional: `heartbeat()` now returns the message so the client can derive `HeartbeatResult{cancelRequested}`, keeping the architecture pull-only (no second polling loop, no inbound HTTP surface on the agent)
- Adds `POST /threads/:threadId/messages/:id/cancel` (registered before `/:id`) to stamp `cancelRequestedAt`; agent-side abort reuses the existing kill path as a third trigger alongside the two timeout timers, with `ClaudeAbortedError` checked before timeout/exit-code classification and excluded from retry
- Fixes `processThread`'s catch swallowing failures (message stayed claimed forever with no reaper) and a `streamIncomplete` run posting an empty reply — the poller now always replies with a non-empty body and an `errorKind`
- Converts `MessageService.reply()` from `Promise.all` to a real `prisma.$transaction` so a partial failure can no longer leave `repliedAt` set with no assistant message
- Renders the new `cancelled`/`incomplete`/`stalled` error kinds in the admin chat UI with a Retry action that resends the originating message

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Cancel aborts the in-flight run within ~3s via the heartbeat response | MET |
| 2 | Abort listener removed after every run — verified by a counted addEventListener double | MET |
| 3 | Abort wins over a racing timeout; aborted runs are not retried | MET |
| 4 | The poller ALWAYS replies: errorKind on runner failure, and a non-empty body on streamIncomplete | MET |
| 5 | reply() is a real prisma.$transaction | MET |
| 6 | New errorKinds cancelled/incomplete/stalled render in the UI with a Retry action | MET |

## Test Plan
- `chat-poller.unit.test.ts` — heartbeat-triggered abort, `deriveReply()` covering cancelled/thrown/streamIncomplete/empty-clean-finish paths
- `claude.unit.test.ts` — counted-listener test proving no leak on clean finish or across resume-retry; abort-vs-timeout race classification; no retry on `ClaudeAbortedError`
- `message-service.unit.test.ts` / `.integration.test.ts` — `$transaction` call sequence and atomic partial-failure behavior; `requestCancel`
- `admin-ui-pages.unit.test.ts` — error badge + Retry button rendering for all three new error kinds
- `messages.smoke.test.ts` — new `/cancel` route wiring


dodizzle (approved):
Verdict: APPROVE — Abort/timeout/exit-code race is correctly ordered and well-tested, the reply-path fixes (always-reply, transactional reply()) close real bugs. One low-confidence suggestion noted in the review body: the new cancel endpoint has no UI trigger yet.

PR #2940, opened 2026-08-31
feat: reap stalled chat claims after 300s of silence

dodizzle:
## Summary
- Add `StallReaper` (`chat/src/stall-reaper.ts`), a background sweep that terminalizes claimed, unreplied user messages with `errorKind: "stalled"` once they've gone silent past `CHAT_STALLED_AFTER_MS` (default 300s / 5min) — protects against a dead agent wedging a message forever.
- Falls back to `claimedAt` when `heartbeatAt` is null (agent died before its first beat). Reuses the existing `MessageService.reply()` (not new write logic) — its idempotent "returns null if already replied" contract makes the sweep race-safe against a resumed agent and safe across multiple replicas. The reaper only marks stalled; it never unclaims for retry, since silently re-running a prompt whose side effects may have already landed (files edited, a PR opened, a deploy triggered) is unacceptable.
- Wired into `chat/src/main.ts` via `setInterval` (60s cadence) — never inside `app.ts`, which stays side-effect-free for smoke tests. No new Prisma index (would break HOT updates on the frequent `heartbeatAt` writes).
- Documented `CHAT_STALLED_AFTER_MS` in both `docs/chat.md` and `docs/configuration.md`.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Stalled claims terminalize with errorKind 'stalled' after 300s of silence | MET |
| 2 | Sweep is idempotent and does not clobber an agent that resumes | MET |
| 3 | Falls back to claimedAt when heartbeatAt is null | MET |
| 4 | One bad row does not abort the sweep | MET |
| 5 | Env-documented in docs/chat.md so check-config-docs passes | MET |

## Test Plan
- `chat/src/stall-reaper.unit.test.ts` — 10 tests covering heartbeat-based reap, claimedAt fallback, idempotency (reply() returning null), one-bad-row-doesn't-abort-sweep, custom `stalledAfterMs` cutoff, and exact WHERE-clause shape. 100% line/branch coverage on `stall-reaper.ts`.
- `task typecheck` — all 7 workspaces pass.
- `bun scripts/check-config-docs.ts` — passes (40 env vars documented).
- `bun plugins/shipwright/scripts/check-banned-strings.ts` — clean.
- `bun run scripts/sync-version.ts --check` — passes.


dodizzle (commented):
Verdict: APPROVE — Clean, well-tested implementation that mirrors the existing task-store stale-claim-reaper pattern; no blocking findings. Posted as COMMENT since GitHub blocks self-APPROVE.

PR #2964, opened 2026-09-01
feat: web push notifications for agent replies (CFB-4.2)

dodizzle:
## Summary
- Adds Web Push notifications for agent replies: a new `PushSubscription` model (per-device) and `ChatThreadWatch` model (targets exactly the users watching a thread, avoiding a fan-out to the whole agent's membership).
- Hand-rolls RFC 8291 (aes128gcm encryption) and RFC 8292 (VAPID) crypto on Bun's `node:crypto` — a ~20-line spike confirmed `web-push` works under Bun but the dependency was heavier than warranted, so correctness is instead pinned to the RFC 8291 Appendix A test vector.
- Adds a content policy (`push-content.ts`) that defaults to a generic "Your agent replied" notification, with `title`/`preview` opt-in levels capped by an operator-set `SHIPWRIGHT_ADMIN_PUSH_MAX_DETAIL` ceiling — and scrubs id/repo/path/cost shapes from any user-authored text before it can reach a locked screen.
- Extends the CFB-4.1 PWA service worker with `push`/`notificationclick` listeners, and adds a client-side toggle that requests permission only inside a click handler, never re-prompts once denied, and gives iOS users install guidance instead of a false "unsupported" error.
- Push is gated end-to-end on `SHIPWRIGHT_ADMIN_VAPID_{PUBLIC_KEY,PRIVATE_KEY,SUBJECT}` all being set — otherwise the service is `undefined`, the routes 503, and the toggle renders nothing, so the page is byte-for-byte the existing CFB-3.2 chat page.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Dedicated unit test asserts no payload at any detail level contains an agentId, threadId, repo name or path | MET |
| 2 | web-push under Bun validated by a ~20 line spike BEFORE committing to the dependency; fallback is hand-rolled RFC 8291/8292 against the spec test vectors | MET |
| 3 | PushSender takes fetchImpl as a constructor argument — no global.fetch override | MET |
| 4 | Subscriptions pruned immediately on 404/410 | MET |
| 5 | Toggle requests permission only inside a click handler; denied state shows guidance and never re-prompts | MET |
| 6 | With push unavailable, denied, or disabled server-side the page degrades to exactly the CFB-3.2 page | MET |

## Test Plan
- `push-crypto.unit.test.ts` — RFC 8291 Appendix A test vector reproduced exactly.
- `push-content.unit.test.ts` — detail-level resolution + the AC-1 leak test across all three levels.
- `push-sender.unit.test.ts` — injected `fetchImpl` (no global override), 404/410 pruning, transient-failure non-pruning, `sendToMany` isolation.
- `push-service.unit.test.ts` — watcher→subscription fan-out, no-watcher no-op, ceiling capping.
- `push-toggle.unit.test.ts` — click-only permission request, denied-never-reprompt, iOS install education, server-side degrade gate.
- `pwa.unit.test.ts` — new `push`/`notificationclick` service-worker listener assertions.
- New admin modules (`push-content.ts`, `push-crypto.ts`, `push-sender.ts`, `push-service.ts`, `push-toggle.ts`) all at 100% line/function coverage.
- [x] `task lint`, `task typecheck`, `task check-strings`, `task check-config-docs`, `task check-version-sync` all pass.
- [x] `task test:coverage` has 31 pre-existing failures (metrics env-leak tests, task-store smoke test) confirmed present on `main` as well — unrelated to this change.
- `task helm:check` couldn't run in this sandbox (`helm-unittest` plugin not installed) — this PR makes no chart changes, so it's not applicable here.


dmcaulay (commented):
Verdict: COMMENT — Solid feature (crypto is validated against the RFC test vector, injected fetchImpl avoids global overrides, content-scrubbing is conservative by design), but the new webhook's deploy docs contain two factual errors about its actual runtime behavior (503 vs. 401 on unset token; wrong response field names), and none of the three new Hono routes have smoke-test coverage despite an existing convention/file for this app.

docs/deploy-kubernetes.md
```diff
@@ -977,6 +977,91 @@ These map to the agent voice env vars (`WHISPER_SERVICE_URL`,
...
+        secretKeyRef:
+          name: shipwright-secrets
+          key: shipwright-admin-push-webhook-token
+```
+
+**Graceful degradation:** when any VAPID config is missing, the `POST /admin/push/notify`
+webhook returns `503 Service Unavailable` (safe for the calling chat service — it
+retries), and the chat page's notification toggle does not render (users see the
+poll-only experience instead). When `SHIPWRIGHT_ADMIN_PUSH_WEBHOOK_TOKEN` is unset,
+the webhook returns `401 Unauthorized` and the chat service's notification trigger
```
dmcaulay: This says the webhook returns `401 Unauthorized` when `SHIPWRIGHT_ADMIN_PUSH_WEBHOOK_TOKEN` is unset, but `admin/src/admin-ui.ts`'s `POST /admin/push/notify` handler checks `!pushEnabled || !pushService || !pushWebhookToken` first and returns `503` in that case — the 401 branch only fires when the token *is* configured but the presented value doesn't match. This will mislead an operator debugging an unexpected 503.

docs/agent.md
```diff
@@ -107,13 +107,15 @@ The admin console is an installable Progressive Web App (PWA). All authenticated
...
+| POST | `/admin/push/notify` | bearer token (`SHIPWRIGHT_ADMIN_PUSH_WEBHOOK_TOKEN`) | Web Push notification trigger — called by the chat service when an agent posts a reply. Request body: `{threadId, agentId, title?: string \| null, preview?: string \| null}`. Response: `{ok: true, deliveryCount, failureCount}`. Returns `503` if Web Push is disabled (VAPID keys not configured) or push service is unavailable. Returns `401` if the bearer token does not match. Returns `400` if body is malformed or missing required fields. Notification content policy is governed by `SHIPWRIGHT_ADMIN_PUSH_MAX_DETAIL` (operator ceiling) and the user's subscription-level opt-in preference (effective level = min of both). |
```
dmcaulay: The documented response shape `{ok: true, deliveryCount, failureCount}` doesn't match the implementation. `push-service.ts`'s `notifyThreadReply()` returns `{delivered, pruned}`, and `admin-ui.ts` spreads that directly into the response — so the actual shape is `{ok: true, delivered, pruned}`. Neither `deliveryCount` nor `failureCount` appears anywhere in the code.

admin/src/admin-ui.ts
```diff
@@ -3432,13 +3505,110 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
...
+      console.error("[push] unsubscribe delete failed:", err);
+    }
+    return c.json({ ok: true });
+  });
+
+  // Inbound webhook the chat service calls when an agent posts a reply, so the
+  // watching user(s) get notified. Authenticated by a shared bearer token
+  // (SHIPWRIGHT_ADMIN_PUSH_WEBHOOK_TOKEN) — NOT a user session — since the
+  // caller is a service, not a browser.
+  app.post("/admin/push/notify", async (c) => {
```
dmcaulay, line 3580: None of the three new routes added in this PR (`push/subscribe`, `push/unsubscribe`, `push/notify`) have smoke-test coverage, even though `admin-ui.smoke.test.ts` already exists and covers the rest of this Hono app per CLAUDE.md's test-conventions table. The 503-vs-401 doc mismatch flagged above is exactly the kind of route-contract issue a smoke test (token unset/mismatched/matching) would have caught.

dmcaulay (approved):
Verdict: APPROVE — Both prior documentation-accuracy findings (503-vs-401 status code, response shape) are verifiably fixed, smoke-test coverage now exists for all three new push routes with meaningful status-code assertions, and no new findings clear the confidence threshold.

PR #2981, opened 2026-09-01
feat: optional outbound reply-notification webhook from chat service (CFB-4.3)

dodizzle:
## Summary
- Adds an env-gated outbound reply-notification webhook from the chat service, following the same pattern as the existing scope resolver (`chat/src/auth.ts`'s `createScopeResolver`) but with its own dedicated credential rather than reusing `SHIPWRIGHT_CHAT_AGENTS_API_KEY`
- Fires fire-and-forget (5s timeout, never throws) after a reply persists in `POST /:id/reply`, calling the already-shipped admin `POST /admin/push/notify` webhook (from CFB-4.2) with `{threadId, agentId, title}` only — no message body/preview text crosses the wire
- Adds the shared `ReplyNotificationEvent` type + `parseReplyNotificationEvent()` validator in `lib/chat-notify.ts`, importable by both services

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Notifier fires exactly once per persisted reply, never per progress tick | MET |
| 2 | Reply still returns 201 when the injected notifier throws | MET |
| 3 | Inert with no env set; a console log states on/off like the scope resolver | MET |
| 4 | Payload contains no message body text | MET |
| 5 | Shared ReplyNotificationEvent type plus a parse validator in lib/ | MET |

## Test Plan
- `chat/src/reply-notifier.unit.test.ts` — payload shape, null-title passthrough, 5s timeout, never-throws on network failure / non-2xx
- `chat/src/messages.smoke.test.ts` — fires exactly once on a persisted reply; payload uses the thread's `agentId` (not the caller's); 201 still returned when the notifier throws; 201 when no notifier configured; confirms no firing on heartbeat/claim/create/duplicate-reply
- `lib/chat-notify.unit.test.ts` — validator accepts well-formed events, rejects malformed shapes


dmcaulay (commented):
Verdict: COMMENT — Solid implementation of the reply-notification webhook (fire-and-forget, text-free payload, well-isolated tests), but `SHIPWRIGHT_CHAT_PUSH_WEBHOOK_TOKEN` is documented as a secret in `docs/configuration.md` without being added to `lib/secret-env-vars.ts`'s `SECRET_ENV_VARS`, so it won't be masked in the admin envs UI/API or scrubbed from Sentry logs. Please add it there (and update the exact-count test) before merging.

dmcaulay (approved):
Verdict: APPROVE — The prior finding (SHIPWRIGHT_CHAT_PUSH_WEBHOOK_TOKEN missing from SECRET_ENV_VARS) is fully resolved in lib/secret-env-vars.ts, with the exact-count test updated accordingly. All 5 acceptance criteria are met, and no new issues surfaced in this pass.

PR #2983, opened 2026-09-01
feat: emit TaskEvent rows from StaleClaimReaper (TCS-1.3)

dmcaulay:
## Summary
- `StaleClaimReaper.reap()`'s Task branch now writes a `TaskEvent` audit row per changed field per reaped task (`method: "reap"`, `actor: null`), so a reap shows up in the same audit trail as claim/release/complete transitions.
- Switched the Task branch from a bulk `$executeRaw` UPDATE to a `$transaction` that pre-selects candidate stale rows (captures "before"), runs `UPDATE ... RETURNING` (captures "after"), then diffs each pair and writes events — race-safe since pre-select and update run in one transaction.
- Extracted `writeTaskEvents()` in `task-transition-diff.ts` as a shared diff+write helper, and refactored `TaskService.recordTaskTransition()` to delegate to it (behavior-identical) so the reaper and the per-row service mutations can't drift.
- The PullRequest branch of `reap()` is unchanged — out of scope for this task.

## Acceptance Criteria

| 1 | Reaped tasks each produce TaskEvent rows showing the in_progress -> pending / claim-fields-nulled transition | MET | `writeTaskEvents` called per reaped row inside the reap transaction; integration test asserts status/claimedBy/claimedAt/startedAt event rows with correct old/new values, `method: "reap"`, `actor: null`. |
| 2 | Integration test: a stale in_progress task reaped by StaleClaimReaper produces the expected event rows | MET | New `describeOrSkip("StaleClaimReaper Task reaping (integration)")` block seeds a stale + fresh task, runs `reap()` against real Postgres, asserts exact TaskEvent rows and a shared `at` timestamp. |
| 3 | Existing stale-claim-reaper tests updated only where the query-shape change requires it — no coverage removed | MET | Unit test double updated to model the new `$transaction`/`$queryRaw`/`taskEvent.create` shape; every pre-existing assertion preserved, 3 new unit tests added for TaskEvent generation. |

## Test Plan
- `bun test task-store/src/stale-claim-reaper.unit.test.ts task-store/src/task-transition-diff.unit.test.ts` — 38/38 pass
- New integration test verified against a real local Postgres instance (not skipped) — red before the fix, green after
- `task typecheck` — clean across all 7 workspaces
- `task lint` — clean
- `stale-claim-reaper.ts` and `task-transition-diff.ts` at 100% line/function coverage

### DIFF
--- a/task-store/src/stale-claim-reaper.integration.test.ts
+++ b/task-store/src/stale-claim-reaper.integration.test.ts
@@ -28,6 +28,137 @@ function makePrisma(): PrismaClient {
 // test tracks any future bump to DEFAULT_CLAUDE_TIMEOUT_MS automatically.
 const DEFAULT_TTL_MS = DEFAULT_CLAIM_TTL_MS;
 
+describeOrSkip("StaleClaimReaper Task reaping (integration)", () => {
+  const NOW = new Date("2026-09-01T12:00:00.000Z");
+  const STALE = new Date(
+    NOW.getTime() - DEFAULT_TTL_MS - 5 * 60_000,
+  ).toISOString();
+  const FRESH = new Date(NOW.getTime() - 60_000).toISOString();
+
+  let prisma: PrismaClient;
+
+  beforeEach(async () => {
+    prisma = makePrisma();
+    // TaskEvent's FK is ON DELETE RESTRICT (TCS-1.1) — clear event rows
+    // before their parent Task rows, in case another test file sharing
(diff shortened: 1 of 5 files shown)


dodizzle (approved):
Verdict: APPROVE — Race-safe transaction (pre-select + UPDATE...RETURNING share the same predicate), writeTaskEvents() extraction genuinely removes drift risk between the reaper and TaskService's per-row mutations, and tests follow this repo's isolation conventions with no coverage removed. No blocking issues.

PR #2984, opened 2026-09-01
fix: derive metrics dashboard feature-grouping from tasks, not pr.taskId (PTL-2.2)

dodizzle:
## Summary
- `groupPrsByPrefix()` now originates from `tasks` (task.id-derived prefix + task.pr) instead of reading the unreliable `pr.taskId` field, which is populated only ~10% of the time (0% in several repos) — this dashboard grouping was silently incomplete for the large majority of PRs.
- Each task's own `.pr` is matched against the fetched PR list by `(repo, prNumber)`, so a PR shared by multiple tasks (bundle case) now correctly appears under each of those tasks' feature buckets instead of only one.
- `TaskRecord`/`PrRecord` gained the `pr`/`prNumber` fields needed to do that matching.

## Acceptance Criteria

| 1 | groupPrsByPrefix derives groupings from task records, not pr.taskId | MET | `groupPrsByPrefix(tasks, prs)` loops over tasks, derives prefix from `t.id`, matches PR via `(t.repo, t.pr)` |
| 2 | PR with taskId null still groups via matching task.pr | MET | Integration + unit tests set `taskId: null` with matching `task.pr` and assert correct grouping |
| 3 | Bundle PR (shared by 2 tasks) appears under each task's bucket | MET | Unit test asserts a shared PR appears under both feature buckets |
| 4 | Unit tests added for both cases; stale pr.taskId-only fixtures retired | MET | New `task-store-provider.unit.test.ts`; integration fixtures changed from `taskId: "QS-1.1"` to `taskId: null` + `prNumber` |
| 5 | Safe to deploy standalone (no schema change) | MET | Only dashboard grouping logic + test fixtures changed |

## Test Plan
- `bun test src/providers/task-store-provider.unit.test.ts src/providers/task-store-provider.integration.test.ts` — 28 pass, 0 fail
- `tsc --noEmit` — clean

### DIFF
--- a/metrics/src/providers/task-store-provider.ts
+++ b/metrics/src/providers/task-store-provider.ts
@@ -531,7 +531,17 @@ export class TaskStoreProvider implements MetricsProvider {
       const prefix = featurePrefix(t.id);
       if (!prefix) continue;
       if (t.pr == null) continue;
-      const pr = prs.find((p) => p.repo === t.repo && p.prNumber === t.pr);
+      // `Task.repo` is nullable (a legal state for unscoped tasks; see
+      // docs/task-store.md), whereas `PullRequest.repo` is always a non-null
+      // string. A strict `p.repo === t.repo` therefore silently drops every
+      // null-repo task's PR — the same silent-drop class this method exists to
+      // fix. Treat a null/undefined `t.repo` as unscoped and match on prNumber
+      // alone; otherwise require the repo to match so two PRs sharing a number
+      // across repos can't be confused.
+      const pr = prs.find(
+        (p) =>
+          (t.repo == null || p.repo === t.repo) && p.prNumber === t.pr,
+      );
       if (!pr) continue;
       const arr = map.get(prefix) ?? [];
       arr.push(pr);
--- a/metrics/src/providers/task-store-provider.ts
+++ b/metrics/src/providers/task-store-provider.ts
@@ -512,15 +512,30 @@ export class TaskStoreProvider implements MetricsProvider {
     return map;
   }
 
-  private groupPrsByPrefix(prs: PrRecord[]): Map<string, PrRecord[]> {
+  /**
+   * Groups PR records by feature prefix, originating from `tasks` rather than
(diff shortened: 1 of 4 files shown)


metrics/src/providers/task-store-provider.ts
```diff
@@ -512,15 +512,30 @@ export class TaskStoreProvider implements MetricsProvider {
...
-    for (const pr of prs) {
-      const id = pr.taskId ?? "";
-      const p = featurePrefix(id);
-      if (!p) continue;
-      const arr = map.get(p) ?? [];
+    for (const t of tasks) {
+      const prefix = featurePrefix(t.id);
+      if (!prefix) continue;
+      if (t.pr == null) continue;
+      const pr = prs.find((p) => p.repo === t.repo && p.prNumber === t.pr);
```
dmcaulay: `p.repo === t.repo` compares `PullRequest.repo` (schema-required, always a non-null string) against `Task.repo` (schema-optional/nullable — `docs/task-store.md:127,568` document `repo: null` as valid for tasks "not scoped to a specific repository"). For any task with `repo: null`, this comparison is always `someString === null` → `false`, so that task's own PR is never matched even when `t.pr`/`p.prNumber` line up — silently dropping exactly the class of PRs this PR is meant to stop dropping. No fixture (unit or integration) covers a null-repo task with a matching PR.

Suggest treating a null/undefined `task.repo` as unscoped and matching on `prNumber` alone in that case.

dmcaulay (commented):
Verdict: COMMENT — Solid fix for the pr.taskId sparsity bug and bundle-PR handling, but the new (repo, prNumber) match silently drops tasks with repo: null (a documented, legal Task state) since PullRequest.repo is always non-null — reintroducing a silent-drop gap of the same shape this PR fixes. See inline comment.

dmcaulay (approved):
Verdict: APPROVE — The prior null-repo matching bug is fixed (`t.repo == null || p.repo === t.repo`) and directly covered by a new unit test (`task-store-provider.unit.test.ts:149-181`). No new issues found.

PR #2997, opened 2026-09-01
feat: migrate SlackProgress to agents.sessions.setStatus (SAV-1.1)

dmcaulay:
## Summary
- Migrates `agent/src/slack-progress.ts` off the deprecated `assistant.threads.setStatus` onto `agents.sessions.setStatus` (Slack's Agent Sessions API, rolled out 2026-08-20)
- `start()` now sets status `"processing"`, `finish()` sets `"active"` (idle/ready — threads stay open for follow-ups, so `"closed"` is never used here); `onProgress()` is now a documented no-op since the new API has no per-phase text slot
- Removes `labelForPhase()`, `DEFAULT_STATUS_LABEL`, and the `PROGRESS_LABELS` import from this file (unused now); `lib/progress-phases.ts` itself is untouched and its exports remain used elsewhere
- Bumps `agent/package.json`'s `@slack/web-api` from `^7.16.0` to `^8.1.0`; `@slack/bolt` stays `^4.7.3` unchanged (it constructs its own nested `@slack/web-api` dependency, no coupling)
- Updates all Slack mock client fixtures across the test suite (`slack-progress.unit.test.ts`, `slack.integration.test.ts`, `slack-chat-conversation.integration.test.ts`, `slack.unit.test.ts`) to the new `agents.sessions.setStatus` shape

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | start() calls agents.sessions.setStatus with status: "processing" | MET |
| 2 | finish() calls agents.sessions.setStatus with status: "active" | MET |
| 3 | onProgress no-ops; labelForPhase/DEFAULT_STATUS_LABEL/PROGRESS_LABELS import removed | MET |
| 4 | lib/progress-phases.ts unmodified, exports still consumed elsewhere | MET |
| 5 | @slack/web-api ^8.1.0+, @slack/bolt unchanged, bun.lock updated | MET |
| 6 | slack-progress.unit.test.ts rewritten for new call shape/lifecycle values | MET |

## Test Plan
- `bun test agent/src/slack-progress.unit.test.ts agent/src/slack.integration.test.ts agent/src/slack-chat-conversation.integration.test.ts agent/src/slack.unit.test.ts` — 193/193 pass, `slack-progress.ts` at 100% line/function coverage
- `task typecheck` clean across all 7 workspaces
- `task lint` clean (676 files)
[...]

PR #3003, opened 2026-09-01
fix: catch synchronous throw when Slack client lacks agents namespace

dmcaulay:
## Problem

Three Slack-enabled Shipwright agents in production (vitals-os namespace) were repeatedly logging:

```
TypeError: undefined is not an object (evaluating 'this.client.agents.sessions')
  at setStatus (agent/src/slack-progress.ts:63)
An unhandled error occurred while Bolt processed (type: event_callback, ...)
```

`SlackProgress.setStatus()` (#2997) calls `this.client.agents.sessions.setStatus(...)`, Slack's new Agent Sessions API. The `client` passed into Bolt event handlers is built from Bolt's own bundled `@slack/web-api` dependency (`^7.16.0`, resolved to `7.19.0`), which predates that API — it has no `agents` namespace at all.

The bug: only the `setStatus()` *promise* was wrapped in `.catch(() => {})`. Accessing `this.client.agents.sessions` throws **synchronously**, before `.setStatus()` is even called and before `.catch()` ever attaches to anything. That throw becomes an unhandled rejection in the surrounding `async` function, propagating up through `start()`/`finish()` into Bolt's event pipeline — visibly breaking Slack thread status updates and spamming error logs on every Slack-triggered agent run.

## Fix

- Wrap the whole call — property access included — in `try/catch`, matching the class's stated invariant that a progress failure must never throw or break the real reply.
- Log the swallowed error via `console.warn` (previously it vanished silently, even on the pre-existing rejection path) — matches the swallowed-error convention already used by `mapReposTolerant` in `check-helpers.ts`. A genuine `setStatus` failure (rate limit, bad token, etc.) now stays visible instead of disappearing without a trace.

## Testing

- Added regression tests: a client with no `agents` namespace at all (reproduces the exact production failure) and a rejecting `setStatus` call, both asserting the warning is logged with the expected message.
- `bun test agent/src/slack-progress.unit.test.ts` — 9 pass.
- `bunx biome check` — clean on both touched files.
- `tsc --noEmit` in `agent/` shows pre-existing, unrelated errors (stale generated Prisma client / missing `playwright`/`yaml` deps) present identically on `main` before this change — confirmed via `git stash`.

https://claude.ai/code/session_018pNRtq4qTN2Zz916zJwfQN

### DIFF
--- a/agent/src/slack-progress.ts
+++ b/agent/src/slack-progress.ts
@@ -18,12 +18,14 @@
  * stay open for follow-up messages — "closed" is for ending a conversation
  * entirely.
  *
- * Every Slack API call is wrapped in .catch(() => {}) — a progress failure
- * must never throw or break the real reply.
+ * Every Slack API call is wrapped in try/catch — a progress failure must
+ * never throw or break the real reply. This covers a client that lacks the
+ * `agents` namespace entirely (property access throws synchronously), not
+ * just a setStatus() call that rejects.
  */
 
-import type { ModelUsage } from "./claude.ts";
 import type { ProgressPhase } from "@shipwright/lib/progress-phases";
+import type { ModelUsage } from "./claude.ts";
 
 // biome-ignore lint/suspicious/noExplicitAny: Bolt client type is complex
 type SlackClient = any;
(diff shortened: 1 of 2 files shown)


dodizzle (approved):
Verdict: APPROVE — Correctly wraps the `agents.sessions` property access (not just the `setStatus()` promise) in try/catch, closing the synchronous-throw gap that caused unhandled rejections in production. Regression tests reproduce the exact failure (client with no `agents` namespace), and the `console.warn` logging/test-isolation approach matches existing repo conventions.

PR #3006, opened 2026-09-01
feat: upgrade @slack/bolt to 5.x, fix broken agent-status updates

dodizzle:
## Summary
- Bump `@slack/bolt` from `^4.7.3` to `^5.0.0` and regenerate `bun.lock`, so Bolt's bundled `@slack/web-api@^8.0.0` dedupes against the top-level `^8.1.0` dependency instead of installing a second, version-skewed copy internally
- Fixes the root cause behind #2997/#3003: `SlackProgress.setStatus`'s call to `this.client.agents.sessions.setStatus(...)` has been silently failing (warn-logged, never crashing) since #2997 merged, because Bolt's internal client was built from a pre-8.0 `@slack/web-api` that predates the Agent Sessions API
- Add `agent/src/slack-bolt-client.integration.test.ts` — a real (unmocked) integration test that constructs an actual `@slack/bolt` App and asserts `app.client.agents.sessions.setStatus` is a function, closing the coverage gap that let the original regression ship (neither `slack-progress.unit.test.ts`'s mocked client nor `slack.integration.test.ts`'s `MockApp` exercise the real resolved dependency tree)

## Acceptance Criteria
| # | Criterion | Status | Evidence |
|---|-----------|--------|----------|
| 1 | `@slack/bolt` bumped to `^5.0.0`, `bun.lock` deduped to single `@slack/web-api@8.x` | MET | `grep '@slack/web-api@' bun.lock` returns exactly 1 line (8.1.1); nested pre-8.0 copies under bolt/oauth/socket-mode removed |
| 2 | New integration test asserts real App has `agents.sessions.setStatus` | MET | `agent/src/slack-bolt-client.integration.test.ts` — real App, dummy creds, construction only (no `.start()`), confirmed RED pre-upgrade / GREEN post-upgrade |
| 3 | No changes to `slack-progress.ts`/unit test/MockApp tests | MET | Diff touches only `agent/package.json`, `bun.lock`, and the new test file |
| 4 | `task ci` passes clean | MET | lint + typecheck pass clean; unrelated sandbox failures (env-var leakage + a pre-existing `task-store/src/routes/prs.ts` bug) reproduce byte-for-byte on pre-upgrade `main` — verified in a scratch worktree — and GitHub Actions CI on `main`'s tip is green |

## Test Plan
[...]

PR #3008, opened 2026-09-01
feat: add spawner injection seam to cron-handler preCheck and de-flake its tests

dodizzle:
## Summary
- Add an optional `spawner` field to `CronHandlerDeps` (mirroring `claude.ts`'s existing `createRunClaude(spawner, ...)` pattern) so the preCheck runner uses `deps.spawner ?? Bun.spawn` instead of calling `Bun.spawn` directly — production behavior is unchanged.
- Move the 15 preCheck-behavior tests out of `cron-handler.smoke.test.ts`'s `"handleCronRequest — preCheck"` describe block into a new `agent/src/cron-handler.unit.test.ts`, rewritten to inject a fake spawner (canned stdout/stderr/exit code) instead of spawning ~15 real bun child processes per test run — this was the confirmed root cause of PR #3006 (SSB-1.1)'s intermittent CI failures under resource contention.
- Keep exactly one true end-to-end test (in `cron-handler.integration.test.ts`) that omits the injected spawner and exercises the real `deps.spawner ?? Bun.spawn` default-fallback wiring with an actual subprocess.

## Acceptance Criteria
| 1 | `CronHandlerDeps` has an optional `spawner` field; preCheck uses it, falling back to real `Bun.spawn`; no other consumer (`index.ts`, `health.ts`) needs changes | MET | `spawner?: typeof Bun.spawn` added, defaulted at destructure, used at the preCheck call site. Zero diff in `index.ts`/`health.ts`. |
| 2 | ~15 preCheck tests moved to `cron-handler.unit.test.ts` w/ fake spawner; deleted from smoke file (no duplicates); exactly one real-subprocess e2e test in `cron-handler.integration.test.ts` | MET | New 604-line unit test file (15 tests, fake spawner mirrors `claude.unit.test.ts`'s pattern). Smoke file's preCheck describe block fully removed (436 lines). One new default-wiring e2e test added to the integration file. |
| 3 | task ci passes clean | MET (scoped) | Lint clean, typecheck clean across all 7 workspaces, all 87 cron-handler tests pass (down from real-subprocess spawning to in-process fakes). The full monorepo `test:coverage` aggregate gate has 30 pre-existing failures/1 error — verified identical on the `main` baseline (`eda95b45`) via a side-by-side run — unrelated to this change (a pre-existing `task-store/src/routes/prs.ts` `validateRepo` null-vs-undefined bug). |

## Test Plan
- `NODE_ENV=test bun test agent/src/cron-handler.unit.test.ts agent/src/cron-handler.smoke.test.ts agent/src/cron-handler.integration.test.ts` — 87 pass, 0 fail (previously spawned ~15 real bun subprocesses; now runs in-process against fake spawner doubles).
- `bunx biome lint` on all four changed source/test files — clean.
- `bun run --filter='*' --sequential typecheck` — clean across all workspaces.


dmcaulay (approved):
Verdict: APPROVE — Clean, minimal DI seam (spawner) mirroring the existing claude.ts pattern, correct test-layer split, no production behavior change. Also fixes the actual root cause of the random CI failures across this PR's history: Bolt 5.x's App constructor makes a real, unawaited auth.test() network call that rejects after its synchronous test finishes, corrupting Bun's shared test-runner state for whatever file happens to be running at that moment — tokenVerificationEnabled: false is the right minimal fix. Verified 5 consecutive clean local full-suite runs post-fix and CI is now green.

PR #3017, opened 2026-09-02
fix(agent): patch @slack/socket-mode's undici ping for Bun compat

dmcaulay:
## Summary

Fixes a live, fleet-wide incident: since `@slack/bolt` was bumped to `5.0.0` (#3006), every Shipwright agent's Slack socket-mode connection has been stuck in a continuous ~5-7s reconnect loop in production, not an intermittent Slack-side issue.

## Root cause

Bolt `5.0.0` bundles `@slack/socket-mode@3.x`, whose ping/pong heartbeat (`SlackWebSocket`'s `monitorPingToSlack` / `monitorPingFromSlack`) depends on:
- the real `undici` package's module-level `ping()` helper, and
- its `undici:websocket:ping` / `undici:websocket:pong` `node:diagnostics_channel` events.

Bun always intercepts the bare `"undici"` specifier with its own built-in polyfill — even when the real package is installed right alongside it — and that polyfill implements neither API. Every client-side ping tick threw `undici_1.ping is not a function` and immediately called `disconnect()`.

This is a confirmed, currently-open Bun bug: [oven-sh/bun#37110](https://github.com/oven-sh/bun/issues/37110), root-caused and reproduced by Bun's own team, with a fix in progress at [PR #37111](https://github.com/oven-sh/bun/pull/37111) (unmerged). It is **not** a bug in `@slack/socket-mode` — its code is correct under real Node.js + real `undici`. It's also not something a Bun version bump fixes today (long-standing, unscheduled gap — see oven-sh/bun#17799, #19748).

## Fix

Patches `@slack/socket-mode@3.0.1` (via `bun patch`, tracked in `patchedDependencies`) so both of its `require("undici")` call sites (`SlackWebSocket.js`, `SocketModeClient.js`) use a subpath require — `require("undici/index.js")` — instead of the bare specifier. Bun's interception matches only the exact string `"undici"`; the subpath falls through to normal filesystem resolution and finds the real, complete `undici` package that already satisfies this file's own `peerDependencies: { undici: "^7.0.0" }`, sitting right there as a sibling in its own resolved `node_modules`. No new dependency needed — an earlier alias-based approach (`"undici-real": "npm:undici@..."`) was considered but turned out to be unnecessary and didn't resolve cleanly under this repo's Bun install layout anyway.

## Verification

- Reproduced the exact production failure locally against a real WebSocket server, both **before** (unpatched: throws `undici_1.ping is not a function`, client disconnects) and **after** (patched: clean, both ping directions delivered, connection survives multiple ping/pong cycles).
- Verified through a full clean-room `bun install --frozen-lockfile` — the exact command `agent/Dockerfile` runs — confirming the patch travels via the committed `patches/*.patch` + `patchedDependencies` and isn't local-only state.
- Added `agent/src/slack-socket-mode-ping.integration.test.ts`, which drives `@slack/bolt`'s actual bundled `SlackWebSocket` (resolved the same way Bolt resolves it in production) against a real local WebSocket server. Confirmed this test fails against the unpatched package and passes against the patched one.
- Full `agent/` suite (1632 tests), lint, and typecheck all green.

## Follow-up

Once oven-sh/bun#37111 merges and lands in a Bun release we take, drop `patches/@slack%2Fsocket-mode@3.0.1.patch` and the `patchedDependencies` entry in one small cleanup PR.

### DIFF
--- a/admin/Dockerfile
+++ b/admin/Dockerfile
@@ -32,6 +32,7 @@ COPY metrics/package.json ./metrics/
 COPY plugins/shipwright/package.json ./plugins/shipwright/
 COPY lib/package.json ./lib/
 COPY mcp-server/package.json ./mcp-server/
+COPY patches/ ./patches/
 
 RUN bun install --frozen-lockfile
 
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -149,6 +149,7 @@ COPY chat/prisma ./chat/prisma/
 COPY plugins/shipwright/package.json ./plugins/shipwright/
 COPY lib/package.json ./lib/
 COPY mcp-server/package.json ./mcp-server/
+COPY patches/ ./patches/
 
 # Install all dependencies (including devDependencies for typecheck)
 RUN bun install --frozen-lockfile
--- a/agent/package.json
+++ b/agent/package.json
@@ -20,7 +20,9 @@
     "playwright": "1.61.1"
   },
   "devDependencies": {
+    "@types/ws": "^8.18.1",
     "bun-types": "^1.3.14",
-    "typescript": "*"
+    "typescript": "*",
+    "ws": "^8.21.3"
   }
 }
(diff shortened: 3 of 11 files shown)


dodizzle (approved):
Verdict: APPROVE — cleanup commit removes the leftover package.json.bak artifact flagged in the last review pass; patch now touches only the two intended require() sites, all 6 Dockerfiles correctly COPY patches/, CI green. No new issues.

PR #3027, opened 2026-09-02
fix: resolve and persist agent.slackId after Slack OAuth completes

dodizzle:
## Summary
- After a successful Slack OAuth code exchange, resolve the bot's own Slack user id via `auth.test` and persist it to `agent.slackId` — this was previously never populated, silently breaking the Sync Manifest button for every agent provisioned through the wizard.
- Add `authTest()` to `SlackProvisioningClient`/`HttpSlackProvisioningClient` and the narrower `AdminUISlackClient` DI interface, folded into `SlackProvisioningService.completeConnect()`'s existing try/catch so a failed `auth.test` is treated the same as a failed OAuth exchange.
- Add `slackId` to `PatchAgentBodySchema` (nullable, optional) so `PATCH /agents/:id` can backfill it for agents that connected Slack before this fix shipped, threaded through `AgentService.updateSelfHosted()`.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | After a successful OAuth code exchange, the bot's own Slack user id is resolved via auth.test and persisted to agent.slackId | MET |
| 2 | The Sync Manifest button appears after completing Slack connect through the current flow, verified by test | MET |
| 3 | PatchAgentBodySchema accepts an optional slackId field so PATCH /agents/:id can backfill it | MET |
| 4 | Test decision — extend connect-slack integration test + add PATCH smoke test | MET |

## Test Plan
- Extended `slack-provisioning-service.integration.test.ts` (recorded-fixture double) with auth.test success/reinstall/failure cases.
- Extended `slack-provisioning-client.integration.test.ts` with a full `authTest` describe block against the HTTP cassette.
- New smoke test in `admin-ui.smoke.test.ts` drives the real `GET /admin/agents/:id/connect-slack/callback` HTTP route end-to-end and asserts the agent detail page renders the "Sync Manifest" button once `slackId` is persisted.
- New smoke tests in `agents-api.smoke.test.ts` for `PATCH /agents/:id` with `slackId` set and cleared (null).

### DIFF
--- a/admin/src/admin-dev-login.smoke.test.ts
+++ b/admin/src/admin-dev-login.smoke.test.ts
@@ -197,6 +197,7 @@ function makeMockDeps(overrides?: Partial<AdminUIDeps>): AdminUIDeps {
         clientSecret: "test-client-secret",
         signingSecret: "test-signing-secret",
       }),
+      authTest: async () => ({ userId: "U0AALR8M69X" }),
       updateAppManifest: async () => {},
       exchangeOAuthCode: async () => ({ botToken: "xoxb-mock-bot-token" }),
     },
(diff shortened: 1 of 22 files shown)


dmcaulay (approved):
Verdict: APPROVE — Clean fix: auth.test is folded into the OAuth exchange try/catch so the flow never half-completes without a resolved slackId, and all four acceptance criteria are genuinely covered end-to-end (including a real-HTTP-route smoke test for the Sync Manifest button). No blocking issues.

PR #3050, opened 2026-09-02
fix: register ChatThreadWatch on the chat UI's actual send route (CFB-4.4)

dodizzle:
## Summary
- Found while doing device verification for CFB-4.4: a real iOS push subscription existed, the agent replied several times, but no push notification ever arrived.
- Root cause: `POST /admin/chat/:agentId/threads/:threadId/messages/upload` — the route the live chat UI's send box actually posts to (`admin-ui-pages.ts`'s `uploadUrl`) — never called `watchThread()`. The two other message-creation routes (`/messages` form POST, `/messages.json`) both call it; this one was missing it entirely, so `ChatThreadWatch` stayed permanently empty and `PushService.notifyThreadReply` always found zero recipients for every thread, with no error anywhere (both failure paths are silent-unless-erroring by design).
- Fix: add the same `if (pushEnabled) { await watchThread(...) }` call this route was missing, mirroring the other two.

## Test plan
- [x] Added 5 new smoke tests in `chat.smoke.test.ts` covering `/messages/upload`: auth gate, happy path, 400 on empty body, and — the actual regression coverage — asserting `chatThreadWatch.upsert` is called with the right `{userEmail, threadId, agentId}` when push is enabled, and NOT called when it's disabled.
- [x] `bun test admin/src/chat.smoke.test.ts` — 45 pass
- [x] `bun run --filter='@shipwright/admin' typecheck` — clean
- [x] Verified live in prod after this fix's logic was manually confirmed against the DB (real `PushSubscription` rows existed, `ChatThreadWatch` was empty across 4+ real agent replies before the fix)

### DIFF
--- a/admin/src/admin-ui.ts
+++ b/admin/src/admin-ui.ts
@@ -3230,6 +3230,7 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
     async (c) => {
       if (!c.var.isAdmin) return new Response("Forbidden", { status: 403 });
 
+      const agentId = c.req.param("agentId");
       const threadId = c.req.param("threadId");
 
       if (!chatClient) {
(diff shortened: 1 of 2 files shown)


dmcaulay (approved):
Verdict: APPROVE — Minimal, surgical fix that adds the missing `watchThread()` call to the `/messages/upload` route, mirroring the identical pattern already used by the two sibling message-creation routes (`/messages` form POST and `/messages.json`). Verified against the base branch and sibling call sites — no breaking changes, no new silent-failure surface (reuses the existing best-effort, non-throwing `watchThread` upsert), and backed by targeted regression tests asserting `chatThreadWatch.upsert` is called with the right args when push is enabled and skipped when disabled.

PR #3010, opened 2026-09-01
refactor: remove superseded self-review heuristics now that ledger covers all cases (PFL-4.1)

dodizzle:
## Summary
- Removes `isResolvedByPriorFindingsStatus` (PVD-1.3) entirely from `compute-unaddressed-findings.ts` — production ledger data confirmed `isResolvedByLedger` (PFL-3.2) now covers every case it caught, and it had no other callers.
- Removes `isSelfCleanApprove` and `isSupersededBySelfReview` as *exclusions inside `hasUnaddressedFindings`* (superseded by `isResolvedByLedger`), but keeps both functions exported — they're still genuinely called by `review.md`'s Step 5.5 to decide what to durably write to the findings ledger in the first place (PFL-2.1), and by `agent/src/check-patch.ts`'s `hasMergeOnlyStaleFindings` for an unrelated staleness check. Deleting them wholesale would have broken both.
- Removes `patch.md`'s manual `reviewState:"pending"` PATCH reset in RPF-1.1 (kept deliberately by PFL-2.2 as a stopgap) — the ledger-write-driven trigger from PFL-3.1 (`check-review.ts`'s `hasFreshLedgerFinding`) is now the sole re-qualification mechanism for no-push rebuttal cycles.
- Removes the now-dead unit tests for the three heuristics (24 tests / ~600 lines); the ledger-based tests added in PFL-3.1/PFL-3.2 already cover the equivalent behavior.
- Refreshes `docs/agent.md`'s stale reference to the removed function.

## Note on acceptance criteria interpretation
Criterion 1 says these three should be "removed from compute-unaddressed-findings.ts and hasUnaddressedFindings." During implementation it became clear `isSelfCleanApprove`/`isSupersededBySelfReview` have real, unrelated callers outside `hasUnaddressedFindings` (confirmed by direct inspection — `check-patch.ts:319` imports and calls `isSelfCleanApprove`). Deleting them wholesale would break `check-patch.ts` compilation and its merge-only-stale-findings check. The fix here removes them specifically as *unaddressed-findings exclusions* (the actual heuristic superseded by ledger coverage per the task's own Feature Overview) while preserving them for their other legitimate uses. Flagging this explicitly for reviewer visibility.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
[...]

PR #3052, opened 2026-09-02
fix: close the self-review ledger fallback gap causing patch to spin (PFL-5.1/5.2/5.3)

dmcaulay:
## Summary

Yesterday's PFL-4.1 (#3010) removed `isSelfCleanApprove`/`isSupersededBySelfReview` as
exclusions in `hasUnaddressedFindings`, relying solely on the task-store findings ledger
(`isResolvedByLedger`). That premise breaks for a PR whose only review ever posted is a
clean self-approve: `review.md`'s Step 5.5 only ledgers a *prior* review at the start of a
*subsequent* review pass, and `check-review.ts`'s own candidacy gate excludes a PR from ever
getting a subsequent pass once nothing has changed. No subsequent pass means no ledger entry
is ever written, so `isResolvedByLedger` stays `false` for that review forever, and
`hasUnaddressedFindings` permanently misclassifies the PR as having an unaddressed finding —
causing `/shipwright:patch` to be repeatedly claimed and dispatched against it (once per
claim-TTL cycle) with nothing to fix each time.

Reproduced directly:
```
$ bun run plugins/shipwright/scripts/compute-unaddressed-findings.ts '{...affected PR data, no findings...}'
{"unaddressedFindings":true}
```
for a PR whose only substantive review is a clean self-approve with no unresolved threads.

The PFL-4.1 removal was properly `hitl: true`-gated in the originating plan
(`planning/pr-findings-ledger/PLAN.md`) — the process worked as designed; the human sign-off
at that gate simply didn't happen to catch this specific PR shape when confirming the bake
period was sufficient.

Full root-cause writeup and design: `planning/pfl-5-ledger-fallback-gap/PLAN.md`.

## Changes (PFL-5.1 / 5.2 / 5.3)

Do both a permanent-fallback fix and a proper ledger fix now, with cleanup deferred until
it's confirmed safe (see PFL-5.4 in the plan doc — filed for tracking, not executed here):

- **PFL-5.1**: Restore `isSelfCleanApprove`/`isSupersededBySelfReview` as fallback exclusions
  in `hasUnaddressedFindings`, layered *alongside* `isResolvedByLedger` — not replacing it.
  Restores the regression tests PFL-4.1 deleted as "now-dead" covering exactly this case.
  This is what immediately fixes the observed spin.
- **PFL-5.2**: Close the gap at the root. `review.md`'s Step 11 now ledgers the review it
  *just* posted, immediately, when it's a self-authored clean-APPROVE — instead of only ever
  ledgering a *prior* review at the start of a subsequent pass. Placed before Step 11b's
  `/complete` call per the existing PR #89 ordering requirement. Scoped to the
  `auto_post_reviews: true` direct-post path (the default); `/shipwright:review-staged`'s
  separate posting call site is flagged as a known follow-up, not silently dropped.
- **PFL-5.3**: `check-patch.ts`'s `getPatchCandidates()` never actually wired the task-store
  ledger into `hasUnaddressedFindings` — `queryPrRecord` was fetched *after* candidacy was
  already decided, and only used for `claimedBy`/`blocked` gating. Move the fetch earlier and
  thread `record.findings` through, so `isResolvedByLedger` isn't dead code at this call site.

All three are additive/restorative. No removal of the ledger path, no regression to the
production-confirmed cases PFL-3.1–3.2 cover.

## Test plan

- [x] `bun test agent/src/check-patch.unit.test.ts` — 69/69 pass (includes 3 new PFL-5.3 cases)
- [x] `bun test plugins/shipwright/scripts/compute-unaddressed-findings.unit.test.ts` — 74/74 pass (restored PFL-5.1 regression tests)
- [x] `bun test plugins/shipwright/commands/review.content.test.ts` — 158/158 pass (new PFL-5.2 content-test assertions)
- [x] Full `bun test agent/src/ plugins/shipwright/` — no failures in any changed file (remaining failures are pre-existing sandbox/environment artifacts unrelated to this diff)
- [ ] `task ci` (lint/typecheck/coverage) — not runnable locally in this environment; deferred to CI on this PR


dodizzle (approved):
Verdict: APPROVE — Restores isSelfCleanApprove/isSupersededBySelfReview as fallback exclusions alongside the task-store ledger check, adds an immediate self-review ledger write at post time, and fixes check-patch.ts's dead-code ledger wiring. Verified the supersession order-safety, the Step 11 gating conditions against Step 10's truth table, and the new unit tests genuinely exercise the restored branches — no blocking issues.

PR #3034, opened 2026-09-02
feat: stream live Thinking Steps for Slack agent runs (STS-1.1)

dmcaulay:
## Summary
- Extends `SlackProgress` (`agent/src/slack-progress.ts`) with an opt-in live Thinking Steps stream: `chat.startStream` on `start()`, one `chat.appendStream` task-update chunk per distinct `ProgressPhase` transition via `onProgress`, and `chat.stopStream` on `finish()` — running in parallel with the existing `agents.sessions.setStatus` lifecycle calls.
- Gated behind a new `SHIPWRIGHT_SLACK_THINKING_STEPS_ENABLED` config value (default `false`), threaded through `config.slack` → `index.ts`'s `createSlackApp(...)` call → all three `SlackProgress` construction sites in `slack.ts` (`app.message`, `app_mention`, `reaction_added`). When disabled, behavior is byte-identical to today.
- Every new Slack API call is wrapped in try/catch and warn-logged, matching the existing `setStatus` fail-open pattern — a Thinking Steps failure never breaks the real reply.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Config flag added, threaded through index.ts/slack.ts, byte-identical when off | MET |
| 2 | start()/onProgress/finish() drive startStream/appendStream/stopStream with phase-collapse | MET |
| 3 | New calls wrapped in try/catch, warn-logged, never thrown | MET |
| 4 | Unit + integration tests extending existing files, no `mock.module()`, fail-open coverage | MET |
| 5 | task ci passes clean | MET |

## Test Plan
- `bun test src/slack-progress.unit.test.ts src/slack.integration.test.ts` — 190/190 passing
- `task lint` — clean
- `task typecheck` — clean across all 7 packages
- [x] Pre-commit checks passing

Generated with [Claude Code](https://claude.com/claude-code)

PR #3057, opened 2026-09-02
fix: repair Slack Thinking Steps stream schema, delivery, status, and queued-message handling

dodizzle:
## Summary
- Fixes the Slack Thinking Steps stream shipped in STS-1.1 (PR #3034), which was broken end-to-end: the `task_update` chunk shape it used (`{type, text}`) is rejected by Slack's real API on every call, producing a permanently blank stream card.
- `task_update` chunks now use the `{type, id, title, status}` shape with one stable `id` reused for the life of a run (a fresh id per update stacked a separate card instead of updating in place — a second confirmed-live bug).
- The final reply is now delivered as its own `markdown_text` chunk via `chat.appendStream` (alongside a `status: "complete"` task_update), instead of `chat.stopStream`'s top-level `markdown_text` param, which Slack rejects with `streaming_mode_mismatch` once `task_display_mode: "timeline"` locks the stream into chunk-based mode. `deliverContent()` returns `undefined` on any failure so all three Slack handlers fall back to the existing `say()`/`blocksConverter` path — the reply is never silently dropped.
- `agents.sessions.setStatus` is now skipped entirely when `thinkingStepsEnabled` is true, since `chat.startStream` already flips the session status to "processing" itself — calling both produced two overlapping "is working" indicators.
- The stream is now seeded with an immediate `in_progress` card on open (no blank gap before the first phase transition), and the terminal card is explicitly sent `status: "complete"` (previously left `"in_progress"`, which renders as an error/warning triangle in Slack even on success). Every `PROGRESS_LABELS` entry now ends in "…"; the completion title ("Done") does not.
- Every message now gets its own independent `chat.startStream`/`appendStream`/`stopStream` lifecycle regardless of `ThreadStatusTracker`'s refcount. The refcount now only picks the seed title ("Queued…" vs "Thinking…") and the `session_status` `stopStream` closes with ("processing" vs "active") — it no longer gates whether `start()`/`finish()` are called in streaming mode. `enter()`/`exit()` are called unconditionally by all three handlers.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `task_update` chunk shape `{type,id,title,status}`, stable id reused | MET |
| 2 | Final answer via `markdown_text` chunk + `appendStream`; `deliverContent()` → `undefined` on failure, falls back to `say()` | MET |
| 3 | `agents.sessions.setStatus` skipped entirely when streaming | MET |
| 4 | Immediate seed card, explicit `status:"complete"` on close, labels end in "…", completion text doesn't | MET |
| 5 | Independent stream lifecycle per message; refcount only informs seed text/session_status | MET |
| 6 | Smoke-appropriate tests extending existing `*.unit.test.ts`/`*.integration.test.ts` files, incl. 2 named regression tests | MET |
| 7 | Manual live-Slack markdown rendering verification | **Not independently re-verified in this session** — see task description, which states this was confirmed live against a running agent (Bodhi) on 2026-09-02. This session had no live Slack access to re-check it. |

## Test Plan
- Extended `agent/src/slack-progress.unit.test.ts` and `agent/src/slack.integration.test.ts` (no new test files) covering every acceptance criterion, including the two named regression tests: a burst of queued messages each getting its own stream with correct Queued/Thinking seed labels, and the first-to-finish message of a burst keeping `session_status: "processing"` while a later one is still in flight — both fail without their respective fix and pass with it.
- Extended `lib/progress-phases.unit.test.ts` to enforce the ellipsis convention.
- `NODE_ENV=test bun test` — all 259 Slack/progress-related tests pass (5 files). Full repo suite: 30 pre-existing failures on `main` (verified identical before/after this change — unrelated env-var-leak and task-store `validateRepo` null-repos issues), zero new failures introduced.
- `task lint`, `bun run --filter='*' --sequential typecheck` — clean across all workspaces.
- Coverage on touched files: `agent/src/slack-progress.ts` 100%/100% (lines/funcs), `lib/progress-phases.ts` 100%/100%, `agent/src/slack.ts` 94.23% lines.


dmcaulay (approved):
Verdict: APPROVE — Well-scoped, correctly-tested fix for the confirmed-live broken Thinking Steps stream (STS-1.1). The startProgress()/finishProgress() refactor preserves non-streaming behavior byte-for-identical across all 3 handlers, deliverContent() never silently drops content, and test coverage genuinely exercises every acceptance criterion including concurrent-burst regressions. No blocking issues.

PR #3064, opened 2026-09-02
feat: mark Thinking Steps card error on thrown run

dodizzle:
## Summary
- Added `SlackProgress.reportError()` — sends a `status: "error"` task_update to the Thinking Steps card, guarded (no-op) when `thinkingStepsEnabled` is off or no stream is open, and marks `completed` so `finish()` doesn't send a redundant `status: "complete"` card afterward.
- Wired `progress.reportError()` into all three Slack event handler catch blocks (DM/thread, `app_mention`, `reaction_added`) in `agent/src/slack.ts`, called before the `finally` block's `finishProgress()`.
- `reportError()` adds no new logging — the existing `sentryClient.captureException(err)` calls remain the sole log point per the repo's `error_handling` principle.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | On a caught run error, if a stream is open, send a `status: "error"` task_update before `finish()` closes it | MET |
| 2 | `sentryClient.captureException(err)` remains the single log point; new status update is UI-state only | MET |
| 3 | Add a case to `slack.integration.test.ts` asserting the card's `status: "error"` update fires when `mockRunClaude` rejects with `thinkingStepsEnabled: true` | MET |

## Test Plan
- Added 3 new tests to `agent/src/slack.integration.test.ts` (DM, `app_mention`, `reaction_added` handlers) asserting an error `task_update` chunk fires on a thrown run with `thinkingStepsEnabled: true`, and that no redundant `status: "complete"` chunk follows.
- `bun test agent/src/slack.integration.test.ts`: 176 pass, 0 fail.
- `task typecheck`: clean.
- `bunx biome lint` on touched files: clean.
- Coverage on touched files: `slack-progress.ts` 96.18% lines, `slack.ts` 91.33% lines (both above the 90% threshold).
- `task ci`'s `test:coverage` step fails on 31 pre-existing, unrelated failures in `metrics/` and `task-store/` — confirmed present on unmodified `main` before this change.


agent/src/slack.ts
```diff
@@ -697,6 +697,10 @@ export function createSlackApp(
         text: formatRunErrorForSlack(err),
         thread_ts: msg.thread_ts ?? msg.ts,
       });
+      // Mark the Thinking Steps card status:"error" (STS2-4.1 AC #1) — a
+      // UI-state signal only; sentryClient.captureException(err) above
+      // remains the single log point for this error.
+      progress.reportError();
```
dmcaulay, line 702: If the `await say(...)` above throws (a real Slack API failure — rate limiting, `invalid_auth`, `channel_not_found`), this line never runs, `SlackProgress.completed` stays `false`, and the `finally` block's `finishProgress()` sends a `status: "complete"` ("Done") card — misrepresenting a run that both errored and failed to notify the user as successful. The `app_mention` handler below has the same ordering (line 886). The `reaction_added` handler avoids this because `reportError()` is its catch block's first statement. Consider calling `progress.reportError()` before the `say()` call, or wrapping `say()` in its own try/catch.

dmcaulay (commented):
Verdict: COMMENT — In the DM and app_mention handlers, `progress.reportError()` runs after the catch block's own `say()` call, so a failed `say()` (rate limit, invalid_auth, etc.) skips it and leaves the card falsely marked "Done".

dmcaulay (approved):
Verdict: APPROVE — Prior ordering bug fixed: reportError() now runs before say() in both DM and app_mention handlers, with say() wrapped in try/catch, plus new regression tests covering the say()-also-throws case. No new issues.

PR #3065, opened 2026-09-02
fix: close the Thinking Steps stream cleanly on a silent response

dodizzle:
## Summary
- `finish()` now closes the stream's dangling `in_progress` card even when the reply was suppressed (e.g. a `[silent]` marker), instead of leaving no completion signal.
- The fallback title is `"Ack"` (not the generic `"Done"`) when the run's reply was suppressed, so the card doesn't imply a real reply went out while still signaling the run completed rather than silently breaking.
- All three Slack event handlers (DM, `app_mention`, `reaction_added`) now compute and pass `wasSuppressed` through to `finishProgress()`.

## Acceptance Criteria

| 1 | Silent suppression sends `status: "complete"` task_update titled "Ack", no markdown_text, before stream closes | MET | `SlackProgress.finish()` uses `SILENT_COMPLETE_TITLE = "Ack"` when `opts?.silent` is true |
| 2 | No dangling "in_progress" card left visible after a silent response completes | MET | All 3 call sites pass `wasSuppressed` through `finishProgress()` → `progress.finish({ silent })`, closing the card unconditionally |
| 3 | Smoke test in DM-routing describe block: `thinkingStepsEnabled: true` + `[silent]` marker asserts "Ack" title | MET | New test in `slack.integration.test.ts` asserts `say()` not called, 0 markdown_text chunks, complete chunk title === "Ack" |

## Test Plan
- `agent/src/slack-progress.unit.test.ts`: `finish({ silent: true })` sends "Ack" title; regression test confirms `finish()`/`finish({ silent: false })` still send "Done"
- `agent/src/slack.integration.test.ts`: DM-routing smoke test confirms a `[silent]` response closes the stream with "Ack" and posts no content


dmcaulay (approved):
Verdict: APPROVE — Clean fix. `wasSuppressed` is correctly hoisted and threaded through all three Slack handlers (DM, app_mention, reaction_added) including error/early-return paths, and the DM `dmOverride` case is handled correctly (tracks `shouldSuppress`, not raw `isSilent`). New unit + integration tests genuinely exercise the new "Ack" title path and pass. No blocking issues.

PR #3069, opened 2026-09-03
fix: seed renderedIds from upload response to prevent duplicate chat bubble

dodizzle:
## Summary
- Seeds `renderedIds` from the upload response's `data.message.id` in `sendText()`'s fetch success branch, so the poll loop's subsequent `renderServerBubble()` call skips re-rendering the just-uploaded message.
- Fixes a duplicate chat bubble: the optimistic bubble drawn immediately by `addUserBubble()` and the server-echoed bubble from the poll loop both stayed in the DOM because `renderedIds` never learned the id from the upload response.
- No changes to the upload endpoint's response shape — `admin-ui.ts` already returns `{ message }` with `id`.

## Acceptance Criteria

| 1 | `sendText()`'s upload fetch success branch parses the JSON response and sets `renderedIds[data.message.id] = true` when present, before the poll loop can render a duplicate | MET | `admin/src/admin-ui-pages.ts:4897-4904` |
| 2 | Unit test added mirroring the existing "inline JS uses an id-based renderedIds dedupe set" test style; no existing tests retired | MET | `admin/src/admin-ui-pages.unit.test.ts:6591-6598` |
| 3 | No changes to the upload endpoint's response shape | MET | `admin-ui.ts` diff is empty |

## Test Plan
- Added unit test asserting the generated inline JS contains `renderedIds[data.message.id]`
- `bun test admin/src/admin-ui-pages.unit.test.ts` — 660/660 pass
- `bunx biome lint .` — clean
- `task typecheck` — clean
- `check-strings`, `check-config-docs`, `check-version-sync` — pass

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -4894,6 +4894,13 @@ export function renderChatThreadPage(
           container.appendChild(errBubble);
           enableSend();
         });
+      } else {
+        // Success: parse response and seed renderedIds to prevent duplicate rendering
+        return r.json().then(function(data) {
+          if (data && data.message && data.message.id) {
+            renderedIds[data.message.id] = true;
+          }
+        });
       }
     }).catch(function() {
       // POST failed — Layer 1 ticker keeps going; poll loop still starts.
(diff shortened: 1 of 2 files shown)


dmcaulay (approved):
Verdict: APPROVE — Minimal, well-tested fix for the duplicate chat bubble bug; seeds renderedIds from the upload response using the existing dedupe pattern, no breaking changes, no regressions.

PR #3070, opened 2026-09-03
docs: migrate investigate-cron SKILL.md's task lookup off stored PR.taskId

dodizzle:
## Summary
- Replaces investigate-cron `SKILL.md`'s stored `.taskId` fallback with a live `GET /tasks?repo=&pr=` lookup, since the `PullRequest.taskId` column/API field is being removed by a follow-up task (PTL-3.1)
- Loops over the lookup results to correctly handle the multi-task bundle case (a PR mapping to zero, one, or many tasks) instead of assuming a single match
- Drops `taskId` from the diagnostic `jq` print of the PR record, and updates surrounding prose to describe the new live-lookup behavior
- Rewrites the matching assertions in `SKILL.content.test.ts` and `SKILL.unit.test.ts` to check for the live-lookup snippet instead of the old `taskId` string

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | SKILL.md section 2b no longer reads `.taskId` off `PR_RECORD_JSON` — uses live `GET /tasks?repo=&pr=` lookup, looping over results for the bundle case | MET |
| 2 | Diagnostic jq print of the PR record no longer includes `taskId` | MET |
| 3 | Surrounding prose in section 2b describes the live-lookup behavior | MET |
| 4 | `SKILL.content.test.ts` and `SKILL.unit.test.ts` rewritten to assert the new live-lookup snippet's content instead of `taskId` presence | MET |

## Test Plan
- `NODE_ENV=test bun test plugins/shipwright/skills/investigate-cron/SKILL.content.test.ts plugins/shipwright/skills/investigate-cron/SKILL.unit.test.ts` — 120 pass, 0 fail
- `bun run --filter='*' --sequential typecheck` — passes
- `task ci` — pre-existing failures unrelated to this change (task-store `prs.ts` validateRepo null-repos bug, metrics env test flakes, github-app-auth integration tests); none touch investigate-cron/SKILL.md

### DIFF
--- a/plugins/shipwright/skills/investigate-cron/SKILL.md
+++ b/plugins/shipwright/skills/investigate-cron/SKILL.md
@@ -306,8 +306,12 @@ if [[ "$ITEM_ARG" =~ ^[A-Z]+-[0-9]+\.[0-9]+$ ]]; then
   else
     echo "$TASK_JSON" | jq '{id, status, hitl, dependencies}'
   fi
-elif [ -n "$PR_RECORD_JSON" ]; then
-  # ITEM_ARG is a PR reference — fetch associated tasks via live lookup
+else
+  # ITEM_ARG is a PR reference — fetch associated tasks via live lookup.
+  # Gated on ITEM_ARG's shape, not on PR_RECORD_JSON being non-empty: a
+  # freshly-opened PR not yet claimed by review/patch/deploy has no
+  # PullRequest record yet, but real Task records (Task.repo/Task.pr) can
+  # already exist — the live lookup below must still run for it.
   TASKS_JSON=$(curl -sf -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
     "$SHIPWRIGHT_TASK_STORE_URL/tasks?repo={org}/{repo}&pr={pr}" 2>/dev/null)
   if [ -z "$TASKS_JSON" ]; then
--- a/plugins/shipwright/skills/investigate-cron/SKILL.md
+++ b/plugins/shipwright/skills/investigate-cron/SKILL.md
@@ -293,18 +293,33 @@ PR_RECORD_JSON=$(curl -sf -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN
 if [ -z "$PR_RECORD_JSON" ]; then
   echo "No task-store PR record for {org}/{repo}#{pr} — not yet tracked, or never a PR. Skip this signal."
 else
-  echo "$PR_RECORD_JSON" | jq '{claimedBy, hitl, reviewState, readyForReviewAt, readyForPatchAt, readyForDeployAt, taskId, patchCycles, commitSha}'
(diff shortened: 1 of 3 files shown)


plugins/shipwright/skills/investigate-cron/SKILL.md
```diff
@@ -293,26 +293,43 @@ PR_RECORD_JSON=$(curl -sf -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN
...
   TASK_JSON=$(curl -sf -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
-    "$SHIPWRIGHT_TASK_STORE_URL/tasks/$TASK_ID_TO_FETCH" 2>/dev/null)
+    "$SHIPWRIGHT_TASK_STORE_URL/tasks/$ITEM_ARG" 2>/dev/null)
   if [ -z "$TASK_JSON" ]; then
-    echo "No task-store task record for $TASK_ID_TO_FETCH — skip this signal."
+    echo "No task-store task record for $ITEM_ARG — skip this signal."
   else
     echo "$TASK_JSON" | jq '{id, status, hitl, dependencies}'
   fi
+elif [ -n "$PR_RECORD_JSON" ]; then
```
dmcaulay: Gating the live task lookup on `PR_RECORD_JSON` being non-empty conflates "task-store already has a tracked `PullRequest` record" with "`ITEM_ARG` is a PR reference." `PullRequest` records are only created later, on the first `/prs/claim` by review/patch/deploy — not when `dev-task` opens the PR. The task-store `Task` model has its own `repo`/`pr` fields independent of `PullRequest` (`task-store/prisma/schema.prisma:50,56`), so a freshly-opened, not-yet-claimed PR has real Task records but empty `PR_RECORD_JSON` here — and this block silently skips the lookup even though `{org}`/`{repo}`/`{pr}` are already known from `ITEM_ARG` (used two lines above to build the `PR_RECORD_JSON` query itself).

Suggest gating on `ITEM_ARG` shape instead: `elif [ -n "$ITEM_ARG" ] && ! [[ "$ITEM_ARG" =~ ^[A-Z]+-[0-9]+\.[0-9]+$ ]]` — anything non-empty that isn't a bare task id is a PR reference, regardless of whether `PR_RECORD_JSON` came back non-empty.

dmcaulay (commented):
Verdict: COMMENT — The live task lookup for PR references is gated on `PR_RECORD_JSON` existing, but that only reflects whether the task-store's `PullRequest` record has been created yet (which happens on the first review/patch/deploy claim, not on PR open). A freshly-opened PR not yet picked up for review/patch/deploy will have real `Task` records (via `Task.repo`/`Task.pr`) but empty `PR_RECORD_JSON`, so the whole task-lookup block silently skips — exactly the case investigate-cron is meant to diagnose. See inline comment.

dmcaulay (approved):
Verdict: APPROVE — Prior finding (task lookup gated on stale PR_RECORD_JSON instead of ITEM_ARG shape) is cleanly fixed: the lookup now branches on ITEM_ARG's shape and runs unconditionally for PR references, matching the suggested fix. Tests updated to match. No new issues.

PR #3074, opened 2026-09-03
feat: subscribe to and handle Slack's agent_session_stopped event

dodizzle:
## Summary
- Subscribes the agent's Slack manifest to the `agent_session_stopped` event — Slack's native Stop/session-cancellation event for the Agent Sessions API — fixing the `missing_agent_session_stopped_event_subscription` log line seen on nearly every run.
- `createSlackApp()` now tracks live Thinking Steps streams in a closure-scoped registry (same per-call lifecycle as `ThreadStatusTracker`) and registers a Bolt listener for `agent_session_stopped` that marks the matching stream externally-stopped (further `appendStream`/`stopStream` calls become no-ops) and independently resets the thread's status via `agents.sessions.setStatus({status: "active", ...})` — fixing the permanently-stuck "is working" status.
- **Human follow-up required:** existing agents need `POST /admin/agents/:id/sync-manifest` (with a fresh `xoxp-` token) to pick up the new event subscription — new agents get it automatically via `apps.manifest.create` at provisioning time. This PR does not run sync-manifest.
- **Known limit, out of scope (confirmed with Dan):** this does not eliminate the duplicate-message pattern on long-running turns — once Slack stops a stream server-side it never accepts further writes to that message, so the original card stays frozen and the real answer still lands via the existing `say()` fallback as a second message. This task only stops the agent from wasting further failed API calls against the dead stream and prevents the status from getting stuck.

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | `buildAgentManifest()` adds `agent_session_stopped` to `bot_events`; unit test updated | MET |
| 2 | `createSlackApp()` registers a Bolt listener for `agent_session_stopped` | MET |
| 3 | Closure-scoped registry (`channel:ts` → `SlackProgress`), per-`createSlackApp()` lifecycle | MET |
| 4 | Registry lookup marks stream externally-stopped; further stream calls become no-ops | MET |
| 5 | Direct `agents.sessions.setStatus('active')` call, independent of stream close path, failure caught | MET |
| 6 | Duplicate-message pattern explicitly out of scope | MET (not touched) |
| 7 | Smoke-layer test coverage (registry lookup, no-op, setStatus reset) | MET |

## Test Plan
- `agent/src/slack-progress.unit.test.ts` — `streamKey()`, `getThreadTs()`, `markExternallyStopped()` no-op guards on `appendStream`/`deliverContent`/`reportError`/`stopStream`
- `agent/src/slack.unit.test.ts` (smoke layer, mocked Bolt app) — registry lookup on `agent_session_stopped`, no-op on further stream calls after the event fires, direct `setStatus('active')` reset with correct `thread_ts`, unknown-key no-op, `setStatus` failure swallowed, registry cleanup on stream close
- `admin/src/slack-provisioning-client.unit.test.ts` — `bot_events` includes `agent_session_stopped`
- [x] `bun test` — all new/changed tests pass
- [x] `task typecheck` — clean
- [x] `task ci` — clean aside from 31 pre-existing failures unrelated to this diff (confirmed identical on a clean `main` checkout — env-var/test-isolation pollution in `metrics/src/lib/env.unit.test.ts`, `errors.unit.test.ts`, and a `task-store` repo-scope smoke test; none touch files in this change)


dmcaulay (approved):
Verdict: APPROVE — All 7 acceptance criteria met; registry lifecycle and no-op guards are correct with no leaks or races found.

agent/src/slack.ts
```diff
@@ -1030,6 +1064,63 @@ export function createSlackApp(
...
+
+    // Look up the matching live stream and mark it externally-stopped —
+    // any further appendStream/stopStream call against it becomes a no-op
+    // instead of a failed API call. An unknown/already-cleared channel:ts
+    // (stream already closed via its own success/failure path, or a
+    // duplicate delivery of the same event) is a safe no-op — never throws.
+    // Also recovers the run's actual thread_ts (distinct from the stream's
+    // own streaming_message_ts) for the direct status reset below, when a
+    // match was found.
+    let threadTs = ev.streaming_message_ts;
```
dmcaulay: In the `agent_session_stopped` handler, `threadTs` starts as `ev.streaming_message_ts` and is only overwritten with the real `progress.getThreadTs()` value inside the `if (progress)` block above. When the `liveStreams` registry lookup misses — which happens for any duplicate or late-arriving event for a run whose entry was already cleared by `unregisterStream` in the handler's `finally` block — `setStatus` below is still called, but with `thread_ts` left as the stream's own `streaming_message_ts` rather than the actual thread ts (per `SlackProgress.getThreadTs()`'s own doc comment, these are explicitly distinct values). Consider moving the `setStatus` call inside the `if (progress)` block, or omitting `thread_ts` on a registry miss rather than falling back to `streaming_message_ts`.

dmcaulay (commented):
Verdict: COMMENT — Registry lifecycle and no-op guards are solid, but the agent_session_stopped handler sends the wrong thread_ts to setStatus when the registry lookup misses (e.g. a duplicate/late event for an already-finished run).

dmcaulay (approved):
Verdict: APPROVE — Prior finding (wrong thread_ts on registry miss) is verifiably fixed with a passing regression test; no new issues found.

PR #3095, opened 2026-09-03
fix: preserve org/repo format in plan-session.md repo auto-detect

dodizzle:
## Summary
- Fix `plan-session.md`'s repo auto-detect to preserve the full `owner/repo` value instead of stripping it to a bare repo name, which was the confirmed root cause of a production 400 from task-store's `validateRepo()` when the bare value flowed into Step 6b's task JSON `repo` field.
- Introduce a separate `{repo-slug}` value (last path segment of `repo`, lowercased) for local filesystem path references, mirroring the pattern already used in `entropy-fix/SKILL.md`'s 6q.1.
- Add a content-test assertion guarding against regression to a bare-repo-name auto-detect instruction.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Arg description example (line 5) updated to org/repo format (e.g. "app-vitals/shipwright") | MET |
| 2 | Auto-detect logic preserves full owner/repo value, separately derives {repo-slug} | MET |
| 3 | Local filesystem path references (Step 1, confirmation banner) use {repo-slug} instead of {repo} | MET |
| 4 | Step 6b's task JSON template continues writing the full org/repo value into "repo" | MET |
| 5 | Content-test assertion added to plan-session.content.test.ts guarding against regression | MET |

## Test Plan
- `bun test plugins/shipwright/commands/plan-session.content.test.ts` — 21/21 pass (14 pre-existing + 7 new)
- `task typecheck` — all 7 workspaces pass
- `task ci` run in full; 31 pre-existing failures confirmed unrelated (reproduce identically on `main`: env-var config defaults, GH App credential checks, Sentry init — all environment-contamination issues, none touching plan-session.md or repo-format logic)

### DIFF
--- a/plugins/shipwright/commands/plan-session.content.test.ts
+++ b/plugins/shipwright/commands/plan-session.content.test.ts
@@ -109,6 +109,57 @@ describe("plan-session.md — Step 6b template omits requiresHumanApproval (RHA-
   });
 });
 
+describe("plan-session.md — repo auto-detect preserves org/repo format (PRF-1.1)", () => {
+  function extractArgsAndAutoDetectSection(md: string): string {
+    const match = md.match(/^---[\s\S]*?Wait for user confirmation before continuing to Step 1\./);
+    expect(match).not.toBeNull();
+    return match?.[0] ?? "";
+  }
+
+  it("the repo arg description example is org/repo format, not a bare repo name", () => {
+    const section = extractArgsAndAutoDetectSection(content);
+    const argDescMatch = section.match(/- name: repo\n\s*description: (.+)/);
+    expect(argDescMatch).not.toBeNull();
+    const argDesc = argDescMatch?.[1] ?? "";
+    expect(argDesc).toMatch(/e\.g\.,\s*[\w.-]+\/[\w.-]+/);
+    expect(argDesc).not.toMatch(/e\.g\.,\s*shipwright\)/);
+  });
+
+  it("does not instruct stripping to the bare repo name in auto-detect", () => {
+    const section = extractArgsAndAutoDetectSection(content);
+    expect(section.toLowerCase()).not.toContain("bare repo name");
+  });
+
+  it("documents preserving the full owner/repo value from git remote parsing", () => {
+    const section = extractArgsAndAutoDetectSection(content);
+    const lower = section.toLowerCase();
+    expect(lower).toMatch(/preserve|do not strip|full owner\/repo/);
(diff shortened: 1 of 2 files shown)


dmcaulay (commented):
Verdict: COMMENT — plan-session.md's org/repo auto-detect fix is correct and well-tested (21/21 content-test assertions pass, mirrors the entropy-fix repo/repo-slug pattern). One important note: dev-task.md consumes the same task-store `repo` field as a raw filesystem path segment (`${SHIPWRIGHT_REPO_DIR}/{repo}`, ~12 call sites, no repo-slug derivation). Before this PR, auto-detect wrote a bare repo name there, which happened to be path-correct; after this PR it writes the full org/repo value, so dev-task's worktree setup would resolve to a nonexistent nested path for tasks queued via auto-detect. This is already tracked separately as PRF-1.4 (in_progress, branch feat/prf-1-4-dev-task-repo-slug) — not a defect to fix in this diff, but worth confirming PRF-1.4 lands before/alongside this PR deploys to avoid a live gap, since PRF-1.1 has no dependency on it.

dodizzle:
Confirmed — this PR (PRF-1.1) intentionally does not touch dev-task.md's consumption of the `{repo}` field. That gap is tracked separately as PRF-1.4 (branch `feat/prf-1-4-dev-task-repo-slug`), which adds repo-slug derivation to dev-task.md's local filesystem path references (`${SHIPWRIGHT_REPO_DIR}/{repo}` call sites, ~12 occurrences).

Verified against the task store: PRF-1.4 exists and is currently `in_progress` (claimed, heartbeat fresh as of 2026-09-03T09:20:54Z), with acceptance criteria that explicitly cover deriving `{repo-slug}` from the fetched task's org/repo `{repo}` field and replacing every local-path reference while leaving task-store API calls on the full org/repo value.

There is no code dependency between the two tasks, but PRF-1.4 should land before or alongside this PR deploys to avoid a live path-resolution gap for tasks queued via plan-session's auto-detect (which, after this PR, writes the full org/repo value into the task-store `repo` field). No code change needed in this PR.

dmcaulay (approved):
Verdict: APPROVE — Re-reviewing at the same commit after your reply. Confirmed: the dev-task.md `{repo}`-path gap I flagged is correctly out of scope for this PR and already tracked separately as PRF-1.4 (in_progress). No new issues found — clean, well-tested fix.

PR #3132, opened 2026-09-03
feat: check-patch.ts candidate path for allowlisted authors (DBR-1.4)

dodizzle:
## Summary
- Added `listAllowlistedOpenPrs(repo)` to `check-patch.ts`'s `CheckPatchDeps` — queries `gh pr list --author {login}` for each login in `patchAuthorAllowlistRef.get()`, skipped entirely (no `gh` calls) when the allowlist is empty
- `getPatchCandidates` now merges `listOwnOpenPrs` and `listAllowlistedOpenPrs` results, deduplicated by `(repo, PR number)` via `candidateId`, before scope-filtering and the existing per-PR qualification loop
- Unlike `check-review.ts`'s `isAuthorAllowed` (a fail-open filter on an already-broad pool), this is an additive source: an empty allowlist contributes nothing, preserving exact self-authored-only behavior as the default
- Refreshed `docs/agent-api.md` and `docs/configuration.md` to reflect that `patchAuthorAllowlist` enforcement is now active (previously documented as synced-but-unenforced from DBR-1.3)

## Acceptance Criteria

| 1 | listAllowlistedOpenPrs(repo) dep queries gh pr list --author {login} per allowlist entry, skipped entirely when empty | MET | check-patch.ts early return on empty allowlist, no fail-open |
| 2 | getPatchCandidates merges listOwnOpenPrs + listAllowlistedOpenPrs, deduped by (repo, PR number) | MET | candidateId-based dedup Set |
| 3 | Self-authored-only behavior unchanged when allowlist empty (default) | MET | `?? []` fallback + dedicated test |
| 4 | Unit tests cover merge/dedup + empty-allowlist default; no existing tests retired | MET | 5 new tests added, all 74 pre-existing + new tests pass |

## Test Plan
- `NODE_ENV=test bun test agent/src/check-patch.unit.test.ts` — 74/74 pass, including 5 new tests covering empty-allowlist default, additive merge, dedup, and the two `buildProductionDeps` gh-call-shape tests
- Coverage on `check-patch.ts`: 93.94% lines / 100% functions (above the 90% threshold)
- `bunx biome lint` and `tsc --noEmit` clean on touched files


agent/src/check-patch.ts
```diff
@@ -394,9 +412,24 @@ export async function getPatchCandidates(
   const scopeSynced = deps.hasScopeSynced();
   const scopedRepos = new Set(deps.getScopedRepos());
   const allOwnPrs = await deps.listOwnOpenPrs("default");
+  // Additive allowlisted-author source (DBR-1.4) — merged with allOwnPrs and
+  // deduplicated by (repo, PR number) via candidateId before scope-filtering.
+  // deps.listAllowlistedOpenPrs is optional (undefined for callers/fixtures
+  // that predate DBR-1.4) and, per its own contract, returns [] when the
+  // allowlist itself is empty — either way this defaults to [], preserving
+  // exact self-authored-only behavior when no allowlist is configured.
+  const allowlistedPrs = (await deps.listAllowlistedOpenPrs?.("default")) ?? [];
```
dmcaulay, line 451: `listAllowlistedOpenPrs` fetches PRs per allowlisted `login`, but that login is never attached to the returned `OwnPr` (no author field exists on `OwnPr`), and `getPatchCandidates` never builds a `prAuthor` onto `reviewData` before calling `hasUnaddressedFindings(reviewData, currentUser)` below (or before `hasMergeOnlyStaleFindings`'s own `isAddressedByAuthorReply(r, comments.nodes, currentUser)` call at line 383).

In `compute-unaddressed-findings.ts:456`, `prAuthor` defaults to `currentUser` when absent — and that file's own RAS-1.1 doc comment says this default is only safe because "`patch` only ever acts on the authenticated user's own open PRs." This PR is exactly what breaks that precondition.

Concrete impact for an allowlisted (non-self-authored) PR: `isAddressedByAuthorReply`/`isThreadAddressedByAuthorReply` compare `comment.author.login` against the agent's own login instead of the real author, so a genuine reply from the actual PR author addressing a finding is never recognized — the PR can stay a patch candidate forever. Conversely, a comment the agent itself posted on that PR could be misattributed as an author reply, wrongly suppressing a real finding.

The two new DBR-1.4 tests below both set `mergeStatus.isDirty: true`, which short-circuits `needsPatch = true` before this code path is reached, so there's no test coverage exercising `hasUnaddressedFindings`/`hasMergeOnlyStaleFindings` for an allowlisted PR — this shipped without a failing test to catch it.

Suggest adding an author/login field to `OwnPr` (from the `login` loop variable in `listAllowlistedOpenPrs`), threading it into `reviewData.prAuthor` here, and passing it (not `currentUser`) into `hasMergeOnlyStaleFindings`'s `isAddressedByAuthorReply` call too.

dmcaulay (commented):
Verdict: COMMENT — Real bug: the actual PR author's login is never threaded through for allowlisted (non-self) candidate PRs, so hasUnaddressedFindings/hasMergeOnlyStaleFindings's author-reply-addressed checks silently default prAuthor to the agent's own login instead of the real author (see inline comment on check-patch.ts:421). This can leave allowlisted PRs stuck as permanent patch candidates, or wrongly suppress genuine findings. Also, docs/agent-api.md:53 still says patchAuthorAllowlist is unenforced, contradicting this PR's own corrected text elsewhere in the same file.

dodizzle:
Addressed — this was fixed in commit fae1c165 (already on this branch before this patch cycle):

- `OwnPr.author` now carries the queried login from `listAllowlistedOpenPrs`, and `getPatchCandidates` threads it into `reviewData.prAuthor` (`pr.author ?? currentUser`) at check-patch.ts:530.
- Both `hasUnaddressedFindings` and `hasMergeOnlyStaleFindings` resolve `prAuthor` as `data.prAuthor ?? currentUser` (check-patch.ts:396), so an allowlisted (non-self) PR's real author reply is now recognized correctly instead of being compared against the agent's own login.
- 4 unit tests were added exercising the review-findings path for allowlisted PRs (non-dirty), covering the case that previously had no coverage.
- The stale `docs/agent-api.md:53` "patchAuthorAllowlist is unenforced" text is also corrected — verified again during this cycle's merge-conflict resolution that no stale "not yet enforced" language survives in docs/, plugins/, or site/.

Inline thread already resolved; leaving this comment so the review is recognized as addressed.

dmcaulay (approved):
Verdict: APPROVE — Both the prAuthor-threading bug and the stale-docs finding from my prior review are fixed and well-tested (4 new regression tests covering the real-author-reply, agent-comment, hasMergeOnlyStaleFindings, and self-authored-unchanged cases). No new issues found.

PR #3142, opened 2026-09-03
feat: switch hitl dev loop to auto permission mode

dmcaulay:
## Summary

Switches the `hitl` dev loop's interactively-spawned Claude Code session (`scripts/hitl.ts`'s `runLoop`, around line 1200) from `--permission-mode acceptEdits` to `--permission-mode auto`.

**Why:** `auto` is now Claude Code's default permission mode, and the `hitl` loop already has a human operator attached (`stdin: "inherit"`, run via `/shipwright:hitl`) who is available to unblock anything `auto` mode declines to auto-approve. This moves the loop from pinning to the old default onto tracking the current one, rather than introducing a new safety posture from scratch.

**What actually changes, mechanically** (per Claude Code's permission-mode docs): both modes route tool calls that match `--allowedTools` (`HITL_ALLOWED_TOOLS`, unchanged by this PR — see `scripts/hitl.ts:76-83`) straight through without a prompt. The difference is in what happens to calls that *don't* match the allowlist — which is where Bash and Agent land, since the comment above `HITL_ALLOWED_TOOLS` explains they're deliberately left off:
- Under `acceptEdits`, an out-of-allowlist Bash call always stops and prompts the human operator before running.
- Under `auto`, an out-of-allowlist Bash/Agent call is instead evaluated by Claude Code's background classifier. If the classifier judges it safe it runs with no human prompt at all; the human operator is only pulled back in when the classifier blocks the action (a notification + retry-with-manual-approval), or after repeated blocks (3 in a row / 20 total) trip a fallback to interactive prompting.

So `HITL_ALLOWED_TOOLS` itself is unchanged — Bash and Agent are still off the allowlist — but the nature of the gate behind that exclusion changes: from "every out-of-allowlist call prompts a human" to "every out-of-allowlist call is classifier-reviewed, with the human as a backstop for what the classifier flags or can't decide." That's consistent with this PR's stated rationale (human-in-the-loop as the unblock mechanism), but it is a real narrowing of per-action human visibility, not a no-op — worth calling out explicitly rather than asserting the old boundary is preserved unchanged.

This is a **local dev-tool change only**. No other file is touched by this diff, and the production agent path (`agent/src/claude.ts`, which spawns Claude non-interactively with its own `--permission-mode acceptEdits`) is untouched.

## Related issue

N/A

## Verification

No automated test covers this line — it's a single CLI argument to an interactively-spawned (`stdin: "inherit"`) `claude` process in a manual dev tool, not independently testable behavior. Verification is via the diff itself: a one-line permission-mode string change (`acceptEdits` → `auto`) with no other code paths affected. `bunx biome lint scripts/hitl.ts` passes clean with no changes required, and CI is green: https://github.com/app-vitals/shipwright/actions/runs/33811085029

## Closing Checklist

- [ ] Tests land in this PR, at the correct layer (unit / integration / smoke / e2e / content) — no "tests later" — N/A: single CLI-arg string change to a manually-run dev tool, not independently testable behavior; see Verification above
- [x] Verification-command output (or a CI run link) is pasted into the PR body above, before merge
- [x] Conventional Commit title (`feat:`, `fix:`, `docs:`, `refactor:`, `test:`, `chore:`, `ci:`)
- [x] No secrets, tokens, client names, or proprietary content
- [x] CI is green (lint, typecheck, tests)

### DIFF
--- a/scripts/hitl.ts
+++ b/scripts/hitl.ts
@@ -1200,7 +1200,7 @@ async function runLoop(): Promise<void> {
         "claude",
         command,
         "--permission-mode",
-        "acceptEdits",
+        "auto",
         "--allowedTools",
         ...HITL_ALLOWED_TOOLS,
       ],


dodizzle (commented):
Verdict: COMMENT — This alters the hitl dev loop's approval-prompt boundary (--permission-mode acceptEdits → auto) with no PR description or verification for what the mode change actually does; the surrounding code's own comment says Bash/Agent are deliberately kept behind that prompt. Small, low-blast-radius change (local dev tool only, production agent path untouched) — worth a quick clarifying comment before merge, not a block.

scripts/hitl.ts
```diff
@@ -1200,7 +1200,7 @@ async function runLoop(): Promise<void> {
         "claude",
         command,
         "--permission-mode",
-        "acceptEdits",
+        "auto",
```
dodizzle, line 1203: This switches `--permission-mode` from `acceptEdits` to `auto` for the interactively-spawned (`stdin: "inherit"`) claude session in the hitl dev loop. The comment above `HITL_ALLOWED_TOOLS` (scripts/hitl.ts:77-79) says Bash/Agent are deliberately excluded from the allowlist to keep them "behind the approval prompt" — permission-mode is what controls whether that prompt fires for out-of-allowlist tool calls. The PR has no description/rationale, and no doc or test in this repo defines what `auto` changes relative to `acceptEdits`, so it isn't verifiable here whether this preserves that safety boundary.

Could you add a short note on what `auto` changes vs `acceptEdits` and confirm it still gates Bash/Agent-equivalent actions behind an approval prompt?

dmcaulay:
Good catch — added rationale to the PR description, and here's the direct answer.

**Why `auto`:** it's now Claude Code's default permission mode. The `hitl` loop already has a human operator attached to this session (`stdin: "inherit"`, run via `/shipwright:hitl`), and that operator is the backstop for anything `auto` mode declines to auto-approve — so this just moves the loop from pinning the old default to tracking the current one.

**Does it preserve the Bash/Agent boundary?** `HITL_ALLOWED_TOOLS` is unchanged by this PR — Bash and Agent are still off the allowlist, per the comment. But I looked into what "off the allowlist" actually gates you into under each mode, since that's the real question:

- Under `acceptEdits`, an out-of-allowlist Bash call always stops and prompts the human before running.
- Under `auto`, an out-of-allowlist Bash/Agent call instead goes to Claude Code's background classifier. If the classifier judges it safe, it runs with **no human prompt at all**; the human is only pulled back in when the classifier blocks the action (notification + retry-with-manual-approval), or after repeated blocks (3 in a row / 20 total) trip a fallback to interactive prompting.

So this isn't a no-op for that boundary — it's a real shift from "every out-of-allowlist call prompts a human" to "every out-of-allowlist call is classifier-reviewed, human is the backstop for what gets flagged." That's consistent with the stated rationale (human-in-the-loop as the unblock mechanism), but it does reduce per-action human visibility, which is worth having explicit rather than assuming unchanged. Given this is the local `hitl` dev tool only (production agent path in `agent/src/claude.ts` is untouched and still pins `acceptEdits`), I think that's an acceptable tradeoff, but flagging it plainly per your ask rather than glossing over it.

Full detail is in the updated PR description.

dodizzle (approved):
Verdict: APPROVE — Prior clarification request (what `auto` permission mode changes vs `acceptEdits`, and whether it still gates Bash/Agent-equivalent calls) was substantively answered in the PR description and a follow-up comment: `auto` routes non-allowlisted tool calls through Claude Code's background classifier instead of an unconditional prompt, escalating to the human on a block. Confirmed `agent/src/claude.ts` (production, non-interactive path) is untouched and still pins `acceptEdits`. Given this tool only ever runs with a human physically attached (`stdin: "inherit"`), that's a reasonable posture for a local dev tool. No new issues found.

PR #3149, opened 2026-09-04
fix: pass recipient_team_id/recipient_user_id on non-DM Thinking Steps streams (STS-2.1)

dmcaulay:
## Summary
- Pass `recipient_team_id`/`recipient_user_id` on `chat.startStream()` for non-DM (channel/group) threads — Slack's own SDK types document both as "required when starting a streaming conversation outside of a DM." Their absence was causing 100% of channel-thread Thinking Steps streams to fail with `missing_recipient_team_id`, while DMs (which don't need them) worked fine.
- `finish()` now falls back to a direct `agents.sessions.setStatus` call whenever `startStream()` never obtained a `ts`, instead of calling `chat.stopStream` (guaranteed to fail with no valid `ts`) — so a stream-open failure can never permanently strand Slack's "is working" indicator again, regardless of the specific error.
- Changed the terminal task_update card's title from `"Done"` to `""` — the "Done" text was getting prepended to the response preview in Slack's UI and adding confusion. `SILENT_COMPLETE_TITLE` (`"Ack"`) is unchanged.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Non-DM `chat.startStream` calls include valid `recipient_team_id`/`recipient_user_id`, no more `missing_recipient_team_id` | MET |
| 2 | DM `chat.startStream` calls omit both fields (unchanged behavior) | MET |
| 3 | `finish()` still clears session status via `agents.sessions.setStatus` when `startStream` fails | MET |
| 4 | Non-silent terminal card title is empty instead of `"Done"`; `"Ack"` unchanged | MET |
| 5 | New unit/integration tests cover all of the above | MET |
| 6 | No regression to existing slack-progress/slack tests | MET |

## Test Plan
- New unit tests in `agent/src/slack-progress.unit.test.ts`: non-DM streams include recipient fields, DM streams omit them even when supplied, `finish()` fallback to `agents.sessions.setStatus` when `startStream` never opens (including no-`chat`-namespace and no-`agents`-namespace cases), empty completion title.
- New integration tests in `agent/src/slack.integration.test.ts`: mention handler includes recipient fields; message handler DM path omits them; fallback to `agents.sessions.setStatus` when `chat.startStream` fails outright.
- `bunx biome lint`/`check` clean on all changed files.
- `bun run --filter='*' --sequential typecheck` passes for every package.
- Full `bun test`: 7595 pass / 1 fail (pre-existing, unrelated — `claude.unit.test.ts`'s `extraAllowedTools`, reproduced identically with this diff stashed out) / 468 skip.

Root-caused live from a running agent's own Slack event logs (repeated `missing_recipient_team_id` since STS-1.1 shipped, confirmed channel-thread-only — DMs unaffected).


dodizzle (approved):
Verdict: APPROVE — Correct root-cause fix for the missing_recipient_team_id channel-thread failure, with a solid defense-in-depth fallback and full test coverage against every acceptance criterion. No blocking concerns.

PR #3182, opened 2026-09-04
feat: add weekly-cron collapsing to Queue & Activity

dodizzle:
## Summary
- Group non-`shipwright-loop` cron activity in the Queue & Activity page's "Past" section under a collapsed `<details>` disclosure by default (one group per cron), reusing the `.more-filters` disclosure styling already added by AXR-1.1
- `shipwright-loop` runs (and any legacy run with no cron attribution) stay in the primary, always-visible table
- Added a pure, exported `partitionCronsForActivityDisplay` helper that classifies each cron as visible vs. collapsed, independently unit-tested against fixture `AgentCronJob` records with varying schedules
- Deduped the repeated 10-column table header markup into a shared `CRON_RUN_TABLE_HEAD` constant used by both the primary table and each collapsed group's table

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Weekly (non-shipwright-loop) crons are grouped under a collapsed disclosure by default, expandable on click | MET |
| 2 | shipwright-loop cron activity remains visible by default, not collapsed | MET |
| 3 | Test decision: unit test covering the grouping logic given a fixture set of AgentCronJob records with varying schedules | MET |

## Test Plan
- `bun test admin/src/admin-ui-pages.unit.test.ts` — 688/688 passing (12 new tests: 4 direct `partitionCronsForActivityDisplay` unit tests + 8 grouping-behavior tests on `renderQueueActivityPage`)
- `task typecheck` — clean
- `task lint` — clean
- `task ci`'s remaining 32 failures (metrics/env, metrics/errors, task-store/prs) are pre-existing on `main`, unrelated to this diff — reproduced independently on a clean `main` checkout


dmcaulay (approved):
Verdict: APPROVE — Clean implementation, all three acceptance criteria verifiably met, no blocking issues found; dependency risk: merge — bun.lock only bumps this repo's own internal workspace package versions (rebase-onto-main artifact), no external dependency change.

dmcaulay (commented):
Verdict: COMMENT — Grouping/partitioning implementation is solid with strong test coverage, but filtering to a specific non-shipwright-loop cron (via the filter dropdown or the Cron column's own link) makes the primary table falsely report "No runs match the selected filters" while the matching runs exist, collapsed inside a <details> block; dependency risk: merge — bun.lock only bumps this repo's own internal workspace package versions (rebase-onto-main artifact), no external dependency change.

admin/src/admin-ui-pages.ts
```diff
@@ -3812,10 +3862,65 @@ export function renderQueueActivityPage(opts: {
...
+      const group = collapsedGroups.get(r.cron.id);
+      if (group) {
+        group.push(r);
+      }
+    }
+  }
+
   const bodyRows =
-    runs.length === 0
+    visibleRuns.length === 0
```
dmcaulay, line 3830: Filtering to a specific non-`shipwright-loop` cron (via the filter dropdown, or via the "Cron" column's own `?cronId=` link that `row()` renders for every run) makes this always report "No runs match the selected filters." even when matching runs exist — they all land in `collapsedGroups` instead of `visibleRuns` once `filters.cronId` narrows the server-side query to a single non-loop cron. Verified by calling `renderQueueActivityPage` directly with `filters.cronId` set to a non-loop cron and `runs` containing only that cron's data: primary table renders the empty-state while a `<details>` block holds the real match.

Suggestion: when `filters.cronId` is set, either skip the visible/collapsed split and render the filtered runs directly in the primary table, or auto-`open` the matching `<details>` block.

dodizzle:
Reviewed the finding above and did not implement additional changes — premise did not hold against the current HEAD:

- Cron-visibility split falsely reporting "No runs match the selected filters" when filtering to a single non-shipwright-loop cron: this is already fixed by commit 0104a28d on this branch. `renderQueueActivityPage` in `admin/src/admin-ui-pages.ts` now computes `const isCronFiltered = Boolean(filters.cronId)` (line 3804) and skips the visible/collapsed split entirely whenever it's true — `visibleRuns` becomes the full (already server-filtered) `runs` array instead of being partitioned into `visibleCronIds`/`collapsedCronIds` (lines 3806-3809), and the `collapsedGroups` population is skipped too (line 3814). This exactly matches the suggestion in the original inline thread ("skip the visible/collapsed split ... when filters.cronId is set").
- Test coverage: `admin/src/admin-ui-pages.unit.test.ts` has a test named "filtering to a non-loop cron renders its runs in the primary table, not a collapsed block" (line 5150) that calls `renderQueueActivityPage` directly with `filters.cronId` set to a non-loop cron (`OTHER_CRON`) and asserts the run appears in the primary `<tbody>`, that `"No runs match the selected filters."` is NOT present, and that no `<details class="more-filters">` block is rendered. I ran the full unit suite (`bun test admin/src/admin-ui-pages.unit.test.ts`) and confirmed all 696 tests pass, including this one.
- The inline thread on this same finding (`admin/src/admin-ui-pages.ts:3900`, thread `PRRT_kwDOSy4Ihs6fS9UH`) is already `isResolved: true` — confirmed via `gh api graphql`, no action needed there.
- bun.lock: confirmed via `git diff origin/main...HEAD -- bun.lock` that the only changes are internal workspace package version bumps (1.233.0 → 1.245.0) picked up from the main-merge, no external dependency changes — matches the reviewer's own "merge" risk classification.

PR #3192, opened 2026-09-04
feat: add slide-over drawer to task board cards

dodizzle:
## Summary
- Adds a slide-over detail drawer to Tasks board cards, reusing the existing `.chat-drawer-toggle` checkbox-driven off-canvas pattern (CFB-3.2) instead of a new JS library/framework
- Clicking a card now opens its own drawer (title, id, status, repo, session, agent, description, joined PR state/reviewState/blocked) in place, rather than navigating away; title/id links still deep-link to the full detail page
- Closes via scrim click, an in-drawer close control, or Escape — all through the same checkbox toggle. Opening a new card's drawer also closes any other drawer left open, since multiple `.task-drawer` panels share the same fixed position

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Clicking a board card opens a slide-over drawer showing full task + joined PR detail, using the checkbox-driven pattern (no new JS framework/library) | MET |
| 2 | The drawer closes via the same checkbox-toggle mechanism used by the existing chat-drawer pattern | MET |
| 3 | Unit test asserting the drawer markup (checkbox id, labelled toggle, drawer content section) renders per card | MET |

## Test Plan
- 9 new unit tests in `admin/src/admin-ui-pages.unit.test.ts` covering checkbox id/DOM order, scrim label binding, drawer content (title/id/PR state), per-card unique ids, readOnly omission, and the close-other-open-drawers fix
- `NODE_ENV=test bun test admin/src/admin-ui-pages.unit.test.ts` — 719 pass, 0 fail
- `bunx biome lint` — clean

### DIFF
--- a/admin/src/admin-ui-pages.ts
+++ b/admin/src/admin-ui-pages.ts
@@ -1994,6 +1994,25 @@ const TASKS_PAGE_EXTRA_STYLES = `
     .badge-hitl { background:#fff7ed;color:#c2410c;border:1px solid #fed7aa; }
     .badge-dep { background:#fefce8;color:#a16207;border:1px solid #fde047; }
     .alert-warning { background:#fefce8;color:#854d0e;border:1px solid #fde047;border-radius:6px;padding:10px 14px;margin-bottom:16px;font-size:13px; }
+    /* ─── Task board card slide-over drawer (AXR-1.4) ────────────────────── */
+    .task-drawer-toggle { position:absolute;opacity:0;width:1px;height:1px;pointer-events:none }
+    .task-drawer-scrim {
+      display:none;position:fixed;inset:0;z-index:55;
+      background:rgba(0,0,0,0.4);cursor:pointer;
+    }
+    .task-drawer-toggle:checked ~ .task-drawer-scrim { display:block }
+    .task-drawer {
+      position:fixed;top:0;right:0;bottom:0;z-index:60;
+      width:100%;max-width:400px;min-width:300px;
+      background:#fff;border-left:1px solid #e8e8ee;
+      overflow-y:auto;
+      transform:translateX(100%);
+      transition:transform 0.2s ease;
+    }
+    .task-drawer-toggle:checked ~ .task-drawer { transform:translateX(0) }
+    @media (prefers-reduced-motion: reduce) {
+      .task-drawer { transition:none }
+    }
   `;
 
 export function renderTasksPage(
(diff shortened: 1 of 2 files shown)


dmcaulay (commented):
Verdict: COMMENT — Clean, well-tested implementation of the checkbox-driven drawer pattern; one fixable accessibility gap: the hidden per-card toggle checkbox lacks tabindex="-1", so a keyboard user can focus and check it directly (bypassing the close-other-drawers click handler) and end up with two stacked drawers.

admin/src/admin-ui-pages.ts
```diff
@@ -2537,6 +2556,47 @@ function renderTasksBoard(args: {
...
+            <div style="font-size:12px;color:#6b7280;margin-bottom:4px"><strong>State:</strong> ${escapeHtml(joinedPr.state)}</div>
+            <div style="font-size:12px;color:#6b7280;margin-bottom:4px"><strong>Review State:</strong> ${escapeHtml(joinedPr.reviewState)}</div>
+            ${joinedPr.blocked ? `<div style="font-size:12px;color:#dc2626"><strong>Blocked:</strong> ${escapeHtml(joinedPr.blockedReason ?? "Yes")}</div>` : ""}
+          </div>
+        ` : ""}
+      </div>
+    `;
+
+    return `<div class="task-card-wrap">
+      <input type="checkbox" id="task-drawer-toggle-${escapeHtml(t.id)}" class="task-drawer-toggle" aria-hidden="true">
```
dmcaulay: The `.task-drawer-toggle` checkbox is visually hidden (`opacity:0`, `pointer-events:none`, 1x1px) plus `aria-hidden="true"`, but none of those remove it from the keyboard tab order — `aria-hidden` only affects the accessibility tree, not focusability. Since there's one such checkbox per board card (unlike the single chat-drawer this pattern was copied from), a keyboard user can Tab to a card's hidden checkbox directly and press Space to check it, bypassing the card's `click` handler — the only place that closes any other already-open drawer. Two simultaneously-checked toggles would render two `.task-drawer` panels, both `position:fixed` at identical coordinates, stacked.

Suggest adding `tabindex="-1"` here to remove it from tab order (mouse/label-driven toggling is unaffected).

dmcaulay (approved):
Verdict: APPROVE — Prior finding (hidden drawer-toggle checkbox missing tabindex="-1") is fixed and test-verified; no new issues.

PR #3219, opened 2026-09-05
docs(planning): add dependency-patch-flow plan

dmcaulay:
## Summary
- Plans a fix for `/shipwright:patch` having no handling for dependency-risk findings — review's Step 5.8 (DBR-3.2) detects and reports these (e.g. vitals-os#5121's real js-yaml v5 transitive-dependency clash), but patch has zero dependency-specific logic, so these findings just sit unaddressed and skip forever.
- Queues `DBP-1.1` (new `references/dependency-patch.md` remediation reference — reproduce-before-fixing protocol + bounded remediation catalog) and `DBP-1.2` (wire detection + the reference into `patch.md`'s existing findings pipeline).
- Detection is designed to not assume review and patch share an agent/session — it parses the dependency-risk clause already present in review bodies patch's Step 3a fetches, falling back to independently re-deriving it via the same script/reference review uses.

## Test plan
- [ ] `dev-task` lands `DBP-1.1` with `references/dependency-patch.content.test.ts`
- [ ] `dev-task` lands `DBP-1.2` with extended `commands/patch.content.test.ts` assertions

### DIFF
--- a/planning/dependency-patch-flow/PLAN.md
+++ b/planning/dependency-patch-flow/PLAN.md
@@ -1,0 +1,181 @@
+# Plan Session: dependency-patch-flow
+
+Repo: `app-vitals/shipwright`
+
+## Background
+
+Triggered by investigating `vitals-os` PR #5121 (a Renovate js-yaml v4→v5 bump): the
+Shipwright review pipeline correctly detected and reported a real, specific
+dependency-risk finding — a transitive-dependency clash breaking a named build script,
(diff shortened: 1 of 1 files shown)


dodizzle (approved):
Verdict: APPROVE — Plan document is internally consistent; every checkable claim about patch.md/review.md step numbers, script names, and reference-file shapes was spot-checked against the current codebase and confirmed accurate. No factual or logical issues found.

PR #3221, opened 2026-09-05
feat: drop Claimed/Done from default board query, raise active-column cap to 200

dodizzle:
## Summary
- Scope the admin Tasks board's default (`/admin/tasks`, no `?view=table`) outbound task-store query to exclude fully-closed statuses (`state=open`) and raise its limit from 50 to 200, so fast-churning Done/Claimed tasks stop crowding out older Queued/In Progress/Blocked-HITL tasks in the shared top-N-by-recency window.
- Drop Claimed and Done from `TASK_BOARD_COLUMNS`, the default board's rendered column set — both remain fully reachable via `?view=table` filters, which are completely unaffected by this change.
- `bucketTaskColumn()` itself is untouched (still a valid 5-value bucketing function); only the default board's rendered/queried subset shrinks to Queued / In Progress / Blocked-HITL.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `/admin/tasks` (board, default view) no longer renders Claimed or Done columns | MET |
| 2 | `/admin/tasks?view=table&state=...` / `?status=...` continue to work exactly as before | MET |
| 3 | Board's outbound query derived from `bucketTaskColumn`'s active-column definitions, raised to 200 | MET |
| 4 | Existing board filters (repo/org/session/source/agent/hitl) still apply | MET |
| 5 | Board-column tests updated for 3-column output; query-param test added; table-view smoke test added; stale 5-column/limit=50 assertions retired | MET |

## Test Plan
- `NODE_ENV=test bun test admin/src/admin-ui-pages.unit.test.ts admin/src/admin-ui.smoke.test.ts` — 1042/1042 pass
- Full `admin/` suite passes except 2 pre-existing failures (`task-store/src/routes/prs.ts` `validateRepo` null-repos bug) that reproduce identically on unmodified `main`
- `bunx biome lint` and `tsc --noEmit` clean on all 4 touched files


admin/src/admin-ui-pages.ts
```diff
@@ -3094,8 +3100,6 @@ export const TASK_BOARD_COLUMNS: { key: TaskBoardColumn; label: string }[] = [
   { key: "queued", label: "Queued" },
```
dmcaulay, line 3132: `renderTasksBoard()`'s "More filters" panel still lists closed statuses (`done`, `merged`, `deploying`, `deployed`, `cancelled`) in its Status dropdown (unchanged by this PR). Selecting one of these on the board hits `admin-ui.ts`'s `if (status) { params.set("status", status) }` branch, which bypasses the new `state=open` fallback — task-store correctly returns matching tasks, but `bucketTaskColumn()` maps every closed status to the `"done"` bucket, and `TASK_BOARD_COLUMNS` (this diff) no longer includes `"done"`. Since the column loop only iterates `TASK_BOARD_COLUMNS`, those tasks never render anywhere — the board silently shows "No tasks" in every column. On `main`, `TASK_BOARD_COLUMNS` included `done`/`claimed`, so this filter worked before; this diff introduces the regression. Consider restricting the board's Status dropdown to non-closed statuses, or hiding it on the board view in favor of `?view=table`.

dmcaulay (commented):
Verdict: COMMENT — board's own Status filter can select a closed status (done/merged/deployed/deploying/cancelled) that's no longer a rendered board column, silently producing an empty board even though matching tasks exist. Everything else (state=open semantics, table-view isolation, test coverage for the stated acceptance criteria) checks out.

dmcaulay (approved):
Verdict: APPROVE — the prior finding (board's Status filter could select a closed status and silently empty the board) is resolved: the board's statusOptions now filters out SESSION_CLOSED_STATUSES, correctly scoped so ?view=table is untouched, and covered by two new targeted unit tests. No new issues in this commit.

PR #3227, opened 2026-09-05
fix(review): gh api --jq doesn't support --arg, breaks Step 14 precheck

dodizzle:
## Summary
- `/shipwright:review`'s Step 14 precheck built a `gh api graphql` call as `-f query='...' --arg currentUser "$CURRENT_USER" --jq '...'`. `gh api` has no `--arg` flag — that's a `jq`-binary-only concept for injecting external variables into a filter — so the call fails immediately with `unknown flag: --arg`.
- Confirmed via `gh api --help` (2.100.0): valid flags are `-f/--raw-field`, `-F/--field`, `-q/--jq`; no `--arg`.
- Found via a learn-dream mining pass across the 2026-09-03 20:00 UTC → 2026-09-04 UTC transcript window: the identical failure recurred in 10+ independent `/shipwright:review` sessions (each self-corrected mid-session by switching to a jq pipe, wasting a round-trip every time since each cron-triggered session starts fresh with no memory of the prior fix).
- Fix: drop `--arg`/`--jq` from the `gh api` call and pipe its raw JSON response into the real `jq` binary, which does support `--arg`. Verified the filter logic is unchanged and produces the same `{headRefOid, terminal}` shape against a representative fixture.

## Test plan
- [x] Verified `gh api --help` confirms no `--arg` flag exists
- [x] Ran the corrected `jq --arg currentUser ... '...'` filter against a representative GraphQL response fixture locally — produces the expected `{headRefOid, terminal}` output
- [ ] Next live `/shipwright:review` invocation should complete Step 14's precheck without the `unknown flag: --arg` error

### DIFF
--- a/plugins/shipwright/commands/review.md
+++ b/plugins/shipwright/commands/review.md
@@ -1396,10 +1396,15 @@ precheck=$(gh api graphql -f query='
       }
     }
   }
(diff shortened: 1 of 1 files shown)


dmcaulay (approved):
Verdict: APPROVE — Small, well-verified fix: gh api's --jq flag doesn't support --arg, so Step 14's precheck always errored; piping into standalone jq instead (filter unchanged) fixes it, and no sibling call sites share the bug.

PR #3222, opened 2026-09-05
feat: add dependency-patch remediation reference file

dodizzle:
## Summary
- Adds `plugins/shipwright/references/dependency-patch.md`, a bounded reference for remediating a dependency-bump PR that `dependency-risk-analysis.md` has already flagged as `review`/`hold` — it defines the reproduce-before-fixing protocol, a two-strategy bounded remediation catalog (transitive-dependency override/resolution pin; removed/renamed first-party API call-site update), the re-verify step, and an explicit no-safe-strategy exit to "leave as hold".
- Pure reference content — no caller wiring in this task.
- Adds `plugins/shipwright/references/dependency-patch.content.test.ts` (content layer), mirroring `dependency-risk-analysis.content.test.ts`'s assertion style.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `dependency-patch.md` exists, documents Inputs/Output/protocol sections matching `dependency-risk-analysis.md`'s structure and cross-references it directly | MET |
| 2 | Reproduce-before-fixing protocol requires running the reasoning's named verification command before any fix, and re-running it after, before claiming fixed | MET |
| 3 | Remediation catalog is explicitly bounded (exactly two in-scope strategies); anything outside, or unreproducible/unverifiable, exits to "leave as hold" | MET |
| 4 | `dependency-patch.content.test.ts` added (content layer) asserting the required sections/language, mirroring the existing content test's style | MET |

## Test Plan
- `bun test plugins/shipwright/references/dependency-patch.content.test.ts` — 17/17 pass
- `bunx biome lint` on both new files — clean
- `task ci` — lint, check-strings, typecheck, check-config-docs, check-version-sync all pass. `test:coverage` fails at the repo-aggregate level due to 31 pre-existing failures in `metrics/src/lib/env.unit.test.ts`, `metrics/src/lib/errors.unit.test.ts`, and `task-store/src/routes/prs.openapi.smoke.test.ts` — confirmed these reproduce identically on an untouched worktree of `main` at the same commit (no diff applied), and GitHub's own CI is green on the latest merged main commit. Unrelated to this change (this PR only touches `plugins/shipwright/references/`).
- [x] Pre-commit checks passing

Generated with [Claude Code](https://claude.com/claude-code)

PR #3235, opened 2026-09-05
feat: wire dependency-risk detection and remediation into patch.md

dodizzle:
## Summary
- Adds Step 3a.5 to `patch.md`: detects a dependency-risk finding on a PR either by parsing a review body's `## Dependency Risk Analysis` section (already fetched by Step 3a) or, when absent, independently re-deriving it via `resolve-dependency-watched-paths.ts` + `references/dependency-risk-analysis.md` — mirroring `review.md`'s Step 5.8 without assuming a worktree exists yet.
- Wires the result into Step 5b's fix-subagent prompt: when a dependency-risk finding recommends `review`/`hold`, injects `references/dependency-patch.md`'s reproduce → catalog-fix → re-verify protocol as an additive section ahead of (not replacing) the existing `[A.5]` ACCEPT/MODIFY/REJECT classification.
- Extends `patch.content.test.ts` with coverage for the new step and prompt section.

Implements task `DBP-1.2` (depends on `DBP-1.1`, already merged — PR #3222).

## Test plan
- [x] `bun test plugins/shipwright/commands/patch.content.test.ts` — 177 pass
- [x] `task lint`
- [x] `task typecheck`
- [x] `task check-strings`
- [x] `task test` — pre-existing failures only (env-var assertion tests conflicting with real env vars in this shell; unrelated to this change)

### DIFF
--- a/plugins/shipwright/commands/patch.md
+++ b/plugins/shipwright/commands/patch.md
@@ -390,24 +390,16 @@ Step 5.
 
 ### Step 3a.5: Dependency-Risk Detection (DBP-1.2)
 
-For each PR, determine whether it has a dependency-risk finding, independently of whether
-review and patch share an agent/session — any agent can claim either phase, and a review's
-dependency-risk analysis is only ever persisted as text inside the GitHub-posted review
-body, never in a shared in-memory or task-store field. Store the result as
-`DEPENDENCY_RISK_FINDING` (a nullable `{recommendation, flags, reasoning}` shape) for use by
-Step 5b. A `"hold"` or `"review"` recommendation additionally routes the PR into **List A**
-directly — see step 3 below — because it has no ordinary review finding for Step 3a's
-criteria to catch on its own.
+For each PR, determine whether it has a dependency-risk finding, deriving it independently
+from the PR itself rather than reading it back off a prior review — any agent can claim either phase,
+and no review artifact that reaches GitHub carries the full `{recommendation, flags,
+reasoning}` shape forward to a later patch run (see "Why this is re-derived" below). Store
+the result as `DEPENDENCY_RISK_FINDING` (a nullable `{recommendation, flags, reasoning}`
+shape) for use by Step 5b. A `"hold"` or `"review"` recommendation additionally routes the
+PR into **List A** directly — see step 2 below — because it has no ordinary review finding
+for Step 3a's criteria to catch on its own.
 
-1. **Scan the review bodies Step 3a already fetched.** For each review in `reviews.nodes[]`
-   (already in memory from Step 3a's GraphQL query — do not re-fetch), check its `body` for
-   a `## Dependency Risk Analysis` section — the exact heading `review.md`'s Step 9 template
-   emits. When found, parse `{recommendation, flags, reasoning}` from that section's
-   `**Recommendation**:`, `**Flags**:`, and `**Reasoning**:` lines. Set
-   `DEPENDENCY_RISK_FINDING` from this parse and skip step 2 below — a review already did
-   this analysis, so there's nothing to re-derive.
-
-2. **Independent fallback, when no fetched review body contains the clause.** Mirror
+1. **Derive the finding from the PR's own diff.** Mirror
    `review.md`'s Step 5.8 exactly, substituting `gh` calls for that step's worktree-relative
    file reads — Step 3 runs before any worktree exists (worktree setup happens later, in
    Step 4a/5a, scoped only to PRs already classified into List C/A):
@@ -444,8 +436,8 @@ criteria to catch on its own.
    d. **When not triggered** (no changed file matches): `DEPENDENCY_RISK_FINDING` stays
       unset for this PR.
 
-3. **Route a `hold`/`review` recommendation into List A.** When `DEPENDENCY_RISK_FINDING`
-   is set (from step 1 or step 2 above) and its `recommendation` is `"hold"` or `"review"`,
+2. **Route a `hold`/`review` recommendation into List A.** When `DEPENDENCY_RISK_FINDING`
+   is set (from step 1 above) and its `recommendation` is `"hold"` or `"review"`,
    add this PR to **List A** — regardless of whether Step 3a's own criteria (unresolved
    inline threads, a non-excluded COMMENTED/CHANGES_REQUESTED review body) independently
    found a finding for this PR. A `"merge"` recommendation has nothing to remediate and does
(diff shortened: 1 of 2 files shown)


plugins/shipwright/commands/patch.md
```diff
@@ -388,6 +388,61 @@ If a PR has unaddressed findings, add it to **List A**. Store the unresolved thr
...
 Step 5.
 
+### Step 3a.5: Dependency-Risk Detection (DBP-1.2)
+
+For each PR, determine whether it has a dependency-risk finding, independently of whether
+review and patch share an agent/session — any agent can claim either phase, and a review's
+dependency-risk analysis is only ever persisted as text inside the GitHub-posted review
+body, never in a shared in-memory or task-store field. Store the result as
+`DEPENDENCY_RISK_FINDING` (a nullable `{recommendation, flags, reasoning}` shape) for use by
+Step 5b — this step never changes List A/C/D membership on its own; Step 3a's criteria above
```
dmcaulay: Step 3a.5 explicitly never changes List A/C/D membership, but Step 5b — the only consumer of `DEPENDENCY_RISK_FINDING` — only ever runs "For each qualifying PR in List A" (Step 5, patch.md:744). A dependency-bump PR with a `hold`/`review` recommendation and no other ordinary findings gets a clean `Verdict: APPROVE` from review.md (the dependency-risk clause is appended on the same line as `Verdict: ...` and explicitly does not feed into `compute-review-verdict.ts`'s event computation — review.md:1118-1125), so Step 3a's clean-APPROVE exclusion keeps it out of List A. The fully-independent fallback (option 2, meant to work with no prior review session) has the same problem — detecting a finding doesn't add the PR to List A either. Net effect: the new remediation protocol only fires when the PR also happens to have an unrelated, unaddressed ordinary review finding — not the dependency-bump-only case this feature targets. Consider giving a `hold`/`review` finding its own path into Step 5's loop, or having review.md's verdict computation factor in the recommendation so these PRs land in List A the normal way.

dmcaulay (commented):
Verdict: COMMENT — DEPENDENCY_RISK_FINDING is computed in Step 3a.5 but never routes a PR into List A, so Step 5b's new remediation protocol is unreachable for a dependency-bump-only PR (the primary scenario) — see inline comment.

plugins/shipwright/commands/patch.md
```diff
@@ -388,6 +388,80 @@ If a PR has unaddressed findings, add it to **List A**. Store the unresolved thr
...
+For each PR, determine whether it has a dependency-risk finding, independently of whether
+review and patch share an agent/session — any agent can claim either phase, and a review's
+dependency-risk analysis is only ever persisted as text inside the GitHub-posted review
+body, never in a shared in-memory or task-store field. Store the result as
+`DEPENDENCY_RISK_FINDING` (a nullable `{recommendation, flags, reasoning}` shape) for use by
+Step 5b. A `"hold"` or `"review"` recommendation additionally routes the PR into **List A**
+directly — see step 3 below — because it has no ordinary review finding for Step 3a's
+criteria to catch on its own.
+
+1. **Scan the review bodies Step 3a already fetched.** For each review in `reviews.nodes[]`
```
dmcaulay: Step 3a.5 step 1 checks `reviews.nodes[].body` (the actual GitHub-posted review, fetched via Step 3a's GraphQL query) for a `## Dependency Risk Analysis` heading and parses `**Recommendation**:`/`**Flags**:`/`**Reasoning**:` lines from it. But that heading/format only ever appears in `review.md`'s Step 9 output — the local file `state/reviews/PR_REVIEW_{pr}.md` (review.md:782), which is never posted to GitHub. What actually gets posted (review.md:1118-1126, folded into `pr_review_{pr}.json`'s `body`) is a condensed one-liner appended after the `Verdict: ...` label, e.g. `"Verdict: COMMENT — {summary}; dependency risk: hold — {reasoning}"` — no `## Dependency Risk Analysis` heading, and `flags` isn't encoded there at all. So this "primary" fast path can never match real data: every dependency-risk-eligible PR always falls through to the costlier independent re-derivation in step 2 (contradicting "a review already did this analysis, so there's nothing to re-derive"), and `flags` is unrecoverable via this path entirely.

Either parse the condensed one-line clause actually posted in the Verdict body (accepting `flags` won't be recoverable this way), or drop step 1 and always use the independent fallback.

plugins/shipwright/commands/patch.md
```diff
@@ -388,6 +388,80 @@ If a PR has unaddressed findings, add it to **List A**. Store the unresolved thr
...
+      definitions rather than restating them here.
+   c. **When triggered** (at least one changed file matches): apply
+      `references/dependency-risk-analysis.md`'s heuristics — using `gh pr diff {pr} --repo
+      {org}/{repo}` for the diff, the PR's `body` and `PR_AUTHOR` (Step 2), and
+      `gh pr view {pr} --repo {org}/{repo} --json labels` for labels — to produce
+      `{recommendation, flags, reasoning}`. Set `DEPENDENCY_RISK_FINDING` from the result.
+   d. **When not triggered** (no changed file matches): `DEPENDENCY_RISK_FINDING` stays
+      unset for this PR.
+
+3. **Route a `hold`/`review` recommendation into List A.** When `DEPENDENCY_RISK_FINDING`
```
dmcaulay: This unconditionally routes a PR into List A whenever `DEPENDENCY_RISK_FINDING.recommendation` is `hold`/`review`, with no exclusion analogous to Step 3a's clean-APPROVE/reply-addressed/superseded-self-review rules or Step 5a.7's second-round-disagreement escalation. When the fix subagent hits the no-safe-strategy exit in `references/dependency-patch.md`'s protocol (outcome: held) and classifies REJECT in `[A.5]`, the next patch cycle recomputes the same recommendation from the same diff/heuristics (the independent-fallback path is synthesized fresh from the diff every time, with no dependency on prior review/reply state) and re-adds the PR to List A again — repeating the reproduce→no-safe-strategy→REJECT cycle and reposting a rebuttal comment indefinitely. This contradicts `plugins/shipwright/CLAUDE.md`'s Independence Principle #6 ("Skills Are Idempotent").

Consider an exclusion mirroring Step 3a's reply-addressed rule — e.g. skip re-routing to List A when a PR-author/agent comment already reports the finding held/rejected and no new commits have landed since.

dmcaulay (commented):
Verdict: COMMENT — the List A routing gap from the prior review is fixed, but Step 3a.5 introduces a dead-code fast path (its review-body scan can never match what review.md actually posts to GitHub) and lacks an idempotency exit for a repeatedly-held dependency-risk finding.

dodizzle:
Reviewed the finding(s) above and did not implement changes — premise did not hold:

- **List A routing gap (04:27:43 review body)**: already fixed by a subsequent commit — the current Step 3a.5 already routes `hold`/`review` recommendations into List A directly, so Step 5b's remediation protocol is reachable for a dependency-bump-only PR. Confirmed by the 04:50:55 follow-up review, which states the routing gap is fixed.

Both of the follow-up review's inline findings were ACCEPTED and fixed in `bbbc48b`:

- **Dead-code fast path (patch.md:402)**: dropped Step 3a.5's review-body scan entirely rather than reparsing the condensed clause. Verified against `review.md`: the `**Recommendation**`/`**Flags**`/`**Reasoning**` block only ever lands under the `## Dependency Risk Analysis` heading in the local `state/reviews/PR_REVIEW_{pr}.md` narrative (review.md:782), which is never posted; the posted body carries only the condensed `; dependency risk: {recommendation} — {reasoning}` clause (review.md:1118-1126). Since `references/dependency-patch.md` treats `flags` (`breakingChange` especially) as a real protocol input, parsing that clause would hand Step 5b a permanently `flags`-less finding — so the independent derivation is now the only path, and a new "Why this is re-derived" note documents the reasoning so the fast path doesn't get re-added.

- **Missing idempotency exit (patch.md:447)**: added an already-held exclusion as Step 3a.5 step 3. It reads the PR record's findings ledger for a `disposition: "rejected"` entry under the ref `dependency-risk@{headRefOid}` and skips the List A routing when one exists. Keying the ref on the HEAD SHA makes the exclusion self-expiring (a new commit produces a ref with no ledger entry, so a genuinely changed bump re-routes), mirroring how Step 3a's reply-addressed exclusion stops applying once a newer review post-dates the reply. It fails open — a missing PR record, empty result, or failed lookup routes normally so a first-round remediation is never suppressed — and an excluded PR still reaches List A on Step 3a's own criteria, just without the DEPENDENCY-RISK REMEDIATION PROTOCOL block. Step 5b's protocol block and Step 5c's ledger-write instructions were wired to the same ref so both ends of the exclusion agree on one identifier.

plugins/shipwright/commands/patch.md
```diff
@@ -935,7 +1058,10 @@ curl -s -o /dev/null -X POST \
 
 Dispatch a `general-purpose` subagent via the Agent tool, passing `model: PATCH_MODEL`
 (resolved once in Step 2.1) so the fix subagent runs at the escalated tier, with this
-prompt:
+prompt. When Step 3a.5 set `DEPENDENCY_RISK_FINDING` for this PR with recommendation
```
dmcaulay: Step 3a.5 step 3 promises that when `HELD_DEP_FINDINGS` is non-zero, patch should "omit Step 5b's DEPENDENCY-RISK REMEDIATION PROTOCOL block for it" — with a parenthetical explicitly anticipating the case where the PR stays in List A anyway via "a still-unresolved *ordinary* finding on the same PR." But Step 5b's actual gating condition here (and again in the rendered prompt template at line ~1099) is only "When Step 3a.5 set `DEPENDENCY_RISK_FINDING` for this PR with recommendation `review` or `hold`, include the ... block" — it never references `HELD_DEP_FINDINGS` or any already-held state. `DEPENDENCY_RISK_FINDING` is set once in step 1 regardless of step 3's outcome; step 3 only describes skipping the List A *routing* in step 2, with no mechanism threading its "omit the block" promise through to Step 5b. So for exactly the scenario step 3's own parenthetical anticipates — a PR independently in List A via an ordinary Step 3a finding, plus an already-rejected dependency-risk finding at the same SHA — Step 5b would still include the REMEDIATION PROTOCOL block, re-running reproduce → no-safe-strategy → REJECT and re-posting the same rebuttal comment. Same Independence Principle #6 ("Skills Are Idempotent") violation the already-held exclusion was added to close, just in a case the fix doesn't cover.

Consider having step 3 unset `DEPENDENCY_RISK_FINDING` (or set a separate `ALREADY_HELD` flag) when `HELD_DEP_FINDINGS` is non-zero, and have Step 5b's gate check that instead of the raw derived value.

dmcaulay (commented):
Verdict: COMMENT — both prior findings are fixed for the scenarios they described, but the new already-held exclusion doesn't actually reach Step 5b's gate, leaving a narrower repeat of the same idempotency gap when the PR is independently in List A via an unrelated ordinary finding.

dmcaulay (approved):
Verdict: APPROVE — all three prior findings are now genuinely resolved; the already-held exclusion clears the shared DEPENDENCY_RISK_FINDING variable, so Step 3a.5's List A routing and Step 5b's remediation-block gate can no longer disagree. No new issues found.

PR #3237, opened 2026-09-05
fix: validate GET /tasks status query param, return 400 instead of 500

dodizzle:
## Summary
- `GET /tasks?status=<value>` previously passed the raw query string straight into `taskService.list()` → `prisma.task.findMany({ where: { status } })`. An invalid/non-enum value (e.g. `claimed`, from Sentry issue [7603167547](https://app-vitals-llc.sentry.io/issues/7603167547/)) threw an unhandled `PrismaClientValidationError`, which fell into the generic `onError` 500 branch and got reported to Sentry.
- Added a validation guard in `task-store/src/routes/tasks.ts`'s `GET /tasks` handler: any `?status=` value not in the `TaskStatus` enum (derived from the existing `OPEN_STATUSES`/`CLOSED_STATUSES` exports in `statuses.ts`, avoiding a third hand-written literal list) now throws `BadRequestError`, returning a clean 400 — matching the existing `status is required` precedent already used on `POST /tasks`.
- Omitted `?status=` continues to apply no filter; every valid enum value continues to filter exactly as before.

## Acceptance Criteria

| 1 | Invalid/non-enum `status` returns 400, not 500 | MET | Guard throws `BadRequestError` before reaching `taskService.list()`; covered by smoke tests for `claimed`, `bogus-status`, empty string, and casing mismatch |
| 2 | Valid TaskStatus enum values keep working (200, filtered) | MET | All 10 enum values tested individually, plus a filter-capture assertion confirming the value passes through unchanged |
| 3 | Omitted `status` param keeps working (200, no filter) | MET | Guard only fires when `statusRaw !== undefined`; test asserts no filter is applied when omitted |
| 4 | 400 behavior covered by an automated test | MET | New `task-store/src/tasks.status-validation.smoke.test.ts` (7 tests, correct `*.smoke.test.ts` layer for an HTTP route contract) |

## Test Plan
- `bun test task-store/src/tasks.status-validation.smoke.test.ts` — 7/7 pass
- `bun test task-store/src` — 641 pass, 0 fail (full task-store suite, no regressions)
- `task typecheck` — passes across all workspaces
- `task lint`, `task check-strings` — pass
- Coverage on the changed file: `task-store/src/routes/tasks.ts` at 96.51% lines (target 90%)

### DIFF
--- a/task-store/src/routes/tasks.ts
+++ b/task-store/src/routes/tasks.ts
@@ -40,6 +40,7 @@ import { readJson } from "@shipwright/lib/http";
 import type { TaskStoreAuthEnv } from "../auth.ts";
 import { BadRequestError, ForbiddenError, NotFoundError } from "../errors.ts";
 import type { Prisma } from "../index.ts";
+import { CLOSED_STATUSES, OPEN_STATUSES } from "../statuses.ts";
 import {
   BulkInsertBodySchema,
   BulkInsertResponseSchema,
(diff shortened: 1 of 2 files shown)


dmcaulay (approved):
Verdict: APPROVE — Guard runs immediately before the only call site consuming the raw status, BadRequestError correctly maps to HTTP 400, no breaking change for any known caller, and the fix lands with adequate same-PR smoke test coverage.

PR #3239, opened 2026-09-05
fix: add status getter alias to ApiError for Sentry defaultShouldHandleError

dodizzle:
## Summary
- Add a `status` getter alias to `ApiError` in both `task-store/src/errors.ts` and `admin/src/errors.ts`, returning `this.statusCode`.
- `@sentry/hono`'s `sentry()` middleware independently captures any error left on `context.error` after the route's own `onError` has already run, gated by `defaultShouldHandleError` — which reads `error.status`, not `error.statusCode`. Since our `ApiError` only exposed `statusCode`, `error.status` was always `undefined`, so `defaultShouldHandleError` always returned `true` and every already-handled 3xx/4xx `ApiError` got captured and alerted anyway, even when the app's own `onError` deliberately chose not to report it.
- Add `errors.unit.test.ts` in both services asserting `status === statusCode` on the base class and representative subclasses (including a 5xx one, `BadGatewayError`, to confirm 5xx faults — which `onError` DOES intentionally report — are unaffected).

Fixes the root cause behind 8 of this week's regressed Sentry issues (all normal application-level rejections working as designed): VITALS-OS-5G, VITALS-OS-5J, VITALS-OS-E, VITALS-OS-5, VITALS-OS-8, VITALS-OS-A, VITALS-OS-Y, VITALS-OS-J.

## Acceptance Criteria
No formal acceptance criteria were attached to this task; the fix guidance in the task description is the spec:
- [x] `ApiError.status` returns `statusCode` in both `task-store/src/errors.ts` and `admin/src/errors.ts`
- [x] 5xx `ApiError` instances remain unaffected (still reported by `onError`'s own capture path)
- [x] No changes needed to either `onError` handler or the `sentry()` mount options — this is a minimal, additive fix on the shared error base class

## Test Plan
- [x] New unit tests in both services lock in `status === statusCode` for the base class and representative subclasses (`bun test task-store/src/errors.unit.test.ts`, `bun test admin/src/errors.unit.test.ts`)
- [x] Existing Sentry smoke tests (`task-store/src/app.sentry.smoke.test.ts`, `admin/src/agents-api.sentry.smoke.test.ts`) still pass unchanged
- Note: a full `app.request()` smoke test of the `@sentry/hono` middleware's own capture behavior is not feasible here — `buildSentryInitOptions` returns `undefined` whenever `NODE_ENV === "test"` (which `bun test` sets automatically), so the `sentry()` middleware never mounts under test, and `@sentry/hono`'s `defaultShouldHandleError` is not part of its public export surface (deep-importing it is rejected by Node/Bun's `package.json` exports resolution). The unit tests above are the correct, testable boundary for this fix.
- [x] `bun test` run repo-wide: no new failures introduced (pre-existing baseline has 31-32 unrelated failures on `main`, confirmed identical before/after this change — env-var pollution in `metrics/src/lib/env.unit.test.ts` and an openapi fixture issue in `task-store/src/routes/prs.openapi.smoke.test.ts`, both unrelated to this change)
- [x] Typecheck clean for both `@shipwright/task-store` and `@shipwright/admin`
- [x] Lint clean on all changed files

### DIFF
--- a/admin/src/errors.ts
+++ b/admin/src/errors.ts
@@ -14,6 +14,15 @@ export class ApiError extends Error {
     super(message);
     this.name = "ApiError";
   }
+
+  /**
+   * Alias for `statusCode` so `@sentry/hono`'s `defaultShouldHandleError`
+   * (which reads `error.status`, not `error.statusCode`) correctly recognizes
+   * already-handled 3xx/4xx responses and skips its own independent capture.
+   */
+  get status(): number {
+    return this.statusCode;
+  }
 }
 
 // ─── HTTP error subclasses ────────────────────────────────────────────────────
--- a/admin/src/errors.unit.test.ts
+++ b/admin/src/errors.unit.test.ts
@@ -1,0 +1,52 @@
+/**
+ * admin/src/errors.unit.test.ts
+ *
+ * Locks in the `status` getter alias on `ApiError` (obs-sentry-status-mismatch).
+ *
+ * `@sentry/hono`'s `sentry()` middleware (mounted in main.ts whenever SENTRY_DSN
+ * is set) independently captures any error left on `context.error` after the
+ * route's own `onError` has already run, gated only by `defaultShouldHandleError`:
+ *
+ *   const status = error.status;
+ *   return !(typeof status === "number" && status >= 300 && status < 500);
+ *
+ * `ApiError` exposed its HTTP status only as `statusCode`, so `error.status`
+ * was always `undefined` — `defaultShouldHandleError` always returned `true`,
+ * and every ApiError with statusCode < 500 (including ones agents-api.ts's own
+ * onError deliberately does NOT report — see agents-api.sentry.smoke.test.ts)
+ * got captured and alerted anyway.
+ *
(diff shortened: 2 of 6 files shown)


dmcaulay (commented):
Verdict: COMMENT — The admin/task-store fix is correct and well-tested (a minimal `status` getter alias so @sentry/hono's defaultShouldHandleError stops double-reporting already-handled 3xx/4xx errors), but the identical bug is still live in metrics/src/lib/errors.ts — metrics mounts the same sentry() middleware via the same buildSentryInitOptions helper and has the same 5xx-only onError reporting gate, so it will keep spuriously alerting on handled errors after this merges.

dmcaulay (approved):
Verdict: APPROVE — The metrics/ gap flagged in the prior review is now closed: `metrics/src/lib/errors.ts` gets the identical `status` getter alias plus matching unit test coverage. Verified the fix is complete across every service that mounts `@sentry/hono` (admin, task-store, metrics) — `agent/` and `chat/` don't use that middleware so are correctly out of scope — and that no `onError` handler branches on `.status`, so this is a safe, purely additive change.

PR #3251, opened 2026-09-05
fix: return 400 (not 500) for malformed JSON request bodies in task-store

dodizzle:
## Summary
- Fixes Sentry issue VITALS-OS-44: POSTing a malformed-JSON body to a task-store route (e.g. `POST /tasks`) returned a 500 and fired a Sentry `captureException` alert, even though this is a client error.
- Root cause: `@hono/zod-openapi`'s request validator throws a bare `HTTPException(400, ...)` from `hono`'s own JSON-parse-failure path (not an `ApiError` instance), so it fell through `task-store/src/app.ts`'s `onError` generic "else" branch, which ignored the exception's own status and always returned 500 + alerted Sentry.
- Added an `err instanceof HTTPException` branch in the shared `onError` hook (between the existing `ApiError` branch and the generic fallback): status `< 500` responds with the exception's own status and skips Sentry capture; status `>= 500` (not raised by this parse path, handled defensively) preserves the existing capture-and-500 behavior. Mirrors the existing `ApiError`-by-`statusCode` split already used in `admin/src/agents-api.ts`.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Malformed JSON body returns 400 (not 500) with a JSON error body | MET |
| 2 | No Sentry `captureException` call for malformed JSON | MET |
| 3 | No regression: valid-JSON-wrong-schema still 400/no-Sentry; genuine 500s still capture+500 | MET |
| 4 | New smoke test covers malformed-JSON → 400, no-Sentry-capture | MET |

## Test Plan
- New test in `task-store/src/app.sentry.smoke.test.ts`: POSTs `"{not valid json"` to `/tasks`, asserts `res.status === 400` and zero Sentry captures. Written first, confirmed failing (500 + 1 capture) against the unmodified code, then confirmed passing after the fix.
- Full `task-store/src` suite (640+ tests) passes; all pre-existing `app.sentry.smoke.test.ts` cases (ApiError handling, genuine unhandled-error 500+Sentry, caller-label logging) still pass — no regression.
- `task lint` and `task typecheck` clean.
- Confirmed via a controlled before/after diff (stash-based baseline comparison) that a set of 31 locally-failing tests (metrics env/error unit tests, `prs.openapi.smoke.test.ts`, a `check-test-readiness` fs error) are pre-existing and unrelated to this change — identical failure count with and without this diff. GitHub Actions' `CI` workflow is green on `main` at the branch's base commit, confirming these are local-sandbox-only flakiness, not real CI failures.

### DIFF
--- a/task-store/src/app.sentry.smoke.test.ts
+++ b/task-store/src/app.sentry.smoke.test.ts
@@ -160,6 +160,29 @@ describe("onError — Sentry capture wiring", () => {
   });
 });
 
+describe("onError — malformed JSON request body", () => {
+  it("returns 400 (not 500) and does not call sentryClient.captureException", async () => {
+    const sentryClient = fakeErrorCapturingClient();
+    const app = createTaskStoreApp({
+      taskService: apiErrorTaskService(),
+      tokenService: fakeTokenService(),
+      sentryClient,
+    });
+
+    const res = await app.request("/tasks", {
+      method: "POST",
+      headers: {
+        Authorization: `Bearer ${VALID_TOKEN}`,
+        "Content-Type": "application/json",
+      },
+      body: "{not valid json",
+    });
+
+    expect(res.status).toBe(400);
+    expect(sentryClient.capturedErrors.length).toBe(0);
+  });
+});
+
 describe("onError — caller label logging (AOB-3.3)", () => {
   it("includes the admin caller label in the unhandled-error console.error line", async () => {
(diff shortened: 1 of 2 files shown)


dmcaulay (approved):
Verdict: APPROVE — Correctly scoped fix for the malformed-JSON 500/Sentry-alert bug; verified the HTTPException branch, the >=500 fallback status handling, and the new smoke test locally (6/6 passing). No blocking issues.

PR #3253, opened 2026-09-05
feat: make SlackProgress call agents.sessions.setStatus eagerly (ISW-1.1)

dodizzle:
## Summary
- `SlackProgress.start()` now calls `agents.sessions.setStatus({channel_id, thread_ts, status: "processing"})` eagerly, before `startStream()` runs, wrapped in try/catch like every other Slack API call in this file — a failure never blocks streaming or the eventual reply.
- `SlackProgress.finish()` now unconditionally calls `setStatus` with the correct active/processing status, independent of whether a stream ever opened, and separately still runs `stopStream()` whenever a stream did open — the two are no longer either/or.
- `docs/agent.md`'s Slack progress-tracking paragraphs updated to drop the now-incorrect "`agents.sessions.setStatus` is never called" claim and describe the new eager, authoritative behavior.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | start() calls setStatus with {channel_id, thread_ts, status:"processing"} before startStream(), both isDM cases | MET |
| 2 | finish() always calls setStatus (per stillInFlight), independent of streamTs; stopStream() still fires when stream opened | MET |
| 3 | setStatus failure at start() caught/logged, doesn't block startStream()/seed card | MET |
| 4 | Old NOT-called tests removed/rewritten; new failure-at-start test added | MET |
| 5 | docs/agent.md updated to match new behavior | MET |

## Test Plan
- `bun test agent/src/slack-progress.unit.test.ts agent/src/slack.unit.test.ts agent/src/slack.integration.test.ts` — 262 pass
- `bunx biome lint .` — clean


dmcaulay (approved):
Verdict: APPROVE — Clean eager/unconditional setStatus change; all 5 acceptance criteria verified against the diff, tests meaningfully cover the new behavior (incl. call-order and failure-path assertions), docs updated to match, no unaddressed findings.

PR #3270, opened 2026-09-07
fix: fail-closed isAdmin default in getSessionUser (admin-ui.ts)

dodizzle:
## Summary
- `getSessionUser` in `admin/src/admin-ui.ts` now computes `isAdmin` as `payload.isAdmin === true` (fail-closed) instead of `payload.isAdmin !== false` (fail-open), matching the fail-closed pattern used by every other `c.var.isAdmin` check in this file.
- A session token missing the `isAdmin` claim, or with a non-boolean value for it, is now treated as non-admin rather than defaulting to admin.
- Not currently exploitable — the sole token-issuing path (`createSessionToken`) always sets `isAdmin` explicitly and tokens are HMAC-signed — but this closes a latent fail-open default that would silently grant admin if any future code path ever minted a session token without explicitly setting `isAdmin`.

## Acceptance Criteria
| 1 | Line 501's `isAdmin` computation changed from `payload.isAdmin !== false` to `payload.isAdmin === true` | MET | `admin/src/admin-ui.ts:501` |
| 2 | Test asserting a session token with `isAdmin` omitted or set to a non-true value is treated as non-admin | MET | Two new smoke tests in `admin/src/admin-ui.smoke.test.ts` asserting 403 for an omitted claim and for a non-boolean truthy claim |

## Test Plan
- New smoke tests: `GET /admin/prs returns 403 when the session token omits the isAdmin claim entirely` and `GET /admin/prs returns 403 when the session token's isAdmin claim is a non-boolean truthy value` — both confirmed RED against the original fail-open code, GREEN after the fix.
- Full `admin-ui.smoke.test.ts` suite: 299/299 pass, no regressions to existing admin/session tests (real admin sessions with `isAdmin: true` still work).
- `task typecheck`: clean.
- `task ci` fails with 31 pre-existing test failures in `metrics/src/lib/env.unit.test.ts`, `metrics/src/lib/errors.unit.test.ts`, and `task-store/src/routes/prs.openapi.smoke.test.ts` — verified identical failure count reproduces on a clean `origin/main` checkout, unrelated to this change.

### DIFF
--- a/admin/e2e/agents-page.e2e.ts
+++ b/admin/e2e/agents-page.e2e.ts
@@ -98,7 +98,7 @@ async function mintSession(
 ): Promise<string> {
   const nowSec = Math.floor(Date.now() / 1000);
   return sign(
-    { userId, email, iat: nowSec, exp: nowSec + 3600 },
+    { userId, email, isAdmin: true, iat: nowSec, exp: nowSec + 3600 },
     SESSION_SECRET,
     "HS256",
   );
--- a/admin/e2e/chat-degraded.e2e.ts
+++ b/admin/e2e/chat-degraded.e2e.ts
@@ -99,7 +99,7 @@ async function mintSession(
 ): Promise<string> {
   const nowSec = Math.floor(Date.now() / 1000);
   return sign(
-    { userId, email, iat: nowSec, exp: nowSec + 3600 },
+    { userId, email, isAdmin: true, iat: nowSec, exp: nowSec + 3600 },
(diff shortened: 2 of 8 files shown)


dmcaulay (approved):
Verdict: APPROVE — Correct, narrowly-scoped fail-closed fix for a latent isAdmin default (not currently exploitable, since createSessionToken always sets isAdmin explicitly); consistent e2e/fixture companion updates and adequate new smoke-test coverage.

PR #3275, opened 2026-09-08
fix: attach truncated response body to task-store query() 500 errors

dodizzle:
## Summary
- `query()` in `agent/src/check-helpers.ts`'s `createTaskStoreClient()` now attaches a truncated (500-char) response body snippet to the `Error` it throws on a non-2xx task-store response, instead of a status-code-only message.
- This is an observability-only change — no polling behavior, cadence, or dispatch logic was touched.

## Background
Sentry issue [VITALS-OS-4P](https://app-vitals-llc.sentry.io/issues/7621911681/) shows the agent's dev-task candidate poll (`GET /tasks?ready=true`) intermittently 500ing, but the thrown error carried no detail beyond the status code, making the task-store-side root cause undiagnosable from the event alone.

The task's leading root-cause hypothesis is that this is a downstream symptom of the task-store DB-connectivity issue tracked in `error-7603012949-db-unreachable` ("Can't reach database server at 127.0.0.1:5432" — 45.9k+ events, still open/pending, HITL). That correlation could **not** be confirmed here: this session has no Sentry API credentials to pull exact event timestamps for either issue and cross-reference them. Going forward, the next occurrence of this 500 will carry the task-store's own response body in the error message, so a human (or a future agent with Sentry access) can confirm or rule out that correlation directly from the error text — no cross-service timestamp comparison needed.

## Acceptance Criteria
No formal `acceptanceCriteria` were set on this task; verified instead against the human-approved scope note on the task record:

| 1 | `query()` in `agent/src/check-helpers.ts` attaches a truncated response body to the thrown Error | MET | try/catch around `res.text()`, sliced to 500 chars, appended to the error message on non-ok response |
| 2 | Do not change polling behavior | MET | No changes to `check-dev-task.ts`, `loop-orchestrator.ts`, or any polling cadence/dispatch logic |
| 3 | Correlate with `error-7603012949` timestamps in the PR description | UNVERIFIABLE | No Sentry API access in this session — see Background above |

## Test Plan
- Added unit tests covering: body snippet attached on failure, long bodies truncated, and body-read failures falling back to the status-only message.
- `NODE_ENV=test bun test agent/src/check-helpers.unit.test.ts` — 161/161 pass.
- `bun run --filter='*' --sequential typecheck` — all packages pass.
- `bunx biome lint` on changed files — clean.
- `task check-strings`, `task check-config-docs`, `task check-version-sync`, `task secret-scan` — all pass.
- `task test:coverage` (full suite) has 30 pre-existing failures identical on `main` (test-order pollution in `metrics/src/lib/{errors,env}.unit.test.ts` and a known `task-store/src/routes/prs.ts` `validateRepo` null-repos bug) — confirmed unrelated to this change by running the same suite on `main`.

### DIFF
--- a/agent/src/check-helpers.ts
+++ b/agent/src/check-helpers.ts
@@ -535,8 +535,21 @@ export function createTaskStoreClient(opts?: { fetchFn?: FetchFn }): {
   return {
     async query(params: URLSearchParams): Promise<Task[]> {
       const res = await doFetch(`${baseUrl}/tasks?${params}`, { headers });
-      if (!res.ok)
-        throw new Error(`task-store GET /tasks?${params} → ${res.status}`);
+      if (!res.ok) {
+        // Attach a truncated response body snippet so a 500 (e.g. task-store
+        // DB unreachable) is diagnosable from the thrown error alone, without
+        // cross-referencing task-store's own logs/Sentry events by timestamp.
+        let bodySnippet = "";
+        try {
+          bodySnippet = (await res.text()).slice(0, 500);
+        } catch {
+          // Body unreadable (already consumed, network cut mid-read, etc.) —
+          // fall back to the status-only message below.
+        }
+        throw new Error(
+          `task-store GET /tasks?${params} → ${res.status}${bodySnippet ? ` — ${bodySnippet}` : ""}`,
+        );
+      }
       const data = (await res.json()) as unknown;
       // Temporary: accept legacy bare Task[] from older task-store instances
       if (Array.isArray(data)) return data as Task[];
--- a/agent/src/check-helpers.ts
+++ b/agent/src/check-helpers.ts
@@ -473,6 +473,75 @@ type FetchFn = (
 ) => Promise<Response>;
 
 /**
+ * Max chars of a non-ok response body scanned when redacting. Deliberately
+ * larger than the emitted snippet so a secret straddling the emit cut is still
+ * matched and masked in full, rather than half of it slipping through.
+ */
+const BODY_SCAN_LIMIT = 2000;
+
+/** Max chars of the redacted body snippet folded into a thrown error message. */
+const BODY_SNIPPET_LIMIT = 500;
+
+/**
+ * Secret-shaped patterns masked out of another service's error body before any
+ * of it is attached to a thrown Error.
+ *
+ * `docs/observability.md` documents that request/response bodies are never
+ * attached to a Sentry event, and `lib/sentry.ts`'s scrub hook only redacts
+ * *this* process's own live `SECRET_ENV_VARS` values by exact string match —
+ * neither does anything for credential material embedded in a *different*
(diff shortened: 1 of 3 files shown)


agent/src/check-helpers.ts
```diff
@@ -535,8 +535,21 @@ export function createTaskStoreClient(opts?: { fetchFn?: FetchFn }): {
...
+        // cross-referencing task-store's own logs/Sentry events by timestamp.
+        let bodySnippet = "";
+        try {
+          bodySnippet = (await res.text()).slice(0, 500);
+        } catch {
+          // Body unreadable (already consumed, network cut mid-read, etc.) —
+          // fall back to the status-only message below.
+        }
+        throw new Error(
+          `task-store GET /tasks?${params} → ${res.status}${bodySnippet ? ` — ${bodySnippet}` : ""}`,
```
dmcaulay, line 622: This embeds up to 500 chars of the raw task-store response body into the thrown Error's message. `docs/observability.md` explicitly documents "Request or response bodies | No — never attached to an event" for this repo's Sentry captures, and this PR's own stated purpose is to make exactly this content visible in the Sentry issue that reported the original 500.

`lib/sentry.ts`'s scrub hook only redacts `Authorization`/`Cookie` headers and this process's own `SECRET_ENV_VARS` values by exact string match — it does nothing for secret material embedded in a *different service's* response body. `task-store/src/app.ts`'s `onError` returns `{ error: err.message }` verbatim for any unhandled 500 and also calls `sentryClient?.captureException(err)` with that same raw message, so whatever task-store's own dependencies put into `err.message` (Prisma connection-string-validation errors like `P1013` are documented to sometimes echo the full `DATABASE_URL`, credentials included) flows straight through: task-store 500 body → `res.text()` → `bodySnippet` → this thrown Error → the agent's own `captureException` call in `loop-orchestrator.ts` — completely unredacted, since `check-dev-task.ts`'s `getReadyTasks` (which calls this `query()`) feeds the dispatch/failure-capture path.

Could you scrub/allowlist the body snippet before attaching it, or attach it as Sentry structured "extra" context with its own redaction pass, instead of folding raw body text straight into the Error message?

dmcaulay (commented):
Verdict: COMMENT — solid implementation and tests, but the response-body snippet this PR attaches to the thrown Error contradicts docs/observability.md's documented "never attach response bodies to Sentry events" policy, with a concrete path for credential leakage via task-store's own unhandled-error messages (see inline comment).

dmcaulay (approved):
Verdict: APPROVE — Follow-up review at c33b7f15: the credential-leak finding from the prior review is genuinely resolved. redactBodySnippet() adds a real, well-tested scrub layer (URI creds, token prefixes, Bearer/Basic auth, secret-shaped key/value pairs) before the task-store response body is folded into the thrown Error, with a 2000/500-char scan/emit split so a secret can't be half-exposed at the truncation boundary. docs/observability.md documents the exception. No new issues.

PR #3276, opened 2026-09-08
fix: retry once on Claude connection-lost-mid-response in _spawn

dodizzle:
## Summary
- Adds a targeted, single retry in `agent/src/claude.ts`'s `_spawn` for the "connection lost mid-response" failure mode reported in Sentry issue 7616867192 (VITALS-OS-47)
- Detection is a substring match on the CLI's own error text ("Connection lost mid-response"), scoped via a new `_isConnectionLostMidResponse` type guard so unrelated `is_error` failures (auth, rate limits, etc.) are untouched
- The existing `_spawn` body is renamed to `_spawnOnce`; `_spawn` is now a thin wrapper that catches this one failure mode, logs it via `sentryClient.captureMessage` for visibility, and re-runs the whole spawn exactly once (discarding the partial response) — a second failure of the same kind propagates normally, with no further retries
- This is independent of `_runClaude`'s existing session-resume retry, which only kicks in when there's an existing session id to resume — the new retry applies even to a brand-new (first-message) session

## Acceptance Criteria
Task had no formal `acceptanceCriteria` entries — the task `note` field (quoted above) is the de facto spec.

| 1 | Retry exactly once on connection-lost-mid-response | MET | `_spawn` wrapper calls `_spawnOnce` a second time only inside the `catch`, and the second call bypasses the wrapper entirely |
| 2 | Discard the partial response, re-run the whole spawn | MET | Second `_spawnOnce` call uses the same `args` from scratch — no resume, no reuse of partial state |
| 3 | No further retries | MET | Test: "does not retry a second time when the retry also hits connection-lost-mid-response" — exactly 2 spawn calls |
| 4 | Log the retry so it's visible | MET | `sentryClient?.captureMessage?.(...)` fired before the retry; test asserts it's captured exactly once |
| 5 | Change point is `_spawn`, other failure modes unaffected | MET | Test: "does not retry an unrelated is_error failure" — exactly 1 spawn call, no log |

## Test Plan
- `bun test agent/src/claude.unit.test.ts` — 3 new tests added under `describe("connection-lost-mid-response retry")`, all passing; only a pre-existing, unrelated `extraAllowedTools` test fails (confirmed failing identically on `main`, caused by ambient `AGENT_ALLOWED_TOOLS` env in this sandbox)
- `task typecheck` — passes across all workspaces
- `bunx biome lint` / `biome format` — clean
- [x] Pre-commit checks passing (scoped to this change; unrelated pre-existing repo-wide test/env failures confirmed present on `main` baseline too)

### DIFF
--- a/agent/src/claude.ts
+++ b/agent/src/claude.ts
@@ -179,6 +179,23 @@ export class ClaudeRunError extends Error {
 }
 
 /**
+ * Substring match against the Claude CLI's own error text for a transient
+ * network drop between the CLI process and the Anthropic API mid-stream
+ * (Sentry issue 7616867192 / VITALS-OS-47, e.g. "API Error: Connection lost
+ * mid-response. The response above may be incomplete.") — distinct from
+ * other `is_error` failures (auth, rate limits, etc.), which are not
+ * eligible for `_spawn`'s single connection-lost retry.
+ */
+const CONNECTION_LOST_MID_RESPONSE = "Connection lost mid-response";
+
+function _isConnectionLostMidResponse(err: unknown): err is ClaudeRunError {
+  return (
+    err instanceof ClaudeRunError &&
+    err.resultMessage.includes(CONNECTION_LOST_MID_RESPONSE)
+  );
+}
+
+/**
  * Thrown when a run is cancelled via its AbortSignal (e.g. a chat cancel
  * request surfaced through the heartbeat tick). Mirrors ClaudeTimeoutError /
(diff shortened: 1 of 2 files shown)


dmcaulay (approved):
Verdict: APPROVE — Retry logic matches the human-decision spec exactly (retry once, discard partial response, no resume, log before retrying, no further retries); control flow correctly prevents a double-retry, no resource leaks, no external contract change, and solid convention-compliant unit test coverage.

PR #3287, opened 2026-09-09
fix: graceful SIGTERM shutdown for admin, task-store, and chat

dmcaulay:
## Summary
- Add `lib/graceful-shutdown.ts`: a shared helper that registers SIGTERM/SIGINT once, stops the server (letting in-flight requests finish), runs ordered async cleanup (`prisma.$disconnect()`), then exits — with a 20s hard deadline (below the 30s pod grace period) that forces `server.stop(true)` + `exit(1)` if cleanup hangs. Idempotent on a second signal.
- Wire it into `admin/src/main.ts`, `task-store/src/main.ts`, and `chat/src/main.ts`, capturing each service's `Bun.serve()` return value (previously discarded in admin and chat).
- Gate `/health/ready` on the new `isShuttingDown()` flag in admin and task-store (chat has no `/health/ready` route today — only unauthenticated `/health` liveness — so nothing to gate there).

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | SIGTERM stops new connections, finishes in-flight requests, disconnects Prisma, exits 0 within grace period | MET |
| 2 | `/health/ready` returns 503 after shutdown signal | MET |
| 3 | Helper covered by `*.unit.test.ts` with injected fakes; `task ci` passes with coverage gate met | MET (see caveat below) |
| 4 | No change to request handling during normal operation | MET |

## Test Plan
- `lib/graceful-shutdown.unit.test.ts` (8 tests, fully DI'd fakes — no `mock.module()`/global overrides): stop-before-cleanup ordering, cleanup order preserved, `exit(0)` on success, forced `stop(true)` + `exit(1)` at a manually-fired fake deadline, second-signal no-op, SIGINT parity, synchronously-readable `isShuttingDown()`, throwing cleanup fn doesn't block later cleanup/exit.
- `admin/src/main.smoke.test.ts`: new case confirming `/health/ready` returns 503 immediately (skipping the DB round-trip) once `shuttingDown` is set.
- **Manual verification**: booted `task-store` and `admin` locally (`bun run src/main.ts`) and sent a real `kill -TERM`. Both logged `received SIGTERM, shutting down gracefully` → `graceful shutdown complete` and exited within ~1.3s. `chat` was not manually verified (same code path, not independently booted).
- [x] `task typecheck` — passes across all packages
- [x] `task lint` — passes on all changed files
- [x] `task ci` chain: lint → check-strings → typecheck → check-config-docs → check-version-sync all pass; `test:coverage` fails only on 3 pre-existing test failures and a repo-wide coverage shortfall (85.3%/85.9% vs 90%/89% threshold) — **both independently confirmed to reproduce identically on a clean `origin/main` worktree**, unrelated to this diff. `secret-scan` couldn't run locally (`gitleaks` not installed in this sandbox); CI enforces that gate remotely.


dodizzle (approved):
Verdict: APPROVE — Clean, well-tested graceful-shutdown implementation across admin/task-store/chat; shutdown state machine, readiness-gate wiring, and test coverage all check out with no blocking issues. Dependency risk: merge — lib/package.json change is an internal export-map addition, not a dependency version bump.

PR #3290, opened 2026-09-09
fix(site): recut Devin deployment claim after Outposts, refresh verified date

dodizzle:
Closes task `MKT-VS-DEVIN-REVERIFY-1`. Approved by Dan.

## What moved

`/vs/devin` carried `verifiedDate = "July 14, 2026"` alongside the sentence *"Re-checked immediately before this page ships."* Cognition shipped **Devin Outposts on 2026-07-22** — eight days after that date — described in their release notes as *"a new capability for running Devin workloads in your own environment."*

That makes our clause **"never inside your own infrastructure" indefensible as written.** It is the same category of overreach `brand/MESSAGING.md`'s never-print list already exists to prevent, so it was escalated rather than rewritten freehand.

## The recut

Outposts relocates the **execution sandbox**, not the **agent loop**. Devin's deployment overview still states its workspace *"operates under the control of its brain, which always resides within Cognition's Cloud."*

So the claim moves from *where the work runs* to *where the control plane runs* — the distinction that still holds, and the one `MESSAGING.md` §1 already draws with "runs entirely in your own cloud — **control plane included**."

> Cognition-hosted control plane at every tier. Devin Outposts can run the execution sandbox in your own environment, but the agent loop — the brain — always resides in Cognition's cloud.

This doesn't weaken the differentiator. It sharpens it, and it's better sourced than before.

## Re-verification results

| Claim | Verdict |
|---|---|
| Deployment | **Recut** — see above |
| Pipeline — "Spec through PR. The deploy step is not part of the agent's scope." | **Holds.** No deploy/CD capability shipped in 2026. Closest entries are PR-workflow UX (auto-merge from Devin Review, 2026-04-08; PR auto-close removed, 2026-04-29). Merging is not deploying. |
| Economics | **Holds**, with "vendor-hosted" dropped as imprecise post-Outposts. Still price-free per D5. |
| License | **Holds.** Still proprietary and commercial. |
| Header prose — "a strong benchmark track record" | **Removed.** |

On that last one: it can't be fixed by citing it. The only evidence that would substantiate "strong benchmark track record" is a SWE-bench Verified number, which the never-print list bans (invalidated Feb 2026). Uncited vague praise for a competitor also violates D10's every-fact-must-be-cited rule. No figure introduced.

## Scope beyond the task

Three things the task didn't ask for, but that leaving alone would have made this a partial fix:

1. **`/self-hosted` carried the identical stale claim.** Fixed there too — shipping only `/vs/devin` would have left the same indefensible sentence live one page over.
2. **`/compare` carried the same uncited benchmark phrase.** Removed.
3. **`brand/MESSAGING.md` encoded the stale wording as the approved "safe framing."** That's the root cause — left alone, the next comparison pass reintroduces it. Now marked superseded, with the control-plane-scoped replacement recorded and dated.

## Dates

`verifiedDate` bumped to September 9, 2026 on `/vs/devin` **only**. `/self-hosted` (Aug 11) and `/compare` (Sep 5) keep theirs — only their Devin claims were re-verified, not their other vendors. Bumping those would claim a freshness sweep that didn't happen.

## Verification

- `bunx astro build` — passes, 27 pages (matches CI; note `npm run build` fails on `main` for an unrelated pre-existing peer conflict, astro@7 vs `@astrojs/tailwind@6`)
- `npm run brand-lint` — all pages on-brand
- `check-banned-strings` — clean
- Rendered `dist/` greps: zero occurrences of the three stale phrases; new claims present on both pages; zero "SWE-bench"; both source links render on the Deployment row

Sources: [deployment overview](https://docs.devin.ai/enterprise/deployment/overview) · [2026 release notes](https://docs.devin.ai/release-notes/2026)

Full dated write-up in `drafts/competitive/compare-vs-pages-audit-2026-09-05.md`.

### DIFF
--- a/brand/MESSAGING.md
+++ b/brand/MESSAGING.md
@@ -37,7 +37,7 @@ The old blanket rule ("never name a competitor") is **amended, not deleted**. Co
 
 **Never-print list (binding on every surface):**
 
-- "Devin has no private deployment" / "no VPC option" (three overreaching versions of this claim were refuted in verification — the safe framing is "Even Devin's dedicated single-tenant VPC tier is Cognition-hosted, never in your environment.")
+- "Devin has no private deployment" / "no VPC option" / **"never inside your own infrastructure"** (three overreaching versions of this claim were refuted in verification 2026-07-14; the fourth was refuted 2026-09-09 by Cognition's own release notes). **Superseded safe framing — do not reuse:** "Even Devin's dedicated single-tenant VPC tier is Cognition-hosted, never in your environment." That sentence went stale on **2026-07-22**, when Cognition shipped **Devin Outposts**, "a new capability for running Devin workloads in your own environment" (https://docs.devin.ai/release-notes/2026). **The current safe framing is control-plane-scoped, not environment-scoped:** "Cognition-hosted control plane at every tier. Devin Outposts can run the execution sandbox in your own environment, but the agent loop — the brain — always resides in Cognition's cloud." Outposts moved the workload, not the brain — Devin's deployment overview still states its workspace "operates under the control of its brain, which always resides within Cognition's Cloud." Claim the control plane, never the environment.
 - Any pricing figure — ours or a competitor's (see §3, price-free rule)
 - ACU pricing presented as a self-serve fact (Devin's ACU pricing is Enterprise-only, rates undisclosed)
 - Any SWE-bench Verified number (invalidated Feb 2026)
--- a/site/src/pages/compare.astro
+++ b/site/src/pages/compare.astro
@@ -450,7 +450,7 @@ helm install shipwright shipwright/shipwright</code></pre>
     <h2 class="mt-4">Shipwright vs Devin</h2>
     <p class="mt-3 max-w-2xl">
       Devin is the highest-profile autonomous coding agent — a managed
-      commercial service with a strong benchmark track record. Choose
+      commercial service. Choose
       Shipwright when you want the full pipeline, deploy stage included, on
       infrastructure you own and operate; choose Devin when a fully managed
       service with no infrastructure to run yourself is what you want.
--- a/site/src/pages/self-hosted.astro
+++ b/site/src/pages/self-hosted.astro
@@ -21,6 +21,7 @@ const src = {
   openhandsLicense: "https://github.com/OpenHands/OpenHands/blob/main/LICENSE",
   coder: "https://coder.com/docs",
   devinDeployment: "https://docs.devin.ai/enterprise/deployment/overview",
+  devinReleaseNotes: "https://docs.devin.ai/release-notes/2026",
   factoryDeployment: "https://docs.factory.ai/enterprise/network-and-deployment",
   factoryPricing: "https://factory.ai/pricing",
 };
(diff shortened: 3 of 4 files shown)


dmcaulay (commented):
Verdict: COMMENT — Solid, well-sourced recut of the Deployment claim and a consistent MESSAGING.md update, but the PR's own stated reasoning for dropping "vendor-hosted" (imprecise post-Outposts) isn't applied consistently on the same page: the Economics table row now says "proprietary", but `site/src/pages/vs/devin.astro` still says "vendor-hosted" in the meta description (line 115), the "Choose Devin when" card (line 210), and the static Economics-section prose right below the fixed table row (line 238). Worth a follow-up pass for internal consistency, or an explicit note on why those usages are intentionally retained.

dodizzle:
Addressed in 46ddb2da — thanks, the inconsistency was real and I confirmed all three usages were still live.

I went with making them consistent rather than documenting an exception, because `brand/MESSAGING.md`'s never-print entry that this PR itself updated is fairly explicit: **"Claim the control plane, never the environment."** `vendor-hosted` is an environment-scoped term, and Outposts is exactly what made it imprecise — so retaining it anywhere on the page would have been retaining the claim shape the PR set out to retire.

Changes, all in `site/src/pages/vs/devin.astro`:

| Location | Before | After |
|---|---|---|
| meta description | `self-hosted vs vendor-hosted` | `self-hosted vs vendor-controlled` |
| "Choose Devin when" card | `a fully managed, vendor-hosted service with no infrastructure to run yourself` | `a fully managed service with no infrastructure to run yourself` |
| Economics prose | `a commercial, vendor-hosted service` | `a commercial, proprietary service` |

Two notes on the specific wording:

- The meta description became `vendor-controlled` rather than `proprietary` because that slot is contrasting *deployment* (`self-hosted vs …`), not licensing — the adjacent clause already says `MIT licence vs proprietary`, so reusing it there would have made the description claim licensing twice and deployment never. `vendor-controlled` keeps it a deployment contrast while making it control-scoped, which is the version that survives Outposts.
- Dropping `vendor-hosted` from the "Choose Devin when" card also lands it on the same phrasing `/compare` already uses for the identical claim (`a fully managed service with no infrastructure to run yourself`), so the two pages now agree.

I also added a line to the file's header comment recording *why* the term is retired, so the next comparison pass doesn't reintroduce it — same intent as the MESSAGING.md supersession note.

On the neighbouring pages: I checked `/compare` and `/self-hosted` for the same inconsistency and both are already clean — neither contains `vendor-hosted` — so I left them untouched, consistent with this PR scoping the re-verification round to `/vs/devin`. No claim was broadened, no figures or citations introduced, and `verifiedDate` is unchanged, since this is an internal-consistency pass over claims already verified on 2026-09-09 rather than a new verification round.

Verified: `bunx biome lint .` clean, `astro build` green, and the full `vs-devin` / `compare` / `self-hosted` Playwright specs pass (52/52). Confirmed `vendor-hosted` renders zero times in the built `/vs/devin` HTML.

dmcaulay (approved):
Verdict: APPROVE — The prior internal-consistency finding (leftover "vendor-hosted" phrasing surviving in three spots on /vs/devin after the Economics row was recut to "proprietary") is fully resolved in 46ddb2da: all three occurrences are now consistent (vendor-controlled/proprietary/dropped). No new issues found.

PR #3310, opened 2026-09-11
feat: extract sendToUsers/notifySession in push-service.ts

dmcaulay:
## Summary
- Extracts the "subscriptions for emails → group by detail level → sendToMany → prune" core out of `PushService.notifyThreadReply` into a reusable `sendToUsers(emails, buildPayload)`.
- Adds `PushService.notifySession(session, level, kind)`, a new entry point for session-lifecycle notifications that resolves subscriptions for a given set of emails and delivers via the extracted core. The `session` parameter is a plain local shape (`{ slug, emails }`) since there's no real `Session`/`SessionFollow` Prisma model yet (that lands with sibling tasks `SES-1.1`/`SES-6.1`); the real payload content policy is deferred to a later task (`SESH-7.2`).
- `notifyThreadReply`'s external contract, error-swallowing behavior, and existing tests are unchanged — it now delegates to `sendToUsers` internally.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | notifyThreadReply's existing behavior and its existing unit tests pass unchanged after the extraction | MET |
| 2 | notifySession() correctly resolves subscriptions for a set of emails and sends via the extracted core | MET |
| 3 | Unit tests add direct coverage for sendToUsers()/notifySession(); no existing test retired | MET |

## Test Plan
- Added `describe("PushService.sendToUsers", ...)` — fan-out/group/prune happy path, and no-match-returns-zero.
- Added `describe("PushService.notifySession", ...)` — resolves + delivers for given emails, no-emails returns zero, swallows a downstream failure without throwing.
- All pre-existing `push-service.unit.test.ts` tests pass unmodified.
- `admin/src/push-service.ts` at 100% line/function coverage.
- `task typecheck`, `task lint`, `task check-strings` all pass. Full-suite `task ci` has 3 pre-existing, unrelated failures (`extraAllowedTools`, `check-downstream-compat.sh`) reproduced on clean `main`.


admin/src/push-service.ts
```diff
@@ -168,6 +218,30 @@ export class PushService {
...
+   * separate, later task.
+   */
+  async notifySession(
+    session: NotificationSession,
+    level: PushDetailLevel,
+    kind: SessionNotificationKind,
+  ): Promise<{ delivered: number; pruned: number }> {
+    try {
+      if (session.emails.length === 0) return { delivered: 0, pruned: 0 };
+      return await this.sendToUsers(session.emails, () =>
```
dodizzle: `sendToUsers` resolves `level = resolveDetailLevel(this.maxDetail, s.detailOptIn)` per subscription and calls `buildPayload(level)` with it — that's the system's documented privacy invariant (payload = min(operator ceiling, subscriber opt-in)). `notifyThreadReply`'s lambda correctly consumes that per-subscription `level`, but this lambda takes no parameter and instead closes over `notifySession`'s own outer `level` argument, discarding the resolved per-subscriber level entirely. Every subscriber on `session.emails` gets the same payload regardless of their own `detailOptIn`, and it's never capped against `this.maxDetail` either — this silently defeats the ceiling/opt-in policy this file otherwise treats as load-bearing. Suggest accepting the per-subscription level in this lambda (e.g. `(subLevel) => JSON.stringify({ kind, slug: session.slug, level: subLevel })`) so session notifications honor the same policy as thread-reply notifications.

dodizzle (commented):
Verdict: COMMENT — solid extraction of sendToUsers, but notifySession's buildPayload lambda ignores the per-subscription detail level sendToUsers resolves, bypassing the operator-ceiling/opt-in policy (see inline).

admin/src/push-service.ts
dmcaulay: Agreed on the premise — fixed in 2faf34ea, with one deviation from the suggested lambda.

Taking `(subLevel) => ({ ..., level: subLevel })` verbatim would leave `notifySession`'s `level` parameter entirely unused (the signature `notifySession(session, level, kind)` is fixed by SESH-7.1 and documented in `docs/agent-key-files.md`), so it would also trip the unused-parameter lint. Instead `level` is now explicitly a **per-call ceiling**, and the payload level is `resolveDetailLevel(level, subLevel)` — i.e. `min(caller ceiling, operator ceiling, subscription opt-in)`. A caller can ask for less detail, never more, and the per-subscriber resolution is no longer discarded.

Added two unit tests in `push-service.unit.test.ts`: one asserting the three-way min, and one with two subscribers on different `detailOptIn` values asserting they get different payloads (`generic` vs `title`) even when the caller requests `preview`.

dodizzle (approved):
Verdict: APPROVE — prior finding (notifySession bypassing the operator-ceiling/opt-in privacy invariant) is fixed correctly: payload level is now min(caller ceiling, operator ceiling, subscriber opt-in), with tests covering the per-subscriber behavior. No new issues.

PR #3315, opened 2026-09-11
fix: await stub server stop and use ephemeral ports in test teardown

dodizzle:
## Summary
- Fixes a stub-server port race in `agent/src/cron-run-reporter.integration.test.ts`, `chat-token-reporter.integration.test.ts`, `loop-orchestrator.integration.test.ts`, and `cron-handler.smoke.test.ts`
- `afterEach` hooks now `await server.stop(true)` instead of firing-and-forgetting it, so the socket is fully released before the next test's `beforeEach` runs
- All four files switched from hardcoded fixed ports (19960/19961/19962/19965, plus a second hardcoded block in `cron-handler.smoke.test.ts`) to OS-assigned ephemeral ports (`port: 0`), deriving each test's base URL from the server's actual bound `server.port`
- Test-infrastructure-only change — no production code touched, no new tests added

## Acceptance Criteria
| 1 | afterEach in all four files awaits server.stop(true) | MET | All 4 files' afterEach hooks are now async and await stop() |
| 2 | All four files use port:0, derive base URL from server.port | MET | Hardcoded PORT/REPORTER_PORT constants removed; base URL derived from `server.port` after bind |
| 3 | No new tests added, existing assertions unmodified | MET | Diff is setup/teardown only — no new test() blocks, all expect() assertions unchanged |
| 4 | Full bun test suite run ≥3x to build confidence | MET | Full suite run 3x: 8001 pass / 506 skip / 2 pre-existing unrelated fails, consistent all 3 runs |

## Test Plan
- Ran the four target test files individually — 114/114 pass
- Ran the full `bun test` suite 3 consecutive times — 8001 pass / 506 skip each run, no flake in the touched files
- `task lint` and `task typecheck` pass clean
- The only 2 failures seen under `task ci` (`test-store/src/routes/prs.openapi.smoke.test.ts`) are a known pre-existing bug (`validateRepo` treats `undefined` repos differently than `null`) unrelated to this change — confirmed unaffected by this diff and pre-existing on main

### DIFF
--- a/agent/src/chat-token-reporter.integration.test.ts
+++ b/agent/src/chat-token-reporter.integration.test.ts
@@ -95,8 +95,7 @@ describe("HttpChatTokenReporter", () => {
   // biome-ignore lint/suspicious/noExplicitAny: Server type param varies by bun version
   let server: ReturnType<typeof Bun.serve<any>>;
   let state: StubState;
-  const PORT = 19961;
-  const BASE_URL = `http://localhost:${PORT}`;
+  let baseUrl: string;
   const AGENT_ID = "agent-abc";
   const API_KEY = "test-api-key";
   const CLOCK = FixedClock(new Date("2026-01-15T12:00:00.000Z"));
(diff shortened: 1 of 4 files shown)


dmcaulay (approved):
Verdict: APPROVE — Clean, complete fix for the FLK-1.1 stub-server port race: every afterEach now awaits server.stop(true), every hardcoded port was migrated to ephemeral port:0 with correct post-bind server.port derivation, no test behavior changed.

PR #3317, opened 2026-09-11
feat: SessionService upsert + wrap Task.create/bulk in transaction

dodizzle:
## Summary
- Add `task-store/src/session-service.ts` with `SessionService.upsert()` — v1 scope, upsert only (get/list/update/purge land in later tasks)
- Wrap `TaskService.create()`/`bulk()` in `$transaction` so a non-blank `session` on a task write atomically upserts the corresponding `Session` row (create-if-missing, never overwrite `title`/`createdAt`)
- Writing into an archived session clears `archivedAt` (un-archive), logged via `console.log` per the reaper convention
- `isBlankSession()` treats `null`, `undefined`, and empty/whitespace-only strings alike as "no session" — a pure no-op, no DB write
- The upsert is implemented as a single atomic `INSERT ... ON CONFLICT (slug) DO UPDATE` (Prisma's native `session.upsert()`) rather than a `findUnique`-then-`create`/`update`, closing a race window where two concurrent writes into the same brand-new session slug could otherwise collide

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `POST /tasks` and `POST /tasks/bulk` with `session: "x"` result in a Session row `x` existing; a second write does not change `createdAt` or overwrite an existing title | MET |
| 2 | The upsert hook treats `session` as absent for `null`, `undefined`, and empty/whitespace-only string alike | MET |
| 3 | Writing a task into an archived session clears `archivedAt` | MET |
| 4 | `Task.create()`/`bulk()` remain functionally unchanged for tasks with `session: null` or empty-string session | MET |
| 5 | Integration tests (real Postgres) covering upsert-on-create, upsert-on-bulk, un-archive-on-write, and no-op for both `session: null` and `session: ""`; >=90% line coverage on `session-service.ts` and touched `task-service.ts` paths | MET |

## Test Plan
- `session-service.unit.test.ts` — `isBlankSession()` pure-logic coverage (null/undefined/empty/whitespace/non-blank)
- `session-service.integration.test.ts` (real Postgres, `DATABASE_URL_SHIPWRIGHT_TASK_STORE_TEST`) — upsert-on-create, no-overwrite of title/createdAt, un-archive-on-write, bulk() parity, per-item P2002 skip semantics preserved, blank-session no-op (`null`/`""`/`"   "`), and a concurrency regression test racing two `create()` calls into a fresh slug
- `task-service.unit.test.ts` — updated `bulk()` mock to supply `$transaction`
- Verified locally against a real local Postgres instance: full `bun test` suite 867 pass / 0 fail; coverage on `session-service.ts` 26/26 lines (100%), and every touched line in `task-service.ts`'s `create()`/`bulk()` is hit (470/503 file-wide, 93.4%)
- `task typecheck` / `task lint` clean
- Repo-wide `task ci` coverage gate (85.25%/85.91%) is below the 90%/89% threshold, but this is a pre-existing gap — confirmed against `origin/main` baseline (85.06%/85.69%, also failing); this branch is not a regression

### DIFF
--- a/task-store/src/session-migration.integration.test.ts
+++ b/task-store/src/session-migration.integration.test.ts
@@ -12,7 +12,14 @@
  * Requires DATABASE_URL_SHIPWRIGHT_TASK_STORE_TEST to be set; skips otherwise.
  */
 
-import { afterEach, beforeEach, describe, expect, it } from "bun:test";
+import {
+  afterAll,
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  it,
+} from "bun:test";
 import { readFileSync } from "node:fs";
 import { join } from "node:path";
 import { PrismaClient } from "../prisma/client/index.js";
@@ -89,9 +96,7 @@ async function hasSessionColumn(
 /**
  * Get column info for the Session table (type, nullable, defaults).
  */
-async function getSessionColumnInfo(
-  prisma: PrismaClient,
-): Promise<
+async function getSessionColumnInfo(prisma: PrismaClient): Promise<
   Array<{
     column_name: string;
     data_type: string;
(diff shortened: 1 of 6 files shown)


dmcaulay (approved):
Verdict: APPROVE — SessionService.upsert() correctly uses Prisma's native atomic upsert to close the concurrent-create race, all acceptance criteria are met, and coverage is thorough (unit + real-Postgres integration incl. a concurrency regression test). No blocking issues.

PR #3321, opened 2026-09-11
feat: add GET /sessions and GET /sessions/:slug routes

dmcaulay:
## Summary
- Add `GET /sessions` and `GET /sessions/:slug` read routes to the task-store service, mirroring `routes/tasks.ts`'s `createRoute` + Zod schema pattern
- Add `SessionService.list()`/`get()` — thin Prisma reads flattened with `computeSessionRollup()` output, scoped by agent token the same way `TaskService.distinct()` scopes (admin unrestricted; agent tokens limited to tasks assigned to them or in their repo scope)
- Add `PullRequestService.lookupBlockedPrNumbers()` — a single-query batched (repo, prNumber) → blocked-flag lookup feeding `computeSessionRollup()`'s `prBlockedSet`
- Add `SessionSchema`/`SessionListQuerySchema`/`SessionListResponseSchema`/`SessionSlugParamSchema` to `openapi-schemas.ts`, regenerate `task-store/openapi.json`, and document both routes in `docs/task-store.md`

## Acceptance Criteria
| Criterion | Status |
|---|---|
| GET /sessions returns 200 with `{ sessions, total, limit, offset }`; default state filter excludes closed and archived | MET |
| `sort=waitingSince` places the oldest waiting session first, non-waiting sessions after | MET |
| Agent-scoped token can't see a session with no task in its repos or assigned to it — 404 on `/sessions/:slug`, absent from the list | MET |
| `bun run generate:task-store-spec` output committed; docs document every `/sessions` route | MET |
| Smoke tests (injected `SessionServiceLike` double) cover scoping/filtering/sorting | MET |

## Test Plan
- `routes/sessions.smoke.test.ts` (new): route-contract coverage via an injected `SessionServiceLike` double — state variants, `sort=waitingSince`, `agentId`/`repo`/`q` filters, agent-token scoping, 200/404 on `GET /:slug`
- `session-service.integration.test.ts`: added real-DB coverage for `SessionService.list()`/`get()` (state filtering, `waitingSince` sort, agentScope visibility, `repo` filter, flattened shape) — closes a gap where the smoke tests' double never executed the real Prisma-backed logic
- `pull-request.integration.test.ts`: added real-DB coverage for `PullRequestService.lookupBlockedPrNumbers()` (blocked match, non-blocked exclusion, cross-product/repo-mismatch guard, batching, empty input)
- Verified against a real local Postgres instance: `task-store/src/session-service.ts`, `routes/sessions.ts`, and the new `lookupBlockedPrNumbers()` helper all sit at 100% line coverage
- [x] `task lint` / `task typecheck` / `task check-strings` all clean
- [x] Full task-store suite green (923 pass, 0 fail, against a real DB)
- [x] Repo-wide `bun run --filter='*' --sequential typecheck` clean across all 7 workspaces

### DIFF
--- a/docs/task-store.md
+++ b/docs/task-store.md
@@ -313,13 +313,16 @@ Query params:
 Returns `{ sessions: Session[], total: number, limit: number, offset: number }`. `total` is the
 count of all sessions matching the filters, independent of `limit`/`offset`.
 
-**Agent token visibility:** a repo-scoped agent token (`agentId` set, non-empty resolved repo scope
-— the same condition `/tasks` uses to decide whether to apply `agentScope`) only sees sessions where
-at least one of that session's own Task rows satisfies `task.assignee === agentId OR (task.repo !==
-null AND task.repo is in the agent's scope)`. A session with zero tasks can never satisfy this OR,
-so it is never visible under a scoped token, and a session none of whose tasks match is silently
-absent from the list (no error, no partial rollup). Admin tokens (`agentId === null`) — and agent
-tokens with no resolved repo scope — see every session matching the query filters, unrestricted.
+**Agent token visibility:** *every* agent token (`agentId` set) is scoped — the scope is applied
+whenever `agentId` is non-null, regardless of whether the token's resolved repo list is empty. A
+scoped token only sees sessions where at least one of that session's own Task rows satisfies
+`task.assignee === agentId OR (task.repo !== null AND task.repo is in the agent's resolved repo
+scope)`. A session with zero tasks can never satisfy this OR, so it is never visible under an agent
+token, and a session none of whose tasks match is silently absent from the list (no error, no
+partial rollup). An agent token with **no** resolved repo scope (empty repo list) degrades to
+assignee-only matching — the repo half of the OR can never match, so it sees only sessions
+containing a task assigned directly to it — it is **not** granted unrestricted visibility. Only
+admin tokens (`agentId === null`) see every session matching the query filters, unrestricted.
 
 #### Get session
 
@@ -328,7 +331,7 @@ GET /sessions/:slug
 ```
 
 Returns the same flattened session+rollup shape as the list response. Returns `404` if the session
-doesn't exist, or if a repo-scoped agent token has no qualifying task in it (same visibility rule as
+doesn't exist, or if an agent token has no qualifying task in it (same visibility rule as
 the list route above) — the two cases are indistinguishable to the caller by design.
 
 ### Task status lifecycle
--- a/docs/task-store.md
+++ b/docs/task-store.md
@@ -282,6 +282,55 @@ POST /tasks/:id/skip/reset
 
 Resets `skipCount` back to 0 and clears `lastSkippedAt`. Called after a human reviews and unblocks a task that was auto-blocked by skip threshold. Returns `200` with the updated task.
 
+### Sessions
+
+A `Session` row (`slug` is the primary key) is upserted automatically whenever a task write sets a
+non-blank `session` field — see `SessionService.upsert()`. These two read routes let a caller list
+and inspect sessions without re-deriving their rollup state client-side: each session's Task rows
+are summarized into a single rollup (`state`, `waitingSince`, `lastActivityAt`, per-status `counts`,
+distinct `agentIds`/`repos`, and the list of currently-`waitingTasks`) via `computeSessionRollup()`,
+then flattened together with the Session row's own fields (`slug`, `title`, `createdAt`, `updatedAt`,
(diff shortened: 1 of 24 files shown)


task-store/src/routes/sessions.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  // ─── List ──────────────────────────────────────────────────────────────────
+  // biome-ignore lint/suspicious/noExplicitAny: service returns Prisma-shaped types; JSON serialization handles Date→string correctly at runtime
+  app.openapi(listRoute, async (c): Promise<any> => {
+    const agentId = c.get("agentId");
+    const repos = c.get("repos");
+
+    // Same useAgentScope condition as routes/tasks.ts's listRoute: only build
+    // an auth scope for agent tokens with a known, non-empty repo scope.
+    const useAgentScope =
```
dodizzle: Critical: when `useAgentScope` is false because `repos` is `[]` (fail-safe-restrictive per `auth.ts` — resolver failure or a legitimate zero-repo agent), this omits `agentScope` from the filters entirely, and `SessionService.list()` treats a missing `agentScope` as unrestricted admin visibility. `routes/tasks.ts`'s `listRoute` and `TaskService.distinct()` both use this exact same `useAgentScope` condition but fall back to an assignee-only restriction (`{ assignee: agentId }`) instead — not to unrestricted. As written, an agent token with `repos: []` sees every session across every repo/agent, violating this PR's own acceptance criterion ("Agent-scoped token can't see a session with no task in its repos or assigned to it"). `hasQualifyingTask()` already degrades correctly to assignee-only matching when `repos` is `[]`, so the fix is just to always build `agentScope` for any non-null `agentId`: `agentId !== null ? { agentScope: { agentId, repos } } : {}`. No test covers an agent token with `repos: []` here.

task-store/src/routes/sessions.ts
```diff
@@ -0,0 +1,144 @@
...
+    const result = await sessionService.list(filters);
+    return c.json(result, 200);
+  });
+
+  // ─── Get one ───────────────────────────────────────────────────────────────
+  // biome-ignore lint/suspicious/noExplicitAny: service returns Prisma-shaped types; JSON serialization handles Date→string correctly at runtime
+  app.openapi(getOneRoute, async (c): Promise<any> => {
+    const agentId = c.get("agentId");
+    const repos = c.get("repos");
+    const useAgentScope =
```
dodizzle: Same issue as the list route above (line 98) — `GET /sessions/:slug` passes `undefined` instead of a restrictive `{ agentId, repos: [] }` scope when `repos.length === 0`, granting unrestricted single-session lookup to any agent token in that state.

dodizzle (commented):
Verdict: COMMENT — Agent tokens with an empty/unresolved repo scope get unrestricted session visibility in the new GET /sessions and GET /sessions/:slug routes, diverging from the fail-safe-restrictive fallback used by the existing /tasks route for the identical condition.

docs/task-store.md
```diff
@@ -282,6 +282,55 @@ POST /tasks/:id/skip/reset
...
+Returns `{ sessions: Session[], total: number, limit: number, offset: number }`. `total` is the
+count of all sessions matching the filters, independent of `limit`/`offset`.
+
+**Agent token visibility:** a repo-scoped agent token (`agentId` set, non-empty resolved repo scope
+— the same condition `/tasks` uses to decide whether to apply `agentScope`) only sees sessions where
+at least one of that session's own Task rows satisfies `task.assignee === agentId OR (task.repo !==
+null AND task.repo is in the agent's scope)`. A session with zero tasks can never satisfy this OR,
+so it is never visible under a scoped token, and a session none of whose tasks match is silently
+absent from the list (no error, no partial rollup). Admin tokens (`agentId === null`) — and agent
+tokens with no resolved repo scope — see every session matching the query filters, unrestricted.
```
dodizzle: This paragraph still describes the pre-fix behavior. `sessions.ts` now always builds `agentScope: { agentId, repos: repos ?? [] }` for any non-null `agentId`, and `hasQualifyingTask()` degrades an empty `repos` array to assignee-only matching — not unrestricted. Please update this sentence to match: only `agentId === null` (admin) tokens are unrestricted; an agent token with no resolved repo scope now sees only sessions it's directly assigned to.

dodizzle (commented):
Verdict: COMMENT — The critical agent-scope visibility bug from the prior review is genuinely fixed and well-tested, but docs/task-store.md still documents the old, incorrect unrestricted-fallback behavior.

dodizzle (approved):
Verdict: APPROVE — Both previously-flagged issues (agent-scope visibility bypass and stale docs) are verifiably fixed with test coverage; no new findings.

PR #3330, opened 2026-09-11
feat: add POST /admin/sessions/:slug/{follow,unfollow} routes

dmcaulay:
## Summary
- Add `POST /admin/sessions/:slug/follow` and `POST /admin/sessions/:slug/unfollow` to the admin UI, in a new `admin-ui-session-follow.ts` module mirroring the existing `admin-ui-sessions.ts` shape.
- Follow enforces session visibility for non-admin members via `session-scope.ts`'s `visibleAgentIdsFor`/`isSessionVisible` (admins always pass, no session lookup); an invisible session (or one that can't be resolved) returns 404 with no `SessionFollow` row written. Unfollow skips the visibility check — removing your own state doesn't require re-proving visibility.
- Extend `SessionFollowService.unfollow()` to also delete the caller's matching `SessionAlertState` row, so a stale alert-cooldown timestamp doesn't linger after unfollowing.
- Wire a real `fetchTaskStoreSession` fetcher into `main.ts` (`GET /sessions/:slug` against the task-store), following the existing `fetchTaskStorePrById` pattern.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Follow/unfollow write and remove the SessionFollow row respectively; a member following an invisible session gets 404 and no row written | ✅ MET |
| 2 | Unfollow deletes the matching SessionAlertState row | ✅ MET |
| 3 | Test decision: smoke tests via app.request() with injected session-follow-service and session-scope doubles, covering admin-follow, member-follow-visible, member-follow-invisible-404, and unfollow-clears-alert-state | ✅ MET |

## Test Plan
- `admin/src/admin-ui-session-follow.smoke.test.ts` (new) — 9 smoke tests via `app.request()` against an isolated `Hono<AdminUIEnv>`, injecting doubles for `sessionFollowService`/`agentMemberService`/`agentService`/`fetchTaskStoreSession`, exercising the real (undoubled) `session-scope.ts` functions end to end.
- `admin/src/session-follow-service.integration.test.ts` — extended with 3 cases covering `SessionAlertState` deletion on unfollow, idempotency, and per-user scoping (real Postgres, skipped without `DATABASE_URL_ADMIN_TEST`).
- `task typecheck`, `task lint` clean. `task ci`'s full suite: 8094 pass / 3 pre-existing sandbox failures unrelated to this change (`extraAllowedTools` — documented flake; 2x `check-downstream-compat.sh` — missing `python3` in this sandbox), confirmed unaffected by this diff.


dodizzle (commented):
Verdict: COMMENT — memberCanSeeSession's catch block swallows fetchTaskStoreSession failures with no logging, silently degrading an outage into indistinguishable 404s on the visibility-decision path (admin/src/admin-ui-session-follow.ts:58). Everything else — fail-closed visibility logic, test coverage across smoke/integration layers, additive API surface — checks out clean.

admin/src/admin-ui-session-follow.ts
```diff
@@ -0,0 +1,105 @@
...
+  slug: string,
+  userEmail: string,
+  deps: SessionFollowRoutesDeps,
+): Promise<boolean> {
+  if (!deps.fetchTaskStoreSession) return false;
+
+  let session: SessionForVisibility | null;
+  try {
+    session = await deps.fetchTaskStoreSession(slug);
+  } catch {
```
dodizzle: `catch { return false; }` swallows any error from `fetchTaskStoreSession` (network failure, timeout, 5xx, or a real bug) with zero logging — the follow route 404s exactly as if the session were legitimately invisible. This is inconsistent with this codebase's own convention for swallowed errors on a similar path: `push-service.ts`'s comparable catch-and-continue methods `console.error(...)` before returning. A task-store outage would silently degrade every non-admin member's follow attempts into indistinguishable 404s with no operational signal, on what is an authorization-decision code path.

Suggestion: `catch (err) { console.error("[session-follow] fetchTaskStoreSession failed:", err); return false; }` — keep the fail-closed behavior, just log before it.

dodizzle (approved):
Verdict: APPROVE — Prior finding (unlogged swallowed error in memberCanSeeSession's catch block) is fixed and covered by a new smoke test. No new issues found; visibility logic, test coverage, and API surface check out clean.

PR #3335, opened 2026-09-11
feat: add Sessions list page + toolbar link + member-scoped detail route

dodizzle:
## Summary
- Add `GET /admin/sessions` — a new Sessions list page grouping all sessions into Waiting on you / Active / Closed (plus an Archived filter), with `state`/`repo`/`agent`/`q`/`sort` query filters and limit/offset pagination, backed by a new `fetchTaskStoreSessions` dependency on `AdminUIDeps`.
- Add member visibility scoping to both the new list page and the existing `GET /admin/sessions/:id` route, reusing SESH-4.1's `session-scope.ts`: an admin sees every session; a member sees only sessions tied to one of their agents or those agents' repos, and a member with zero memberships sees an empty page (list) or a 404 (detail) rather than an error. The detail route's old flat `if (!isAdmin) 403` gate is replaced with this scope-aware check (non-visible → 404, not 403, so existence isn't leaked).
- Add the Sessions link to the shared toolbar (`lib/web/toolbar.ts`), between Tasks and PRs.

**Note on file naming:** the brief called for `admin/src/admin-ui-sessions.ts`, but that filename is already used by an unrelated, already-merged feature (the per-user notification-settings page, SESH-6.3). To avoid a collision, the new list page lives in `admin/src/admin-ui-sessions-list.ts` instead; the pre-existing file is untouched.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | GET /admin/sessions as admin renders all sessions grouped Waiting/Active/Closed, waiting sorted oldest-first | MET |
| 2 | Member of agent A only sees only sessions tied to A/A's repos; zero-membership member sees empty page, no error | MET |
| 3 | /admin/sessions/:id returns the page for a visible member, 404 when not visible | MET |
| 4 | Sessions toolbar link present and highlighted on /admin/sessions* | MET |
| 5 | Smoke tests with injected doubles covering admin/member/no-membership + :id gate; toolbar tests extended; old flat-403 test replaced | MET |

## Test Plan
- New `admin/src/admin-ui-sessions-list.smoke.test.ts`: admin sees all three sections (sorted `waitingSince`); archived sessions excluded by default, shown only under `?archived=true`; member scoping via agent/repo match; zero-membership member never calls the task store and gets an empty page; degraded banner when the fetcher is absent or throws; Follow stub renders.
- `admin/src/admin-ui.smoke.test.ts` extended: `:id` scope gate — visible member (200), non-visible member (404), zero-membership member (404), admin always (200).
- `lib/web/toolbar.unit.test.ts` extended: Sessions link present between Tasks and PRs, highlights on `/admin/sessions` and `/admin/sessions/:slug`.
- `task typecheck`, `task lint` clean. `task ci` runs the full gate through `test:coverage`; the only 2 failing tests there are pre-existing on `main` (`task-store/src/routes/prs.ts`'s `validateRepo` null-vs-undefined guard), unrelated to this change and not touched by this diff.


admin/src/admin-ui.ts
```diff
@@ -3212,6 +3241,37 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
...
+      c.var.userEmail,
+      agentMemberService,
+      agentService,
+    );
+    if (scope.agentIds !== "all") {
+      const derived: SessionForVisibility = {
+        agentIds: [
+          ...new Set(
+            tasks
+              .flatMap((t) => [t.assignee, t.claimedBy])
```
dmcaulay: This derives `agentIds` as `tasks.flatMap(t => [t.assignee, t.claimedBy])` (both fields, unioned), but `task-store/src/session-rollup.ts` — which produces the `agentIds` the *list* route's visibility check consumes — uses `t.claimedBy ?? t.assignee` (one id per task, `claimedBy` wins). This route's set is always a superset. Example: a task with `assignee: "agent-a"`, `claimedBy: "agent-b"` yields rollup `agentIds: ["agent-b"]` (list page correctly hides it from agent-a's member after reassignment), but this route's derived set still includes `"agent-a"`, so that member can view the session directly here. The comment above (admin-ui.ts:397) claims both routes "apply the exact same visibility rule" — this contradicts that. Suggest deriving with the same `t.claimedBy ?? t.assignee` precedence, ideally via a shared helper.

admin/src/admin-ui.ts
```diff
@@ -3212,6 +3241,37 @@ export function createAdminUIApp(deps: AdminUIDeps): Hono<AdminUIEnv> {
...
     }
 
+    // SESH-4.2: replaces the old flat `if (!isAdmin) 403` gate with a
+    // session-scope.ts-based visibility check — an admin always sees the
+    // page; a member sees it only when one of the session's own tasks is
+    // assigned/claimed by one of their agents, or is in one of those
+    // agents' repos. Derived from the tasks already fetched above (no
+    // second fetch). A session outside the caller's scope 404s rather than
+    // 403s, so its existence isn't leaked to callers who can't see it.
+    const scope = await resolveVisibilityScope(
```
dmcaulay: This visibility check runs unconditionally regardless of `degraded` (set above when `fetchTaskStoreTasks` is absent/throws). When degraded, `tasks` is `[]`, so the derived scope is always empty and every non-admin member gets 404'd on every session — even ones they legitimately own — during a transient task-store outage. Every other degraded-mode route in this file (task list, PR list, public task board) instead renders 200 with an "unavailable" banner rather than denying access. Consider skipping this check (or showing the degraded banner) when `degraded` is true, since scope can't be verified either way.

admin/src/admin-ui-sessions-list.ts
```diff
@@ -0,0 +1,415 @@
...
+    // fetched page, on top of (not instead of) the user's own repo/agent/q
+    // filters already forwarded above.
+    if (scope.agentIds !== "all") {
+      sessions = sessions.filter((s) =>
+        isSessionVisible({ agentIds: s.agentIds, repos: s.repos }, scope),
+      );
+    }
+
+    return deps.html(
+      renderSessionsListPage(sessions, filters, degraded, userEmail, {
```
dmcaulay, line 434: For a scoped member, `sessions` was just filtered client-side by `isSessionVisible`, but `total`/`limit`/`offset` here are still the task-store's pre-filter values. The rendered "X–Y of Z" summary and Prev/Next controls will misstate the actual visible-row count for members, and pagination can land on pages with few or zero visible rows while implying more exist. Consider recomputing total/pagination from the filtered set for scoped members, or relabeling the count/controls to avoid implying an inaccurate total.

dmcaulay (commented):
Verdict: COMMENT — the detail route's `agentIds` derivation diverges from the canonical session-rollup logic used by the list route (allowing over-broad access for reassigned tasks), and its degraded-mode handling 404s legitimate members during transient task-store outages instead of degrading gracefully like every other route in this file. List-page pagination metadata also isn't adjusted for member-side filtering. Well tested otherwise — inline comments below.

dmcaulay (approved):
Verdict: APPROVE — all three prior findings (agentIds precedence, degraded-mode 404, pagination metadata) are resolved with verifiable evidence and matching regression tests. Independent re-review of the fix commit found no new issues.

PR #3337, opened 2026-09-11
feat: add session-alert-sweeper.ts and wire into main.ts

dmcaulay:
## Summary
- Adds `admin/src/session-alert-sweeper.ts` — `SessionAlertSweeper`, the admin service's first background loop. Every tick it fetches waiting/closed sessions from the task-store, materializes auto-follows, sends immediate/reminder/completed push notifications with per-user-per-day dedup, and prunes alert-state rows for followers who've lost visibility.
- Registers the sweeper via `setInterval` in `admin/src/main.ts` (default 60s, configurable via `SHIPWRIGHT_ADMIN_SESSION_ALERT_INTERVAL_MS`), gated on push being enabled and the task-store admin client being configured.
- Mirrors `chat/src/stall-reaper.ts`'s pattern: injected `Clock`, per-row try/catch so one bad session/follower can't abort the sweep, and a counted return value.

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Session entering waiting → exactly one push per subscribed follower on first tick, none on next | MET |
| 2 | Clock past reminderHourLocal next day → one reminder; second same-day tick → none | MET |
| 3 | Closed session → one completed push, no further pushes; fetch rejection → tick() returns all-zero, sends nothing | MET |
| 4 | Follower loses membership visibility → no push, alert-state row pruned | MET |
| 5 | Unit tests over injected doubles + FixedClock, no mock.module()/global.fetch; notifyThreadReply regression guard untouched | MET |

## Test Plan
- 21 new unit tests in `admin/src/session-alert-sweeper.unit.test.ts` — one per acceptance-criteria branch plus resilience/isolation cases. 100% line and function coverage on the new file.
- 3 new unit tests in `admin/src/main.unit.test.ts` for `resolveSessionAlertIntervalMs`.
- `notifyThreadReply` and its existing test suite (SES-7.1) are untouched — confirmed still green.
- [x] `task lint` clean
- [x] `task typecheck` clean


dodizzle (commented):
Verdict: COMMENT — Solid sweeper (DST-correct date math, substantive test coverage, no breaking changes), but two behavioral issues and one config-plumbing gap are worth resolving before merge: (1) auto-follow materializes on any visible waiting session, not just "sessions I start" as the toggle claims; (2) the unguarded setInterval lets overlapping ticks double-send pushes, violating AC1's exactly-once guarantee at runtime; (3) SHIPWRIGHT_ADMIN_PUSH_MAX_DETAIL=preview is silently ignored for session alerts since detailLevel isn't passed through. See inline comments.

admin/src/session-alert-sweeper.ts
```diff
@@ -0,0 +1,552 @@
...
+
+    const alertByEmail = new Map(alertRows.map((row) => [row.userEmail, row]));
+    const followers = followRows.map((row) => ({
+      userEmail: row.userEmail,
+      muted: row.muted,
+    }));
+    const followedEmails = new Set(followers.map((f) => f.userEmail));
+
+    // Materialize auto-follows for users who can see this session.
+    for (const userEmail of ctx.autoFollowEmails) {
```
dodizzle, line 442: Auto-follow materializes for every user with `autoFollowSessions = true` on any waiting session merely *visible* in their agent/repo scope — not sessions they started. The toggle reads "Automatically follow sessions I start" (`admin/src/admin-ui-sessions.ts:127`) and the schema comment defines it the same way, but this loop never reads `autoFollowSince` (which the schema comment says exists specifically to distinguish "always on" users from users who opted in later, for exactly this kind of backfill). Since `autoFollowSessions` defaults to `true`, the first tick after deploy will auto-follow and immediately push every eligible user for the entire pre-existing backlog of waiting sessions, and going forward for every session started by anyone else in their scope.

Suggest gating on `autoFollowSince` (skip sessions already waiting before opt-in), or updating the toggle copy if "any visible session" is the intended behavior.

admin/src/main.ts
```diff
@@ -663,6 +688,46 @@ async function startServer(): Promise<void> {
...
+            `task-store GET /sessions?state=${state} → ${res.status}`,
+          );
+        }
+        const body = (await res.json()) as { sessions?: SessionForAlert[] };
+        return body.sessions ?? [];
+      },
+      ...(adminTz ? { timezone: adminTz } : {}),
+    });
+    const sessionAlertIntervalMs = resolveSessionAlertIntervalMs(process.env);
+    setInterval(() => {
```
dodizzle, line 737: This `setInterval` has no re-entrancy guard, so a tick that runs longer than `intervalMs` overlaps with the next one. Unlike `chat/src/stall-reaper.ts` (whose docstring explicitly calls out that its sweep is idempotent/race-safe and safe to run concurrently), this sweeper's dedup is a read-then-write across network I/O (`sessionAlertState.findMany()` → `pushService.notifySession()` → `stampAlertState()`), so two overlapping ticks can both observe `lastAlertedAt = null` and both push — violating AC1's "exactly one push per follower" at runtime, even though the sequential-only unit tests can't catch it. The injected `fetchSessions` has no timeout, so a slow task-store response is a realistic overlap trigger; this would also double-send at >1 admin replica.

Suggest adding an in-flight guard around the tick callback and a timeout on the task-store fetch.

admin/src/main.ts
```diff
@@ -663,6 +688,46 @@ async function startServer(): Promise<void> {
...
   root.route("/", adminUIApp);
 
+  // Session alert sweeper (SESH-7.4) — the admin service's only background
+  // loop. Registered HERE, never inside createAdminUIApp(), which must stay
+  // side-effect-free (same rule as chat/src/stall-reaper.ts's registration).
+  // Gated on BOTH push being enabled and the task-store client being
+  // configured: without either, every tick would be a guaranteed no-op (or a
+  // guaranteed error log), so we simply don't start it.
+  if (pushService && taskStoreUrl && taskStoreAdminToken) {
+    const sessionAlertSweeper = new SessionAlertSweeper({
```
dodizzle, line 706: `pushMaxDetail` (resolved from `SHIPWRIGHT_ADMIN_PUSH_MAX_DETAIL` at line 642 and passed to `PushService`) isn't passed here as `detailLevel`, so the sweeper falls back to `DEFAULT_MAX_DETAIL` ("title"). Since `notifySession()` computes `resolveDetailLevel(level, subLevel)`, that hardcoded "title" becomes a stricter cap than the operator's configured ceiling — an operator set to `preview` still only gets `"title"` on session pushes, contradicting the documented `min(ceiling, opt-in)` contract in `docs/configuration.md`. Currently latent since `notifySession`'s payload is a placeholder, but will surface once real content lands.

Suggest passing `detailLevel: pushMaxDetail as PushDetailLevel` here, same as `PushService` receives it.

dmcaulay:
Addressed all three review findings in b5973d9a. One note where I deviated from the suggested prescription:

**Auto-follow scope (`session-alert-sweeper.ts:345`)** — the premise held, so I did both halves of the suggestion rather than picking one, because neither alone was sufficient:

- *Gate on `autoFollowSince`*: implemented via a new pure helper `startedWaitingBefore(session, since)`, which skips auto-following any session already waiting before the user's opt-in. But `autoFollowSince` was never actually written by anything — `updatePrefs()` only ever set `autoFollowSessions`/`reminderHourLocal`, so the column was permanently null and the gate would have been dead code. So `SessionFollowService.updatePrefs()` now stamps it (via an injected `Clock`) on a genuine off→on transition; re-saving the form with the box already ticked does not move an existing boundary. A null `autoFollowSince` stays the documented "always on / never explicitly opted in" cohort and keeps today's behavior.
- *Update the toggle copy*: "sessions I start" isn't implementable — the task-store `Session` model has no creator/owner field, so "who started it" isn't knowable from `SessionForAlert`. The label now reads "Automatically follow new sessions I can see", with helper text spelling out that it covers any session in the agents and repos you have access to. The `schema.prisma` and `docs/agent.md` descriptions were corrected to match.

The other two findings were implemented as suggested: `tick()` now carries an in-flight guard (returns all-zero rather than starting a concurrent sweep, with the docstring contrasting this against `stall-reaper.ts`'s idempotent sweep), `main.ts` bounds the task-store fetch with `AbortSignal.timeout(SESSION_ALERT_FETCH_TIMEOUT_MS)` (10s), and `pushMaxDetail` is threaded through as `detailLevel` — normalized in the constructor rather than cast, since it's an unvalidated env string.

New coverage: 5 cases for `startedWaitingBefore`, 3 for the sweeper-level opt-in boundary, 2 for re-entrancy, 2 for detail-level plumbing, and a new `session-follow-service.unit.test.ts` with 5 cases for the `autoFollowSince` stamping rule.

admin/src/session-alert-sweeper.ts
```diff
@@ -0,0 +1,638 @@
...
+      if (followedEmails.has(userEmail)) continue;
+      // Opt-in boundary: a user who turned auto-follow on at `autoFollowSince`
+      // is not retro-followed onto sessions that were already waiting before
+      // that moment. schema.prisma documents the column for exactly this
+      // ("so a backfill can distinguish 'always on' users from users who
+      // opted in later") — a null value is the always-on cohort and has no
+      // boundary to apply.
+      if (
```
dodizzle, line 449: The `autoFollowSince` gate only fires when the value is non-null, but it's only ever stamped by `updatePrefs()` on an explicit off→on save. `autoFollowSessions` defaults to `true` and `autoFollowSince` defaults to `null` in `schema.prisma`, and a prefs row is created just by loading the settings page (`admin-ui-sessions.ts:181`) — so essentially every push-subscribed user is in the null/"always-on" cohort with no boundary at all. On the first tick after deploy, every visible waiting session back-follows and immediately pushes each of them — the exact backlog-burst outcome this fix's own test comments (`session-alert-sweeper.unit.test.ts:914-915`) argue against, for users who never explicitly opted in.

Consider backfilling `autoFollowSince = now()` for existing prefs rows in a migration, or treating a null value as "boundary = first sweep" rather than "no boundary."

dodizzle (commented):
Verdict: COMMENT — re-review after b5973d9a: all three prior findings (auto-follow scope, tick re-entrancy, detailLevel plumbing) are genuinely fixed with tests. One new finding: the opt-in boundary only applies to users who explicitly re-saved the toggle — the default "always-on" cohort (autoFollowSince still null) has no boundary, so first tick after deploy can back-follow and immediately push the entire pre-existing waiting backlog to every push-subscribed user.

dmcaulay:
Addressed the auto-follow boundary finding in 4792ad07. Confirmed the premise: `autoFollowSince` is only stamped on an explicit off→on save in `updatePrefs()`, `autoFollowSessions` defaults to `true`, and a prefs row is created just by loading the settings page (`admin-ui-sessions.ts:181`) — so essentially every push-subscribed user had no boundary.

I took **both** suggested options rather than either/or, because each alone leaves a hole:

- **Migration** (`20260911000000_backfill_auto_follow_since`) — data-only `UPDATE ... SET autoFollowSince = CURRENT_TIMESTAMP WHERE autoFollowSince IS NULL`. Pins pre-existing rows' boundary at deploy time, which is stricter than anything derivable from existing data and kills the one-time backlog burst.
- **`autoFollowBoundary()` in the sweeper** — falls back to the prefs row's own `createdAt` when `autoFollowSince` is null. Without this, the migration only fixes rows that exist *today*: the next user to open the settings page gets a fresh null row and the same burst, just scoped to them. `createdAt` is the earliest moment auto-follow can meaningfully be "on" for that user, so it's the natural implicit boundary.

Net effect: every auto-follow user has a boundary at all times, and `null` at the `startedWaitingBefore()` layer now only means "no prefs row at all" — a user who can't be in `autoFollowEmails` in the first place.

Explicit follows are untouched by either change; the boundary still only gates *materializing* an auto-follow, never an alert on an existing follow.

Tests: added `autoFollowBoundary` unit coverage (stamp wins / createdAt fallback / undefined → null) and a sweeper-level test that a null-`autoFollowSince` user with a prefs row created mid-backlog follows only the newer session. Reframed the old "always-on cohort" sweeper test, which still passes — its prefs row is seeded at EPOCH, so its fallback boundary predates both sessions.

`task lint` clean; `task test` has 3 pre-existing sandbox failures (2 × `check-downstream-compat` needing `python3`, 1 × the known `extraAllowedTools` flake), all reproduced with my changes stashed.

dodizzle (approved):
Verdict: APPROVE — Re-review after 4792ad07: both prior findings (auto-follow scope/re-entrancy/detailLevel plumbing, and the unstamped always-on cohort's missing opt-in boundary) are genuinely fixed with matching regression tests (createdAt fallback in autoFollowBoundary() + a backfill migration stamping existing rows). No new issues found; all acceptance criteria verified against the test file; no breaking changes.

PR #3352, opened 2026-09-11
feat: dev-task guard (fallback safety net) for PRD-shaped tasks

dmcaulay:
## Summary
- Adds a "PRD-Shaped Task Guard" to `dev-task.md`'s Step 1: detects tasks that are actually PRD-shaped (misrouted to `dev-task` instead of `plan-session`) via three heuristics — task id prefix `prd-`, description opening with the known plan-session handoff marker, or an empty-acceptance-criteria task on `main`.
- On match, PATCHes the task to `status: blocked`, `hitl: true`, `blockedReason: "misrouted_needs_plan_session_not_dev_task"` and stops before Step 2's claim — mirroring the existing "no branch field" guard's PATCH-then-stop pattern.
- This is a fallback safety net for legacy/unflagged PRD tasks; once PDR-2.1/PDR-2.2 ship, properly-flagged `autonomousPlanSession:true` tasks are excluded upstream at the `?ready=true` filter and never reach this guard — no special-casing of that flag was added here.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Detection check added right after the status == "pending" validation in dev-task.md Step 1, before Dependency Check/claim | MET |
| 2 | On match: PATCH sets status: "blocked", hitl: true, blockedReason: "misrouted_needs_plan_session_not_dev_task"; command stops, no claim, no build attempted | MET |
| 3 | A properly-flagged autonomousPlanSession: true task is never reached by this guard once PDR-2.2 ships -- no special-case needed here | MET |
| 4 | Test: extend dev-task.content.test.ts pinning the detection heuristic and exact blockedReason string (content layer, matching this file's existing convention) | MET |

## Test Plan
- Added 11 content-layer tests to `dev-task.content.test.ts` pinning: section placement, all three detection conditions, the exact PATCH body fields, curl shape, and the printed warning message.
- `bun test plugins/shipwright/commands/dev-task.content.test.ts` — 85/85 pass.
- `task lint`, `task typecheck`, `task check-config-docs`, `task check-version-sync` all pass.

### DIFF
--- a/plugins/shipwright/commands/dev-task.content.test.ts
+++ b/plugins/shipwright/commands/dev-task.content.test.ts
@@ -684,3 +684,80 @@ describe("dev-task.md Step 1 — repo-slug derivation for local paths (PRF-1.4)"
     );
   });
 });
+
+describe("dev-task.md Step 1 — PRD-shaped task guard (fallback safety net) (PDR-1.1)", () => {
+  const getGuardSection = () => {
+    const guardIdx = content.indexOf("### PRD-Shaped Task Guard");
+    expect(guardIdx).toBeGreaterThan(-1);
+    const dependencyCheckIdx = content.indexOf("### Dependency Check (pending tasks only)");
+    expect(dependencyCheckIdx).toBeGreaterThan(guardIdx);
+    return { guardIdx, dependencyCheckIdx, section: content.slice(guardIdx, dependencyCheckIdx) };
+  };
+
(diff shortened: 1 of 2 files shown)


dodizzle (approved):
Verdict: APPROVE — Clean, well-scoped fallback safety net; heuristics are grounded in a real incident and safe against false positives, tests pin the intended behavior, no blocking concerns.

PR #3361, opened 2026-09-12
feat: wire autonomous plan-session dispatch phase into loop-orchestrator

dodizzle:
## Summary
- Adds `agent/src/check-plan.ts`, a candidate provider for the autonomous plan-session phase, querying `GET /tasks?autonomousPlanSession=true&status=pending` and mirroring `check-dev-task.ts`'s Deps/mapper/buildProductionDeps shape.
- Wires a 5th `"plan"` `LoopPhase` into `agent/src/loop-orchestrator.ts` — the `LoopPhase` union, `PHASE_COMMANDS`, the phase-toggle read logic (`loop-cron-classifier.ts`), and the candidate-merge/strict-FIFO selection (`work-selector.ts`) all updated to account for it, dispatching `/shipwright:plan-session {repo} {session} --autonomous {task-id}` via the same atomic `POST /tasks/{id}/claim` mechanics as the other phases.
- Entirely gated behind `SHIPWRIGHT_AGENT_AUTONOMOUS_PLAN_SESSION_ENABLED` (code-level kill switch, ANDed with the manifest's `shipwright-plan` cron toggle) — with the env var unset, behavior is byte-for-byte unchanged from today.
- Ships `enabled: false` in `agent-types/coding/manifest.yaml` per the "new system crons always ship disabled" convention.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Env var unset → byte-for-byte unchanged; existing tests pass unmodified | MET |
| 2 | Env var set → flagged task claimed & invoked exactly once, atomic claim under concurrency | MET |
| 3 | New phase ships enabled: false in manifest.yaml | MET |
| 4 | Test: extended loop-orchestrator.unit.test.ts + new check-plan.unit.test.ts | MET |

## Test Plan
- New `agent/src/check-plan.unit.test.ts` mirroring `check-dev-task.unit.test.ts`'s shape.
- Extended `agent/src/loop-orchestrator.unit.test.ts` with a new "autonomous plan phase" suite covering: env-var-unset no-op, env-var-set dispatch/claim/concurrency, FIFO merge with dev-task, and defensive candidate-skipping.
- Extended `work-selector.unit.test.ts` and `loop-cron-classifier.unit.test.ts` for the new phase field/toggle.
- `task typecheck` / `task lint` — clean.
- `bun test` — 8308 pass, 2 pre-existing failures unrelated to this diff (confirmed by file-scope: `task-store/src/routes/prs.ts` validateRepo null-repos bug, and an env-dependent/pre-existing `agent/src/claude.unit.test.ts` flake — neither file touched by this branch).
- Coverage: `loop-cron-classifier.ts`/`work-selector.ts` 100%, `loop-orchestrator.ts` 91.6%, `check-plan.ts` 60.9% (mirrors `check-dev-task.ts`'s existing ~52% baseline for the same untested `buildProductionDeps` wiring pattern). Repo-wide aggregate coverage gate is below threshold pre-existing to this change (DB-backed task-store services skip locally).


agent/src/loop-orchestrator.ts
```diff
@@ -937,6 +1009,25 @@ export function createLoopOrchestrator(
...
+        // age-based FIFO below picks the winner across both. Gated behind
+        // BOTH the shipwright-plan cron toggle and the
+        // SHIPWRIGHT_AGENT_AUTONOMOUS_PLAN_SESSION_ENABLED kill switch (read
+        // fresh every iteration, never cached, so an operator's env change
+        // lands on the very next tick). With the env var unset this
+        // condition is always false, so zero new candidates ever enter the
+        // pool and every pre-PDR-4.1 code path below is untouched.
+        const planEnabled =
+          toggles.plan && readBooleanEnv(AUTONOMOUS_PLAN_SESSION_ENV);
+        if (planEnabled && getPlanCandidates) {
```
dmcaulay: Plan candidates duplicate dev-task candidates in the merged pool. `check-plan.ts` queries `?autonomousPlanSession=true&status=pending`, but `check-dev-task.ts`'s `?ready=true` query never passes `autonomousPlanSession`, so the task-store doesn't filter it out there either — a flagged, pending, hitl-false task with satisfied deps is returned by both providers into the same `tasks` array. Since both copies share the identical `createdAt` and `selectNextWorkItem` uses a strict `<` comparison, the first occurrence (dev-task, pushed above this block) always wins the tie, so the flagged task gets dispatched as `/shipwright:dev-task` and never reaches plan-session. Consider adding `autonomousPlanSession: "false"` to check-dev-task.ts's ready query, or dedupe the merged pool by id with the plan-tagged copy winning.

agent/src/check-plan.ts
```diff
@@ -0,0 +1,112 @@
...
+  if (!agentId) {
+    process.stderr.write("error: SHIPWRIGHT_AGENT_ID is required\n");
+    process.exit(1);
+  }
+
+  return {
+    getAutonomousPlanTasks: () =>
+      client.query(
+        new URLSearchParams({
+          autonomousPlanSession: "true",
```
dmcaulay, line 131: No hitl/blocked gate here. Every other candidate provider refuses to dispatch a human-escalated task — check-review.ts/check-patch.ts/check-deploy.ts apply `isTaskBlockedForDispatch()`, and check-dev-task.ts gets the equivalent for free via `ready.ts` dropping `hitl === true`. This query (`autonomousPlanSession=true&status=pending`) doesn't filter on hitl, so a task a human escalated with `hitl: true` while leaving it `pending` would still be claimed and dispatched into an autonomous plan session. Suggest adding `hitl: "false"` to the query (or filtering with `isTaskBlockedForDispatch()` in the mapper), plus a unit test for the hitl:true case.

dmcaulay (commented):
Verdict: COMMENT — AC1 (byte-for-byte unchanged when unset) and AC3 (manifest ships disabled) are cleanly met, but AC2 isn't: enabling the phase 400s the admin work-queue snapshot (phase enum doesn't include "plan"), plan-flagged tasks get raced away by dev-task's FIFO tie-break, and check-plan.ts has no hitl gate. Details inline + in the review body.

dodizzle:
Addressed in ee32914f. Both findings were correct; I deviated from the suggested *prescription* in both cases for the same underlying reason, so flagging it explicitly:

- **`Task.autonomousPlanSession` and `Task.hitl` are both nullable** (`Boolean?`, no default — see `task-store/prisma/schema.prisma:77-78`), and the task-store filters both by strict equality in *both* code paths (`matchesTaskFilters`'s `!==` for the ready/blocked branches, and the Prisma `where` for the plain list branch). So `?autonomousPlanSession=false` on check-dev-task's ready query would have dropped every task whose column is still `NULL` — i.e. the entire pre-PDR-2.1 queue — and `?hitl=false` on check-plan's query would have done the same there.

Applied instead:

1. **Duplicate candidates** — took the alternative you offered: dedupe the merged pool by id in `loop-orchestrator.ts`, plan-tagged copy winning. Dev-task candidates are now filtered against the plan candidates' id set before the two lists are concatenated.
2. **hitl gate** — took your parenthetical: `getPlanCandidates` now filters with `isTaskBlockedForDispatch()` in the mapper, same helper `check-review`/`check-patch`/`check-deploy` use. Covers `hitl === true` and `status === "blocked"` (the latter defense-in-depth, since the query is already `status=pending`).
3. **Work-queue snapshot 400** — confirmed and fixed: `"plan"` added to `RankedWorkItemSchema`'s phase enum (`admin/src/openapi-schemas.ts`), with `admin/openapi.json` + `lib/admin-types.ts` regenerated, plus the admin-ui `WorkQueueItem` type/badge palette and `docs/agent-api.md`.

Tests added: 3 hitl/blocked cases in `check-plan.unit.test.ts`, 4 dedupe cases in `loop-orchestrator.unit.test.ts` (dispatch routing, single claim, snapshot shape, non-duplicate untouched), and a `phase: "plan"` 200 case in `agents-api.smoke.test.ts`. `task lint`, `task typecheck`, and `bun test` all clean (the 2 `prs.openapi.smoke.test.ts` failures in a full-suite run are the known pre-existing order-dependent `validateRepo` null-repos issue on `main` — they pass in isolation and this PR doesn't touch `task-store/`).

dmcaulay (commented):
Verdict: COMMENT — both prior findings (candidate-pool dedupe, hitl gate) plus the unmentioned admin work-queue phase-enum 400 are genuinely fixed and test-backed; new issue: docs/agent-ops.md and cron-jobs.mdx claim the standalone-cron case is silently inert for all five phases, but plan-session.md has no no-argument [silent] guard like the other four commands.

docs/agent-ops.md
```diff
@@ -51,21 +51,23 @@ On the agent's next config sync, Claude will no longer have access to shell exec
...
-for its own work — each responds `[silent]` and exits immediately if invoked with no target.
+The five pipeline crons (`shipwright-plan`, `shipwright-dev-task`, `shipwright-review`,
+`shipwright-patch`, `shipwright-deploy`) are **loop-driven, item-addressed executors, not
+self-discovering standalone crons.** They are linked as child crons of `shipwright-loop` (via
+`parentCronId`), which is the sole supported driver for these five phases: its in-process
+candidate providers (`agent/src/check-plan.ts`, `check-dev-task.ts`, `check-review.ts`,
+`check-patch.ts`, `check-deploy.ts`) select a winning candidate each tick and
+`loop-orchestrator.ts` dispatches the matching command with an explicit task id or
+`org/repo#number` embedded in the prompt. None of the five commands scans for its own work —
+each responds `[silent]` and exits immediately if invoked with no target.
```
dmcaulay, line 62: This claims "each responds `[silent]` and exits immediately if invoked with no target" for all five pipeline commands, and `site/src/content/docs/cron-jobs.mdx` adds "Inert if `shipwright-loop` is disabled" for the new `shipwright-plan` row. True for the four pre-existing phases (`plugins/shipwright/commands/dev-task.md` has an explicit no-argument `[silent]` guard), but `plugins/shipwright/commands/plan-session.md` has no such guard — with zero arguments it falls into its auto-detect path (parses `git remote get-url origin`, prints a warning, waits for user confirmation) rather than going inert. The new classifier test at `agent/src/loop-cron-classifier.unit.test.ts:205` itself asserts a top-level `shipwright-plan` row with no parent still gets `dispatch: "generic"` — exactly the untargeted-dispatch path this doc claims can't happen. Reachability is low today (manifest reconciliation always sets `parentCronId`, which the classifier skips), so this is a doc/safety-invariant gap rather than a live bug, but it's the exact wording an operator reads before flipping the toggle. Suggest either adding a no-argument `[silent]` guard to `plan-session.md` (matching `dev-task.md`'s), or narrowing this claim to exclude `shipwright-plan`.

dmcaulay (commented):
Verdict: COMMENT — both prior review cycles' findings are fully resolved and test-backed (all 4 acceptance criteria now met, including the no-argument [silent] guard on plan-session.md); two new documentation-drift findings: plugins/shipwright/CLAUDE.md still describes four loop phases/candidate providers (and is cited by check-plan.ts itself as the governing contract), and docs/agent-key-files.md's work-selector.ts row doesn't mention the widened WorkTaskCandidate/RankedWorkItem phase field.

docs/agent-key-files.md
```diff
@@ -68,6 +68,7 @@
...
...nt; status === "pending" and dependencies guaranteed server-side) and a list of ready PRs (pre-filtered by the check-* collectors' own `?ready=true`/claim-filtering, per LPF-2.2), returns the single oldest item across both types by age — strict age-based FIFO, no phase-priority bias, no local status/claim re-validation on either side. Also exports `rankWorkItems(tasks, prs) → RankedWorkItem[]`, which ranks ALL ready items (both tasks and PRs) oldest-first, returning a flat array of `RankedWorkItem` objects; ordering matches calling `selectNextWorkItem()` repeatedly and removing each winner from its source list (both apply the same age-based comparison). Tie-breaks on equal age by tasks-before-PRs, then original input order within each group (deterministic, stable-sort-based). Exports types `WorkTaskCandidate` (id, createdAt, title?), `WorkPrCandidate` (id, age, phase?, title?, commitSha), `WorkItem` (discriminated union: `{ type: "task"; task }` or `{ type: "pr"; pr }`), and `RankedWorkItem` (type, id, title?, phase, age). Unit-testable via fixtures; `selectNextWorkItem()` is used by `loop-orchestrator.ts` for single-winner selection after redispatch-cooldown filtering (CBD-2.2). |
```
dmcaulay: This PR adds `phase?: "dev-task" | "plan"`, `repo?`, and `session?` to `WorkTaskCandidate` and widens `RankedWorkItem["phase"]` to include `"plan"` (`agent/src/work-selector.ts`), but this row still documents the old shapes verbatim — "Exports types `WorkTaskCandidate` (id, createdAt, title?)" and "`RankedWorkItem` (type, id, title?, phase, age)" — with no mention of the plan phase. It also still describes `selectNextWorkItem`'s task input as exclusively "ready tasks (pre-filtered by task-store's `?ready=true` endpoint)", which is no longer true once the plan pool (sourced from `?autonomousPlanSession=true&status=pending`, deliberately not `ready`-filtered) is merged in. Worth updating since this same PR edits an adjacent row (`check-plan.ts`) two lines above.

dodizzle:
Addressed in 10fb3da5. Both round-3 findings verified against the code before implementing; both premises held.

**`plugins/shipwright/CLAUDE.md` — four phases / four providers.** Correct, and the sharpest version of the problem is the one you named: `check-plan.ts:24` cites this file's Candidate Selection Contract as the authority on what qualifies, while the contract it points at didn't know the plan phase existed. Updated the Terminology note, Independence Principles, Principle 5, the Candidate Selection Contract, and System Cron Changes from four to five, and added `check-plan.ts` to the provider list.

Two facts had no place in the four-phase text, so they got new paragraphs rather than a find-and-replace: (a) two of the five providers return *task* candidates and are merged into one pool deduped by id with the `phase: "plan"` copy winning — the non-obvious part being *why* the dedupe is needed at all (`?ready=true` never excludes flagged tasks, and both copies share a `createdAt`, so the untagged copy would win the FIFO tie); and (b) the plan phase is double-gated behind both its cron toggle and `SHIPWRIGHT_AGENT_AUTONOMOUS_PLAN_SESSION_ENABLED`, unlike the other four.

One deliberate non-change: I did **not** fold `plan-session` into the "explicit-target-only executors" sentence. It shares the no-argument `[silent]` guard, but a human passing a single `session` argument still gets the auto-detect-and-confirm flow — only the `--autonomous` mode the loop dispatches is fully explicit-target. Flattening that would have replaced one inaccuracy with another, so the paragraph now states the distinction explicitly.

**`docs/agent-key-files.md` — `work-selector.ts` row.** Correct on both counts. Documented `WorkTaskCandidate`'s `phase?`/`repo?`/`session?` and `RankedWorkItem`'s widened union (including both fallback defaults), and corrected the `?ready=true`-only claim to note the plan pool is deliberately not ready-filtered.

I extended slightly past the row you flagged: the adjacent `loop-orchestrator.ts` row carried the same drift for code this PR changes — "resolves four phase toggles", a `LoopPhase` union missing `"plan"`, and "dispatches ... with the winning item's ID appended", which is precisely what the plan phase does *not* do. Those are the same finding one row down, so fixing the flagged row and leaving them would have been odd. Added `buildPlanCommandArgs` to that row's exports while there.

**Round-2 `docs/agent-ops.md:62`** was already fixed at HEAD by 493829c3 (the no-argument `[silent]` guard on `plan-session.md`, with `plan-session.content.test.ts` pinning the guard, its placement before the auto-detect block, and consistency with the four siblings). The doc claim is accurate as written — the thread was simply never resolved. Resolved now, no further change.

No tests added: both files are prose reference docs with no content-test harness pinning them, and inventing one for a design-constitution narrative would pin wording rather than behavior. The behavioral claims these docs now describe are already covered by `loop-orchestrator.unit.test.ts`, `check-plan.unit.test.ts`, and `work-selector.unit.test.ts`.

`task lint` clean. `task test`: 8324 pass / 2 fail — both the known pre-existing `prs.openapi.smoke.test.ts` `validateRepo` null-repos failure that reproduces on `main`, untouched by this PR.

docs/agent-key-files.md
```diff
@@ -68,17 +68,18 @@
...
 | `agent/src/loop-cron-classifier.ts` | Pure cron dispatch helpers (zero I/O, mirrors `work-selector.ts` pattern): `classifyCronJobsForScheduling(jobs)` → `ScheduledCronJob[]` (decides which jobs get an independent node-cron schedule and with which `dispatch` kind: `"loop"` for the shipwright-loop job, or `"generic"` for standalone crons); when shipwright-loop is enabled, pipeline phase jobs (dev-task, review, patch, review-patch, deploy) are loop-config-only and excluded from independent scheduling; `resolveLoopPhaseToggles(jobs, loopCronId)` → `LoopPhaseToggles` (resolver for the loop orchestrator's phase-enable flags — reads dev-task, review, patch, deploy exclusively from child rows whose `parentCronId` equals `loopCronId`, matched by job name; a same-named top-level row or a child of a different parent is ignored; ignores review-patch; LPC-2.1). Exports types `CronJobLike` (minimal job view, decoupled from Prisma), `CronDispatchKind` (dispatch classification), `ScheduledCronJob<T>` (job+dispatch pair), `LoopPhaseToggles` (four-phase toggle state). Unit-testable; used by `agent/src/index.ts` cron-sync loop to classify and schedule jobs. |
```
dmcaulay: This diff adds `plan: boolean` to `LoopPhaseToggles` and a fifth `resolveLoopPhaseToggles()` read (`agent/src/loop-cron-classifier.ts`), and updates the sibling `work-selector.ts`/`loop-orchestrator.ts` rows in this same table to match — but this row still says `resolveLoopPhaseToggles` "reads dev-task, review, patch, deploy exclusively from child rows" and exports "`LoopPhaseToggles` (four-phase toggle state)". Worth updating to five phases while the adjacent rows are already being touched.

dmcaulay (commented):
Verdict: COMMENT — all three prior review cycles' findings are genuinely fixed and test-backed (verified against code, not just the author's narrative). The round-3 doc-drift fix itself left new four→five drift: docs/agent-key-files.md's loop-cron-classifier.ts row (inline comment) plus root CLAUDE.md, docs/architecture.md, and the public site's loop docs outside this diff's scope still describe four pipeline phases; plugins/shipwright/CLAUDE.content.test.ts also wasn't extended to guard the rewritten five-phase sections the way plan-session.content.test.ts and manifest.content.test.ts were.

dodizzle:
Addressed the round-4 findings (2026-09-12T03:21:49Z) in 668015f9 — the four→five phase doc sweep is now complete, not partial.

**Doc drift (all five locations named, plus three siblings in the same drift class):**

| File | What was stale |
|---|---|
| `docs/agent-key-files.md` | `loop-cron-classifier.ts` row (the unresolved inline thread) — `PIPELINE_PHASE_JOB_NAMES` and `resolveLoopPhaseToggles` now both list `plan`, and `LoopPhaseToggles` is "five-phase toggle state". Also added the `LOOP_PHASE_JOB_NAMES` export and **why it deliberately stays at four** (it is the unreconciled-agent reconciliation-gap probe; including `shipwright-plan` would make it permanently true). |
| `CLAUDE.md` (root) | Execution-loop paragraph's phase enumeration + the double-gate note. |
| `docs/architecture.md` | "the four loop-driven phases" / "these four phases" → five, named explicitly (the adjacent "five item-addressed executors" count is a *different* set that includes `merge`, so naming them was necessary to avoid a new ambiguity). Also made `merge`'s not-loop-driven status explicit. |
| `docs/agent-ops.md` | The `shipwright-loop` table row still said "(dev-task, review, patch, deploy)" — round 2 fixed the prose block above it, not this row. |
| `docs/agent-types.md` | "The core shipwright loop has four phases" + why `plan` gets no `enabled: true` exception. |
| `site/src/content/docs/the-shipwright-loop.mdx` | The public loop docs — intro, the "Read the four phase toggles" StepFlow step, the qualify/merge steps (now covering the plan candidate pool and the dedupe), and the dispatch step's command list. |
| `site/src/content/docs/day-to-day-operations.mdx` | Pipeline-cron checklist enumeration. |
| `plugins/shipwright/references/user-guide.md` | "The Shipwright loop, in one paragraph" + two incidental "all four phases" references. |
| `plugins/shipwright/skills/investigate-cron/SKILL.md` | Three "across all four phases" references. |

`docs/test-readiness/*`'s "four phases" and `plugins/shipwright/TESTING.md`/`check-test-readiness.ts`'s "4 phase artifacts" are the **test-readiness** pipeline's phases, not the loop's — deliberately left alone. Likewise `manifest.yaml`'s "the four phases below consume" comment, which literally refers to the four cron entries listed after it.

**Test coverage (`plugins/shipwright/CLAUDE.content.test.ts`):** extended from 4 to 14 tests, pinning the three rewritten sections the way `plan-session.content.test.ts` and `manifest.content.test.ts` pin theirs — the Independence Principles preamble is now sliced out and asserted on too. Guards: all five phases appear in every phase-enumerating section; both "None of the five self-scans/scans" claims state the set size as *five* (asserted positively, via a capture group, rather than by banning the string "four" — the Candidate Selection Contract legitimately says plan is gated "unlike the other four"); the terminology note lists `plan-session`; `check-plan.ts`/`getPlanCandidates`, the `--autonomous` dispatch form, the plan/dev-task dedupe, the `SHIPWRIGHT_AGENT_AUTONOMOUS_PLAN_SESSION_ENABLED` double gate, and `shipwright-plan`'s `parentCron` + ships-disabled status are all pinned.

**Validation:** `task lint` clean · `task test` 8334 pass / 2 fail (both the known pre-existing `prs.openapi.smoke.test.ts` `validateRepo` null-repos failures on `main`, untouched by this diff) · `task check-strings` and `task check-config-docs` clean · `astro build` clean (28 pages, both edited MDX files render).

dmcaulay (commented):
Verdict: COMMENT — all four prior review cycles' findings are genuinely fixed and test-backed. But the "four→five phase doc sweep is now complete, not partial" claim doesn't hold: three residual "four pipeline phases" references survive (investigate-cron/SKILL.md:19, user-guide.md:233, day-to-day-operations.mdx:91 — two leaving their own file self-contradictory), plus a stale admin-console.mdx phase enumeration and an unextended site e2e spec (docs-the-shipwright-loop.spec.ts:42). See review body for details.

dmcaulay (approved):
Verdict: APPROVE — all five prior review cycles' findings are genuinely resolved and test-backed; the substantive code (check-plan.ts, loop-orchestrator.ts, loop-cron-classifier.ts, work-selector.ts) is correct and all four acceptance criteria are met. One below-threshold nit for awareness: site/src/content/docs/cron-jobs.mdx:184 still says "the four pipeline crons" while line 84 of the same file (rewritten by this PR) says "the five pipeline crons".

PR #3373, opened 2026-09-12
fix(admin): alert allowlisted admins on followed sessions and add a Send test notification button

dodizzle:
## Summary
An admin who followed a session on the Sessions tab never received the "A session needs you" push, and nothing said why. Found while live-testing the session-management feature: the squadron session flipped to `waiting` (HITL task IH-6.1) and the sweeper silently skipped the only follower.

- **Sweeper admin fast path.** `SessionAlertSweeper` evaluated every follower as a non-admin member, so an operator on `SHIPWRIGHT_ADMIN_ALLOWED_EMAILS` with no `AgentMember` row was never visible and never pushed. It now scopes allowlisted emails to `"all"`, the same rule the Sessions tab and Follow button already apply. `main.ts` passes the allowlist through.
- **Outcome logging.** One line per push sent (`<kind> → <email> for <slug>: delivered=N pruned=M`, where `delivered=0` means the follower never enabled push on a device) and one per follower skipped as not visible, via an injected `log` so tests assert on it. A follow that never becomes a notification is no longer silent.
- **Send test notification.** New `POST /admin/push/test` (`admin-ui-push-test.ts`) sends a fixed, detail-free payload to the caller's own subscriptions and returns `{ok, delivered, pruned}`; body is ignored so nobody can page another user. Rendered as a button under the push toggle on `/admin/settings/notifications`, with a hint that following a session does not subscribe the device.

## Tests
- `session-alert-sweeper.unit.test.ts`: admin without memberships is alerted (case-insensitive), non-admin without membership is skipped and logged, admin path never consults memberships.
- `admin-ui-push-test.smoke.test.ts`: payload + target user, `delivered=0` case, caller-supplied email ignored, 503 when push disabled or unwired.
- `admin-ui-sessions.smoke.test.ts`: control rendered only when push is enabled.

`task lint`, admin `typecheck`, `check-config-docs`, `check-version-sync` pass. `bun test`: 4 failures identical on clean `main` (local browser-launch / WORKSPACE_PATH env), none in touched files. `check-strings` / `secret-scan` hits are pre-existing local `state/` and planning files, none in this diff.

https://claude.ai/code/session_01NxhcLzqkhTUbN6Ln1haw2b


admin/src/admin-ui-push-test.ts
```diff
@@ -0,0 +1,101 @@
...
+  delivered: number;
+  pruned: number;
+}
+
+/** Registers POST /admin/push/test onto `app`. */
+export function registerPushTestRoute(
+  app: Hono<AdminUIEnv>,
+  deps: PushTestRouteDeps,
+): void {
+  app.post(PUSH_TEST_PATH, deps.requireAuth, async (c) => {
```
dmcaulay, line 51: This route is registered with only `deps.requireAuth` — it never checks `c.var.isAdmin`. Every other route in `registerSessionSettingsRoutes` (admin-ui-sessions.ts:202,219,275) and the structurally identical push subscribe/unsubscribe routes (admin-ui.ts:4058,4087) explicitly 403 non-admins after `requireAuth`. `completeLogin` confirms non-admin sessions are real (any authenticated `AgentMember` gets `isAdmin: false`), so an authenticated non-admin can call this endpoint directly even though the settings page it's rendered on 403s them. Impact is limited — the body is correctly ignored so it can only push-test the caller's own subscriptions — but it's a clear deviation from this file's own convention, and the new smoke test hardcodes `isAdmin: true` in every case so this path is untested. Suggest adding the same `if (!c.var.isAdmin) return new Response("Forbidden", { status: 403 });` guard plus a non-admin 403 test.

dmcaulay (commented):
Verdict: COMMENT — the sweeper allowlist fix is correct and well-tested, but POST /admin/push/test is missing the isAdmin check every sibling route in the same file enforces.

dmcaulay (approved):
Verdict: APPROVE — the previously flagged missing isAdmin check on POST /admin/push/test is fixed (admin-ui-push-test.ts:52) and covered by a new 403 test. No new issues in the rest of the diff.

PR #3375, opened 2026-09-12
feat: fire task.write from single-task mutation paths

dmcaulay:
## Summary
- Wires the TSW-1.1 injected `webhookDispatcher` into `TaskService`'s eight single-task mutation methods (`create`, `update`, `claim`, `complete`, `fail`, `release`, `recordSkip`, `resetSkip`) — each fires `type: "task.write", data: [task]` right before its `$transaction` callback returns.
- A thrown `WebhookDeliveryError` propagates uncaught out of the transaction callback, so Prisma automatically rolls back the write and the original call rejects with the same error (mapped to HTTP 502 by the existing error handler).
- `heartbeat()` and `remove()` are explicitly left unmodified, each annotated with a one-line comment explaining the exclusion (heartbeat: hottest-volume path, no audit-worthy change; remove: no surviving row to send).
- Documents the observed Prisma connection-pool sizing tradeoff in `main.ts` (default pool, no explicit `connection_limit`) and adds `WebhookDeliveryError` to `docs/task-store.md`'s error table.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | All eight methods call the dispatcher with `("task.write", [task])` right before returning, inside their transaction | MET |
| 2 | A dispatcher failure rolls back the underlying DB write (verified against real Postgres) and the API call surfaces 502 | MET |
| 3 | `heartbeat()` and `remove()` are unmodified, each with a one-line comment explaining the exclusion | MET |
| 4 | Peak concurrent-transaction count vs. Postgres pool size, with timeout headroom, documented for human review | MET |
| 5 | Integration tests (real Postgres) for rollback on create/update/claim; smoke test for 502 on ≥1 call site; no existing tests retired | MET |

## Test Plan
- `task-store/src/task-service.integration.test.ts`: new tests against real Postgres proving a `WebhookDeliveryError` rolls back `create`/`update`/`claim` (no row survives / row unchanged), plus a payload-shape assertion for `complete()`.
- `task-store/src/api.smoke.test.ts`: new test confirming `POST /tasks/:id/claim` surfaces HTTP 502 when the injected dispatcher throws `WebhookDeliveryError`.
- Full `task-store` suite: 8596 pass / 3 fail — all 3 failures (2x `check-downstream-compat.sh` python3/mise sandbox issue, 1x `extraAllowedTools` flake) confirmed pre-existing on clean `origin/main`, unrelated to this change.
- `task typecheck`, `bunx biome lint`, `task check-strings` all clean.


task-store/src/main.ts
```diff
@@ -119,6 +119,16 @@ async function startServer(): Promise<void> {
 
   await runMigrations();
 
+  // No explicit `connection_limit` override — this uses Prisma's default
```
dodizzle, line 132: `DEFAULT_WEBHOOK_TIMEOUT_MS` (line 31, from TSW-1.1) defaults to 5000ms — the same as Prisma's own default interactive-transaction `timeout` (5000ms, `maxWait` 2000ms; unchanged in the pinned `@prisma/client@^6.19.2`), and none of the `$transaction(...)` calls this PR wires the dispatcher into pass an explicit `{ timeout, maxWait }` override.

Since the webhook call now happens *inside* the transaction, after other queries (`findUnique`/`update`/`recordTaskTransition`) have already consumed part of that same 5s budget, Prisma's transaction-timeout clock (which starts when the transaction opens, not when the webhook call starts) will fire at or before the dispatcher's own `AbortSignal.timeout(timeoutMs)` for any webhook receiver that's slow (roughly ≥4.9s, depending on prior query time). When that happens, Prisma throws its own transaction-closed error — not `WebhookDeliveryError` — which `translateNotFound()` doesn't recognize (it only rewrites P2025) and which falls through `app.ts`'s generic `onError` to an unhandled 500, not the documented 502.

This undercuts AC4 ("timeout headroom … documented for human review" — this comment documents pool-size headroom but not this timeout-vs-timeout collision) and AC2 ("the API call surfaces 502") for the slow-webhook case.

Suggestion: lower `DEFAULT_WEBHOOK_TIMEOUT_MS` well below Prisma's transaction timeout (e.g. 3000ms), and/or pass an explicit `{ timeout }` to the affected `$transaction` calls that's comfortably larger than the webhook timeout.

dodizzle (commented):
Verdict: COMMENT — Solid, well-tested wiring, but the webhook dispatch timeout (5000ms) has no headroom against Prisma's default transaction timeout (also 5000ms), which can surface a 500 instead of the documented 502 for slow (not just failing) webhook receivers.

dodizzle (approved):
Verdict: APPROVE — The prior finding (webhook dispatch timeout had no headroom against Prisma's default transaction timeout) is fully resolved: an explicit 10000ms `$transaction` timeout is now threaded through all eight dispatching write paths, the default webhook timeout is lowered to 3000ms, and a new startup buffer check with correct boundary math + solid unit/integration/smoke coverage backs it up. No new issues found.

PR #3378, opened 2026-09-12
feat: make bulk() atomic with single task.write event and document behavior change

dmcaulay:
## Summary
- `TaskService.bulk()` now runs the whole batch in one `$transaction` instead of one transaction per item — a mid-batch `id` collision (P2002) is translated to `ConflictError` (409) and rolls back every insert already made in that call, instead of silently skipping the colliding task and continuing.
- The outbound `task.write` webhook now fires exactly once per successful `bulk()` call, carrying every created row (previously it wasn't wired into `bulk()` at all). A webhook delivery failure mid-batch also rolls back the entire transaction (502).
- `docs/task-store.md`'s bulk-insert section and troubleshooting note now document this breaking, all-or-nothing behavior explicitly; `BulkInsertResponseSchema.skipped`'s doc comment is updated to reflect that it's now always `[]` on success (kept only for response-shape backward compatibility).

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | `bulk()` runs as one transaction; a mid-batch P2002 rolls back every insert and returns 409, not a partial result | MET |
| 2 | Exactly one `task.write` event fires per successful `bulk()` call, containing every created row | MET |
| 3 | `docs/task-store.md`'s bulk endpoint section documents the new atomic behavior | MET |
| 4 | `BulkInsertResponseSchema.skipped` is kept (empty array on success) with an updated doc comment | MET |
| 5 | Integration tests for full-batch success + single event, mid-batch collision rollback (409), and webhook-failure rollback (502); existing partial-success tests rewritten in place | MET |

## Test Plan
- Rewrote `task-store/src/task-service.unit.test.ts`'s bulk test to assert `ConflictError` + halted control flow instead of partial success.
- Rewrote `task-store/src/session-service.integration.test.ts`'s "P2002 skip" test to assert full-batch rollback (including already-processed session upserts), and added 3 new integration tests against real Postgres:
  - full-batch success fires exactly one `task.write` event with all rows
  - mid-batch collision rolls back every row (verified via direct Postgres query)
  - mid-batch webhook failure rolls back every row and rejects with `WebhookDeliveryError`
- `task-store/openapi.json` regenerated (`bun run generate:task-store-spec`) after adding 409/502 responses to the `/tasks/bulk` OpenAPI route.
- [x] `task ci` — full suite passes (8626 pass / 307 skip); 3 pre-existing failures confirmed unrelated to this diff (sandbox-only flakes: `check-downstream-compat` missing `python3`, `extraAllowedTools` flake) and reproduce independent of the touched files.
- [x] Integration tests ran against a real local Postgres instance, not skipped.


task-store/src/task-service.ts
```diff
@@ -645,42 +645,64 @@ export class TaskService implements TaskServiceLike {
...
+        await this.sessionService.upsert(tx, created.session);
+        rows.push(created);
       }
-    }
-    return { inserted, updated: 0, skipped };
+      // Fires once for the whole batch, after every row has landed, still
+      // inside this transaction — see JSDoc above.
+      await this.webhookDispatcher("task.write", rows);
+      return rows;
+    }, WEBHOOK_TX_OPTIONS);
```
dodizzle: `WEBHOOK_TX_TIMEOUT_MS` (10s) is a fixed constant sized for a single-item transaction (per its own JSDoc). Before this PR, `bulk()` ran one `$transaction` per item, so batch size never affected any single transaction's budget. Now the whole array (N sequential creates + session upserts, before the webhook call even fires) runs inside one transaction under the same fixed 10s budget, with no batch-size cap anywhere. A large batch can now exceed the budget in a way it couldn't before — and per the JSDoc, a Prisma transaction-timeout error here isn't a `WebhookDeliveryError`, so it'd surface as a generic 500 instead of the documented 409/502. The TSW-1.2 timeout-conformance unit test suite doesn't include `bulk()` in its parametrized cases, so this gap isn't covered. Consider capping batch size in `BulkInsertBodySchema` or scaling the timeout with `tasks.length`.

task-store/src/session-service.integration.test.ts
```diff
@@ -172,38 +180,123 @@ describeOrSkip(
...
+      expect(dispatched).toHaveLength(3);
+      expect(dispatched.map((t) => t.title).sort()).toEqual([
+        "batch task 1",
+        "batch task 2",
+        "batch task 3",
+      ]);
+    });
+
+    it("bulk(): a mid-batch P2002 collision rolls back every row in the batch (proven via direct Postgres query, not just the rejected promise)", async () => {
+      await prisma.task.create({
```
dodizzle: Minor: this seeds the colliding row via a raw `prisma.task.create(...)` rather than `service.create(...)`, diverging from this repo's API-first seeding convention (direct DB writes are meant to be reserved for corrupt-state/legacy-data cases the service layer can't express). The adjacent rewritten test in this same file uses `service.create(...)` for an equivalent seed.

dodizzle (commented):
Verdict: COMMENT — solid atomic-bulk() implementation, but the AC-required docs/migration.md breaking-change entry is missing and the reused fixed-size transaction timeout doesn't scale with batch size.

dodizzle (approved):
Verdict: APPROVE — both prior findings (missing docs/migration.md entry, non-scaling transaction timeout) are verifiably fixed with a solid batch-size-scaled timeout + MAX_BULK_TASKS cap and thorough tests/docs; fresh review of the full diff found no new blocking issues. dependency risk: merge — bun.lock's only change is an internal workspace-package version sync (1.260.0 → 1.265.0), no external dependency bump.

PR #3381, opened 2026-09-13
docs: document the webhook feature

dodizzle:
## Summary
- Update `SHIPWRIGHT_TASK_STORE_WEBHOOK_URL`'s description in `docs/configuration-agent.md` to reflect the TSW-1.2 trigger wiring and link to the new task-store docs section
- Add a new "Outbound webhook" section to `docs/task-store.md` covering the `{ type, data }` envelope, the `task.write` event's trigger scope (which methods fire it and which deliberately don't, and why), the fail-closed/no-retry guarantee, the `X-Shipwright-Signature` header, and an explicit scope boundary noting integration-specific logic (Jira/Linear/etc.) belongs in downstream services, not task-store

## Acceptance Criteria

| 1 | Both env-var docs and the docs/task-store.md section exist and match the shipped behavior from TSW-1.1-1.3 | MET | `docs/configuration-agent.md` already documents all four webhook env vars (from TSW-1.1); new "Outbound webhook" section in `docs/task-store.md` documents envelope, trigger table, fail-closed guarantee, and signature header matching `task-store/src/webhook-dispatcher.ts` and `task-store/src/task-service.ts` |
| 2 | `task check-config-docs` passes with the new env vars documented | MET | `check-config-docs: all 40 env var(s) are documented.` |
| 3 | Test decision: none — content-only change | MET | No test added; `check-config-docs` CI gate is the existing enforcement mechanism |

## Test Plan
- [x] `task check-config-docs` passes
- [x] `task lint` passes
- [x] `task typecheck` passes
- [x] `task check-strings` passes
- [x] `task check-version-sync` passes


dmcaulay (approved):
Verdict: APPROVE — Docs-only PR; every factual claim (trigger table, payload shapes, remove()/heartbeat() exclusions, signature construction, timeout buffer) verified against the shipped source and matches exactly. No blocking issues.

PR #3385, opened 2026-09-13
feat(admin): upgrade admin service to Prisma 7

dodizzle:
## Summary
- Bumps admin's Prisma stack from 6.19.2 to 7.10.0: generator provider switched from `prisma-client-js` to `prisma-client` (ESM, driver-adapter based), and the datasource block no longer carries `url`
- Adds `admin/prisma.config.ts` for CLI/Migrate (schema path, migrations path, connection string), and a single `admin/src/prisma-client.ts` factory (`createAdminPrismaClient`) that wires a `PrismaPg` driver adapter over a `pg.Pool` with an explicit 10s connect timeout (v7 dropped v6's implicit 5s Postgres default)
- Every existing call site (`main.ts`, dev/CI scripts, every `*.integration.test.ts`) now goes through the same factory instead of instantiating `PrismaClient` directly, so pool tuning and disconnect semantics are consistent everywhere
- Only `task-store` and `chat` remain on Prisma 6 and are unaffected — no consumer-facing API changes

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Prisma/adapter deps bumped to 7.10.0, `@prisma/adapter-pg` added, lockfile updated, `bun install` succeeds | MET |
| 2 | Generator switched to `prisma-client`, regenerated into `./prisma/client`, all import paths resolve | MET |
| 3 | `admin/prisma.config.ts` added, pointing at schema + migrations | MET |
| 4 | `main.ts` constructs a `PrismaPg` adapter over a `pg.Pool` reading `DATABASE_URL_SHIPWRIGHT_ADMIN`, explicit connect timeout, passes `{ adapter }` into `new PrismaClient()` | MET |
| 5 | No existing test rewritten/deleted; one new integration test added asserting the adapter-backed client boots and round-trips | MET |
| 6 | `task ci` passes for admin, including `test:coverage`, against a real Postgres instance | MET |
| 7 | Existing `--schema=` invocations in `ci.yml`, `Taskfile.yml`, `admin/Dockerfile`, `scripts/hitl.ts`, `scripts/dev-tmux.ts` still work | MET — `Taskfile.yml`/`scripts/hitl.ts`/`scripts/dev-tmux.ts` already `cd admin` before invoking Prisma, so `prisma.config.ts` is auto-discovered via cwd and needed no changes; `ci.yml` was updated to `cd admin` for the same reason; `admin/Dockerfile` was updated to copy `admin/prisma.config.ts` into the deps stage (needed for `postinstall: prisma generate`) |
| 8 | Safe to deploy standalone | MET — internal dependency bump only |

## Test Plan
- `bun test` (admin, against local Postgres 16): **2562 pass, 0 fail**, including the two new `prisma-client.integration.test.ts` cases
- `bun test --coverage --coverage-reporter=lcov`: same 2562/2562, package coverage 92.90% lines / 97.42% functions
- `tsc --noEmit` (both `tsconfig.json` and `tsconfig.type-parity.json`): clean
- `bunx biome check` on all changed files: clean
- `bunx prisma migrate deploy --schema=prisma/schema.prisma` (run from `admin/`, loading `prisma.config.ts`): applies cleanly against a local test DB


admin/src/main.ts
```diff
@@ -378,8 +378,14 @@ async function startServer(): Promise<void> {
...
 
-  // Construct PrismaClient once at boot
-  const prisma = new PrismaClient();
+  // Construct PrismaClient once at boot. Prisma 7 has no built-in query
+  // engine, so this is a PrismaPg driver adapter over a pg.Pool built from
+  // DATABASE_URL_SHIPWRIGHT_ADMIN — see prisma-client.ts. The `?? ""` keeps
+  // the unset-URL path identical to runMigrations' (warn and carry on, then
+  // fail on first query) rather than throwing before the server ever binds.
+  const prisma = createAdminPrismaClient(
+    process.env.DATABASE_URL_SHIPWRIGHT_ADMIN ?? "",
```
dmcaulay, line 390: Empty `DATABASE_URL_SHIPWRIGHT_ADMIN` silently falls back to pg's default connection instead of failing loudly.

`createAdminPrismaClient(process.env.DATABASE_URL_SHIPWRIGHT_ADMIN ?? "")` passes `""` when the var is unset. `pg` only parses `connectionString` when it's truthy (`if (config.connectionString) { ... }`), so an empty string is silently ignored and `pg.Pool` falls back to `PGHOST`/`PGUSER`/`PGPASSWORD`/`PGDATABASE` env vars, or `localhost:5432` as the OS user, if present. Under Prisma 6, `new PrismaClient()` threw a named error (`Environment variable not found: DATABASE_URL_SHIPWRIGHT_ADMIN`) immediately. Now the process boots and either connects to an unintended database or fails later with a generic `ECONNREFUSED 127.0.0.1:5432` with no signal the env var is what's missing.

The comment above justifies this as matching `runMigrations()`'s "warn and continue", but `runMigrations()` actually prints an explicit `[admin] DATABASE_URL_SHIPWRIGHT_ADMIN not set` warning and skips all work — it doesn't silently point at a different database the way this path can.

Suggest throwing (or logging an explicit fatal) when the var is unset, rather than defaulting the connection string to `""`.

dmcaulay (commented):
Verdict: COMMENT — Solid, well-verified Prisma 7 migration (2562/2562 tests, all --schema= call sites confirmed working, complete call-site migration to the new factory). One important finding: an unset DATABASE_URL_SHIPWRIGHT_ADMIN now silently falls back to pg's default connection (PG* env vars / localhost) instead of failing loudly like Prisma 6 did — see inline comment. One non-blocking suggestion on a now-stale CI comment/query-param for admin's test pool sizing. Dependency risk: review — major version bump on a production dependency, author isn't a known dependency-bot, but the migration itself is thoroughly tested.

admin/src/prisma-client.ts
```diff
@@ -0,0 +1,60 @@
...
+  // a bare `ECONNREFUSED 127.0.0.1:5432` that names no env var. Prisma 6's
+  // `new PrismaClient()` threw `Environment variable not found:
+  // DATABASE_URL_SHIPWRIGHT_ADMIN` immediately; this restores that.
+  if (databaseUrl.trim() === "") {
+    throw new Error(
+      "DATABASE_URL_SHIPWRIGHT_ADMIN is not set — the admin service requires a Postgres connection string (postgresql://user:password@host:5432/database)",
+    );
+  }
+
+  const pool = new pg.Pool({
```
dmcaulay, line 52: External `pg.Pool` is created with no `error` listener — an idle-connection error can now crash the admin process.

node-postgres emits an `error` event when a backend error or network partition hits an *idle* client. Because `Pool` is an `EventEmitter`, an `error` event with no listener is rethrown as an uncaught exception that terminates the process. Under Prisma 6 the Rust query engine owned the pool and absorbed this; now the pool is plain application code. The admin service runs behind Cloud SQL / a proxy sidecar that can drop idle connections, and `pg`'s default `idleTimeoutMillis` (10s) keeps idle clients around long enough to hit this. Nothing in the new test suite exercises this path.

Suggest attaching `pool.on("error", (err) => console.error("[admin] idle pg pool client error:", err))` in the factory before constructing the adapter.

admin/src/prisma-client.ts
```diff
@@ -0,0 +1,60 @@
...
+    connectionString: databaseUrl,
```
dmcaulay, line 53: Prisma-only connection-string params are now silently dropped by `pg`, changing operator-facing behavior for BYO-Postgres deployments.

The connection string is now parsed by `pg`/`pg-connection-string` instead of Prisma's engine. Prisma-only query params (`connection_limit`, `pool_timeout`, `pgbouncer`, `schema=`) are silently ignored at runtime — a `?schema=foo` URL would migrate schema `foo` (CLI still honors it via `prisma.config.ts`) while the runtime client queries `public`, a silent split. Separately, `sslmode=require` no longer means "encrypt without verifying": `pg` leaves `rejectUnauthorized` at Node's default `true`, so a private-CA/self-signed Postgres that connected fine under Prisma 6 can now fail to connect at runtime after migrations already succeeded. This is scoped to the documented `externalDatabase`/BYO-Postgres path, and `docs/migration.md` gets no new entry for it.

Suggest adding a `docs/migration.md` entry telling BYO-Postgres operators to drop Prisma-only query params from `DATABASE_URL_SHIPWRIGHT_ADMIN` and use `sslmode=no-verify` where they previously relied on Prisma's non-verifying `require`.

dmcaulay (commented):
Verdict: COMMENT — Prior finding fully fixed (DATABASE_URL_SHIPWRIGHT_ADMIN now fails loudly when unset, verified against main.ts's fatal-exit catch and a new unit test). Two new important findings on owning the pg.Pool directly: no pool 'error' listener (an idle-connection drop can now crash the process, where Prisma 6's engine used to absorb it), and Prisma-only connection-string params (schema=, sslmode=require semantics) now silently reinterpreted by pg for BYO-Postgres operators — see inline comments. Dependency risk: review — major version bump on a production dependency, author isn't a known dependency-bot, but the migration itself is thoroughly tested.

dmcaulay (commented):
Verdict: COMMENT — Both prior findings (unset-URL silent fallback; pg.Pool error listener + undocumented BYO-Postgres connection-string semantics) are fully resolved. One new important finding: the new `admin/prisma.config.ts` is excluded from `task typecheck` (admin/tsconfig.json's `include` is `["src"]`), diverging from the already-merged chat Prisma 7 migration which added `prisma.config.ts` to its tsconfig include. Dependency risk: review — major version bump on a production dependency by a non-bot author, but the migration is thoroughly tested and has been reviewed across three rounds with fixes applied each time.

dodizzle:
Addressed @dmcaulay's review-body finding about `admin/prisma.config.ts` being excluded from `task typecheck`.

**Verified the premise, then fixed it:** `admin/tsconfig.json` had `"include": ["src"]`, and `prisma.config.ts` sits at `admin/prisma.config.ts` — outside `src` — so `tsc` never saw it.

**Fix** (7ab4f20) mirrors the already-merged chat migration's exact pattern (`chat/tsconfig.json`):

```json
"include": ["src", "prisma.config.ts"],
```

**Confirmed it actually takes effect** — `tsc --noEmit --listFiles` from `admin/` now emits `admin/prisma.config.ts` in the compilation (it was absent before):

```
/…/admin/prisma.config.ts
```

Left `admin/tsconfig.type-parity.json` alone — it's a separate narrow config with an explicit 3-file `include` for the type-parity check, unrelated to this.

**Validation** (all from the worktree root):
- `task typecheck` — clean across all 7 workspaces, including admin's `tsc --noEmit && tsc --noEmit -p tsconfig.type-parity.json`. No new errors from newly type-checking `prisma.config.ts`.
- `task lint` — 751 files, no findings.
- `task test` — 8399 pass / 0 fail / 560 skip across 355 files.

One note on the test run, since it could confuse a future reader: two tests in `agent/src/claude.unit.test.ts` (`does not pass --fallback-model when not configured`, `empty extraAllowedTools produces identical args…`) initially failed locally. That's environment leakage, not a code defect — those tests read `process.env.ANTHROPIC_FALLBACK_MODEL` / `ANTHROPIC_MODEL` / `ANTHROPIC_EFFORT_LEVEL` / `AGENT_ALLOWED_TOOLS` without clearing them, and my local runner has them set. With those four unset the full suite is 0 fail, and `agent/` is byte-identical to `origin/main` on this branch, so CI's clean env is unaffected.

**No new test added** — this is a tsconfig include-list change with no runtime behavior; the meaningful verification is that `typecheck` now covers the file, which is asserted by `task typecheck` itself passing with the file in the compilation.

dmcaulay (approved):
Verdict: APPROVE — All three prior review rounds' findings (unset-URL silent fallback, missing pg.Pool error listener, undocumented BYO-Postgres connection-string semantics, prisma.config.ts excluded from typecheck) are now fully resolved; the migration is complete, well-tested, and every acceptance criterion is satisfied. Dependency risk: review — major version bump (Prisma 6.19.2 -> 7.10.0) on a production dependency by a non-bot author, but the migration itself is thoroughly vetted across three review rounds.

dodizzle:
Reviewed the outstanding findings against the current HEAD (`dcb62649bfa500b022ec6bf07091fce960c4aa92`) and made no code changes — each one's premise no longer holds. Details below.

**Dependency-risk finding — `dependency-risk@dcb62649bfa500b022ec6bf07091fce960c4aa92` — REJECTED (not reproducible)**

The finding is a risk *classification* (`recommendation: review`, `breakingChange=true`, `productionImpact=true`), auto-derived from this PR touching `admin/package.json` — a watched dependency-manifest path — because the author (`dodizzle`) isn't a known dependency-bot login, so Dependabot major-bump heuristics were applied as a best-effort fallback.

Per the dependency remediation protocol, step 1 is to extract and run the verification command named or implied by the finding's reasoning. There isn't one: the reasoning asserts no failing build, test, or runtime behavior — it flags a major version bump of `@prisma/client` (6.19.2 → 7.10.0) in production `dependencies` for human awareness. With nothing broken to reproduce, step 2's "does not reproduce as described" branch applies directly, and the protocol's bounded fix catalog (transitive-dependency override pin, or removed/renamed first-party API call-site update) has no applicable entry. Fabricating a verification command to chase would be guessing, which the protocol explicitly forbids.

The substance behind the flag has already been handled by the normal review process rather than by a patch:

- All 9 CI checks are green on this exact HEAD, including `lint / typecheck / test`, `admin docker build`, and `e2e (admin UI)`.
- Three full review rounds by @dmcaulay, each producing a finding that was fixed and verified: the unset-`DATABASE_URL_SHIPWRIGHT_ADMIN` silent-fallback guard, the missing `pool.on('error', ...)` listener, and `admin/prisma.config.ts` being excluded from `task typecheck`.
- The connection-string semantics change that motivates `breakingChange=true` is documented for operators in `docs/migration.md` ("Breaking: admin's `DATABASE_URL_SHIPWRIGHT_ADMIN` is now parsed by `pg`, not Prisma").
- The PR is currently `APPROVED`.

Recording the ref above so a later run doesn't re-derive this same finding from this same unchanged diff and re-report it indefinitely.

**Three prior review-round findings — REJECTED (already fixed; premise no longer true)**

Independently re-verified in the worktree at this HEAD rather than taken on trust. No action needed — re-implementing any of them would duplicate an existing fix:

- **Unset `DATABASE_URL_SHIPWRIGHT_ADMIN` silently falls back to `pg`'s default connection.** Fixed. `createAdminPgPool` in `admin/src/prisma-client.ts` throws on a blank/whitespace connection string, naming the env var explicitly — restoring the loud Prisma 6 failure.
- **No pool `error` listener; an idle-client error would crash the process.** Fixed. `admin/src/prisma-client.ts` registers `pool.on("error", ...)` and logs, letting the pool discard the broken client.
- **`admin/prisma.config.ts` excluded from `task typecheck`.** Fixed. `admin/tsconfig.json`'s `include` is `["src", "prisma.config.ts"]`.

No inline threads required resolution — all threads on this PR were already resolved.

PR #3428, opened 2026-09-14
feat: rewrite docs-sync.md to flag and propose, don't regenerate

dmcaulay:
## Summary
- Rewrite `.claude/commands/docs-sync.md` to consume SDR-2's page-scoped freshness signal (`check-site-docs-freshness.ts`) instead of iterating a fixed, hardcoded Section Registry — removes the old "regenerate the whole page" flow entirely.
- New flow: **flag → diff → propose → confirm**, per flagged page — diff its mapped source(s) (`site/docs-source-map.json`) since its own last-synced anchor (`state/site-docs-last-synced.json`), then propose a section-scoped edit (per `doc-refresh-recipe.md` Part 2) that preserves the page's structure/voice. Interactive mode waits for explicit human confirmation before editing; `--auto`/cron mode never edits — it files a `hitl: true` proposal task via `/tasks/bulk` instead.
- Public-repo scrub rules, frontmatter/nav-chain conventions, and the `npm run build:check` validation step are preserved verbatim from the current command.
- Follow-up fix: `bun test`'s recursive walker silently skips dot-prefixed directories, so the new `.claude/commands/docs-sync.content.test.ts` never ran under plain `bun test`/`task ci`. Added `scripts/find-hidden-dir-tests.ts` (uses `Bun.Glob`'s `dot: true` option) and wired it into `task test`/`task ci` — a general fix so any future test under a hidden directory runs automatically, not a one-off for this file.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Old Section Registry table and full-regenerate language removed; flag→diff→propose→confirm flow documented | ✅ Met |
| 2 | Public-repo scrub rules, frontmatter/nav-chain conventions, and build:check step preserved verbatim | ✅ Met |
| 3 | Cron/auto mode files a hitl:true proposal task; interactive mode waits for human confirmation | ✅ Met |
| 4 | New `docs-sync.content.test.ts` added asserting new flow present and old language removed | ✅ Met (20 assertions; now wired into `task ci` via the hidden-dir test fix) |

## Test Plan
- `.claude/commands/docs-sync.content.test.ts` — 20 assertions, verified red against the original file and green against the rewrite.
- `scripts/find-hidden-dir-tests.unit.test.ts` — 7 assertions covering the hidden-dir filter (node_modules/.git exclusion, existing bunfig ignore patterns, sorting, empty input).
- Full repo `bun test`: same pre-existing failures with or without this change (`agent/src/browser.integration.test.ts` — needs a real Chromium binary, unrelated to this PR).
- `bunx biome check` clean on all new/changed files; `check-banned-strings.ts` clean.


.claude/commands/docs-sync.md
```diff
@@ -1,158 +1,134 @@
...
+- `order` (required): A number indicating the section's position in the navigation chain. Use increments of 1 (1, 2, 3, ...) or 10 (10, 20, 30, ...) for flexibility. Earlier sections should have lower order numbers.
+- `prev` (optional): The name of the previous section in the navigation chain (use the `section` value, not the filename). E.g., `prev: Getting Started`.
+- `next` (optional): The name of the next section in the navigation chain (use the `section` value, not the filename). E.g., `next: Configuration`.
 
-1. Run `npm run build:check` from the `site/` directory to validate the Astro content collection against the schema.
-2. If validation passes, report success for each regenerated section.
-3. If validation fails, print the error and halt with a non-zero exit code (do not silently ignore schema violations).
+**Navigation chain**: Using the section registry order, set `prev` and `next` to create a continuous chain through the docs. Use the `section` value (from the Section Name column) for prev/next references. For example:
```
dodizzle: This sentence — "Using the section registry order, set `prev` and `next` to create a continuous chain through the docs" — is preserved verbatim from before, but this PR deletes the `## Section registry` heading and table it refers to. `registry` now appears exactly once in the whole rewritten file: right here. The hardcoded example list right below (introduction/getting-started/configuration with order 0/1/3) is also a leftover of the deleted registry-driven flow and doesn't map to anything the new per-page `site/docs-source-map.json`-driven model tracks.

Since this doc is the instruction set a future `/docs-sync` run follows, this is a real dead end — there's no registry left to consult for "the section registry order." Suggest rewording to reference the target page's existing neighboring pages' `section` values directly (read from their current frontmatter) instead, and dropping/genericizing the stale example list. `docs-sync.content.test.ts`'s "preserves the Navigation chain explanation verbatim" assertion would need updating to match.

dodizzle (commented):
Verdict: COMMENT — Solid rewrite (flag→diff→propose→confirm flow is well cross-referenced and the preserved sections check out verbatim), but the "Navigation chain" bullet still says "Using the section registry order" even though this PR deletes the Section Registry table entirely — a dangling reference in a doc that itself drives a future /docs-sync run. See inline comment.

dodizzle (approved):
Verdict: APPROVE — Re-review: the prior dangling "section registry" reference in docs-sync.md's Navigation chain rules is cleanly fixed (now derives prev/next from neighboring pages' frontmatter, with a dedicated regression test), CI is green at head, and no new issues surfaced.

PR #3366, opened 2026-09-12
docs: split agent-api.md — exceeds 200 lines

dmcaulay:
## Summary
- Split `docs/agent-api.md` (543 lines, over the 200-line hard threshold) into core CRUD/auth (`docs/agent-api.md`, now 213 lines) and operational surfaces (`docs/agent-api-ops.md`, new, 341 lines).
- `docs/agent-api.md` keeps Authentication, Agents, Environment variables, and Runtime config, plus a short "Operational APIs" pointer to the new doc.
- `docs/agent-api-ops.md` covers Cron jobs, Cron runs, Tools (allowed-tools list), API tokens, Plugins, Chat token usage, and Work queue snapshot.
- Added a `docs/agent-api-ops.md` entry to CLAUDE.md's Reference section and tightened the existing `docs/agent-api.md` entry's description to match its trimmed scope.

## Acceptance Criteria
None specified on the task (docs-freshness split proposal, human/dev-task-executed).

## Test Plan
- [x] Content moved verbatim — no wording changes to any endpoint doc, only relocation and one new pointer section.
- [x] `task check-strings` passes.
- [x] `task check-config-docs` passes (unaffected — only checks `docs/configuration.md`).
- [x] `task check-version-sync` passes.
- [x] `task lint` passes.
- [x] Verified every other doc referencing `agent-api.md` (`docs/README.md`, `docs/extending.md`, `docs/configuration.md`, `.claude/commands/docs-sync.md`) still points at content that still lives in `docs/agent-api.md` — no broken anchors.

Generated with [Claude Code](https://claude.com/claude-code)

PR #3431, opened 2026-09-15
feat: escalate .claude/** review findings to HITL on first round in patch.md

dmcaulay:
## Summary
- Adds Step 5a.6b to `patch.md`'s Step 5 sequence — a pre-dispatch check that scans List A findings for a remediation requiring a write under `.claude/**` (inline thread `path` starting with `.claude/`, with a fallback text-mention match for review-body-level findings).
- When matched, skips dispatching the fix subagent for that PR this cycle and escalates immediately to HITL via `references/escalation-pattern.md`'s shared sequence, with a distinct `blockedReason` (`"structurally unfixable — finding requires a write under .claude/**, which the Claude Code CLI blocks unconditionally — escalated to HITL"`) and its own site-specific hook resolving all unresolved inline threads before releasing the claim.
- Adds a parallel row to `unblock.md`'s origin-phase table for this pattern.

Rooted in app-vitals/shipwright#3366 — the `.claude/**` write restriction was discovered as a review finding mid-flight, not known at planning time. Without this check, a fix subagent dispatched against such a finding can never succeed, so the PR would loop through the fix dispatch every patch cycle forever instead of ever reaching a human.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | New step inserted between 5a.6 and 5a.7, firing before Step 5b's fix-subagent dispatch, on the first round — no second-round wait, unlike 5a.7 | MET |
| 2 | Distinct `blockedReason` string, not reusing 5a.7's second-round string | MET |
| 3 | Existing Step 5a.7 behavior unaffected — coexistence content test added | MET |
| 4 | `unblock.md`'s origin-phase table gets a new row parallel to the existing second-round-disagreement row | MET |
| 5 | `patch.content.test.ts` cases asserting existence, firing order, distinct blockedReason, and non-interference with 5a.7 | MET |

## Test Plan
- New `describe` block in `patch.content.test.ts` (9 tests) covering: step positioning, first-round firing, path-matching logic (inline + fallback), distinct blockedReason, shared escalation-pattern usage, PR_TASK_ID/PR_RECORD_ID reuse, fix-subagent-skip behavior, and the site-specific thread-resolution hook.
- Coexistence test confirming Step 5a.7's second-round-disagreement logic is untouched.
- `unblock.content.test.ts` updated with the new `blockedReason` pattern.
- `bun test plugins/shipwright/commands/patch.content.test.ts` — 197/197 pass.
- `task lint`, `task typecheck` — clean.
- `task ci` (full coverage gate) — passes except 3 pre-existing full-suite-only test-isolation flakes in unrelated files (`metrics/src/lib/errors.unit.test.ts`, `task-store/src/routes/prs.openapi.smoke.test.ts`), verified to pass cleanly in isolation and untouched by this diff.


plugins/shipwright/commands/patch.md
```diff
@@ -919,7 +920,79 @@ Skip the rest of Step 5 for this PR. Move to the next qualifying PR in List A. I
...
+write requirement is structurally unfixable regardless of how many rounds have passed, so
+there is no reason to wait for a second round to escalate it.
+
+**Scan this cycle's List A findings for a `.claude/**` remediation.** For each finding that
+contributed to this PR's List A membership (Step 3a's `reviewThreads.nodes[]` entries with
+`isResolved == false`, and the qualifying COMMENTED/CHANGES_REQUESTED review bodies):
+
+- **Inline thread finding**: the thread's first comment `path` (from Step 3a) starts with
+  `.claude/`.
+- **Review-body-level finding** (no inline thread — fallback): the review body text
```
dodizzle: This fallback is a bare substring match for `.claude/` anywhere in the review body, with no requirement that the mention actually describes the finding's required remediation. Unlike the inline-thread path (anchored on the structured `path` field of the commented file), a review body that merely cites a `.claude/` path as context (e.g. referencing `.claude/shipwright/principles.md`, a convention this plugin's own review agents routinely cite) but whose actual finding is unrelated and otherwise auto-fixable would incorrectly trip this check — escalating the whole PR to HITL and skipping the fix subagent for the cycle. Consider requiring the `.claude/` mention to co-occur with remediation language (e.g. inside a code span, or near "edit"/"update"/"change") rather than a bare anywhere-in-body substring match.

dodizzle (commented):
Verdict: COMMENT — the review-body-level fallback for the new Step 5a.6b escalation check matches on a bare `.claude/` substring anywhere in a review body, with no requirement it's the actual remediation target — see inline comment.

dodizzle (approved):
Verdict: APPROVE — Prior finding (imprecise `.claude/` substring fallback) is fully addressed: the fallback now requires the path inside an inline code span co-occurring with a remediation verb in the same sentence, and biases genuine ambiguity toward not matching. No new issues found.

PR #3453, opened 2026-09-15
feat: plan-session --autonomous materializes the spec from the task description

dodizzle:
## Summary
Closes the last gap in the autonomous plan-session hand-off. `check-plan.ts` and `loop-orchestrator.ts` already dispatch `/shipwright:plan-session {repo} {session} --autonomous {task-id}` for flagged tasks, but Step 1 only ever read `planning/{session}/PRODUCT-SPEC.md` from the worktree and otherwise fell back to the interactive "What are we building?" prompt — which has no one to answer it under `--autonomous`. External submitters (e.g. a platform gateway) deliver the whole spec in the task's `description`, so the command stalled on a file that nothing had written.

- New Step 1 `--autonomous` Mode: if the spec file is absent, `GET /tasks/{task-id}`, strip the `Commit as PRODUCT-SPEC.md and run /shipwright:plan-session.` preamble, and write the remainder to `planning/{session}/PRODUCT-SPEC.md`
- Empty/missing description → PATCH `blocked` + `hitl: true` with `plan_session_autonomous_no_spec`, then stop
- The interactive fallback is explicitly forbidden under `--autonomous`; the originating task is excluded from Step 1.3's same-session duplicate scan
- TESTING.md Scenario 45; `docs/agent-ops.md` `shipwright-plan` row notes the repo/session requirement and spec source

Pairs with app-vitals/squadron#196, which makes `submit-prd` send `session` and default `autonomousPlanSession: true`.

## Test plan
- [x] `bun test plugins/shipwright/commands/plan-session.content.test.ts` — 54 pass (6 new pins for the Step 1 subsection)
- [x] `task lint`, `task typecheck`, `task check-config-docs`, `task check-version-sync`, `secret-scan` green locally
- [x] Full `bun test`: 8496 pass; the 4 failures are pre-existing/environmental in untouched files (`browser.integration.test.ts` needs a local Playwright Chromium binary; `check-deploy.unit.test.ts` is sensitive to the workspace's `SHIPWRIGHT_*` env) — expecting CI green

https://claude.ai/code/session_019xcqgnE78AJ6LLMVkVsdVB


dmcaulay (commented):
Verdict: COMMENT — the new `--autonomous` Mode subsection is mechanically precise except for one gap: the duplicate-scan exclusion at plan-session.md:99 references "1.3's duplicate scan" but that step already executes earlier in Step 1's numbered list and gets no actual filter (unlike item 4's analogous `select(.session != $s)`). Everything else — blocking behavior, docs/agent-ops.md, TESTING.md Scenario 45, and the new content tests — is clean and consistent.

plugins/shipwright/commands/plan-session.md
```diff
@@ -76,6 +76,28 @@ This is the engineering planning pass. The product spec (what and why) is alread
...
+3. If the description is missing, or blank after stripping, there is nothing to plan and no human to ask. Block the task and stop:
+   ```bash
+   curl -sf -X PATCH -H "Authorization: Bearer $SHIPWRIGHT_TASK_STORE_TOKEN" \
+     -H "Content-Type: application/json" \
+     "$SHIPWRIGHT_TASK_STORE_URL/tasks/{task-id}" \
+     -d '{"status": "blocked", "hitl": true, "blockedReason": "plan_session_autonomous_no_spec: no planning/{session}/PRODUCT-SPEC.md in the worktree and the task description is empty"}' | jq .
+   ```
+   Print `⚠ Task {task-id} has no spec to plan from — blocked for human triage (plan_session_autonomous_no_spec).` and stop. Do not continue to Step 2.
+
+The interactive fallback below — asking **"What are we building?"** — must never run under `--autonomous`; there is no one to answer it. Likewise, in 1.3's duplicate scan, exclude `{task-id}` itself from the existing-tasks list: the originating PRD task shares this session slug because it is the *input* to the session, not a duplicate of the work it produces.
```
dmcaulay: Step 1's numbered list (items 1-5) runs first per this document's own stated order; item 3 is the actual duplicate-scan `curl ... | jq '.tasks'` command against `/tasks?session=$SESSION`, whose output includes the originating `--autonomous` task itself (it carries the session's `repo`/`session` fields, per `check-plan.ts`). This new `### --autonomous Mode` subsection is explicitly anchored to run "between 1.5 and the orientation header" — i.e. after item 3 has already executed and printed its results. Only here, in trailing prose, does it say to "exclude `{task-id}` itself from the existing-tasks list" — with no mechanical filter given, unlike item 4 two lines above in the same Step 1, which achieves an analogous exclusion via an explicit jq filter (`map(select(.session != $s))`).

Since this file is executed verbatim by an autonomous agent with no other source of truth, an instruction that references modifying an already-completed step, with no concrete mechanism, is worth tightening — e.g. add `select(.id != "{task-id}")` (gated on `--autonomous`) directly to item 3's jq filter instead of relying on prose that points backward at an already-executed step.

dodizzle: Fixed in 8265013e. Item 3's command now carries the filter itself — `jq --arg t "${AUTONOMOUS_TASK_ID:-}" '.tasks | map(select(.id != $t))'`, with `AUTONOMOUS_TASK_ID` set to `{task-id}` under `--autonomous` and a no-op otherwise (the empty-string arg matches no id). The `--autonomous` Mode subsection no longer instructs a backward exclusion; it just notes item 3 already handles it. Added a content test pinning the filter to item 3 (between the item 3 and item 4 headings) rather than to trailing prose.

dmcaulay (approved):
Verdict: APPROVE — the prior finding (duplicate-scan exclusion described in prose but not actually implemented in item 3) is genuinely fixed: item 3 now carries a real jq filter (`select(.id != $t)`) gated on `--autonomous`, pinned by a new content test. No new issues.

PR #3360, opened 2026-09-11
feat: exclude autonomousPlanSession tasks from ready query

dmcaulay:
## Summary
- Add an unconditional `if (task.autonomousPlanSession === true) continue;` exclusion to `resolveReadyTasks` in `task-store/src/ready.ts`, mirroring the existing `hitl` exclusion — independent of any dispatch-enable env var, so PRD-shaped tasks with the flag set can never be claimed by dev-task/cron as normal work.
- Add `autonomousPlanSession?: boolean | null` to the `ReadyTaskLike` interface and update the file's doc comment.
- Refresh `docs/task-store.md`'s ready-query reference and troubleshooting section to describe the new exclusion.

## Acceptance Criteria

| # | Criterion | Status | Evidence |
|---|-----------|--------|----------|
| 1 | A pending task with autonomousPlanSession: true never appears in resolveReadyTasks's output | MET | `if (task.autonomousPlanSession === true) continue;` added unconditionally in resolveReadyTasks's loop |
| 2 | Existing hitl exclusion behavior is unchanged | MET | hitl check untouched; new check added after it; `makeTask` defaults `autonomousPlanSession: null` so all 28 pre-existing cases are unaffected |
| 3 | New unit test mirroring the hitl exclusion test | MET | `"excludes a task with autonomousPlanSession === true even if otherwise ready"` added directly after the hitl test, same style |

## Test Plan
- `bun test task-store/src/ready.unit.test.ts` — 29/29 pass, 100% line/function coverage on `ready.ts`
- `task typecheck` — all 7 packages pass
- `task lint` — clean
- [x] Pre-commit checks passing

Generated with [Claude Code](https://claude.com/claude-code)

PR #3459, opened 2026-09-15
feat: introduce TaskKind enum, replace autonomousPlanSession internally

dmcaulay:
## Summary
- Adds a `TaskKind` enum (`dev`, `prd`) + `kind` column to `Task` (default `dev`), superseding the boolean `autonomousPlanSession` flag internally.
- `ready.ts` now excludes `kind === "prd"`; `check-plan.ts` now queries `?kind=prd&status=pending`.
- `POST /tasks` and `POST /tasks/bulk` dual-accept `kind` and the legacy `autonomousPlanSession: true`, normalizing the latter to `kind: "prd"` server-side (`normalizeTaskKind()` in `task-service.ts`). The legacy column, field, and `?autonomousPlanSession=` GET filter remain fully functional during the transition window — nothing is removed.
- Corrects `loop-orchestrator.ts`'s dedupe comment and `plugins/shipwright/CLAUDE.md`'s Candidate Selection Contract paragraph, both of which inaccurately claimed `?ready=true` never excludes flagged tasks (PR #3360 / PDR-2.2 already added that exclusion) — now explains the dedupe as a defensive backstop, not a required workaround.
- Regenerates `lib/task-store-types.ts`, `task-store/openapi.json`, `mcp-server/src/generated-tools.ts`, `docs/mcp-tools.md`; updates prose in `docs/task-store.md`, `docs/agent-ops.md`, `docs/configuration-agent.md`, `docs/agent-key-files.md`, and `plugins/shipwright/commands/plan-session.md`.

## Acceptance Criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | TaskKind enum + kind column via Prisma migration, defaulting to dev; ready.ts filters on kind !== "prd"; check-plan.ts queries ?kind=prd&status=pending | MET |
| 2 | POST /tasks and /tasks/bulk dual-accept kind + legacy autonomousPlanSession, normalizing latter to kind:"prd"; GET ?autonomousPlanSession= filter unchanged | MET |
| 3 | loop-orchestrator.ts dedupe comment + CLAUDE.md Candidate Selection Contract corrected to reference kind, describe dedupe as defensive backstop | MET |
| 4 | Test suite updated to assert on kind/kind=prd; new dual-accept unit test added; no new integration/smoke beyond existing openapi smoke updates | MET |
| 5 | Regenerate lib/task-store-types.ts + docs/mcp-tools.md; update docs/task-store.md, docs/agent-ops.md, docs/configuration-agent.md | MET |
| 6 | Safe to deploy standalone (additive/back-compat) | MET |

## Test Plan
- `task test` — 8529 pass, 1 pre-existing unrelated flake (`extraAllowedTools` in `agent/src/claude.unit.test.ts`, reproduces identically on clean `main`).
- `task typecheck` — clean (after regenerating the Prisma client for the new `kind` column/enum).
- `task lint`, `task check-strings`, `task check-config-docs`, `task check-version-sync` — all pass.
- Coverage gate: 85.65%/86.49% (lines/functions) vs 90%/89% threshold — pre-existing sandbox gap (no test DB for integration tests), not a regression: clean `main` measures 85.61%/86.48% in the same sandbox.
- New unit test coverage: `task-service.unit.test.ts` dual-accept block (legacy-boolean → kind normalization, precedence when both are sent, back-fill in both directions).


task-store/src/task-service.ts
```diff
@@ -132,6 +138,42 @@ export function bulkTxTimeoutMs(taskCount: number): number {
...
+ * would risk clobbering a deliberate write.
+ */
+export function normalizeTaskKind<
+  T extends { kind?: unknown; autonomousPlanSession?: unknown },
+>(data: T): T {
+  const { kind, autonomousPlanSession } = data;
+  if (kind !== undefined) {
+    // Explicit kind wins. Back-fill the legacy flag only when the caller
+    // didn't state it, so the deprecated filter still sees this row.
+    if (typeof kind !== "string" || autonomousPlanSession !== undefined)
```
dodizzle: **Critical (confidence 90):** This function's own JSDoc, the schema comment, the OpenAPI description, and docs/task-store.md all claim the `kind`/`autonomousPlanSession` pair is kept in sync — "each one back-fills the other so a row never says 'prd' in one field and 'dev' in the other." This branch does the opposite when both fields are present: it returns `data` untouched, so `POST /tasks {"kind":"dev","autonomousPlanSession":true}` persists a genuinely contradictory row (`kind='dev'` AND `autonomousPlanSession=true`). That row passes the legacy `?autonomousPlanSession=true` filter *and* `?ready=true` (which only checks `kind`), while `check-plan.ts`'s `?kind=prd` query misses it — a PRD task can get silently dispatched as `/shipwright:dev-task`. The new test "create() lets an explicit kind win over a contradicting legacy flag" (task-service.unit.test.ts:709) only asserts `kind`, never `autonomousPlanSession`, so it passes with the divergent row.

Suggest forcing the legacy flag to match in this branch: `return { ...data, autonomousPlanSession: kind === "prd" }`.

task-store/src/ready.ts
```diff
@@ -80,7 +86,12 @@ export async function resolveReadyTasks<T extends ReadyTaskLike>(
   for (const task of tasks) {
     if (task.status !== "pending") continue;
     if (task.hitl === true) continue;
-    if (task.autonomousPlanSession === true) continue;
+    // TKD-1.1: the PRD slice is never dev work. Supersedes PDR-2.2's
+    // `autonomousPlanSession === true` check — the two are equivalent by
+    // construction (the migration backfilled kind='prd' for every flagged row,
+    // and every write path keeps the pair in sync), except that an explicit
+    // `kind` now wins over a contradicting legacy flag.
+    if (task.kind === "prd") continue;
```
dodizzle: **Important (confidence 80):** This comment justifies replacing the PDR-2.2 `autonomousPlanSession === true` check with a `kind`-only check on the claim that "every write path keeps the pair in sync" — which only holds once the `normalizeTaskKind()` gap (flagged on task-service.ts:169) is fixed. It also fails open during the rolling deploy this PR's AC6 claims is safe: if an old task-store pod is still writing `autonomousPlanSession: true` rows (defaulting `kind='dev'` per the migration) while a new pod's `?ready=true` is already live, that PRD task now qualifies as ready — something the pre-PR check would never allow.

For the transition window, consider checking both: `if (task.kind === "prd" || task.autonomousPlanSession === true) continue;`

agent/src/check-plan.ts
```diff
@@ -116,6 +117,18 @@ export async function getPlanCandidates(
...
+/**
+ * The task-store query that defines this phase's pool (TKD-1.1).
+ *
+ * Exported so its shape is unit-testable without standing up a task-store
+ * client: `kind=prd` is the current spelling of what used to be
+ * `autonomousPlanSession=true`, and getting it wrong silently empties the
+ * whole plan phase.
+ */
+export function buildPrdTaskQuery(): URLSearchParams {
+  return new URLSearchParams({ kind: "prd", status: "pending" });
```
dodizzle: **Important (confidence 78):** `agent/` and `task-store/` deploy independently, and task-store's list-query schema silently ignores unknown params rather than rejecting them. Against an older task-store, `?kind=prd&status=pending` degrades to `?status=pending` — every pending task becomes a plan candidate, and `loop-orchestrator.ts`'s dedupe deliberately lets the plan-tagged copy win ties, so ordinary dev tasks could get dispatched as `/shipwright:plan-session --autonomous`. Blast radius is limited today since the plan phase ships double-gated and disabled by default, but this is a real fail-open direction during rollout.

Consider sending both params during the transition (`kind=prd&autonomousPlanSession=true`) or defensively filtering in the mapper.

dodizzle (commented):
Verdict: COMMENT — `normalizeTaskKind()` doesn't enforce its own documented sync invariant when both `kind` and the legacy `autonomousPlanSession` are sent in one request, letting a row persist as `kind='dev'` with `autonomousPlanSession=true` (or vice versa); `ready.ts` and `check-plan.ts`'s new `?kind=prd` query both fail open rather than closed against that divergence (or agent/task-store version skew during rollout).

docs/agent-key-files.md
```diff
@@ -69,18 +69,18 @@
...
+| `agent/src/check-plan.ts` | Candidate provider for the autonomous plan-session phase (PDR-4.1): `getPlanCandidates(deps: CheckPlanDeps) → Promise<WorkTaskCandidate[]>` and `buildProductionDeps()` factory. Mirrors `check-dev-task.ts`'s shape but queries `?kind=prd&status=pending` (TKD-1.1, replacing the legacy `?autonomousPlanSession=true`) — PRD tasks awaiting autonomous planning, deliberately NOT filtered by `ready`. Tags each candidate `phase: "plan"` and carries the task's `repo`/`session` through, since the dispatched command is `/shipwright:plan-session {repo} {session} --autonomous {task-id}` rather than a bare id; a task missing either field doesn't qualify and is dropped. Exports `buildPrdTaskQuery()` (the `?kind=prd&status=pending` query, exported so its shape is unit-testable) and type `CheckPlanDeps`. |
```
dodizzle: **Important (confidence 80):** This row says `check-plan.ts` "queries `?kind=prd&status=pending` (TKD-1.1, replacing the legacy `?autonomousPlanSession=true`)" and describes `buildPrdTaskQuery()` as "the `?kind=prd&status=pending` query." Both are wrong against the code this PR ships: `buildPrdTaskQuery()` (agent/src/check-plan.ts) returns `{kind: "prd", autonomousPlanSession: "true", status: "pending"}` — the legacy param is deliberately *not* replaced, it rides along for the transition window so a new agent binary hitting an older task-store doesn't silently widen to `?status=pending` (its own doc comment explains why, and `check-plan.unit.test.ts` asserts all three params). This file's own `loop-orchestrator.ts` row (line 83) gets it right, so this doc now contradicts itself. `plugins/shipwright/CLAUDE.md:138` has the identical understatement — that one matters more since it's the design constitution, and `CLAUDE.content.test.ts`'s `toContain("kind=prd")` assertion passes either way, so nothing catches the drift there.

Suggestion: update both lines to state the query is `?kind=prd&autonomousPlanSession=true&status=pending`, retained on purpose until every deployed task-store honors `?kind=`.

dodizzle (commented):
Verdict: COMMENT — All three prior findings (normalizeTaskKind() sync invariant, ready.ts fail-open, check-plan.ts kind-only query) are substantively fixed and independently verified in code. One new important-severity issue: docs/agent-key-files.md:72 and plugins/shipwright/CLAUDE.md:138 still describe check-plan.ts's query as `?kind=prd&status=pending` alone, contradicting the shipped `buildPrdTaskQuery()` (which sends both `kind` and the legacy `autonomousPlanSession` param on purpose) and inviting removal of the param that keeps the transition window rollout-safe.

agent/src/check-plan.ts
```diff
@@ -116,6 +118,36 @@ export async function getPlanCandidates(
...
+ * `/shipwright:plan-session --autonomous`. Sending the legacy flag too keeps
+ * the pool correctly narrowed on an old task-store, and costs nothing on a new
+ * one: normalizeTaskKind() forces the pair to agree on every write path, so
+ * `kind=prd` and `autonomousPlanSession=true` select the same rows.
+ *
+ * Drop the legacy param once every deployed task-store honors `?kind=`.
+ */
+export function buildPrdTaskQuery(): URLSearchParams {
+  return new URLSearchParams({
+    kind: "prd",
```
dodizzle: **Important (confidence 80):** This sends `kind=prd` and `autonomousPlanSession="true"` together, and `TaskService.list()` (`task-store/src/task-service.ts:476-478`) puts both on the same Prisma `where` object — a strict AND, not an OR.

This PR's own `ready.ts:104-108` and `docs/task-store.md` document exactly the row shape where the two diverge: an old task-store pod writes `autonomousPlanSession: true` while `kind` stays at the migration's `dev` default. For that row, `ready.ts:111`'s OR check correctly excludes it from the dev-task pool — but this AND query *also* excludes it from the plan pool, since `kind !== "prd"`. Nothing back-fills `kind` afterward, so the PRD task becomes permanently invisible to both providers with no error and no alert, for exactly the rolling-deploy skew window AC6 claims is safe.

Suggest dropping the `kind: "prd"` conjunct (the legacy flag alone is an exact match on a new store, since every `kind='prd'` row has `autonomousPlanSession=true` by construction, and the only thing that works on an old one) or sending the two as an OR.

dodizzle (commented):
Verdict: COMMENT — buildPrdTaskQuery() sends kind=prd AND the legacy autonomousPlanSession=true as a strict AND (Prisma where clause), which silently orphans exactly the mid-rollout divergence row shape (kind='dev', autonomousPlanSession=true) this PR's own ready.ts and docs document as possible — that PRD task becomes invisible to both the dev-task and plan pools with no recovery path. Prior findings from the last two review passes are confirmed fixed.

dodizzle (approved):
Verdict: APPROVE — All three prior findings across three review passes (normalizeTaskKind() sync invariant, ready.ts fail-open, buildPrdTaskQuery()'s divergence-row orphaning) are independently verified fixed in code. No new defects at or above the confidence threshold. One documented, review-driven deviation from acceptance criterion 1 (buildPrdTaskQuery() sends the legacy autonomousPlanSession param instead of kind, on purpose, to stay safe against an older task-store) should be reconciled in the task record, not the code.

PR #3471, opened 2026-09-16
docs(planning): add sessions-pt-agent-names plan

dmcaulay:
## Summary
- Sessions list page (`/admin/sessions`) timestamps aren't shown in Pacific time and the Agents column shows raw agent ids instead of names — both diverge from every other admin page's established pattern.
- Folded in three related asks from planning discussion: merge the redundant Session/Slug columns, reuse the Tasks page's distinct-values autocomplete for filtering, and add an Org filter (which `GET /sessions` doesn't support yet, unlike `GET /tasks`).

## Test plan
- [ ] N/A — planning doc only, no code changes. Queues SPT-1.1 (task-store: org filter) and SPT-1.2 (admin UI, depends on SPT-1.1) in the task store.


dodizzle (approved):
Verdict: APPROVE — Planning doc only, no code changes; every technical claim about the current codebase checked out exactly against the live source, and the task breakdown/sequencing is sound.

PR #3472, opened 2026-09-16
feat: retry GitHub App auth setup on syncConfig tick (ABF-2.1)

dodizzle:
## Summary
- Adds `agent/src/github-auth-startup.ts`, mirroring `slack-startup.ts`'s ref/guard pattern, so `index.ts`'s `syncConfig()` loop retries GitHub App auth setup on every 60s tick once `GH_APP_ID`/`GH_APP_INSTALLATION_ID`/`GH_APP_PRIVATE_KEY` are newly complete — a GitHub App installed after the pod boots no longer requires a restart to become usable.
- Guards against `setupGitHubAuth()`'s non-idempotency (each call creates an independent `GitHubTokenManager` with its own background-refresh interval) via a `githubAuthActiveRef` "already set up" flag, plus a start-in-flight guard for overlapping ticks — regression-tested to ensure setup runs exactly once.
- Extracts `agent/src/github-auth-deps.ts` (`buildGitHubAuthDeps()`) so `entrypoint-main.ts`'s one-shot boot call and `index.ts`'s retry-on-tick call share identical production wiring instead of duplicating it.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `syncConfig()` retries GitHub App auth setup on each tick, mirroring `startSlackIfPossible()` | MET |
| 2 | Retry entry point is idempotent — no double credential-helper registration or duplicate refresh interval | MET |
| 3 | GitHub App installed after boot becomes usable within one syncConfig tick (60s), no restart required | MET |
| 4 | Unit test coverage added, analogous to the Slack retry test; no existing tests retired | MET |

## Test Plan
- `agent/src/github-auth-startup.unit.test.ts` (new, 18 tests): credential-completeness checks, single successful start, no re-trigger once active (regression test for the duplicate-interval bug), overlapping/concurrent call safety, error propagation without marking active (so a failed attempt can retry).
- `bun test agent/src/github-auth-startup.unit.test.ts agent/src/slack-startup.unit.test.ts` — 31 pass.
- Full agent unit suite (`cd agent && bun test src`) — 1864 pass, 0 fail.
- `task typecheck` — all 7 workspaces clean.
- `task lint` — clean (0 issues in changed files).


dmcaulay (approved):
Verdict: APPROVE — Retry hook extraction is a faithful match of the prior inline wiring, the guard/re-check logic is race-safe, error handling correctly avoids marking auth active on failure, and the new tests cover concurrency/error-recovery with real assertions. All 4 acceptance criteria met, no blocking findings.

PR #3490, opened 2026-09-16
feat: add org filter to GET /sessions

dmcaulay:
## Summary
- Adds `org` as a first-class filter on `GET /sessions`, mirroring the existing `repo` filter's shape exactly (`SessionListFilters.org?: string | string[]`, repeatable query param, in-memory post-rollup filter — no Prisma where-clause).
- Wires it through `SessionService.list()`, `routes/sessions.ts`'s query parsing, and `SessionListQuerySchema` in `openapi-schemas.ts`.
- Purely additive — no Prisma schema change, no migration.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `GET /sessions?org=some-org` returns only sessions with a task whose repo's org prefix matches (case-sensitive exact match) | MET |
| 2 | `GET /sessions?org=a&org=b` matches either org (array-OR semantics) | MET |
| 3 | `GET /sessions` with no `org` param is unaffected (backward compatible) | MET |
| 4 | Test coverage: single org, multiple orgs via array, no match, combined with an existing filter, plus a smoke test for query-param forwarding | MET |

Note: the task's source description named `session-service.unit.test.ts` for the new filter tests, but that file is scoped to pure-logic-only helpers (its own docstring says so, and `SessionService.list()` requires a real Prisma client). The existing `repo` filter test for `list()` already lives in `session-service.integration.test.ts` for the same reason — the new `org` tests were added there instead, following established convention. Coverage is equivalent; only the file name deviates from the literal AC text.

## Test Plan
- 4 new integration test cases in `session-service.integration.test.ts` (single org, multiple orgs via array, no match, org combined with an existing state filter) — verified RED before implementation, GREEN after.
- 1 new smoke test in `routes/sessions.smoke.test.ts` asserting `?org=` is parsed and forwarded into `SessionService.list()`.
- `task lint` / `task typecheck` clean.


docs/mcp-tools.md
```diff
@@ -61,7 +61,7 @@ List sessions
 - **Method:** GET
 - **Path:** `/sessions`
 - **Has body:** No
-- **Parameters:** `state` (query), `sort` (query), `agentId` (query), `repo` (query), `q` (query), `limit` (query), `offset` (query)
+- **Parameters:** `state` (query), `sort` (query), `agentId` (query), `repo` (query), `org` (query), `q` (query), `limit` (query), `offset` (query)
```
dodizzle, line 64: This is a generated file (`<!-- GENERATED FILE — do not edit by hand. -->`, regenerated via `bun run generate:mcp-docs`, which reads from `mcp-server/src/generated-tools.ts`, not `task-store/openapi.json` directly). The PR added `org` to `openapi.json`/`openapi-schemas.ts` but never ran `bun run generate:mcp-tools`, so `generated-tools.ts`'s `sessions_list` entry still lacks `org` (and its schema has `additionalProperties: false`) — an MCP client's `org` param would be rejected/stripped. Running the actual documented regen command (`generate:mcp-docs`) would revert this hand-edit, since it reads from the still-stale `generated-tools.ts`. Please run `bun run generate:mcp-tools && bun run generate:mcp-docs` and commit both regenerated files.

dodizzle (commented):
Verdict: COMMENT — Core org-filter logic and tests are solid, but the PR hand-edits the generated `docs/mcp-tools.md` instead of regenerating `mcp-server/src/generated-tools.ts`, leaving the new `org` param invisible to (and rejected by) the MCP `sessions_list` tool while the doc now falsely claims it's available there.

dodizzle (approved):
Verdict: APPROVE — The prior finding (stale generated MCP tool artifact) is confirmed resolved: I independently regenerated both mcp-server/src/generated-tools.ts and docs/mcp-tools.md and they're byte-for-byte identical to what's committed. Full re-review found no new issues.

PR #3504, opened 2026-09-16
feat: sessions list UI Pacific time, agent names, merged columns, filter autocomplete

dodizzle:
## Summary
- Threads `timezone` (default `America/Los_Angeles`) into the Sessions list page's timestamp rendering, matching every other admin page.
- Resolves session `agentIds` to agent names via `agentService.listByIds` (raw-id fallback when unmatched), and widens `SessionsListAgentService` to also pick `listOptions`.
- Merges the redundant Session/Slug columns into one (8 → 7 columns); the slug renders underneath the title only when it differs from it.
- Adds Org/Repo multiselect + Agent datalist autocomplete to the filter form (`renderRepoOrgFilterFields()`, `fetchDistinctTaskValues()`, `agentService.listOptions()`), with degraded-mode parity when `fetchDistinctTaskValues` is absent, and forwards `org` query values through to the task-store `/sessions` fetch (SPT-1.1).

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Timestamps render in configured timezone (Pacific vs UTC observably different) | MET |
| 2 | Agents column renders resolved names, raw-id fallback | MET |
| 3 | Session/Slug merged into 7-column table, slug shown only when it differs from title | MET |
| 4 | Org/Repo multiselects + Agent datalist autocomplete, degraded-mode parity with Tasks page | MET |
| 5 | Org filter forwarded to task-store `/sessions` fetch | MET |

## Test Plan
- `bun test admin/src/admin-ui-sessions-list.smoke.test.ts` — 18 pass, 0 fail, one new smoke-test assertion per acceptance criterion
- `admin-ui-sessions-list.ts` coverage: 100% lines, 95.24% functions
- `task typecheck` — clean
- `task ci` — 2 pre-existing unrelated failures confirmed present on unmodified `main` (agent env-var leak in `agent/src/claude.unit.test.ts`, pre-existing `validateRepo` null-repos bug in `task-store/src/routes/prs.ts`); no failures in touched files
- [x] Pre-commit checks passing

Generated with [Claude Code](https://claude.com/claude-code)

PR #3510, opened 2026-09-16
docs(planning): add sessions-list-filter-fix plan

dmcaulay:
## Summary
- Adds `planning/sessions-list-filter-fix/PLAN.md`, documenting a live-task-store investigation into two Sessions-list (`/admin/sessions`) filtering reports from Dan.
- **Agent filter** — real bug from #3504: the route forwards a typed/selected agent name straight through as `agentId` to task-store, which expects a real ID. Queues SLF-1.1 to fix it by mirroring the Tasks page's name→ID resolution pattern.
- **Org filter** — not a code bug: confirmed live that the fix (#3490, tagged `task-store-v1.140.0`) simply hasn't rolled out to the running task-store pod yet. No task queued; flagged for a recheck instead.

## Test plan
- N/A — planning doc only, no code change.


dodizzle (approved):
Verdict: APPROVE — Docs-only planning doc; Agent-filter root cause and proposed fix verified accurate against admin-ui-sessions-list.ts/admin-ui.ts, task breakdown matches design, no issues found.

PR #3513, opened 2026-09-17
docs(planning): add chart-bump-tag-drop-fix plan

dmcaulay:
## Summary
- Adds `planning/chart-bump-tag-drop-fix/PLAN.md`, diagnosing why the sessions-list-filter-fix org-filter fix (#3490) took ~6.5 hours to reach production despite releasing instantly.
- Root cause: `auto-bump-chart.yml`'s `collect_batch_tags()` decides which release tags are "new" via a wall-clock comparison against the previous chart-bump PR's merge timestamp, not against what that PR actually credited. A tag batch landing while an earlier, unrelated bump PR is still open gets silently dropped — confirmed this hit `admin-v1.186.0` too, not just task-store.
- `check-chart-drift.yml` already exists specifically to catch this defect class (per its own header comment) but the root cause in `auto-bump-chart.yml` was never fixed — only detected.
- Queues CBT-1.1: replace the timestamp heuristic with a drift-based diff, reusing already-tested primitives from `lib/chart-drift-check.sh` / `lib/chart-tag-utils.sh`.

## Test plan
- N/A — planning doc only, no code change.


planning/chart-bump-tag-drop-fix/PLAN.md
```diff
@@ -0,0 +1,124 @@
...
+The drift-check's own building blocks are the fix: `lib/chart-drift-check.sh`'s
+`read_pinned_tag()` (reads what's currently pinned in `values.yaml`) and
+`lib/chart-tag-utils.sh`'s `highest_semver_tag()` already implement, and have test
+coverage for, exactly the "is this service's tag actually reflected in values.yaml"
+comparison that `collect_batch_tags()` should be doing instead of a wall-clock heuristic.
+
+## Design
+
+**CBT-1.1** — replace timestamp-based tag collection with a drift-based diff:
+- Move `read_pinned_tag()` from `lib/chart-drift-check.sh` into `lib/chart-tag-utils.sh`
```
dodizzle: This and several other references in the Design section (lines 49-53, 64, 70, 76, 92-93, 109) cite `lib/chart-drift-check.sh` / `lib/chart-tag-utils.sh`. The actual files live at `.github/workflows/lib/chart-drift-check.sh` and `.github/workflows/lib/chart-tag-utils.sh` (confirmed via both files' header comments and `check-chart-drift.yml`'s `source` lines). This repo also has a *different*, unrelated top-level `lib/` directory (a TS package — `lib/http.ts`, `lib/pricing.ts`, etc.), so the bare `lib/...` citation isn't just imprecise, it points at the wrong location entirely. Worth fixing before CBT-1.1 execution starts so the engineer doesn't go looking in (or create files in) the wrong `lib/`.

dodizzle (commented):
Verdict: COMMENT — Plan's root-cause analysis and proposed fix are accurate and well-corroborated against the actual codebase, but it cites `lib/chart-tag-utils.sh`/`lib/chart-drift-check.sh` at the wrong path — the real files live under `.github/workflows/lib/`, and this repo has an unrelated top-level `lib/` dir that could mislead whoever executes CBT-1.1.

dodizzle (commented):
Verdict: COMMENT — Prior lib/ path-citation finding is fixed, but the Investigation section mischaracterizes mcp-server-v0.100.0 as dropped by this bug when it was never tracked by auto-bump-chart.yml at all.

planning/chart-bump-tag-drop-fix/PLAN.md
```diff
@@ -0,0 +1,124 @@
...
+Follow-up from the sessions-list-filter-fix investigation (Dan, Slack): the Sessions
+page's Org filter fix (#3490) took ~6.5 hours to reach production after merging, even
+though it was tagged into a release within seconds. Dan asked to diagnose the
+deployment delay.
+
+## Investigation
+
+Traced the full release chain for `task-store-v1.139.0` (the tag containing the org
+filter fix, cut 2026-09-16 15:40:22 UTC — same instant as four other same-day releases:
+`admin-v1.186.0`, `chat-v1.102.0`, `mcp-server-v0.100.0`, `metrics-v1.107.0`).
```
dodizzle: `mcp-server-v0.100.0` shouldn't be grouped with the tags this bug actually drops. `auto-bump-chart.yml`'s tag-push trigger is only `admin-v*`/`metrics-v*`/`agent-v*`/`task-store-v*`/`chat-v*` (no `mcp-server-v*`), and `chart-tag-utils.sh`'s `service_for_tag()`/`values_paths_for_service()` have no `mcp-server` branch — same for `check-chart-drift.yml`'s `SERVICES=(admin metrics agent task-store chat)`. So `mcp-server-v0.100.0` could never be credited by any chart-bump PR under current code, bug or no bug — it's a separate, pre-existing "mcp-server isn't wired into auto-bump-chart.yml" gap, not a symptom of the timestamp-heuristic bug this plan fixes. Worth correcting so whoever writes CBT-1.1's regression test doesn't treat mcp-server as in-scope evidence.

dodizzle (approved):
Verdict: APPROVE — Both prior findings (lib/ path citations, mcp-server-v0.100.0 framing) are fixed; independently re-verified the plan's technical claims against the actual repo and they hold up. No blocking issues.

PR #3514, opened 2026-09-17
fix: replace timestamp-based tag collection in auto-bump-chart.yml with drift-based diff

dmcaulay:
## Summary
- Replaces `auto-bump-chart.yml`'s timestamp-based `collect_batch_tags()` (anchored to "since the last chart-bump commit") with a new `compute_batch_from_drift()` that diffs live git tags against what's currently pinned in `charts/shipwright/values.yaml` — fixing a confirmed bug where a batch of release tags landing while an earlier, unrelated chart-bump PR was still open got silently dropped (only the tag that won the debounce race survived).
- Moves `read_pinned_tag()` from `lib/chart-drift-check.sh` to `lib/chart-tag-utils.sh` (it's a generic values.yaml reader, not drift-check-exclusive) — `check-chart-drift.yml` needs zero changes since it already sources `chart-tag-utils.sh` first.
- When the recomputed batch is empty (nothing has actually drifted), the workflow now skips opening a chart-bump PR entirely instead of redundantly crediting the trigger tag.
- Replaces `test-auto-bump-chart.sh`'s `collect_batch_tags` test block with unit tests for `compute_batch_from_drift()`, including a regression test reproducing this incident's exact shape (two tag batches landing back-to-back, second predating the first batch's PR-merge time) that asserts every tag in the second batch is still credited.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | `collect_batch_tags()`'s timestamp comparison removed, replaced with `compute_batch_from_drift()` | MET |
| 2 | `read_pinned_tag()` moved to `chart-tag-utils.sh`; `check-chart-drift.yml` unchanged | MET |
| 3 | Back-to-back batch (second predating first's merge) no longer dropped | MET |
| 4 | Empty recomputed batch skips opening a PR | MET |
| 5 | Test suite replaces `collect_batch_tags` tests with `compute_batch_from_drift` unit tests + regression test | MET |

## Test Plan
- `bash .github/workflows/test-auto-bump-chart.sh` — 88 passed, 0 failed (includes the new regression test reproducing the incident shape)
- `bash .github/workflows/test-check-chart-drift.sh` — 40 passed, 0 failed (confirms the `read_pinned_tag()` move is a no-op for this consumer)
- `task ci` — lint/check-strings/typecheck/check-config-docs/check-version-sync all pass; `test:coverage` hit one pre-existing, unrelated sandbox flake (`extraAllowedTools`, documented as a known clean-main flake) — nothing in this diff touches `agent/src/claude.ts`


dodizzle (approved):
Verdict: APPROVE — Drift-based batch computation correctly eliminates the timestamp blind spot; regression test faithfully reproduces the incident, and the read_pinned_tag() move is a verified no-op for check-chart-drift.yml. No blocking issues.

PR #3520, opened 2026-09-18
feat: widen claude retry gate for early session-id capture and ceiling timeouts

dodizzle:
## Summary
- Widen `_runClaude`'s resume-retry gate to retry using ANY known session id — the pre-existing `existingSessionId`, a session id captured mid-attempt off the stream's `system`/`init` line (a call that started fresh), or the session id carried by the thrown error itself — instead of only retrying when `-r` was already passed at the start.
- Split `ClaudeTimeoutError` retry handling by `reason`: `'ceiling'` (the hard wall-clock cap firing on a still-legitimately-working session — ~67 Sentry occurrences over ~2 months) is now retried once; `'idle'` (the genuine-hang signal — zero observed occurrences) stays excluded as a safety rail.
- Add a new optional `onEarlySessionId` callback, threaded through `_consumeStream` → `_spawnOnce` → `_spawn` → `_runClaude` → the returned `runClaude`, fired the instant a session id is captured off the init line — strictly before any `onProgress` fire for the same run.
- Exclude connection-lost-mid-response failures from this new outer retry gate, since `_spawn` already gives them their own dedicated one-shot retry (a prior, separate human decision capped at exactly one retry) — without this exclusion the widened gate would stack a second retry on top.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | A fresh (non-resumed) call that fails after capturing a session id is retried once using `-r <that-id>` | MET |
| 2 | `ClaudeTimeoutError` with `reason: 'idle'` is never retried; `reason: 'ceiling'` is now retried | MET |
| 3 | The new early-session-id callback fires before the existing `onProgress` callback for the same run | MET |
| 4 | Existing Slack-thread resume-retry behavior (`existingSessionId` present at call start) is unchanged | MET |
| 5 | Unit tests in `agent/src/claude.unit.test.ts` follow the existing injected-Clock/fixture pattern — no `mock.module()`/global overrides | MET |

## Test Plan
- `bun test agent/src/claude.unit.test.ts` — 94 pass / 0 fail
- Full suite: `bun test --coverage` — 8650 pass / 0 fail / 563 skip (DB-dependent integration tests self-skip without a local Postgres instance)
- `task lint`, `task typecheck`, `task check-strings`, `task check-config-docs`, `task check-version-sync` all clean
- `agent/src/claude.ts` itself: 100% line coverage (494/494), 94.6% function coverage (35/37)


dmcaulay (approved):
Verdict: APPROVE — small, additive retry-gate widening with a test covering every stated acceptance criterion; no blocking findings.

PR #3526, opened 2026-09-18
feat: dev-task-only resume wiring + early session-id cron-log push

dmcaulay:
## Summary
- Add `CronRunReporter.recordSessionId(cronId, runId, sessionId)` (HttpCronRunReporter + NoopCronRunReporter) — a fire-and-forget partial PATCH of `{sessionId}` only, mirroring the existing `recordProgress` pattern.
- Give cron-dispatched dev-task invocations a stable `sessionKey` (`dev-task:{taskId}`); review/patch/deploy/plan dispatch continue passing `undefined` since those phases must re-validate live state on every repeat dispatch.
- Wire the DTW-1.2 early-session-id callback to `recordSessionId`, so the cron-run log carries the session id before the run reaches a terminal state.
- In `loop-orchestrator.ts`'s `dispatchItem`, after a dev-task dispatch finishes, check the task's live status via the task store; if still `in_progress`, resume immediately with the same `sessionKey` (which the underlying runner turns into `-r <session-id>`) instead of waiting on the next cron tick or the 65-minute StaleClaimReaper. Capped at 3 auto-resumes, each its own `createRun`/`completeRun` pair.

## Acceptance Criteria
| 1 | Dev-task phase sessionKey `dev-task:{taskId}`; review/patch/deploy/plan pass undefined, regression-tested | MET | `loop-orchestrator.ts` dispatchItem computes sessionKey only for `phase === "dev-task"`; integration test explicitly asserts review dispatch gets `undefined` sessionKey and never calls `getTaskStatus` |
| 2 | `recordSessionId` PATCH body = `{sessionId}` only, fires before completion | MET | `cron-run-reporter.ts`'s `recordSessionId()` PATCHes `{sessionId}` only; `onEarlySessionId` fires `recordSessionId` before `completeRun` (call order asserted in unit test) |
| 3 | Integration test: resume loop with fake runner sequence, capped at 3, distinct `createRun`/`completeRun` per attempt | MET | New `loop-orchestrator.integration.test.ts` describe block: cap-at-3 (4 total runner calls), never-terminal-stops-at-3, terminal-after-first-attempt, review-never-resumes |
| 4 | Test decision: integration test in `loop-orchestrator.integration.test.ts` + unit test for `recordSessionId` PATCH shape in `cron-run-reporter`'s test file | MET | Both present as specified |

## Test Plan
- `bun test` on all touched files (375 tests, 0 fail)
- `task ci` — 3 pre-existing failures unrelated to this diff (`extraAllowedTools`/`check-downstream-compat.sh`), reproduced identically on clean `main`
- Coverage on changed files: `check-helpers.ts` 94.2%, `cron-run-reporter.ts` 95.0%, `loop-orchestrator.ts` 97.4% — all above the 90% threshold


agent/src/loop-orchestrator.ts
```diff
@@ -733,64 +871,42 @@ export function createLoopOrchestrator(
...
+    }
+
+    // DTW-1.3 auto-resume loop — dev-task only. After an attempt finishes,
+    // re-read the task's LIVE status: still `in_progress` means the session
+    // ended without finishing the task (the ScheduleWakeup/long-wait case), so
+    // resume it right now with the same sessionKey instead of leaving the
+    // claim to go stale for ~65 minutes and burning a cold session on the next
+    // tick. Capped at MAX_AUTO_RESUMES; a "silent" outcome or a thrown failure
+    // ends the loop (the throw propagates out of dispatchItem unchanged).
+    if (phase === "dev-task" && sessionKey) {
```
dodizzle, line 1008: This auto-resume loop lets a single dev-task dispatch make up to 4 sequential `runner()` calls (1 initial + `MAX_AUTO_RESUMES` = 3), each independently bounded by claude.ts's 30-minute ceiling — so one dev-task iteration can now legitimately run for ~2 hours.

That breaks the invariant `BUSY_STALL_THRESHOLD_MS` (35 min, line 355) relies on: its doc comment assumes a healthy single drain iteration can never run past the 30-minute `runner()` ceiling, since `busySince` is reset once per iteration (line 1079) before `dispatchItem` runs. A concurrent tick firing during a multi-resume dev-task dispatch will now take the escalated branch and emit a false, Sentry-eligible `console.error` claiming the tick is "stuck/wedged." It also means one slowly-resuming task can serialize the whole drain — blocking review/patch/deploy dispatches for hours instead of ≤30 min.

Consider scaling `BUSY_STALL_THRESHOLD_MS` by `(MAX_AUTO_RESUMES + 1)`, or resetting `busySince` before each `runOneAttempt()` call so elapsed time keeps measuring one bounded runner call.

agent/src/loop-orchestrator.ts
```diff
@@ -592,139 +632,237 @@ export function createLoopOrchestrator(
...
-      await cronRunReporter.completeRun(
+    // DTW-1.3: dev-task — and ONLY dev-task — gets a stable session identity
+    // so a follow-up attempt continues the same Claude session instead of
+    // starting cold. The underlying runner persists session ids keyed by this
+    // value and builds `-r <id>` itself on the next call with the same key, so
+    // nothing here has to carry the id around. review/patch/deploy/plan keep
+    // sessionKey undefined — a deliberate product decision: those phases
+    // re-validate live state on every dispatch and must not continue stale
+    // context.
+    const sessionKey = phase === "dev-task" ? `dev-task:${itemId}` : undefined;
```
dodizzle: This session key is derived purely from the task id, and the underlying store (`agent/src/sessions.ts`) is file-backed with a 7-day TTL — nothing in the repo ever calls `sessions.clear("dev-task:...")` for it. So this isn't scoped to just this dispatch's own resume loop: any *later, independent* dev-task dispatch of the same task id within 7 days (next tick after the StaleClaimReaper releases the claim, a dispatch after an agent-process restart, or a human re-run) will also resume the old session via `-r`, carrying a dead session's stale conclusions into what should be a fresh run.

That contradicts `plugins/shipwright/commands/dev-task.md`'s documented contract for exactly this scenario (lines 732-733, 970-971): after the reaper reclaims a stale claim, the next tick should re-dispatch "with no memory of this pipeline's progress — a full context-free re-bootstrap." A stale resumed transcript could conclude "I already opened the PR" and emit `[silent]`, eventually auto-blocking the task via the skip counter.

Consider scoping the key to the dispatch (e.g. a per-dispatch nonce or the cron-run id) or clearing `dev-task:{taskId}` from the session store once the resume loop in `dispatchItem` exits.

dodizzle (commented):
Verdict: COMMENT — the auto-resume loop is well-tested and bounded, but it breaks the busy-stall threshold's single-runner-call-per-iteration invariant (false Sentry-eligible "wedged" errors), and the persistent `dev-task:{taskId}` session key can leak a resume across independent dispatches, contradicting dev-task.md's documented reaper re-bootstrap contract.

agent/src/loop-orchestrator.ts
```diff
@@ -592,139 +639,276 @@ export function createLoopOrchestrator(
...
+    // dispatch's own loop below — all 1 + MAX_AUTO_RESUMES attempts share this
+    // one key, which is what makes `-r` fire on the resumes — while
+    // guaranteeing the next dispatch starts cold.
+    //
+    // A nonce rather than a clear-on-exit: an explicit clear can't cover a hard
+    // process kill mid-resume-loop, whereas an unreachable per-dispatch key is
+    // inert by construction. Abandoned entries age out through the store's own
+    // TTL prune.
+    const sessionKey =
+      phase === "dev-task" ? `dev-task:${itemId}:${randomUUID()}` : undefined;
```
dodizzle, line 747: Per-dispatch sessionKey nonce grows the persistent `sessions.json` without bound. The comment above justifies never clearing this key with "Abandoned entries age out through the store's own TTL prune" — but `createFileSessionStore`'s `prune()` (agent/src/sessions.ts) has zero production call sites anywhere in the repo. A nonce key is written once during its dispatch and never read again, so it's never lazily evicted via `get()` either. Key cardinality goes from bounded (one `dev-task:{taskId}` entry, reused) to one new permanent entry per dev-task dispatch, in a file on the agent's PVC-backed home that also backs Slack thread sessions — so every Slack message now reads/rewrites a growing file too.

Suggest wiring a periodic `sessions.prune()` call, or having the resume loop `clear()` the nonce key once it exits.

agent/src/loop-orchestrator.ts
```diff
@@ -733,64 +917,42 @@ export function createLoopOrchestrator(
...
+        } catch (err) {
+          // Resuming is an optimization — a task-store blip here must not turn
+          // an otherwise-successful dispatch into a failure. Fall back to
+          // today's behavior (stop, let the reaper + next tick handle it).
+          console.warn(
+            `[loop-orchestrator] getTaskStatus failed for ${itemId}: ${String(err)} — not resuming`,
+          );
+          return;
+        }
+        if (liveStatus !== "in_progress") return;
```
dodizzle: The resume gate checks `liveStatus === "in_progress"` only — production wiring (`getTaskStatus: (id) => taskStoreClient.getTask(id).then((t) => t?.status ?? null)`) discards `claimedBy`. Since one dispatch can hold a task across up to 4 sequential `runner()` calls, the window is long enough for the claim TTL to lapse if an attempt stalls past its heartbeat. If the StaleClaimReaper then reaps the claim and a different claimant claims it back to `in_progress` before this loop's next check, the loop reads `in_progress` and resumes its own stale session against a task another session now owns.

`getTask` already returns `claimedBy` at no extra cost — suggest resuming only when the live task is both `in_progress` and still claimed by this agent.

agent/src/index.ts
```diff
@@ -630,11 +631,25 @@ const loopJobsRef = createJobsRef<CronJobLike>();
 // wiring cost, and its construction errors surface at fire time (logged by the
 // cron callback's try/catch) rather than crashing agent startup.
 const getLoopOrchestrator = createLoopOrchestratorGetter({
+  // DTW-1.3: sessionKey is no longer hardcoded undefined for the loop path —
+  // the orchestrator supplies `dev-task:{taskId}` for dev-task dispatches (and
```
dodizzle: This comment still says "the orchestrator supplies `dev-task:{taskId}`" — the fix commit changed the key shape to `dev-task:{taskId}:{per-dispatch nonce}` (loop-orchestrator.ts:672) but this is the one surface that wasn't updated to match.

dodizzle (commented):
Verdict: COMMENT — both prior findings (busy-stall invariant, cross-dispatch session leak) are genuinely fixed with regression tests, but the fix's per-dispatch nonce grows the persistent sessions.json file without bound (its stated TTL-prune mitigation is dead code in production), and the resume gate should check claim ownership, not just task status, to close a narrow reap-then-reclaim race.

agent/src/loop-orchestrator.ts
```diff
@@ -733,64 +963,71 @@ export function createLoopOrchestrator(
...
+    // ScheduleWakeup/long-wait case), so resume it right now with the same
+    // sessionKey instead of leaving the claim to go stale for ~65 minutes and
+    // burning a cold session on the next tick. Capped at MAX_AUTO_RESUMES; a
+    // "silent" outcome or a thrown failure ends the loop (the throw propagates
+    // out of dispatchItem unchanged).
+    if (phase === "dev-task" && sessionKey) {
+      try {
+        if ((await runOneAttempt()) !== "completed") return;
+        let resumeAttempt = 0;
+        while (resumeAttempt < MAX_AUTO_RESUMES) {
```
dodizzle, line 1012: Auto-resume loop holds one task claim across up to 4 Claude sessions with no heartbeat refresh, breaking the claim-TTL sizing invariant.

`lib/claim-ttl.ts` defines `DEFAULT_CLAIM_TTL_MS = DEFAULT_CLAUDE_TIMEOUT_MS (60min) + CLAIM_TTL_BUFFER_MS (5min) = 65min`, explicitly sized so "a claim isn't reaped mid-session" — an invariant that assumes a claim spans exactly one Claude session. `agent/src/config.ts:36-38` confirms production resolves `SHIPWRIGHT_CLAUDE_TIMEOUT_MS` to that same 60-minute default.

This loop lets one dispatch make up to 4 sequential `runner()` calls against the *same* claim (1 initial + `MAX_AUTO_RESUMES` = 3), and neither `dispatchItem` nor `createTaskStoreClient` ever refreshes the claim's heartbeat between attempts — the only heartbeat renewals are prompt-driven, inside the dev-task session itself. A dispatch can now legitimately span close to 4 hours against a 65-minute claim TTL. If any attempt ends without the in-session prompt reaching a heartbeat step (the abnormal-early-exit case this feature targets), the StaleClaimReaper releases the claim mid-dispatch and a sibling agent can claim and start a second, cold dev-task session on the same task/branch while this dispatch's next resume attempt is still in flight. The `claimedBy === agentId` gate correctly stops *this* loop from resuming afterwards, but it can't prevent the duplicate session the reap already allowed — it closes the resume half of the race, not the reap half.

Suggestion: refresh the claim (a heartbeat call) immediately before each resume attempt, or bound the resume loop by wall-clock elapsed-since-claim against the claim TTL rather than by attempt count alone.

dodizzle (commented):
Verdict: COMMENT — both prior findings (sessions.json growth, status-only resume gate) are genuinely fixed with regression tests, but the auto-resume loop can now hold one task claim across up to 4 Claude sessions (~4h) with no heartbeat refresh, breaking the documented claim-TTL sizing invariant (65min TTL assumes exactly one 60min session) and re-opening a duplicate-session exposure via the reap side of the race rather than the resume side.

dodizzle (approved):
Verdict: APPROVE — all three prior review rounds' findings (busy-stall invariant, cross-dispatch session leak, sessions.json growth, status-only resume gate, claim-TTL heartbeat gap) are genuinely resolved with regression tests at the correct layer. The claim is now renewed before every resume attempt, so the claim-TTL sizing invariant (65min TTL ≈ one 60min session) holds even across up to 4 attempts.

PR #3533, opened 2026-09-19
feat: wire sessions list Follow button to SessionFollowService

dmcaulay:
## Summary
- Replace the disabled `follow-toggle-stub` button on the Sessions list page with a live Follow/Unfollow toggle, wired to the existing `SessionFollowService` and `POST /admin/sessions/:slug/{follow,unfollow}` routes that already power the session detail page.
- Batch-resolve per-row follow state via a single `sessionFollowService.listByUser()` call per page render (not one call per row), threaded through `renderSessionsListPage` → `renderSection` → `sessionRow`.
- Wire click handling via one delegated inline `<script>` for the whole page (`classList.contains('session-follow-btn')`) instead of a per-row listener, mirroring the detail page's existing button behavior (in-place label/`data-following` update, no full page reload).
- Refresh `docs/agent-web-ui.md` to describe the now-live Follow button.

## Acceptance Criteria
| # | Criterion | Status |
|---|-----------|--------|
| 1 | Sessions list page renders each row's Follow button enabled, with `data-slug` set to that row's session slug and `data-following` reflecting whether the current user already follows it (label "Follow" / "Following" to match). | MET |
| 2 | Clicking a row's button POSTs to `/admin/sessions/{slug}/follow` or `/unfollow` (based on current state) and updates that row's button label and `data-following` in place on success, without a full page reload. | MET |
| 3 | `admin-ui-sessions-list.smoke.test.ts` replaces the old stub-only test with cases covering enabled rendering, `data-following="true"` for an already-followed session, and `data-following="false"` for one that isn't — all via injected `SessionsListDeps.sessionFollowService` doubles, no real DB. | MET |

## Test Plan
- `bun test admin/src/admin-ui-sessions-list.smoke.test.ts` — 25 pass, covering enabled rendering, followed/unfollowed states, and the delegated click-handler script content.
- `bun test admin/src` (full admin suite) — 2335 pass, 0 fail.
- `bun run --filter='@shipwright/admin' typecheck` — clean.
- `task ci` — 3 pre-existing, unrelated failures (`check-downstream-compat.sh` allowlist tests, `extraAllowedTools`), confirmed via `git stash` to reproduce on unmodified branch tip; not caused by this change.
- Coverage on the changed file: 96.30% lines / 100% branches (threshold 90%/89%).


admin/src/admin-ui-sessions-list.ts
```diff
@@ -536,6 +584,15 @@ export function registerSessionsListRoutes(
       for (const a of agents) agentNames[a.id] = a.name;
     }
 
+    // Batch-resolve which of the fetched rows the current user already
+    // follows via a single listByUser() call — mirrors the agentNames
+    // resolution above (one call per page render, not one per row).
+    const followedSlugs = new Set(
```
dodizzle: Unguarded `sessionFollowService.listByUser()` call — any failure here throws unhandled out of the route handler and 500s the entire sessions list page, including the Waiting/Active/Closed sections that would otherwise render fine.

This is inconsistent with the adjacent, already-shipped precedent for the *same* call: the session detail page's `resolveIsFollowingSession` (`admin-ui.ts:3389-3397`) wraps the equivalent `sessionFollowService.listByUser()` call in try/catch, explicitly documented as "best-effort follow-state lookup... fail open to 'not following' on error since this is just button display state, not a security check." The comment in `sessionRow()` above says this mirrors the detail page's button, but the error-handling half of that mirror was dropped.

Suggest wrapping this in try/catch, falling back to an empty `Set()` (all buttons render as "Follow") on failure.

dodizzle (commented):
Verdict: COMMENT — solid FLW-1.1 implementation meeting all three acceptance criteria, but the new `sessionFollowService.listByUser()` call at admin-ui-sessions-list.ts:590 has no try/catch, unlike the detail page's fail-open precedent for the identical call — a failure here would 500 the whole sessions list page instead of just falling back to "not following".

dodizzle (approved):
Verdict: APPROVE — prior finding (unguarded listByUser() call) is fixed with a try/catch fail-open matching the detail page's precedent, plus a new covering test. All three acceptance criteria met, no new issues found.

PR #3534, opened 2026-09-19
feat: replace Estimation Accuracy KPI with Shipwright PRs Merged count

dodizzle:
## Summary
- Replace the "Estimation Accuracy" KPI tile on the metrics dashboard with a plain integer count of Shipwright-authored merged PRs (`shipwrightPrsMerged`)
- Reuses the existing origin-tracking infra (POM-1.1/1.2/4.1) that already backs the "Merged PRs by repo" panel — no new task-store calls, just a different aggregation (single total instead of per-repo breakdown)
- Trends bucket `shipwrightPrsMerged` per period the same way `reviewsShipIt` is bucketed

## Acceptance Criteria

| 1 | Schema/handler swap `estimationAccuracy`→`shipwrightPrsMerged`, computed from PR list, trends bucketed per period like `reviewsShipIt` | MET | `schemas.ts`, `task-store-provider.ts`, `api.ts` updated; unit+integration tests pass |
| 2 | Dashboard tile relabeled with correct `data-metric`/`id`, `app.js` updated, `fmtEstError` removed, click-to-expand preserved | MET | `dashboard-page.ts`, `index.html`, `app.js` diffs confirmed |
| 3 | Tests updated: `dashboard-page.unit.test.ts` + 5 snapshots, `dashboard.e2e.ts` 3 fixtures + renamed test; `api.public.smoke.test.ts` had no applicable fixture (confirmed via grep — N/A, not a gap) | MET | all 408 unit/smoke/integration tests pass |
| 4 | Out-of-scope files (`plugins/shipwright/commands/metrics.md`, `site/src/pages/index.astro`, `plugins/shipwright/references/metrics-dashboard-demo.html`, `metrics/src/__snapshots__/queries.integration.test.ts.snap`) untouched | MET | `git diff --stat` confirms only the 10 intended files changed |

## Test Plan
- `cd metrics && bun test` — 408 pass, 0 fail, 5 snapshots regenerated
- `task lint` — clean (only pre-existing warnings in an out-of-scope orphaned prototype file)
- Coverage: metrics package unchanged (87.97% → 87.96% lines, rounding noise)


metrics/src/providers/task-store-provider.integration.test.ts
```diff
@@ -319,6 +320,8 @@ describe("TaskStoreProvider (integration)", () => {
     expect(row[colIndex(t, "avg_coverage_delta")]).toBe(2.5);
     // avg_review_iterations = avg(reviewCycles + patchCycles) = avg(2, 3) = 2.5
     expect(row[colIndex(t, "avg_review_iterations")]).toBe(2.5);
+    // Neither fixture PR has origin "shipwright" — shipwright_prs_merged is 0.
+    expect(row[colIndex(t, "shipwright_prs_merged")]).toBe(0);
```
dmcaulay, line 375: This is the only assertion covering the new `shipwright_prs_merged` aggregation, but the `PRS` fixture (line 133) sets neither `origin` nor `state` on either record, so `expect(...).toBe(0)` passes trivially regardless of whether `p.state === "merged" && p.origin === "shipwright"` is even the right predicate. Add a fixture PR with `origin: "shipwright", state: "merged"` and assert a non-zero count here (and in the `trends()` test block) to actually exercise the true-match branch.

dmcaulay (commented):
Verdict: COMMENT — Critical: removing `estimationAccuracy` from `SummaryResultSchema`/`TrendsPeriodSchema` breaks the live marketing site. `site/src/pages/index.astro` fetches `https://proof.shipwrightharness.com/public/metrics/summary` at build time and reads `summary.estimationAccuracy` into a homepage KPI card (confirmed present on `origin/main`, untouched by this diff). Since `site/` deploys independently from `metrics/`, the field disappears from the live endpoint on merge and the card silently renders "—" under a label that still says "Estimation accuracy." Either update `site/src/pages/index.astro` to consume `shipwrightPrsMerged` (or drop/null-guard the card), or keep serving `estimationAccuracy` alongside the new field for a deprecation window. Separately, the new `shipwrightPrsMerged`/`shipwrightPrsMergedInPeriod` filters have no positive-path test — see inline comment.

dodizzle:
Addressed the review. One finding implemented, one rejected on scope grounds — details below.

**Implemented** — the `shipwright_prs_merged` positive-path test gap (inline thread on `task-store-provider.integration.test.ts:324`). Correct call: the shared `PRS` fixture sets neither `origin` nor `state`, so `toBe(0)` passed regardless of the predicate. Added an `ORIGIN_PRS` fixture with two matching PRs plus a near-miss for each half of the conjunct (`merged`+`human`, `open`+`shipwright`) and non-zero assertions in both the `summary()` and `trends()` paths. Mutation-verified: dropping either `state === "merged"` or `origin === "shipwright"` now fails a test.

**Rejected (scope, not correctness)** — `site/src/pages/index.astro` consuming `estimationAccuracy`:

- The technical premise is **correct** and I verified it independently: `origin/main`'s `site/src/pages/index.astro` does fetch `https://proof.shipwrightharness.com/public/metrics/summary` at build time (line 124 type, line 195 `fmtSignedPct(summary.estimationAccuracy)`) and will render `—` under the "Estimation accuracy" label once this merges. The `summaryFallback` snapshot only applies when the fetch *fails*, so it does not mask the missing field on a successful 200.
- It is nonetheless **explicitly out of scope for this task**. MPC-1.1's description and its fourth acceptance criterion both name this file by path: *"Do not touch plugins/shipwright/commands/metrics.md, site/src/pages/index.astro, plugins/shipwright/references/metrics-dashboard-demo.html, or metrics/src/__snapshots__/queries.integration.test.ts.snap — confirmed out of scope with the requester."* Editing it here would violate a stated acceptance criterion, so I have not.
- The suggested alternative — keeping `estimationAccuracy` on the response for a deprecation window — also conflicts with AC #1, which requires `SummaryResultSchema`/`TrendsPeriodSchema` to *drop* the field.

Worth flagging for whoever picks up the follow-up: the planner's rationale for excluding this file called it a *"hardcoded marketing-site demo widget"*, which is inaccurate — it is live-fetched with a hardcoded fallback. So the exclusion decision was made on a slightly wrong premise even though the reviewer's impact analysis is right. The blast radius is a graceful degrade (one KPI card reads `—`; no build break, no crash, no error), not an outage. Per the no-ad-hoc-tasks rule I have not queued a follow-up myself — this needs a planning pass to decide whether the site card should move to `shipwrightPrsMerged` or be dropped.

Resolving the inline thread for the implemented finding; leaving this comment as the record for the rejected one.

dmcaulay (commented):
Verdict: COMMENT — The positive-path test gap from the prior review is fixed: the new `ORIGIN_PRS` fixture in `task-store-provider.integration.test.ts` genuinely exercises the true-match branch of `state === "merged" && origin === "shipwright"` for both `summary()` and `trends()`, with same-day near-misses on each conjunct. Still open: removing `estimationAccuracy` from `SummaryResultSchema`/`TrendsPeriodSchema` breaks the live marketing site — `site/src/pages/index.astro` (untouched by this diff) still reads `summary.estimationAccuracy` from the same `/public/metrics/summary` endpoint at build time, and since `site/` deploys independently from `metrics/`, the homepage KPI card will silently render "—" forever the moment this merges. This task's AC #4 explains why `index.astro` wasn't edited here, but that scoping doesn't resolve the underlying regression on a live, separately-deployed consumer. Recommend filing a fast-follow task to update or null-guard that card before/alongside merging this PR, rather than leaving it as an open-ended silent break.

dodizzle:
This finding was already rebutted once and the review still disagrees after re-evaluating that rebuttal — this looks like a genuine disagreement between the reviewer and the automated fix, not something another automated pass will resolve. Flagging for a human decision instead of rebutting again.

Context: the prior reply rejected the `site/src/pages/index.astro` finding as out-of-scope (MPC-1.1's description and AC #4 both explicitly name this file as excluded, and the technical premise of the finding was independently confirmed correct). The latest review round accepts the scope rationale but still flags the underlying regression as an open concern and recommends a fast-follow task. Both sides are being reasonable here — it's a scope-vs-blast-radius tradeoff, not a factual disagreement, so it needs a human call rather than another automated rebuttal cycle.

dodizzle:
Human decision (HITL escalation resolved): merge as-is, accepting the temporary blast radius on the marketing homepage's "Estimation accuracy" card (it will render `—` until fixed) rather than expanding this PR's scope or holding it open.

Follow-up queued: **MPC-1.2** (depends on MPC-1.1) — updates `site/src/pages/index.astro` to consume `shipwrightPrsMerged` instead of the removed `estimationAccuracy`, or drops the card if that reads better.

Merging via admin override since branch protection requires an approving review and the existing review here is COMMENT-only.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

PR #3545, opened 2026-09-21
fix: use absolute worktree-path fallback in dev-task bundled-branch flow

dmcaulay:
## Summary
- The stale-bundle re-add path in `dev-task.md` (line 437) used relative `${SHIPWRIGHT_REPO_DIR:-repos}` / `${SHIPWRIGHT_WORKTREE_DIR:-worktrees}` fallback defaults, inconsistent with every other worktree add/remove command in the same file (lines 390, 411, 443, 448), which all default to `$HOME/src` / `$HOME/worktrees`.
- `git -C <dir> worktree add <path>` resolves `<path>` relative to `<dir>`, so following this line literally landed the worktree nested at `repos/<repo>/worktrees/...` instead of the intended top-level `worktrees/` dir.
- Found via a nightly `/learn-dream` batch pass mining session transcripts — this exact mistake recurred across 5+ documented sessions (WLS-3.2, PR review #2017, docs-freshness cron, dev-task POM-1.3, dev-task CBT-1.1), always traced back to an agent following this literal example text.

## Test plan
- [x] Diffed against the three sibling worktree add/remove commands in the same file to confirm the fix matches their established absolute-path convention exactly.
- [ ] No functional test possible for a markdown command spec; correctness verified by direct comparison to sibling lines.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3549, opened 2026-09-21
test: add sweeping worktree add/remove absolute-fallback regression guard (T-085)

dmcaulay:
## Summary
- Adds a sweeping regression guard to `dev-task.content.test.ts` that regex-extracts every `git -C ... worktree add`/`worktree remove` invocation in `dev-task.md` and asserts each uses the absolute-fallback form (`${SHIPWRIGHT_REPO_DIR:-$HOME/src}` / `${SHIPWRIGHT_WORKTREE_DIR:-$HOME/worktrees}`), never a bare relative default.
- Widens the file's existing narrower single-instance assertion (~line 712) to cover every invocation, so a future PR reintroducing a relative default anywhere in `dev-task.md` fails CI instead of silently recurring.
- Test-only change — no production file touched. The underlying bug was already fixed in #3545.

## Acceptance Criteria

| 1 | New it() block regex-extracts every git -C ... worktree add/remove invocation | MET | `WORKTREE_INVOCATION_RE` + `getWorktreeInvocations()`, sanity-checked (>0 matches) |
| 2 | Assertion confirms absolute-fallback form, never bare relative default | MET | `BARE_RELATIVE_DEFAULT` negative check + two positive `$HOME/`-prefix checks |
| 3 | No production file touched | MET | `git diff --stat` shows only dev-task.content.test.ts changed |
| 4 | Verification command passes | MET | `bun test plugins/shipwright/commands/dev-task.content.test.ts` → 95 pass, 0 fail |

## Test Plan
- [x] `bun test plugins/shipwright/commands/dev-task.content.test.ts` — 95 pass, 0 fail
- [x] `bunx biome check` on the changed file — clean
- [x] `bun run typecheck` in `plugins/shipwright` — clean

## Note
While building this test, a genuine pre-existing bare-relative-default bug was found outside this task's scope: `dev-task.md` line ~419 uses `${SHIPWRIGHT_REPO_DIR:-repos}` (not the absolute-fallback form) in the `GH_REPO` derivation inside the merged-PR stale-bundle-branch check — not a `worktree add/remove` invocation, so out of scope for this regex, and correctly left untouched. Worth a follow-up task (same bug class as #3545, on an adjacent line #3545 didn't cover).


dodizzle (approved):
Verdict: APPROVE — Test-only regression guard for worktree add/remove absolute-fallback defaults; verified regex against dev-task.md (all 5 real invocations correctly matched and passing), no findings.

PR #3553, opened 2026-09-21
fix(security): bump devalue to 5.9.2+ to patch DoS (CVE-2026-81176)

dodizzle:
## Summary
- Bump `devalue` to `>=5.9.2` in `site/bun.lock` (via a `site/package.json` override) to patch a quadratic-work DoS in `devalue.parse` when given crafted out-of-bounds indices.

## Acceptance Criteria
| 1 | `devalue` bumped to a patched version (>=5.9.2) in `site/bun.lock` | MET | `site/package.json` override pins `devalue@^5.9.2`; `site/bun.lock` resolves to `devalue@5.9.4` |

## Test Plan
- `bun install --frozen-lockfile` in `site/` succeeds with no lockfile drift

Fixes GHSA-9rgm-9g3h-6x36 / CVE-2026-81176 (CVSS 5.3, moderate).


dodizzle:
Reviewed the dependency-risk finding on this PR and did not implement changes — there is nothing to reproduce.

- `dependency-risk@377136942a65ffe3d79fd79cc4e7fe28299a353c`: **REJECT (nothing to reproduce).** `devalue` 5.8.1 → 5.9.4 is a minor-version security patch with no reported build/test breakage. The `review` recommendation was driven purely by the "CVE in body → review minimum" heuristic floor, not by any failing verification command — the finding's own reasoning states there is no associated failing build/test/lint command. Per `references/dependency-patch.md`, the reproduce → attempt-catalog-fix → re-verify protocol terminates at step 2 (does not reproduce as described), so there is nothing to fix under the bounded remediation catalog.

Verification performed:

- **CVE applicability confirmed.** GHSA-9rgm-9g3h-6x36 / CVE-2026-81176 (CVSS 5.3, medium) is `devalue.parse` failing to reject out-of-bounds indices, letting a crafted payload alternate between array representations and produce quadratic work (DoS). OSV reports the affected range as `>=0 <5.9.2`, fixed in 5.9.2. The `^5.9.2` override resolves to `devalue@5.9.4` in `site/bun.lock`, which is remediated.
- **5.9.4 is not itself flagged.** An OSV version-scoped query for `devalue@5.9.4` (npm) returns zero advisories. The nine known `devalue` advisories all have `fixed` versions at or below 5.9.2, so 5.9.4 clears every one.
- **No breaking change / no first-party call sites.** 5.8 → 5.9 is a minor bump, and `devalue` appears only in the `overrides` block of `site/package.json` (a transitive pin via `astro`), with no first-party `import`/`require` of `devalue` anywhere under `site/`. There is no API surface for this repo to break against.
- **Lockfile churn is a no-op re-hoist, not a version change.** The unrelated `site/bun.lock` movement (`satteri` 0.9.5→0.10.5, `@astrojs/markdown-satteri` 0.3.4→0.4.1, `@napi-rs/wasm-runtime` 1.1.6→1.2.4) is bun flattening packages that were *already present on `main`* as nested entries under `astro/@astrojs/markdown-satteri/…`. Same package set, different hoisting. The one substantive effect is that `@astrojs/mdx@7.0.4`'s **optional** peer `@astrojs/markdown-satteri: "^0.3.1"` now resolves to the hoisted 0.4.1 instead of a dedicated 0.3.4 copy — noted for awareness; the `site build / brand-lint / smoke` check passes green on this SHA, so it is not causing a regression here.

Leaving the bump as-is. Flagging for human awareness that this is a security-motivated dependency bump, not a code defect.

Separately: the PR title has been updated to Conventional-Commit form so the `pr-title-lint` check (the only failing check on this PR) can pass.

dmcaulay (approved):
Verdict: APPROVE — Minimal, correctly-scoped devalue override (5.8.1 → 5.9.4) that patches CVE-2026-81176; follows the existing overrides convention, resolves to a non-vulnerable version, no first-party devalue usage, and the incidental satteri/@napi-rs lockfile re-hoisting checks out as the same package set already on main. dependency risk: review — CVE-in-body heuristic floor, but independently verified with no unaddressed concerns.

